Issue #21, opened 2018-12-05
Create a classifier to detect bugs that need QA


marco-c:
Needed for https://github.com/mozilla/relman-auto-nag/issues/227.

To do this, we'll need to collect some labels.
We can automatically create some positive labels by getting bugs that have the `qawanted` keyword or that have `qe-verify` flag.
We can't automatically create negative labels because we can't be sure that QA was not needed when the keyword/flag was not set.

ayush1999:
@marco-c The function for generating labels will return `True` if `qawanted` is a label and `False` otherwise, right?

marco-c:
#22 was the first step here, adding a function to create the labels. Now we need to add some code to run.py to use it.

ayush1999:
@marco-c  The `args.goal` in run.py would need to have a new QA option right? Also, are there some models present at the moment to test these changes?

marco-c:
@ayush1999 yes, there should be a new `qaneeded` option. There isn't a model present yet, since you're just adding the code to train now :smile:

marco-c:
#27 added the code to train.

The result seems too good to be true though, so we need to figure out if there is some feature that should be not considered in the 'qaneeded' case.

ayush1999:
@marco-c Currently we are considering labels that have `qawanted` or `qe-verify` keyword. What should we be experimenting with to check the validity of the model? Should we be checking for a specific keyword that might cause the condition to be false? (So we'd have to remove that condition too)

marco-c:
It's possible that we are using those keywords as features, which clearly would make the problem too easy :smile: 
We have to remove these keywords from the training for this 'qaneeded' case. Currently it's not easy to do it, so this will have to wait for the refactoring from https://github.com/marco-c/bugbug/issues/6.

marco-c:
#6 is done, and c95a6f809d45991489591066ebf5ff0b86ff4d2f made it possible to make the feature extractors configurable.
For the QA wanted case, we should clearly remove `qawanted` from the keywords extractor.

marco-c:
Here's an example for the regression model: https://github.com/marco-c/bugbug/commit/e9d291aaae297c885688db7583a4bfc6200ede8f#diff-5f2351779b955b08e775eb4234c3f2ebR28.

We should do something similar for the qaneeded model, with the only difference being that in that case we want to ignore the `qawanted` keyword.

marco-c:
Actually, we still need to make use of #5 when that's ready.

[...]

PR #22, opened 2018-12-06
Initial QA labels function

ayush1999:
The function `get_qa_tracking_labels()` reads data from the json file and returns a Dict containing id to True/False label. Currently, the label depends on whether the labels `qawanted` or `qe-verify` are present in the data.

Initial changes for #21

### DIFF
--- a/bugbug/labels.py
+++ b/bugbug/labels.py
@@ -34,6 +34,27 @@ def get_tracking_labels():
     return classes
 
 
+def get_qa_needed_labels():
+    classes = {}
+
+    for bug_data in bugzilla.get_bugs():
+        bug_id = int(bug_data['id'])
+
+        for entry in bug_data['history']:
+            for change in entry['changes']:
+                if change['added'].startswith('qawanted'):
+                    classes[bug_id] = True
+                elif 'flags' in entry:
+                    for flag in entry['flags']:
+                        if flag['name'].startswith('qe-verify'):
+                            classes[bug_id] = True
+
+        if bug_id not in classes:
+            classes[bug_id] = False
+
+    return classes
+
+
 def get_bugbug_labels(kind='bug', augmentation=False):
     assert kind in ['bug', 'regression']
 


bugbug/labels.py
```diff
@@ -33,6 +33,25 @@ def get_tracking_labels():
...
+    classes = {}
+
+    for bug_data in bugzilla.get_bugs():
+        bug_id = int(bug_data['id'])
+
+        for entry in bug_data['history']:
+            for change in entry['changes']:
+                if change['added'].startswith('qawanted'):
+                    classes[bug_id] = True
+                elif "flags" in entry.keys():
```
marco-c: Always use single quotes for strings instead of double quotes

bugbug/labels.py
```diff
@@ -33,6 +33,25 @@ def get_tracking_labels():
 
     return classes
 
+def get_qa_tracking_labels():
```
marco-c: Let's call this `get_qa_needed_labels()`

.gitignore
```diff
@@ -2,4 +2,4 @@
 *.model
 
 data/bugs.json
-data/commits.json
+data/commits.json
```
marco-c: Please remove this unrelated change from the PR

bugbug/labels.py
```diff
@@ -33,6 +33,25 @@ def get_tracking_labels():
...
+    classes = {}
+
+    for bug_data in bugzilla.get_bugs():
+        bug_id = int(bug_data['id'])
+
+        for entry in bug_data['history']:
+            for change in entry['changes']:
+                if change['added'].startswith('qawanted'):
+                    classes[bug_id] = True
+                elif 'flags' in entry.keys():
```
marco-c: No need to do `entry.keys()`, you can just do `'flags' in entry`.

marco-c:
Almost perfect, just need to do a small change I suggested and you should fix the flake8 issues (you can see them on Travis, or you can run flake8 locally).

ayush1999:
@marco-c Done!

PR #31, opened 2018-12-12
Add qawanted to keywords

### LINKED ISSUE #21: Create a classifier to detect bugs that need QA

ayush1999:
Added the `qawanted` keyword to `bugfeatures.keywords()` in case of qaneeded classification.

Fixes #21.

### DIFF
--- a/bugbug/models/qaneeded.py
+++ b/bugbug/models/qaneeded.py
@@ -24,7 +24,7 @@ class QANeededModel(Model):
             bug_features.has_str(),
             bug_features.has_regression_range(),
             bug_features.severity(),
-            bug_features.keywords(),
+            bug_features.keywords(set(['qawanted'])),
             bug_features.is_coverity_issue(),
             bug_features.has_crash_signature(),
             bug_features.has_url(),


marco-c:
What are the results of the training after this change?

ayush1999:
```
ayush99@ayush99:~/GitHub/bugbug$ python run.py --train --goal=qaneeded
(441, 1687513) (441,)
(49, 1687513) (49,)
CV Accuracy: 0.89 (+/- 0.04)
Accuracy: 0.8979591836734694
Precision: 0.896551724137931
Recall: 0.9285714285714286
[[18  3]
 [ 2 26]]
```

The accuracy decreases, which is what we wanted. (overfitting reduces)

PR #33, opened 2018-12-13
Cleanup title and comments

ayush1999:
Initial changes : Added a small change for replacing all links with URL token. Will add more changes upon approval.

### DIFF
--- a/bugbug/bug_features.py
+++ b/bugbug/bug_features.py
@@ -179,6 +179,9 @@ class BugExtractor(BaseEstimator, TransformerMixin):
     def fit(self, x, y=None):
         return self
 
+    def cleanup(self, text):
+        return re.sub(r"http\S+", "URL", text)
+
     def transform(self, bugs):
         results = []
 
@@ -208,7 +211,7 @@ class BugExtractor(BaseEstimator, TransformerMixin):
             result = {
                 'data': data,
                 'title': bug['summary'],
-                'comments': ' '.join([c['text'] for c in bug['comments']]),
+                'comments': ' '.join([self.cleanup(c['text']) for c in bug['comments']]),
             }
 
             if self.commit_messages_map is not None:
--- a/bugbug/bug_features.py
+++ b/bugbug/bug_features.py
@@ -211,11 +211,15 @@ class BugExtractor(BaseEstimator, TransformerMixin):
             # TODO: Try simply using all possible fields instead of extracting features manually.
 
             for cleanup_function in self.cleanup_functions:
-                result = {
-                    'data': data,
-                    'title': bug['summary'],
-                    'comments': ' '.join([cleanup_function(c['text']) for c in bug['comments']]),
-                }
+                bug['summary'] = cleanup_function(bug['summary'])
+                for c in bug['comments']:
+                    c['text'] = cleanup_function(c['text'])
+
+            result = {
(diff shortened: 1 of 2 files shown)


bugbug/bug_features.py
```diff
@@ -179,6 +179,9 @@ def __init__(self, feature_extractors, commit_messages_map=None):
     def fit(self, x, y=None):
         return self
 
+    def cleanup(self, text):
+        return re.sub(r'http\S+', 'URL', text)
```
marco-c: This will also match invalid URLs, but it should be fine.

marco-c: I'd define, outside of this class, a list of "text cleanup" classes (just like the classes we have for feature extraction). Then you call them either in this `cleanup` function, or in `transform`.

bugbug/models/bug.py
```diff
@@ -36,6 +36,7 @@ def __init__(self, lemmatization=False):
             bug_features.landings(),
             bug_features.title(),
             bug_features.comments(),
+            bug_features.cleanup_url(),
```
marco-c: Sorry for the confusion, I didn't mean to make it another feature extractor. I just meant we should have "cleanup" functions implemented similarly to feature extractors.

marco-c: So, call the cleanup functions in the place where you were calling it before, but making it more generic.

marco-c: Something like:
```
for cleanup_function in cleanup_functions:
    summary = cleanup_function(summary)
    comments = cleanup_function(comments)
```

marco-c:
> @marco-c I've created a new class that replaces all URLs with the token. Should I do so for remaining changes? (stack traces, file references etc)

Let's add one cleanup function per PR. After you add it, please also comment in the PR with the results of the training before/after your change (with the 'bug' and 'regression' models).

ayush1999:
@marco-c Current results:
Before changes:
```
(1627, 101401) (1627,)
(181, 101401) (181,)
CV Accuracy: 0.89 (+/- 0.05)
Accuracy: 0.9171270718232044
Precision: 0.9642857142857143
Recall: 0.8709677419354839
[[85  3]
 [12 81]]
```
After changes:
```
(1627, 85316) (1627,)
(181, 85316) (181,)
CV Accuracy: 0.90 (+/- 0.04)
Accuracy: 0.9171270718232044
Precision: 0.9642857142857143
Recall: 0.8709677419354839
[[85  3]
 [12 81]]
```

ayush1999:
In case of regression, results are exactly same.

bugbug/bug_features.py
```diff
@@ -205,11 +210,12 @@ def transform(self, bugs):
 
             # TODO: Try simply using all possible fields instead of extracting features manually.
 
-            result = {
-                'data': data,
-                'title': bug['summary'],
-                'comments': ' '.join([c['text'] for c in bug['comments']]),
-            }
+            for cleanup_function in self.cleanup_functions:
+                result = {
```
marco-c: Thanks, the patch is looking better!

A couple of things to fix:
1. The cleanup functions should be applied to both the summary and the comments;
2. The result should not be redefined multiple times, but just once. In the loop you should just apply the cleanup functions to the textual fields.

PR #36, opened 2018-12-15
replace filenames with token

ayush1999:
Added a cleanup function to replace all occurrences of words ending with .py or .json with the token 'FILE_REFERENCE'.

The results are same before and after adding this token. @marco-c Which other file extensions (apart from .py and .json) should we be replacing here?


marco-c:
> The results are same before and after adding this token. @marco-c Which other file extensions (apart from .py and .json) should we be replacing here?

JS files ('.js', '.jsm'), HTML/CSS files ('.html','.css'), C/C++ files ('.c', '.cpp', '.h')

ayush1999:
@marco-c Results:
Before changes:
```
(1627, 85256) (1627,)
(181, 85256) (181,)
CV Accuracy: 0.90 (+/- 0.05)
Accuracy: 0.9116022099447514
Precision: 0.963855421686747
Recall: 0.8602150537634409
[[85  3]
 [13 80]]
```
After changes:
```
(1627, 83732) (1627,)
(181, 83732) (181,)
CV Accuracy: 0.90 (+/- 0.04)
Accuracy: 0.9171270718232044
Precision: 0.9642857142857143
Recall: 0.8709677419354839
[[85  3]
 [12 81]]
```
Tl;dr : Accuracy increases.

bugbug/bug_features.py
```diff
@@ -185,11 +185,15 @@ def cleanup_url(text):
     return re.sub(r'http\S+', 'URL', text)
 
 
+def cleanup_fileref(text):
+    return re.sub(r'\w+\.py\s?|\w+\.json\s?|\w+\.js\s?|\w+\.jsm\s?|\w+\.html\s?|\w+\.css\s?|\w+\.c\s?|\w+\.cpp\s?|\w+\.h\s?', 'FILE_REFERENCE ', text)
```
marco-c: Using `\s?` makes this match a lot of things that shouldn't be matched.
E.g. any string that contains '.c' will be matched. Instead, we should only match strings which end with `.c`.

Maybe just replacing `\s?` with `\b` and removing the whitespace at the end of `FILE_REFERENCE` will be enough.

PR #46, opened 2018-12-18
Code reference url token.

ayush1999:
Added a new "CODE_REFERENCE_URL" to cleanup_url function. It tests for the presence of words starting with `https://hg.mozilla` or `https://searchfox` and replaces them with this token.

### DIFF
--- a/bugbug/bug_features.py
+++ b/bugbug/bug_features.py
@@ -182,6 +182,11 @@ class component(object):
 
 
 def cleanup_url(text):
+    mozilla_text = ['mozilla', 'fox', 'bugzilla']
+    for mozilla_keyword in mozilla_text:
+        if mozilla_keyword in text:
+            return re.sub(r'http\S+', 'CODE_REFERENCE_URL', text)
+
     return re.sub(r'http\S+', 'URL', text)
 
 
--- a/bugbug/bug_features.py
+++ b/bugbug/bug_features.py
@@ -182,11 +182,7 @@ class component(object):
 
 
 def cleanup_url(text):
-    mozilla_text = ['mozilla', 'fox', 'bugzilla']
-    for mozilla_keyword in mozilla_text:
-        if mozilla_keyword in text:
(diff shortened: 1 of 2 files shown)


ayush1999:
Added another option to check for `http` along with `https`.

tests/test_cleanup_functions.py
```diff
@@ -9,6 +9,7 @@
 def test_cleanup_url():
     tests = [
         ('This code lies in https://github.com/marco-c/bugbug', 'This code lies in URL'),
+        ('Another url can be https://hg.mozilla.org/camino/ or https://google.com', 'Another url can be CODE_REFERENCE_URL or URL')
     ]
```
marco-c: Also add a test for the http case.

bugbug/bug_features.py
```diff
@@ -182,6 +182,10 @@ def __call__(self, bug):
 
 
 def cleanup_url(text):
+    mozilla_text = ['hg.mozilla', 'searchfox', 'dxr.mozilla']
+    for mozilla_keyword in mozilla_text:
+        if mozilla_keyword in text:
+            text = re.sub(r'https://{0}\S+|http://{0}\S+'.format(mozilla_keyword), 'CODE_REFERENCE_URL', text)
```
marco-c: Instead of applying multiple regexes, one for each URL, you should be able to have just one regex.
Also, you can make the `s` optional to avoid having the `|` between the https case and the http case.

tests/test_cleanup_functions.py
marco-c: And for the other URLs.

bugbug/bug_features.py
ayush1999: We have two tokens to use: URL and CODE_REFERENCE_URL. How can I use only one regex to make appropriate substitution?

marco-c: I mean one regex for CODE_REFERENCE_URL, currently you're making one per loop iteration (so three)

ayush1999: Ah, got it.

PR #52, opened 2019-01-05
Cleanup responses from text

ayush1999:
Added a function to cleanup responses from text comments.


ayush1999:
Results:
Before changes:
```
(11173, 438922) (11173,)
(1882, 438922) (1882,)
(1118, 438922) (1118,)
CV Accuracy: 0.92 (+/- 0.02)
```
After changes:
```
(11173, 438922) (11173,)
(1882, 438922) (1882,)
(1118, 438922) (1118,)
CV Accuracy: 0.93 (+/- 0.02)
```

marco-c:
The `regression` case is a little better without this, so maybe we have to make it optional.
Let's keep this open for a while until we support making cleanup steps optional.

ayush1999:
There is another similar case for removing code snippets from comments. Should I create a PR for that? (Or we should deal with that later?)

ayush1999:
@marco-c Code snippets, similar to responses, start with `> +` to `\n` . Since this PR removes all text between `>` to `\n`, code snippets are also removed.

ayush1999:
Should I change the function name to something more generic?

marco-c:
> @marco-c Code snippets, similar to responses, start with > + to \n . Since this PR removes all text between > to \n, code snippets are also removed.

Ah no, that's just one of the ways people write code snippets, but it is not a rule.

> Should I change the function name to something more generic?

No, it's OK.

ayush1999:
Ok, I'll let this be open for now.

ayush1999:
@marco-c I've identified a few cases where this works well. Show I add these changes only for those cases? (Similar to #56)

marco-c:
> @marco-c I've identified a few cases where this works well. Show I add these changes only for those cases?

Yes, which cases?

ayush1999:
It works well for bugs and qaneeded. For others, not so much. Should I just add the function for now, similar to the other PRs?

marco-c:
Yes

bugbug/bug_features.py
```diff
@@ -134,6 +134,10 @@ def cleanup_fileref(text):
     return re.sub(r'\w+\.py\b|\w+\.json\b|\w+\.js\b|\w+\.jsm\b|\w+\.html\b|\w+\.css\b|\w+\.c\b|\w+\.cpp\b|\w+\.h\b', '__FILE_REFERENCE__', text)
 
 
+def cleanup_responses(text):
+    return re.sub('>.*?\\n', ' ', text)
```
marco-c: Wouldn't `>.*` or `>[^\n]+` be better? After matching `>`, you'd match all characters until a line terminator.

ayush1999:
Using these will not remove the `\n`. Is that what we want? (Won't it be better to remove `\n` too? Otherwise we'd get a series of consecutive `\n` after the cleanup.)

marco-c:
It's fine if we have multiple `\n`, as they are skipped anyway.

Also, if a `>` line doesn't end with `\n` (e.g. if it's the last line), your regex wouldn't match it. Try to add a test containing a `>` line at the end without `\n`.

PR #57, opened 2019-01-09
Seperate cleanup functions per model

ayush1999:
Added a new `cleanup_functions` per model, similar to feature extractors, which is called by BugExtractor.


marco-c:
Thanks!

PR #56, opened 2019-01-09
Substitute hex numbers in word by token

ayush1999:
Currently, any occurrence of a hex number will be replaced by __HEX_NUMBER__ .

PR #65, opened 2019-01-14
Cleanup .dll words

ayush1999:
Changes made: Added a regex for replacing all `.dll` words with token.


ayush1999:
For --goal=bug,
After changes: 
```
(6528, 240771) (6528,)
(1868, 240771) (1868,)
(653, 240771) (653,)
CV Accuracy: 0.93 (+/- 0.02)
No confidence threshold - 653 classified
[[508  55]
 [  4  86]]

          1       0.99      0.90      0.96      0.95      0.93      0.86       563
          0       0.61      0.96      0.90      0.74      0.93      0.87        90

avg / total       0.94      0.91      0.95      0.92      0.93      0.86       653
.
.
.
Confidence threshold > 0.9 - 444 classified
[[373   6]
 [  0  65]]

          1       1.00      0.98      1.00      0.99      0.99      0.98       379
          0       0.92      1.00      0.98      0.96      0.99      0.99        65

avg / total       0.99      0.99      1.00      0.99      0.99      0.98       444
```

After changes:
```
(6528, 240610) (6528,)
(1868, 240610) (1868,)
(653, 240610) (653,)
CV Accuracy: 0.93 (+/- 0.01)
No confidence threshold - 653 classified
[[506  57]
 [  5  85]]

          1       0.99      0.90      0.94      0.94      0.92      0.84       563
          0       0.60      0.94      0.90      0.73      0.92      0.85        90

avg / total       0.94      0.91      0.94      0.91      0.92      0.85       653
.
.
.
Confidence threshold > 0.9 - 439 classified
[[367   7]
 [  0  65]]

          1       1.00      0.98      1.00      0.99      0.99      0.98       374
          0       0.90      1.00      0.98      0.95      0.99      0.98        65

avg / total       0.99      0.98      1.00      0.98      0.99      0.98       439
```

marco-c:
> The difference isn't much, should this be added?

As in #56, let's add the function only, without using it in any model yet.

bugbug/bug_features.py
```diff
@@ -134,6 +134,10 @@ def cleanup_fileref(text):
     return re.sub(r'\w+\.py\b|\w+\.json\b|\w+\.js\b|\w+\.jsm\b|\w+\.html\b|\w+\.css\b|\w+\.c\b|\w+\.cpp\b|\w+\.h\b', '__FILE_REFERENCE__', text)
 
 
+def cleanup_dll(text):
+    return re.sub(r'\w+\.dll', '__DLL_NAME__', text)
```
marco-c: Let's only replace when either the word ends at `.dll`, or when there is a `@` after `.dll`.

bugbug/bug_features.py
```diff
@@ -134,6 +134,10 @@ def cleanup_fileref(text):
     return re.sub(r'\w+\.py\b|\w+\.json\b|\w+\.js\b|\w+\.jsm\b|\w+\.html\b|\w+\.css\b|\w+\.c\b|\w+\.cpp\b|\w+\.h\b', '__FILE_REFERENCE__', text)
 
 
+def cleanup_dll(text):
+    return re.sub(r'\w+\.dll |\w+\.dll@\w+', '__DLL_NAME__ ', text)
```
marco-c: A word doesn't always finish with a space, but it could finish with other characters (e.g. a comma). You can use `\b` to define a word boundary.
Also, it would be better to have the common part at the beginning and then the variable part in brackets at the end.
E.g. if I want to match file1 and file2, I'd write file[12] and not file1|file2.

ayush1999:
Right, will do that. Also, for words such as scdetour.dll@0x2dd77, (which have .dll but also end in a hex number), should cleanup_dll replace the entire word with the DLL_NAME token, or only the first part should be replaced? (Should it become DLL_NAME or DLL_NAME@0x2dd77)

marco-c:
Only the first part, so it will become __DLL_NAME__@0x...

marco-c:
We should also skip the Firefox DLLs, but we can do that in a follow-up.

ayush1999:
@marco-c Please review.

PR #77, opened 2019-01-18
Cleanup crash stats

ayush1999:
Added the cleanup_crash function with tests.


ayush1999:
For --goal=bug, The CV accuracy increases from `0.92 +/- 0.03` to `0.93 +/- 0.03`. No other significant difference.

marco-c:
Not bad, considering the simplicity of the change!

PR #78, opened 2019-01-19
Only show meaningful features

### LINKED ISSUE #67: Only show "meaningful enough" feature importances
Right now I handpicked the number of most important features to show (the top 20).

We should use an algorithm to select what to show. The simplest two alternatives are:
1. Show all the contributing features that are at least X% of the sum of the contributions.
2. Show the top contributing feature, and all other contributing features that are at least X% of the top one.

yatinmaan:
Show the top contributing feature, and all other contributing features that are at least 15% of the top one.

Fixes #67

### DIFF
--- a/bugbug/model.py
+++ b/bugbug/model.py
@@ -30,7 +30,29 @@ class Model():
     def get_feature_names(self):
         return []
 
-    def train(self):
+    def get_important_features(self, cutoff, shap_values):
+        # Calculate the values that represent the fraction of the model output variability attributable
+        # to each feature across the whole dataset.
+        shap_sums = shap_values.sum(0)
+        abs_shap_sums = np.abs(shap_values).sum(0)
+        rel_shap_sums = abs_shap_sums / abs_shap_sums.sum()
+
+        cut_off_value = cutoff * np.amax(rel_shap_sums)
+
+        # Get indices of features that pass the cut off value
+        top_feature_indices = np.where(rel_shap_sums >= cut_off_value)[0]
+        # Get the importance values of the top features from their indices
+        top_features = np.take(rel_shap_sums, top_feature_indices)
+        # Gets the sign of the importance from shap_sums as boolean
+        is_positive = (np.take(shap_sums, top_feature_indices)) >= 0
+        # Stack the importance, indices and shap_sums in a 2D array
+        top_features = np.column_stack((top_features, top_feature_indices, is_positive))
+        # Sort the array (in decreasing order of importance values)
+        top_features = top_features[top_features[:, 0].argsort()][::-1]
+
+        return top_features
+
+    def train(self, importance_cutoff=0.15):
         classes = self.get_labels()
         class_names = sorted(list(set(classes.values())), reverse=True)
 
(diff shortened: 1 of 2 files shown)


yatinmaan:
Should the cut off percentage be configurable via cli ?

I just chose 15% because it shows ~20 features.

marco-c:
It looks good so far. Could you implement the same feature for `classify`?
I'd move the code to a separate utility function that both `train` and `classify` can use.

> Should the cut off percentage be configurable via cli?

No, let's just add a `importance_cutoff` parameter to train and classify, with 0.15 being the default.

bugbug/model.py
```diff
@@ -73,14 +73,22 @@ def bugs():
...
+
+            cut_off_percentage = 15
+            cut_off_value = (cut_off_percentage / 100) * np.amax(rel_shap_sums)
+
+            top_feature_indices = np.where(rel_shap_sums >= cut_off_value)[0][::-1]
+            top_vals = np.take(rel_shap_sums, top_feature_indices)
+            top_vals = np.column_stack((top_vals, top_feature_indices))
+            top_vals = top_vals[top_vals[:, 0].argsort()][::-1]
+
+            print('Feature ranking (atleast {}% of the top feature):'.format(cut_off_percentage))
```
marco-c: Let's print the number of top features, instead of the fact that it's XX% of the top.

bugbug/model.py
```diff
@@ -73,14 +73,22 @@ def bugs():
...
+            for i, row in enumerate(top_vals):
```
marco-c: Nit: Unpack the row here, so you don't have to do row[0] and row[1] later.

bugbug/model.py
```diff
@@ -73,14 +73,22 @@ def bugs():
...
             shap_sums = np.abs(shap_values).sum(0)
             rel_shap_sums = shap_sums / shap_sums.sum()
-            indices = np.argsort(rel_shap_sums)[::-1][:20]
-            for i, index in enumerate(indices):
-                print('{}. \'{}\' ({})'.format(i + 1, feature_names[index], rel_shap_sums[index]))
+
+            cut_off_percentage = 15
+            cut_off_value = (cut_off_percentage / 100) * np.amax(rel_shap_sums)
+
+            top_feature_indices = np.where(rel_shap_sums >= cut_off_value)[0][::-1]
```
marco-c: No need to reverse the array here.

bugbug/model.py
```diff
@@ -73,14 +73,22 @@ def bugs():
...
             # to each feature across the whole dataset.
             shap_sums = np.abs(shap_values).sum(0)
             rel_shap_sums = shap_sums / shap_sums.sum()
-            indices = np.argsort(rel_shap_sums)[::-1][:20]
-            for i, index in enumerate(indices):
-                print('{}. \'{}\' ({})'.format(i + 1, feature_names[index], rel_shap_sums[index]))
+
+            cut_off_percentage = 15
+            cut_off_value = (cut_off_percentage / 100) * np.amax(rel_shap_sums)
+
```
marco-c: Nit: it'd be nice to have (short) comments explaining the steps here. E.g. `# Get the indices of the features that pass the cutoff value` for the first step.

marco-c:
@yatinmaan are you planning to still work on this?

yatinmaan:
> @yatinmaan are you planning to still work on this?

Yes

bugbug/model.py
```diff
@@ -30,7 +30,29 @@ def __init__(self, lemmatization=False):
...
+        cut_off_value = cutoff * np.amax(rel_shap_sums)
+
+        # Get indices of features that pass the cut off value
+        top_feature_indices = np.where(rel_shap_sums >= cut_off_value)[0]
+        # Get the importance values of the top features from their indices
+        top_features = np.take(rel_shap_sums, top_feature_indices)
+        # Get the shap_sums of the top features
+        top_shap_sums = np.take(shap_sums, top_feature_indices)
+        # Stack the importance, indices and shap_sums in a 2D array
+        top_features = np.column_stack((top_features, top_feature_indices, top_shap_sums))
```
marco-c: Since we're not really using the shap_sums, can we have an array of True/False (1/0) to tell whether the feature is positive or negative?

bugbug/model.py
```diff
@@ -69,14 +91,11 @@ def bugs():
...
-            shap_sums = np.abs(shap_values).sum(0)
-            rel_shap_sums = shap_sums / shap_sums.sum()
-            indices = np.argsort(rel_shap_sums)[::-1][:20]
-            for i, index in enumerate(indices):
-                print(f'{i + 1}. \'{feature_names[index]}\' ({rel_shap_sums[index]})')
+            important_features = self.get_important_features(importance_cutoff, shap_values)
+
+            print(f'\nTop {len(important_features)} Features :')
+            for i, [importance, index, _unused] in enumerate(important_features):
+                    print(f'{i + 1}. \'{feature_names[int(index)]}\' ({importance})')
```
marco-c: Let's show whether the feature contribution is positive or negative here too, like in `classify`.

yatinmaan:
@marco-c Updated.

PR #80, opened 2019-01-20
Initial bug_features test

### LINKED ISSUE #41: Add tests for bug feature extractors

ayush1999:
@marco-c Need your review on this before adding tests for other extractors. Currently, I've randomly added 4 bugs to `tests/data/test_bugs.json`. The test function reads from this json file. Is this approach correct? If yes, then approximately how many bugs should I add to the `test_bugs.json` file?

Fixes #41.


marco-c:
> @marco-c Need your review on this before adding tests for other extractors. Currently, I've randomly added 4 bugs to tests/data/test_bugs.json. The test function reads from this json file. Is this approach correct?

Yes! Nit: Rename the tests/data directory to tests/fixtures, which is a nicer name. I would also split the files for each thing you need to test, to make it easier to change them in the future if needed.

> If yes, then approximately how many bugs should I add to the test_bugs.json file?

You can just add the bugs you need in order to test the fuctions. E.g. for has_str you can add one without STR and one with STR.

ayush1999:
@marco-c Please review.

tests/test_bug_features.py
```diff
@@ -0,0 +1,141 @@
...
+        for line in f:
+            data.append(check(json.loads(line)))
+
+    return data
+
+
+def test_has_str():
+
+    bugs = read('tests/features/has_str.json', has_str)
+    results = ['yes', None]
```
marco-c: There should also be a `no` case

tests/test_bug_features.py
```diff
@@ -0,0 +1,141 @@
...
+
+    bugs = read('tests/features/title.json', title)
+    results = [['fail'], []]
+    assert results == bugs
+
+
+def test_product():
+
+    bugs = read('tests/features/product.json', product)
+    results = ['Core', 'Core']
```
marco-c: Let's get two bugs that have two different products

tests/test_bug_features.py
```diff
@@ -0,0 +1,141 @@
...
+
+    bugs = read('tests/features/has_str.json', has_str)
+    results = ['yes', None]
+    assert results == bugs
+
+
+def test_has_regression_range():
+
+    bugs = read('tests/features/has_regression_range.json', has_regression_range)
+    results = ['yes', None]
```
marco-c: Nit: here and in the other test functions, you can avoid having the `results` variable and just do `assert [...] == bugs`.

tests/test_bug_features.py
```diff
@@ -0,0 +1,141 @@
...
+def test_has_str():
+
+    bugs = read('tests/features/has_str.json', has_str)
+    results = ['yes', None]
+    assert results == bugs
+
+
+def test_has_regression_range():
+
+    bugs = read('tests/features/has_regression_range.json', has_regression_range)
```
marco-c: Nit: change the name of this `bugs` variable, since it's not really a list of bugs. Maybe `results` or `features`.

tests/test_bug_features.py
```diff
@@ -0,0 +1,141 @@
...
+    data = []
+    with open(path, 'r') as f:
+        for line in f:
+            data.append(check(json.loads(line)))
+
+    return data
+
+
+def test_has_str():
+
```
marco-c: Nit: no white line between `def` and the first line of the function

tests/test_bug_features.py
```diff
@@ -0,0 +1,141 @@
...
+from bugbug.bug_features import patches
+from bugbug.bug_features import product
+from bugbug.bug_features import severity
+from bugbug.bug_features import title
+from bugbug.bug_features import whiteboard
+
+
+def read(path, function):
+    check = function()
+    if not os.path.exists(path):
```
marco-c: Do an assertion here, as the file should exist.

tests/test_bug_features.py
```diff
@@ -0,0 +1,141 @@
...
+from bugbug.bug_features import keywords
+from bugbug.bug_features import landings
+from bugbug.bug_features import patches
+from bugbug.bug_features import product
+from bugbug.bug_features import severity
+from bugbug.bug_features import title
+from bugbug.bug_features import whiteboard
+
+
+def read(path, function):
```
marco-c: `function` is actually a class, so I'd call it `cls` or `feature_extraction_cls`.

marco-c: And `check` should be `feature_extractor`.

tests/test_bug_features.py
marco-c: Or even better, make read do the assertions, by passing the expected values to it.

marco-c:
Also move the JSON files under the `tests/fixtures` directory. E.g. severity.json should be in `tests/fixtures/bug_features/severity.json`.

ayush1999:
Done. `test` now takes the expected results as an argument and tests it.

tests/test_bug_features.py
```diff
@@ -0,0 +1,96 @@
...
+
+def read(filename, cls, expected_results):
+    path = 'tests/fixtures/bug_features/' + filename
+    feature_extractor = cls()
+    assert os.path.exists(path)
+
+    data = []
+    with open(path, 'r') as f:
+        for line in f:
+            data.append(feature_extractor(json.loads(line)))
```
marco-c: Nit: you could assert here directly, doing something like:
```Python
results = (feature_extractor(json.loads(line)) for line in f)
for result, expected_result in zip(results, expected_results):
    assert result == expected_result
```

This way you will avoid building the `data` list, and you'll get a nicer message in case of errors.

tests/test_bug_features.py
```diff
@@ -0,0 +1,96 @@
...
+from bugbug.bug_features import landings
+from bugbug.bug_features import patches
+from bugbug.bug_features import product
+from bugbug.bug_features import severity
+from bugbug.bug_features import title
+from bugbug.bug_features import whiteboard
+
+
+def read(filename, cls, expected_results):
+    path = 'tests/fixtures/bug_features/' + filename
```
marco-c: Nit: use os.path.join instead of +

tests/test_bug_features.py
```diff
@@ -0,0 +1,96 @@
...
+    feature_extractor = cls()
```
marco-c: It looks like `cls` is a reserved keyword, let's use `feature_extractor_class` then.

Issue #82, opened 2019-01-20
Create constant sets directly instead of converting lists into sets


marco-c:
Instead of doing something like `set(['val1', 'val2'])`, we should do `{'val1', 'val2'}`.

ayush1999:
@Jeetu95 are you working on this? This involves a few subtle changes, so if you're not, I'll open a quick PR.

Jeetu95:
Yes @ayush1999, I am working on this

marco-c:
I'm adding some rules to the README.md to avoid conflicts. PRs with `reserved-for-beginners` label will be reserved to people who have never committed in the project before. Every issue is to be considered as being unassigned until there is a PR linked to it.

PR #92, opened 2019-01-21
Creation of constant sets directly instead of converting lists into sets

Jeetu95:
Fixes for Issue #82 

@marco-c  Please take a look
Thank you.


marco-c:
#82 is reserved for newcomers (people who haven't committed to the repository yet).

PR #90, opened 2019-01-21
'Is reporter a Mozillian' as a feature #30

poojan124:
Fixes #30 I have updated the code from #86. Please review this PR.

Issue #30, opened 2018-12-12
'Is reporter a Mozillian' as a feature


marco-c:
The first implementation will simply check if an email contains "@mozilla.com" or "@mozilla.org".

marco-c:
1. You'll need to add an additional feature extractor in bugbug/bug_features.py.
2. No, I don't think so. Most of them should be automated accounts. Could you list here all the emails that end with "@mozilla.bugs"?

poojan124:
` 'wptsync@mozilla.bugs' 
 'omphalos@mozilla.bugs'
 'intermittent-bug-filer@mozilla.bugs' `
This are 3 emails that ends with "@mozilla.bugs".

marco-c:
Yes we can ignore them.

PR #95, opened 2019-01-21
test added for #30 is_mozillian feature

poojan124:
This contains a test for `is_mozillian` feature added in #90.
The test is checking for 3 cases. 
* `XXX@YYY`
* `XXX@mozilla.com`
* `XXX@mozilla.org`
So it should return `[False, True, True]`.
Please review this PR and comment if any change needed.


marco-c:
There are some linting failures.

Issue #89, opened 2019-01-21
.so and .dylib should be cleaned up to as DLLs


marco-c:
.so and .dylib are DLLs on Linux and Mac. We should replace them with _DLL_NAME_ as we do for .dll files.

marco-c:
Fixed by #96.

PR #96, opened 2019-01-22
expand cleanup_dll() regex to match .so and .dylib

dbxnr:
Fix for #89 - expand regex to match .so and .dylib


marco-c:
Looks good, but could you expand the test in tests/test_cleanup_functions to test these too?

PR #97, opened 2019-01-22
Delta request-merge

ayush1999:
Changes proposed: A new bug_extractor, that returns the time difference between request date and merge date.

Example:
```
>>> a = delta_request_merge()
>>> a(bug)
('55.0', datetime.datetime(2017, 8, 8, 7, 0, tzinfo=<UTC>))
```

If I'm not wrong, 55 is the required delta here. (right?)
@marco-c Is this similar to what was required?

### DIFF
--- a/bugbug/bug_features.py
+++ b/bugbug/bug_features.py
@@ -4,8 +4,11 @@
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import re
+from datetime import datetime
+from datetime import timezone
 
 import pandas as pd
+from libmozdata import versions
 from sklearn.base import BaseEstimator
 from sklearn.base import TransformerMixin
 
@@ -129,6 +132,17 @@ class component(object):
 class is_mozillian(object):
     def __call__(self, bug):
         return any(bug['creator_detail']['email'].endswith(domain) for domain in ['@mozilla.com', '@mozilla.org'])
+
+
+class delta_request_merge(object):
+    def __call__(self, bug):
+        for change in bug['history']:
+            for ind_change in change['changes']:
+                if ind_change['added'].startswith('approval-mozilla'):
+                    uplift_request_datetime = datetime.strptime(change['when'], '%Y-%m-%dT%H:%M:%SZ').replace(tzinfo=timezone.utc)
+                    return versions.getCloserRelease(uplift_request_datetime)[1] - uplift_request_datetime
+
+        return None
 
 
 def cleanup_url(text):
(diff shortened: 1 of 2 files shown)


bugbug/bug_features.py
```diff
@@ -131,6 +134,12 @@ def __call__(self, bug):
         return any(bug['creator_detail']['email'].endswith(domain) for domain in ['@mozilla.com', '@mozilla.org'])
 
 
+class delta_request_merge(object):
+    def __call__(self, bug):
+        uplift_request_datetime = datetime.strptime(bug['history'][0]['when'], '%Y-%m-%dT%H:%M:%SZ').replace(tzinfo=timezone.utc)
```
marco-c: This is the date of the first modification to the bug, we should get the date of the first uplift request instead.

marco-c:
> If I'm not wrong, 55 is the required delta here. (right?)

55 is the version of Firefox.

You should calculate the difference between the uplift request date and the `datetime.datetime(2017, 8, 8, 7, 0, tzinfo=<UTC>)`

ayush1999:
For the first uplift request, I'd have to iterate through `bug['history']` to search for `added` starting from `approval-mozilla` right? (And return the corresponding date).

marco-c:
Exactly, look at the uplift model for an example.

ayush1999:
Yep, that's where I saw this. After these changes, we get `datetime.timedelta(26, 40819)` as a result for an arbitrary example. (26 days and 40819 seconds). In which format should this be standardized? (Something like 26.X days?)

marco-c:
Yes, I think days would be good.

ayush1999:
Ok, everything gets converted to days now (I'm not sure why. but the travis test takes forever). Also, @marco-c small question: I just noticed that https://github.com/mozilla/bugbug/blob/master/bugbug/bug_features.py#L135 uses the `commit` key in the bug, which isn't available in the bugs data (so I'm guessing this isn't a regular feature extractor.) . Can you tell me how is this supposed to work?

marco-c:
> Can you tell me how is this supposed to work?

The commit data is added to the bugs if you use `commit_data=True`: https://github.com/mozilla/bugbug/blob/b9bb067a90af68ef4fc4b0936e521c2b459cf723/bugbug/bug_features.py#L213.

bugbug/bug_features.py
```diff
@@ -132,6 +135,18 @@ def __call__(self, bug):
         return any(bug['creator_detail']['email'].endswith(domain) for domain in ['@mozilla.com', '@mozilla.org'])
 
 
+class delta_request_merge(object):
+    def __call__(self, bug):
+        for change in bug['history']:
```
marco-c: Nit: Call this `history` and `ind_change` `change`, to keep the same variable names as other parts of the project.

ayush1999:
Travis CI fails due to an error in `test_cleanup_synonyms`. This can be merged now; that test can be fixed in a separate PR.

marco-c:
> Travis CI fails due to an error in test_cleanup_synonyms. This can be merged now; that test can be fixed in a separate PR.

The error should be gone if you rebase on top of current master.

PR #98, opened 2019-01-22
Remove unnecessary formating from json.dump and added model file.

### LINKED ISSUE #93: Unnecessary string formatting when saving the model
https://github.com/mozilla/bugbug/blob/ab007344c5c799c08739b2d56f334afcaf242919/bugbug/model.py#L111 is not needed, we can use the variable directly.

shankarj67:
Removed the unnecessary string from the **json.dump()** and added the model **self.clf** which is required while dumping the model into the local system.

Fixes #93.

### DIFF
--- a/bugbug/model.py
+++ b/bugbug/model.py
@@ -104,7 +104,7 @@ class Model():
             print(metrics.confusion_matrix(y_test_filter, y_pred_filter, labels=class_names))
             print(classification_report_imbalanced(y_test_filter, y_pred_filter, labels=class_names))
 
-        joblib.dump(self.clf, self.__class__.__name__.lower())
+        joblib.dump(self, self.__class__.__name__.lower())
 
     @staticmethod
     def load(model_file_name):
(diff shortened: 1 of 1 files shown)


bugbug/model.py
```diff
@@ -104,7 +104,7 @@ def bugs():
             print(metrics.confusion_matrix(y_test_filter, y_pred_filter, labels=class_names))
             print(classification_report_imbalanced(y_test_filter, y_pred_filter, labels=class_names))
 
-        joblib.dump(self, f'{self.__class__.__name__.lower()}')
+        joblib.dump(self.clf, self.__class__.__name__.lower())
```
marco-c: Why storing only the classifier?

shankarj67: So that It will help us to predict the classes next time without any training. What would you suggest, which value to store in a dump?

marco-c: This is already the case right now, we are storing `self`. `clf` is a property of self, so we are storing `clf` too :)

shankarj67: Thanks a lot, I have changed my code and committed the same.

Issue #104, opened 2019-01-23
Extend the whiteboard feature to consider all possible whiteboard words


marco-c:
https://github.com/mozilla/bugbug/blob/1784a10346d9dedcb2e18e076f3d482e39be93a1/bugbug/bug_features.py#L83

marco-c:
Right now it's only looking for `whiteboard` or `[ux]`. It should instead get all possible whiteboard words.

ayush1999:
@marco-c It's currently looking for `memshrink` or `[ux]` right?. (You mentioned `whiteboard`) above.

marco-c:
Yes, see the comment above :smile:

ayush1999:
I found a list of commonly used whiteboard words here: https://wiki.mozilla.org/ReleaseEngineering/Bugzilla/Whiteboard . Should I add these?

marco-c:
No, unfortunately they are bound to change and there's no complete reference.
We should just use all whiteboard items we find in the whiteboard. The model will automatically discard the ones which are not meaningful.

ayush1999:
Oh, I see. Please correct me if I'm wrong: I'll have to implement an regex that finds all text between `[` and `]` in the whiteboard, extracts only the tags from each (`[memshrink:XXX]` becomes only `memshrink`), and appends it to the returned list. Right?

marco-c:
Correct!

PR #119, opened 2019-01-25
More whiteboard words

ayush1999:
#104

Based on these changes, if whiteboard consists of `[memshrink:XX][qf:p3]`, the function converts them into `memshrink` and `qf`. (Only words containing `:` are extracted for this purpose).


marco-c:
Please also add tests.

marco-c:
It might be simpler and more readable to split the whiteboard string with either ` `, `[`, or `]`, then remove the empty strings from the resulting list and operate on it.

ayush1999:
Done. @marco-c Please review.

tests/test_bug_features.py
```diff
@@ -72,7 +72,7 @@ def test_has_github_url():
 
 
 def test_whiteboard():
-    read('whiteboard.json', whiteboard, [['memshrink'], []])
+    read('whiteboard.json', whiteboard, [['memshrink', 'platform-rel-facebook'], []])
```
marco-c: Please add some additional tests, for example from the two bugs that I mentioned.

bugbug/bug_features.py
```diff
@@ -87,11 +87,12 @@ class whiteboard(object):
     def __call__(self, bug):
         ret = []
 
-        # TODO: Add any [XXX:YYY] that appears in the whiteboard as [XXX: only
-
-        for elem in ['memshrink', '[ux]']:
-            if elem in bug['whiteboard'].lower():
-                ret.append(elem)
+        for ele in bug['whiteboard'].lower().split(']'):
```
marco-c: I'm not so sure this will work, you can verify after you add the tests.

ayush1999:
Thanks, it works now. I guess I forgot to test cases where words outside `[ ]` come before words inside brackets.

bugbug/bug_features.py
```diff
@@ -87,11 +87,14 @@ class whiteboard(object):
     def __call__(self, bug):
         ret = []
 
-        # TODO: Add any [XXX:YYY] that appears in the whiteboard as [XXX: only
-
-        for elem in ['memshrink', '[ux]']:
-            if elem in bug['whiteboard'].lower():
-                ret.append(elem)
+        for whiteboard_word in bug['whiteboard'].lower().split(']'):
```
marco-c: You can use `re` to split by multiple characters at the same time. Split also by space, as otherwise something like `whiteboard1 whiteboard2` won't work.

Also add more cases to the test, creating "fake" bugs with only values in the `whiteboard` field.
- `whiteboard1 whiteboard2`
- `[whiteboard1]whiteboard2`
- `[whiteboard1] whiteboard2`
- `whiteboard1[whiteboard2]`
- `whiteboard1 [whiteboard2]`

marco-c: > You can use re to split by multiple characters at the same time.

In particular, `re.split`.

ayush1999: Splitting at space has a drawback though: In cases like `[AV:Quick Heal][regressed sept 6th][dll version is 3.0.1.*]`.

ayush1999:
Caveat: Dealing with `whiteboard1 whiteboard2`. Splitting at spaces isn't a good idea, since we have examples like `[dll version is 3.0.1.*]` which should not be split at spaces.

marco-c:
> Caveat: Dealing with whiteboard1 whiteboard2. Splitting at spaces isn't a good idea, since we have examples like [dll version is 3.0.1.*] which should not be split at spaces.

Right, this field is a pain :)

marco-c:
We could:
1. Split at '['
2. For each element of the split, if it contains `]` don't split at space, if it doesn't contain `]` split at space.

Something like:
```Python
# Split by '['
paren_splits = whiteboard.split('[')

# Split splits by space if they weren't in [ and ].
splits = []
for paren_split in paren_splits:
    if ']' in paren_split:
        splits.append(paren_split)
    else:
        splits += paren_split.split(' ')

# Remove empty splits
splits = [split for split in splits if split.strip() != '']

# For splits which contain ':', return both the whole string and the string before ':'.
splits += [split.split(':', 1)[0] for split in splits if ':' in split]
```

ayush1999:
Thanks! This works now! (with a few more subtle changes)

bugbug/bug_features.py
```diff
@@ -85,15 +85,20 @@ def __call__(self, bug):
...
 class whiteboard(object):
     def __call__(self, bug):
-        ret = []
 
-        # TODO: Add any [XXX:YYY] that appears in the whiteboard as [XXX: only
+        paren_splits = bug['whiteboard'].lower().split('[')
+        splits = []
+        for paren_split in paren_splits:
+            if ']' in paren_split:
+                paren_split = paren_split.split(']')
```
marco-c: Is this needed? Can't we just remove the ']' characters?

bugbug/bug_features.py
```diff
@@ -85,15 +85,20 @@ def __call__(self, bug):
 
 class whiteboard(object):
     def __call__(self, bug):
-        ret = []
 
-        # TODO: Add any [XXX:YYY] that appears in the whiteboard as [XXX: only
+        paren_splits = bug['whiteboard'].lower().split('[')
```
marco-c: Add the comments to explain the various steps, like I did in the previous comment

bugbug/bug_features.py
ayush1999: @marco-c  In case like `[whiteboard1] whiteboard`, just removing the `]` will result in both the words getting joined. Hence, it's better to split at ']'.

tests/fixtures/bug_features/whiteboard.json
```diff
@@ -1,2 +1,9 @@
...
...ox-56.0a1.en-US.win64.installer.exe [2017.07.02 mozilla central]).  This is in a Windows 8 machine!\n\n\nActual results:\n\nAfter restarting, nothing showed the bars were empty, the content was empty.  However (some?) buttons areas in the navigation bar were useful (tool tips show up).  A click above on the area below the close button made the options panel to show and I was able to restart in safe mode.  Firefox works in Safe Mode.  However, Firefox will not work with a new profile, even after refreshing Firefox.  I reverted to 2017.07.01. The 2017.07.03 does not fix anything for me.\n\n\nExpected results:\n\nFirefox Nightly should work mostly after an update.", "id": 12445163, "creation_time": "2017-07-04T00:08:07Z", "author": "johngraciliano@gmail.com"}, {"text": "\n\n*** This bug has been marked as a duplicate of bug 1377866 ***", "id": 12445279, "creation_time": "2017-07-04T02:36:27Z", "author": "yfdyh000@gmail.com"}, {"text": "John, if you get a chance could you attach your about:support? I will make sure the configuration is included in the fix for bug 1377866.", "id": 12445393, "creation_time": "2017-07-04T05:00:38Z", "author": "dvander@alliedmods.net"}], "attachments": []}
```
marco-c: Let's only add the `whiteboard` field (here and in the following), so the test is more readable

ayush1999:
Done. Should tests be fixed for other cleanup functions too? Probably in a fresh PR? (Can be a good reserved-for-beginners issue)

marco-c:
> Should tests be fixed for other cleanup functions too? Probably in a fresh PR?

Yes, could you file an issue and explain what needs to be done?

marco-c:
@ayush1999 does this make any difference in the `bug` or the `tracking` models?

ayush1999:
There isn't a noticeable difference in the results. For example: in case of `bug`, the CV accuracy goes down slightly from 0.9427 to 0.9327.

> 
> Yes, could you file an issue and explain what needs to be done?

I've opened an issue: #126 .

PR #121, opened 2019-01-25
bug_features.py: Add 'priority' feature

sbeesm:


sbeesm:
In cases where priority is `'--'`, should that be handled by returning `None`?

marco-c:
Could you provide results before and after adding this to the `bug` and `tracking` models?

marco-c:
> In cases where priority is '--', should that be handled by returning None?

Let's leave '--' for now.

sbeesm:
> Could you provide results before and after adding this to the `bug` and `tracking` models?

For some reason I am getting the same CV Accuracy before and after adding the priority feature to `bug` and `tracking` models.

I've added `bug_features.priority()` in `feature_extractors` list in both model files. Do some other files also need to be changed to get the right results?

marco-c:
Could you post the whole result (with the confidence thresholds included) before/after for both models?

marco-c:
Good, let's add the feature to the tracking model then!

sbeesm:
Do I need to create a new PR for changing the tracking model or I can make the changes in this one only?

marco-c:
It's better in this one

sbeesm:
I've made the changes. Waiting for review!

PR #128, opened 2019-01-29
Add number of touched files to commits

### LINKED ISSUE #116: Add number of touched files as a feature

ayush1999:
@marco-c Small question: Once I've made a change in repository.py, how do I update commits.json based on these changes?

Fixes #116.


marco-c:
You will have to clone https://hg.mozilla.org/mozilla-central/ locally, then you can run `repository.py` giving it the path to your local clone.

bugbug/repository.py
```diff
@@ -47,6 +47,7 @@ def _transform(commit):
         'bug_id': bug_id,
         'added': 0,
         'deleted': 0,
+        'touched':0,
```
marco-c: Call this one `files_modified_num`

bugbug/repository.py
```diff
@@ -72,6 +73,7 @@ def _transform(commit):
         else:
             type_ = ext
         obj['types'].add(type_)
+        obj['touched'] = len(stats['touched'])
```
marco-c: `touched` is not the number of files modified, it is the number of lines modified.

marco-c:
Can you try passing the absolute path to the repository?

ayush1999:
I still get the same error.

marco-c:
Can you `cd` into your cloned repository directory and type `hg log` and see if it works?

ayush1999:
That's probably the reason. It works now, though I get a different error: `TypeError: __init__() got an unexpected keyword argument 'initializer'` is this because of some version issues?

marco-c:
`initializer` was introduced in Python 3.7, so you'll need to use at least that version.

marco-c:
Yep!

marco-c:
(It takes a few hours, beware!)

marco-c:
You'll also need to add a function in bug_extractor.py to get this feature.

ayush1999:
> You'll also need to add a function in bug_extractor.py to get this feature.

You mean bug_features.py?

marco-c:
> You mean bug_features.py?

Yes :)

ayush1999:
@marco-c I just noticed that there's keras in the requirements, and this project uses python3.7, although there is no (official) support for tensorflow on python3.7 yet. Would this be a problem when a deep learning model might be tried out in the future?

marco-c:
We can use tensorflow nightly which supports Python 3.7.

PR #130, opened 2019-01-29
Added bug alias as feature

### LINKED ISSUE #127: Add bug alias as a feature
Not the alias by itself, but something like True if `CVE` is in bug['alias'], False otherwise.

Aftaab99:
Fix #127


bugbug/bug_features.py
```diff
@@ -183,6 +183,14 @@ def __call__(self, bug):
         return bug['priority']
 
 
+class bug_has_alias(object):
+    def __call__(self, bug):
+        if bug['alias']:
```
marco-c: All this can be simplified to `return 'CVE' in bug['alias']`.

bugbug/bug_features.py
```diff
@@ -183,6 +183,14 @@ def __call__(self, bug):
         return bug['priority']
 
 
+class bug_has_alias(object):
```
marco-c: Please rename this to `bug_has_cve_in_alias`.

marco-c:
Could you check if this makes any difference in the `bug` model?

Aftaab99:
There was no change in the cross-validation scores in the bug model upon adding this feature.

bugbug/models/bug.py
```diff
@@ -33,6 +33,7 @@ def __init__(self, lemmatization=False):
             bug_features.patches(),
             bug_features.landings(),
             bug_features.title(),
+            bug_features.bug_has_cve_in_alias(),
```
marco-c: Don't add it here if it didn't make any difference

bugbug/bug_features.py
marco-c: You should also add some tests for this

marco-c:
Looks good, but there is still one review comment open.

Aftaab99:
> Looks good, but there is still one review comment open.

Are you referring to the one where you told me to to replace the function body with `return 'CVE' in bug['alias']`? It gave me a error as alias is `None` for some blocks hence I added that `if` as a check before that statement.
Or was it any other review comment?

marco-c:
> Are you referring to the one where you told me to to replace the function body with return 'CVE' in bug['alias']? It gave me a error as alias is None for some blocks hence I added that if as a check before that statement.

Yes, that was it. If bug['alias'] can be None, then you can do:
`return bug['alias'] is not None and 'CVE' in bug['alias']`

PR #131, opened 2019-01-29
Add number and length of comments as a feature

### LINKED ISSUE #107: Add number and length of comments as a feature

puneet29:
Added number of comments and the sum of comments for each bug to the features.

Fixes #107


bugbug/bug_features.py
```diff
@@ -182,6 +182,21 @@ class priority(object):
...
         return bug['priority']
 
+class comment_count(object):
+    def __call__(self, bug):
+        return field(bug, 'comment_count')
+
+
+class comment_length(object):
+    def __call__(self, bug):
+        comments = bug['comments']
```
marco-c: You could implement this more nicely with a [sum](https://docs.python.org/3/library/functions.html#sum).

bugbug/bug_features.py
```diff
@@ -183,6 +183,16 @@ def __call__(self, bug):
...
 
 
+class comment_count(object):
+    def __call__(self, bug):
+        return field(bug, 'comment_count')
+
+
+class comment_length(object):
+    def __call__(self, bug):
+        return sum([len(x['text']) for x in bug['comments']])
```
marco-c: Nit: The argument to `sum` doesn't need to be a list, it can be an iterator.
If you use a list, you'll needlessly spend time creating it. You can do `sum(len(c['text']) for c in bug['comments'])` instead.

puneet29: Got it! @marco-c 
Thank you for your guidance. Appreciate it.

Issue #134, opened 2019-01-30
Print progress bar while downloading bugs


marco-c:
Similar to what we're doing for commits data: 28b83c12c29185c52afb58d94a533a9448969a8a.

EternalLearner42:
Sounds interesting. I am a bit of a newbie, if you could give me some advice on how to do this, I'd be happy to try 👍

marco-c:
You have an example in the first comment :)
You can just do something similar in the function to download bugs in bugbug/bugzilla.py.

akashgosai:
@marco-c are these changes enough?

marco-c:
@akashgosai sorry, the issue is reserved for new contributors (people who haven't done any commit to the project yet).

akashgosai:
Okay.

PR #139, opened 2019-01-30
in response to #134

akashgosai:
Added the tqdm function for printing progress bar


marco-c:
Sorry, that issue is reserved for people who haven't contributed yet.

Issue #137, opened 2019-01-30
Stop parsing bug ID manually and use a hg.log template


marco-c:
We can use the {bugs} template (https://github.com/mozilla/version-control-tools/blob/d5db0980ab004b3aada29afdcf798a8a95520240/hgext/mozext/__init__.py#L137) at https://github.com/mozilla/bugbug/blob/5da534ecc95b375de2e5b4e8262253115725bce1/bugbug/repository.py#L86 and stop parsing the bug ID manually at https://github.com/mozilla/bugbug/blob/5da534ecc95b375de2e5b4e8262253115725bce1/bugbug/repository.py#L35.

akashgosai:
I am a newbie here, can you suggest a bit about how are we supposed to parse manually?

marco-c:
We *are* parsing manually, and we should stop doing it. The first step would be to clone mozilla-central and then read docs about hg log (the template parameter) and try to use it from the cmd line.

akashgosai:
do I have to clone mozilla-gecko-dev ?

marco-c:
You'll have to clone mozilla-central, which is a Mercurial repository, see https://github.com/mozilla/bugbug#running-the-repository-mining-script.

ayir:
@marco-c I managed to get the bug ids from the cmd line using "hg log --template '{bug}\n'" command but I am stuck as how to put the template argument in the hglog python command?

marco-c:
@ayir @ayush1999 is working on that in #148, once that is landed you'll be able to use the function he wrote.

ayir:
Ohh alright thanks !

PR #146, opened 2019-02-01
Mozilla/#137

ayir:


marco-c:
@ayir read https://github.com/mozilla/bugbug#running-the-repository-mining-script for a guide on how to test your change.

What we need to do to fix #137 is not to filter the commits (which you are doing with this PR), but to use a template argument to avoid manually parsing the bug ID.

Issue #105, opened 2019-01-23
Add `author experience` as feature


marco-c:

ayush1999:
@marco-c From where should this data be extracted? (Any specific field that should be used?)

marco-c:
The data should be extracted from commits, using a template when calling hg.log.

marco-c:
I think the first step would be to try to add the `template` parameter to the `hg.log` call and adjust the rest of the code.

E.g. template='{author}|{node}\n'
will give you something like:
```
Masayuki Nakano <masayuki@d-toybox.com>|c650507f7feee04516c2c869a2f0662c2a6dc0c0
Kartikaya Gupta <kgupta@mozilla.com>|47b5e6d95d9eaa421b6aaaccd8f263f32bbc8121
Brian Grinstead <bgrinstead@mozilla.com>|5ea2f61028a5e989f1dbf46e9fe961a7b1c419ca
```

ayush1999:
I just dug a bit into the hglib code, looks like using `hg.log()` returns commit in `'{rev}\\0{node}\\0{tags}\\0{branch}\\0{author}\\0{desc}\\0{date}\\0'` format. Can't we use this directly? `commit.author` and `commit.node` attributes can be used to extract the corresponding values.

marco-c:
We can use it for the author, we can't use it for the reviewer.

ayush1999:
@marco-c I'm trying to do something like:
```
>>> import hglib
>>> hg = hglib.open("mozilla-central")
```

`hg` in this case doesn't have a kwarg called template, according to the docs. What am I doing wrong? (Is there some problem in installing the extension?)

marco-c:
Ahhh, that's a pity. I thought there was a `template` parameter for log.

OK, so it looks like you'll have to write a custom log function, you can take inspiration from https://www.mercurial-scm.org/repo/python-hglib/file/tip/hglib/client.py#l1095.

marco-c:
(you don't need to support all the options, just template)

ayush1999:
The default template is defined in `templates.changeset`, equal to `'{rev}\\0{node}\\0{tags}\\0{branch}\\0{author}\\0{desc}\\0{date}\\0'`, so if I try to change that, I get an empty list in the result, so maybe the template `{author}|{node}\n` isn't right?
```
>>> import hglib
>>> from hglib import templates
>>> templates.changeset = '{author}|{node}\n'
>>> hg = hglib.open('mozilla-central')
[...]

PR #148, opened 2019-02-01
[WIP] Author experience and reviewer experience as features

ayush1999:
Created a new class, derived from `hgclient`, and the log function now takes the `template` parameter. 
Initial fix for #105


marco-c:
Looks good so far. A suggestion: why don't you make a first PR just moving from normal hg.log to the templated hg.log, and then a second PR to actually fix #105?

ayush1999:
Sure, this can be treated as the PR moving to the new class, I reckon?

marco-c:
Yes!

ayush1999:
Ok, I've just added the log function with no arguments for now, so the results are same as before. Changes to this can be made in subsequent PRs.

bugbug/repository.py
```diff
@@ -20,6 +23,16 @@
...
 
 
+class hgclient_template(hgclient):
+    def __init__(self, path):
+        super(hgclient_template, self).__init__(path, None, None)
+
+    def log(self, template=templates.changeset):
+        args = cmdbuilder(b'log', template=template)
+        out = self.rawcommand(args)
+        return out
```
marco-c: I don't think that this will work without parsing the output

bugbug/repository.py
```diff
@@ -20,6 +23,16 @@
 BUG_PATTERN = re.compile('[\t ]*[Bb][Uu][Gg][\t ]*([0-9]+)')
 
 
+class hgclient_template(hgclient):
```
marco-c: Instead of subclassing, we can just create a hg_log function, it's more maintainable as we don't need to worry about the parent at all (e.g. even if upstream changes the constructor, we won't have to update anything if we don't subclass).

ayush1999:
Sure, The function now takes two parameters: `repo_dir` and the `template`.

marco-c:
> Ok, I've just added the log function with no arguments for now, so the results are same as before. Changes to this can be made in subsequent PRs.

bugbug/repository.py
```diff
@@ -20,6 +22,14 @@
 BUG_PATTERN = re.compile('[\t ]*[Bb][Uu][Gg][\t ]*([0-9]+)')
 
 
+def hg_log(repo_dir, template=templates.changeset):
+    hg = hglib.open(repo_dir)
+    args = cmdbuilder(b'log', template=template)
```
marco-c: Hardcode the template here, removing the `template` parameter from hg_log (the parameter is not really supported anyway, given that we are parsing the result assuming the template is always the same).

bugbug/repository.py
```diff
@@ -20,6 +22,14 @@
...
+    x = hg._parserevs(hg.rawcommand(args))
```
marco-c: Implement your own parsing here, as our custom template doesn't have the same values as the default template.

ayush1999: > Implement your own parsing here, as our custom template doesn't have the same values as the default template.

Ah, sorry, I forgot about this.

marco-c:
@ayush1999 this does not work. We are not using author and node only, we are using more information (e.g. the commit message).
Please test the change (with a limit in the log command) before the next update of the PR.

bugbug/repository.py
```diff
@@ -20,6 +25,20 @@
 BUG_PATTERN = re.compile('[\t ]*[Bb][Uu][Gg][\t ]*([0-9]+)')
 
 
+def hg_log(repo_dir):
+    hg = hglib.open(repo_dir)
+    args = cmdbuilder(b'log', template='{rev}\\0{node}\\0{tags}\\0{branch}\\0{author}\\0{desc}\\0{date}\\0')
```
marco-c: Nit: use `hglib.util.cmdbuilder`, so it's clear where it is coming from

bugbug/repository.py
```diff
@@ -20,6 +25,20 @@
...
+    x = hg.rawcommand(args)
+    out = x.split(b('\0'))[:-1]
```
marco-c: No need to use the `b` function, you can just hardcode the result of `b('\0')` (which is `b'\x00'`).

bugbug/repository.py
```diff
@@ -20,6 +25,20 @@
...
+    revs = []
+    for rev in util.grouper(7, out):
+        posixtime = float(rev[6].split(b('.'), 1)[0])
```
marco-c: Same here, you don't need to use the `b` function.

bugbug/repository.py
```diff
@@ -20,6 +25,20 @@
...
+        dt = datetime.fromtimestamp(posixtime)
+        revs.append(revision(rev[0], rev[1], rev[2], rev[3], rev[4], rev[5], dt))
```
marco-c: Just use a normal tuple, why returning a `revision` object?

marco-c: When you remove `revision`, you should also be able to remove the `commits = (tuple(commit) for commit in commits)` line.

bugbug/repository.py
```diff
@@ -20,6 +25,20 @@
 BUG_PATTERN = re.compile('[\t ]*[Bb][Uu][Gg][\t ]*([0-9]+)')
 
 
+def hg_log(repo_dir):
+    hg = hglib.open(repo_dir)
+    args = cmdbuilder(b'log', template='{rev}\\0{node}\\0{tags}\\0{branch}\\0{author}\\0{desc}\\0{date}\\0')
+    x = hg.rawcommand(args)
+    out = x.split(b('\0'))[:-1]
+    revs = []
+    for rev in util.grouper(7, out):
```
marco-c: Use `hglib.util`

bugbug/repository.py
```diff
@@ -84,13 +99,9 @@ def _transform(commit):
...
 def download_commits(repo_dir):
-    hg = hglib.open(repo_dir)
-
-    commits = hg.log()
+    commits = hg_log(repo_dir)
     commits_num = len(commits)
 
-    hg.close()
-
     commits = (tuple(commit) for commit in commits)
```
marco-c: This line should be no longer needed now.

ayush1999:
Ah right, missed that. Fixed now.

PR #150, opened 2019-02-01
Create constant sets directly instead of converting lists into sets

### LINKED ISSUE #82: Create constant sets directly instead of converting lists into sets

Delkhaz:
Fixes #82

### DIFF
--- a/bugbug/bugzilla.py
+++ b/bugbug/bugzilla.py
@@ -87,7 +87,7 @@ def _download(ids_or_query):
 
 
 def download_bugs_between(date_from, date_to, security=False):
-    products = set([
+    products = {
         'Add-on SDK',
         'Android Background Services',
         'Core',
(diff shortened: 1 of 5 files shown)


bugbug/bugzilla.py
```diff
@@ -126,7 +126,7 @@ def download_bugs_between(date_from, date_to, security=False):
 def download_bugs(bug_ids, products=None, security=False):
     old_bug_count = 0
     old_bugs = []
-    new_bug_ids = set(int(bug_id) for bug_id in bug_ids)
+    new_bug_ids = {int(bug_id) for bug_id in bug_ids}
```
marco-c: This change is not needed, we were not converting a list into a set here.

bugbug/models/component.py
```diff
@@ -142,7 +142,7 @@ def get_labels(self):
             classes[bug_id] = bug_data['product']
 
         component_counts = Counter(classes.values()).most_common()
-        top_components = set(component for component, count in component_counts)
+        top_components = {component for component, count in component_counts}
```
marco-c: Same here.

comment_level_labeler.py
```diff
@@ -28,7 +28,7 @@
     next(reader)
     labeled_comments = [(int(r[0]), int(r[1]), r[2]) for r in reader]
 
-already_done = set((c[0], c[1]) for c in labeled_comments)
+already_done = {(c[0], c[1]) for c in labeled_comments}
```
marco-c: Same here.

PR #154, opened 2019-02-05
Add total author experience

ayush1999:
Initial fix for #105 .

The `_transform` function now updates the number of commits per author on each iteration, and `_add_experience` adds this information to the data.


ayush1999:
Is this correct way to be doing this? Would love to get this reviewed.

marco-c:
This won't work for a few reasons:
1) Each process will have its own copy of the number_of_commits map, so no process will have the correct data.
2) You're using information from the future. If you have commit A by author A, commit B by author B and commit C by author A, commit A should have author_experience = 0, commit C should have author_experience = 1. With your current implementation, they both have the same author_experience = 1.

ayush1999:
Oh, right. I confused total commits with commits made throughout the project (past + future). I've fixed that now.

marco-c:
The PR is still affected by problem 1.

Also, we can't be sure about the order in which we are analyzing commits given that we are using multiple processes.

The best solution would probably be to populate this map ahead of time (before creating the process pool), and then use it in _transform.

marco-c:
> The map you're referring to, what would it consist of? We can't have a dict from author names to number of commits, since a single author name will correspond to various values. Various commits also don't have `bug_id`.

For example it can be a map commit -> number of commits by the same author before commit.

ayush1999:
How about adding something like this:
```
def download_commits(repo_dir):
    commits = hg_log(repo_dir)
    commits_num = len(commits)

    global number_of_commits
    for i in range(commits_num):
        number_of_commits[_transform(commits[i])] = sum(commit['author'] == commits[i]['author'] for commit in commits[:i])

    with concurrent.futures.ProcessPoolExecutor(initializer=_init, initargs=(repo_dir,)) as executor:
        commits = executor.map(_transform, commits, chunksize=256)
        commits = executor.map(_add_experience, commits, chunksize=256)
        commits = tqdm(commits, total=commits_num)
        db.write(COMMITS_DB, commits)


def _add_experience(commit):
    commit['num_of_commits'] = number_of_commits[commit]
    return commit
```

(Getting the value before the ProcessPool is called, and adding them via `_add_experience`)

ayush1999:
That can be done, but dict can't have dict as keys. (Since commit is also a dict). We could convert everything into `frozenset`, but then we'll have to recursively convert all corresponding values in the dict to frozensets, because a few fields are lists (which is again unhashable).

marco-c:
> That can be done, but dict can't have dict as keys

You don't have to use the full dict, you can just use the commit hash as a key.

marco-c:
> You don't have to use the full dict, you can just use the commit hash as a key.

Ah, and I think `commit` is a tuple, not a dict, so you should be able to use it.

bugbug/repository.py
```diff
@@ -114,6 +117,10 @@ def download_commits(repo_dir):
     commits = hg_log(repo_dir)
     commits_num = len(commits)
 
+    global number_of_commits
+    for i in range(commits_num):
+        number_of_commits[commits[i]] = sum(commit.author == commits[i].author for commit in commits[:i])
```
marco-c: With this implementation, you're iterating through all commits multiple times.
It's better to iterate only once, having a temporary map where you store the number of commits by the author so far.

ayush1999:
> > You don't have to use the full dict, you can just use the commit hash as a key.
> 
> Ah, and I think `commit` is a tuple, not a dict, so you should be able to use it.

Yes, I was initially adding the experience data after `_transform` call, so everything was a dict. Now I'm taking the commits as namedtuples, which solves this. The changes in latest commit should solve the above mentioned issues.

bugbug/repository.py
```diff
@@ -24,6 +24,8 @@
 
 Commit = namedtuple('Commit', ['node', 'author', 'desc', 'date'])
 
+number_of_commits = {}
```
marco-c: The name of the map is really too generic. Let's call it author experience.

bugbug/repository.py
```diff
@@ -51,6 +53,7 @@ def _transform(commit):
         'deleted': 0,
         'files_modified_num': 0,
         'types': set(),
+        'num_previous_commits': number_of_commits[commit]
```
marco-c: Same here, `author_experience`

bugbug/repository.py
```diff
@@ -114,6 +117,17 @@ def download_commits(repo_dir):
     commits = hg_log(repo_dir)
     commits_num = len(commits)
 
+    temp_commits = {}
```
marco-c: `commits_by_author`

bugbug/repository.py
```diff
@@ -114,6 +117,17 @@ def download_commits(repo_dir):
...
+
+    global number_of_commits
+    for i in range(commits_num):
```
marco-c: Why not iterating through the commits directly? You're never really using `i`

ayush1999:
Done.

marco-c:
Looks good now! You'll also need to add the feature to bug_features.py.

bugbug/bug_features.py
```diff
@@ -224,6 +224,11 @@ def __call__(self, bug, **kwargs):
         return sum(commit['files_modified_num'] for commit in bug['commits'])
 
 
+class commit_author_experience(object):
+    def __call__(self, bug, **kwargs):
+        return sum(commit['author_experience'] for commit in bug['commits'])
```
marco-c: Actually, here we can't sum. If you have two patches from the same author, you don't want the sum of the same experience.
We probably want the average here.

ayush1999: Right, fixed that.

PR #161, opened 2019-02-08
Author experience for previous 3 months

ayush1999:
Fixes : #105 .

For every commit, a loop iterates through all commits made before it, to check for the same author, and if the same author is encountered, counter is incremented based on the 3 month constraint. (If duration is greater that 3 months, it breaks from the loop).

### DIFF
--- a/bugbug/repository.py
+++ b/bugbug/repository.py
@@ -120,24 +120,23 @@ def download_commits(repo_dir):
 
     for ind, commit in enumerate(commits):
         if commit.author not in commits_by_author:
-            commits_by_author[commit.author] = 0
+            commits_by_author[commit.author] = [commit]
         else:
-            commits_by_author[commit.author] += 1
+            commits_by_author[commit.author].append(commit)
 
-        author_experience[commit] = commits_by_author[commit.author]
+        author_experience[commit] = len(commits_by_author[commit.author]) - 1
+        
+    for commit in commits:
+        cut = 0
+        for prev_commit in commits_by_author[commit.author]:
+            if (commit.date - prev_commit.date).days <= 90 :
+                cut += 1
+            else:
+                break
+        
+        commits_by_author[commit.author] = commits_by_author[commit.author][1:]
 
-        commits_by_author_90_days[commit.author] = 0
-        for prev_commit in commits[ind + 1:]:
-            if commit.author == prev_commit.author:
-                res = commit.date - prev_commit.date
-                if hasattr(res, 'months') and res.months <= 3:
-                    commits_by_author_90_days[commit.author] += 1
-                elif hasattr(res, 'days') and res.days <= 90:
-                    commits_by_author_90_days[commit.author] += 1
-                else:
-                    break
-
-        author_experience_90_days[commit] = commits_by_author_90_days[commit.author]
+        author_experience_90_days[commit] = cut
 
     print(f'Mining commits using {multiprocessing.cpu_count()} processes...')
 
--- a/bugbug/bug_snapshot.py
+++ b/bugbug/bug_snapshot.py
@@ -287,7 +287,7 @@ def rollback(bug, when, verbose=True):
                 bug[field] = old_value
 
     # If the first comment is hidden.
-    if dateutil.parser.parse(bug['comments'][0]['creation_time']) > rollback_date:
+    if bug['comments'][0]['count'] != 0:
         bug['comments'].insert(0, {
             'id': 0,
             'text': '',


bugbug/repository.py
```diff
@@ -53,7 +54,8 @@ def _transform(commit):
         'deleted': 0,
         'files_modified_num': 0,
         'types': set(),
-        'author_experience': author_experience[commit]
+        'author_experience': author_experience[commit],
+        'author_experience_3_months': author_experience_3_months[commit]
```
marco-c: For simplicity, let's use 90_days instead of 3_months

bugbug/repository.py
```diff
@@ -118,8 +120,11 @@ def download_commits(repo_dir):
     commits_num = len(commits)
 
     commits_by_author = {}
+    commits_by_author_3_months = {}
 
     global author_experience
+    global author_experience_3_months
+
     for commit in commits:
```
marco-c: This loop can be integrated in the other one, so you can avoid looping twice.

bugbug/repository.py
```diff
@@ -128,6 +133,19 @@ def download_commits(repo_dir):
 
         author_experience[commit] = commits_by_author[commit.author]
 
+    for i in range(commits_num):
```
marco-c: Nicer to do something like `for i, commit in enumerate(commits):`

bugbug/repository.py
```diff
@@ -128,6 +133,19 @@ def download_commits(repo_dir):
...
+        commits_by_author_3_months[commits[i].author] = 0
+        for j in range(i + 1, commits_num):
```
marco-c: And here something like `for prev_commit in commits[i:]`

bugbug/repository.py
```diff
@@ -128,6 +133,19 @@ def download_commits(repo_dir):
...
+            if commits[i].author == commits[j].author:
+                res = commits[i].date - commits[j].date
+                if hasattr(res, 'months') and res.months <= 3:
```
marco-c: Just use 90 days, easier to handle

ayush1999: The difference between two datetime object have either > 90 days or > 3 months. I've added both cases.

bugbug/repository.py
```diff
@@ -53,7 +54,8 @@ def _transform(commit):
         'deleted': 0,
         'files_modified_num': 0,
         'types': set(),
-        'author_experience': author_experience[commit]
+        'author_experience': author_experience[commit],
+        'author_experience_90_days': author_experience_90_days[commit]
```
marco-c: Put a comma at the end, otherwise the next time we update this we'll have to modify two lines.

bugbug/repository.py
marco-c: Mmmmh, the timedelta object should have a `days` property, not a `months`.

marco-c:
Given that the commits are sorted by latest to first, it seems like `author_experience` is wrong.

bugbug/repository.py
```diff
@@ -118,16 +120,32 @@ def download_commits(repo_dir):
     commits_num = len(commits)
 
     commits_by_author = {}
+    commits_by_author_90_days = {}
 
     global author_experience
-    for commit in commits:
+    global author_experience_90_days
+
+    for ind, commit in enumerate(commits):
```
marco-c: The double loop is going to be really expensive, please check how long it takes to run this (you can return before the processpoolexecutor).

ayush1999: I'll check it out. Is there a way to avoid nested loops here? For every commit, we'd have to iterate up to all commits after it to check the time gap, right?

ayush1999:
I just checked the time: It takes about 100 minutes for the loops to run, before the processpoolexecutor starts.

marco-c:
That is a lot of time, we should try to think of an alternative solution which is cheaper to run.

ayush1999:
We need to do this only once though, right? Once we have the updated dataset, we can just upload it to dropbox?

marco-c:
> We need to do this only once though, right? Once we have the updated dataset, we can just upload it to dropbox?

Yep, but still every time we make changes to this script, and periodically to update the database with new commits from mozilla-central.

ayush1999:
Oh. I can't see a way to avoid nested loops though, do you?

ayush1999:
Should we keep this open for now, until we figure out a way to optimize it?

marco-c:
Here's a possible implementation.
Keep a map from author -> list of previous 90 days commits from the author.

For each commit:
1) You get the list of previous commits by the author, and remove any previous commit which is older than commit date - 90 days
2) The 90 days experience is the len of the list of previous commits
3) You add the commit to the list of previous commits

If you keep the list sorted (it's automatically sorted because you append to the list), then the step 2 will be fast (you only have to remove the first commits from the list, until the first which is new enough).

ayush1999:
@marco-c A few observations:

1. Step 1 is similar to what we are currently doing; Iterating for every prev_commit before our commit, and breaking out of the loop whenever time > 90 days, since we are sure that all commits after that will also have time > 90 days (since commits are in chronological order).

2. Step 2 : We're currently incrementing our counter by 1 for every commit within the 90 days time frame, which is similar to appending in a list and returning the length at the end.

If I'm not wrong, this implementation will also have O(n^2) complexity, right?

marco-c:
No, in that implementation you only have one loop on all commits, because you store the previous commits (limited to 90 days) from the same author in a list, so you don't have to reloop through all previous commits.

ayush1999:
> No, in that implementation you only have one loop on all commits, because you store the previous commits (limited to 90 days) from the same author in a list, so you don't have to reloop through all previous commits.

When we store the previous 90 days commits from the same author in a map, that'd need two loops, if I'm not wrong.

```
author_commit_map = {}

for ind, commit in enumerate(commits):
        if commit.author not in author_commit_map:
            author_commit_map[commit.author] = [commit]
        else:
            author_commit_map[commit.author].append(commit)

# author_commit_map now has authors -> all previous commits by them.
```
If we need to get the number of previous 90 days commits per author, per commit, we'd now have to iterate the entire `author_commit_map` and return length of the commits made for each commit in a single key of the map.

marco-c:
No, the solution I described earlier is a little different. Let me describe it again with some approximate pseudocode to make it clearer:
```
for commit in commits:
    cut = None
    for i, prev_commit in enumerate(commits_by_author[commit.author]):
        if commit.date - prev_commit.date <= 90 days:
            break
        cut = i
    if cut is not None:
        commits_by_author[commit.author] = commits_by_author[commit.author][cut:]

    author_experience_90_days[commit] = len(commits_by_author[commit.author])

    commits_by_author[commit.author].append(commit)
```

ayush1999:
I'm away from my laptop right now, but I'll try it out later today. Although, isn't this a bit similar to the current implementation? There are again 2 nested loops here, the only main difference being that we now have separate commit maps for each author, whereas in the current implementation we have an additional check for the author of the commit.

marco-c:
> I'm away from my laptop right now, but I'll try it out later today. Although, isn't this a bit similar to the current implementation? There are again 2 nested loops here, the only main difference being that we now have separate commit maps for each author, whereas in the current implementation we have an additional check for the author of the commit.

There is a big difference between an inner loop going through all of the previous commits (earlier implementation) and an inner loop going through only a few previous commits (not all, because we keep the list bounded by removing old commits) from the same author.

ayush1999:
Right, I'll fix it. Also, for the `author_experience` (total experience) field, the value for the latest commit should be more than the value for an earlier commit by the same author, right? (Because the current implementation that was merged has it the other way).

marco-c:
You haven't really implemented my suggestion :smile: 
You are iterating through all previous commits from the same author.

ayush1999:
@marco-c Is this implementation close? I have two separate maps now, and I use them for iterating from a `count` index (hence not iterating through all commits). (Sorry for the delay, had an assignment to complete.)

marco-c:
@ayush1999 why are you not implementing the solution I proposed? The pseudocode from https://github.com/mozilla/bugbug/pull/161#issuecomment-462677908 is almost complete. If it's not clear for you, or if you noticed something wrong, feel free to tell me.

ayush1999:
The results from the pseudocode aren't correct. For example, suppose there are 5 commits by athe same author, named A1, A2, A3, A4, A5 (all within 90 days of each other). The pseudocode returns author_experience_90 _days as 5,6,7,8,9 respectively. However, the actual result should be 4,3,2,1,0. To overcome this, we'd need to have a map which stores the index upto which we are certain the commits are within the time frame, and increment it according to the condition, which is what I've currently implemented.

marco-c:
> The results from the pseudocode aren't correct. For example, suppose there are 5 commits by athe same author, named A1, A2, A3, A4, A5 (all within 90 days of each other). The pseudocode returns author_experience_90 _days as 5,6,7,8,9 respectively. However, the actual result should be 4,3,2,1,0.

Why?
The pseudocode adds in author_experience_90_days the length of the previous commits in a 90 days range.
When only A1 is in the list, author_experience_90_days[A1] will be 1. Why do you think it would be 5?
When A1 and A2 are in the list, author_experience_90_days[A2] will be 2.
And so on.

ayush1999:
> > The results from the pseudocode aren't correct. For example, suppose there are 5 commits by athe same author, named A1, A2, A3, A4, A5 (all within 90 days of each other). The pseudocode returns author_experience_90 _days as 5,6,7,8,9 respectively. However, the actual result should be 4,3,2,1,0.
> 
> Why?
> The pseudocode adds in author_experience_90_days the length of the previous commits in a 90 days range.
> When only A1 is in the list, author_experience_90_days[A1] will be 1. Why do you think it would be 5?
> When A1 and A2 are in the list, author_experience_90_days[A2] will be 2.
> And so on.

I just tested it out, and I get the same result (5,6,7,8,9). I think it is probably because in this case `cut` is always `None`, and  we're appending the commit to the list, thus increasing the size on each iteration.

ayush1999:
Oh, I've updated the pseudocode now, with a few minor changes.

marco-c:
You can avoid having two loops by having two maps. One containing the total count and one containing the previous commits in the 90 days range.

bugbug/repository.py
```diff
@@ -122,13 +124,28 @@ def download_commits(repo_dir):
...
+    for commit in commits:
+        cut = None
+        for i, prev_commit in enumerate(commits_by_author[commit.author]):
+            if (commit.date - prev_commit.date).days <= 90:
+                break
+            cut = i
+        if cut is not None:
+            commits_by_author[commit.author] = commits_by_author[commit.author][cut:]
+
+        author_experience_90_days[commit] = len(commits_by_author[commit.author]) - 1
```
marco-c: Why the -1? Maybe you need [cut + 1:] in the previous line instead?

ayush1999:
Also, @marco-c , the pseudocode won't work for a general case, when the have commits >= 90 days, since the loop will start considering all commits, even those out side the timeframe. For example: if there are 5 commits by the same author = A = [A1, A2, A3, A4, A5], out of which only A2 and A3 are within 90 days of A1. Now during the first iteration, `cut = None`, it'll break out of the loop and author_exp_90_days[A1] will become len(A) - 1 = 4, which is wrong.

marco-c:
> Also, @marco-c , the pseudocode won't work for a general case, when the have commits >= 90 days, since the loop will start considering all commits, even those out side the timeframe. For example: if there are 5 commits by the same author = A = [A1, A2, A3, A4, A5], out of which only A2 and A3 are within 90 days of A1. Now during the first iteration, `cut = None`, it'll break out of the loop and author_exp_90_days[A1] will become len(A) - 1 = 4, which is wrong. (result should be 2).

I see you still haven't a clear picture about the algorithm I proposed. The best way to understand it is to try to run it with pen and paper with an example.

marco-c:
I'll just implement it myself, it's going to be easier for this one.

PR #165, opened 2019-02-10
Number of backouts as feature

### LINKED ISSUE #110: Add number of backouts as a feature

ayir:
Fixes #110.

### DIFF
--- a/bugbug/bug_features.py
+++ b/bugbug/bug_features.py
@@ -232,7 +232,7 @@ class commit_author_experience(object):
 
 class commit_no_of_backouts(object):
     def __call__(self, bug, **kwargs):
-        return len([commit for commit in bug['commits'] if commit['ever_backedout'] is True])
+        return sum(1 for commit in bug['commits'] if commit['ever_backedout'] is True)
 
 
 def cleanup_url(text):
--- a/bugbug/bug_features.py
+++ b/bugbug/bug_features.py
@@ -230,6 +230,11 @@ class commit_author_experience(object):
         return sum(res) / len(res)
 
 
+class commit_no_of_backouts(object):
+    def __call__(self, bug, **kwargs):
+        return len([commit for commit in bug['commits'] if commit['ever_backedout'] is True])
+
+
 def cleanup_url(text):
     text = re.sub(r'http[s]?://(hg.mozilla|searchfox|dxr.mozilla)\S+', '__CODE_REFERENCE_URL__', text)
     return re.sub(r'http\S+', '__URL__', text)
--- a/bugbug/bug_features.py
+++ b/bugbug/bug_features.py
@@ -232,7 +232,7 @@ class commit_author_experience(object):
 
 class commit_no_of_backouts(object):
     def __call__(self, bug, **kwargs):
-        return sum(1 for commit in bug['commits'] if commit['ever_backedout'] is True)
+        return sum(1 for commit in bug['commits'] if commit['ever_backedout'])
 
 
 def cleanup_url(text):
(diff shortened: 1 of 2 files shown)


bugbug/repository.py
```diff
@@ -19,7 +19,7 @@
 COMMITS_DB = 'data/commits.json'
 db.register(COMMITS_DB, 'https://www.dropbox.com/s/mz3afgncx0siijc/commits.json.xz?dl=1')
 
-Commit = namedtuple('Commit', ['node', 'author', 'desc', 'date', 'bug'])
+Commit = namedtuple('Commit', ['node', 'author', 'desc', 'date', 'bug', 'no_of_backouts'])
```
marco-c: I think a given commit can be backed out only once, so this should be a boolean `ever_backedout`

marco-c:
You also need to actually add the feature to bug_features.py, where you will sum the backouts for all commits linked to the bug.

bugbug/repository.py
ayir: yes as far as I saw the outputs it was giving only one node.And yes I'll change it.

bugbug/bug_features.py
```diff
@@ -230,6 +230,11 @@ def __call__(self, bug, **kwargs):
         return sum(res) / len(res)
 
 
+class commit_no_of_backouts(object):
+    def __call__(self, bug, **kwargs):
+        return len([commit for commit in bug['commits'] if commit['ever_backedout'] is True])
```
marco-c: You can use `sum` to avoid creating a list: `sum(1 for commit in ...)`

bugbug/repository.py
```diff
@@ -92,13 +93,16 @@ def hg_log(repo_dir):
...
+        backed_out = False
+        if rev[5] != b'':
+            backed_out = True
         revs.append(Commit(
             node=rev[0],
             author=rev[1],
             desc=rev[2],
             date=dt,
             bug=rev[4],
+            ever_backedout=backed_out,
```
marco-c: You can just do `ever_backedout=(rev[5] != b'')` and avoid defining the backed_out variable.

bugbug/repository.py
```diff
@@ -92,13 +93,13 @@ def hg_log(repo_dir):
     for rev in hglib.util.grouper(template.count('\\0'), out):
         posixtime = float(rev[3].split(b'.', 1)[0])
         dt = datetime.fromtimestamp(posixtime)
-
```
marco-c: Don't remove this newline (better not to change things that are unrelated to the goal of the patch).

bugbug/bug_features.py
```diff
@@ -230,6 +230,11 @@ def __call__(self, bug, **kwargs):
         return sum(res) / len(res)
 
 
+class commit_no_of_backouts(object):
+    def __call__(self, bug, **kwargs):
+        return sum(1 for commit in bug['commits'] if commit['ever_backedout'] is True)
```
marco-c: Forgot to tell you before, but here you don't need `is True`

marco-c: `return sum(1 for commit in bug['commits'] if commit['ever_backedout'])`

ayir: Ohh thanks!Just removed it.

bugbug/repository.py
```diff
@@ -91,14 +92,15 @@ def hg_log(repo_dir):
     revs = []
     for rev in hglib.util.grouper(template.count('\\0'), out):
         posixtime = float(rev[3].split(b'.', 1)[0])
+
```
marco-c: You added a newline here instead of readding the one you removed :smile:

ayir: Just a minute. sorry

PR #169, opened 2019-02-11
Add alternative component model using a neural network

johngian:


bugbug/model.py
```diff
@@ -27,6 +27,9 @@ def __init__(self, lemmatization=False):
         self.undersampling_enabled = True
         self.cross_validation_enabled = True
 
+        self.calculate_importance = True
+        self.calculate_classification_metrics = True
```
marco-c: This one is important to compare models, so we should try to make it work

bugbug/model.py
```diff
@@ -52,6 +55,9 @@ def get_important_features(self, cutoff, shap_values):
 
         return top_features
 
+    def calculate_labels(self, classes, bugs):
+        return np.array([classes[bug['id']] for bug in bugs()])
```
marco-c: We could make each model directly return the labels in the format they want in `get_labels`

bugbug/models/triage.py
```diff
@@ -0,0 +1,315 @@
...
+            'short_desc_vocab_size': self.short_desc_vocab_size,
+            'short_desc_emb_sz': self.short_desc_emb_sz,
+            'long_desc_maxlen': self.long_desc_maxlen,
+            'long_desc_vocab_size': self.long_desc_vocab_size,
+            'long_desc_emb_sz': self.long_desc_emb_sz
+        }
+        self.clf = TriageClassifier(**kwargs)
+
+    def get_labels(self):
+        products = {
```
marco-c: We can reuse the get_labels from ComponentModel

bugbug/models/triage.py
```diff
@@ -0,0 +1,315 @@
+# -*- coding: utf-8 -*-
```
marco-c: Given that this is implementing the same as the components model but with a different ML algorithm, maybe it's better to call the file something like `component_nn.py` (nn as in neural networks).

run.py
```diff
@@ -19,13 +19,19 @@
...
+                        help='Goal of the classifier',
+                        choices=['bug', 'regression', 'tracking', 'qaneeded', 'uplift', 'component', 'devdocneeded', 'triage'],
+                        default='bug')
     parser.add_argument('--classify', help='Perform evaluation', action='store_true')
     parser.add_argument('--generate-sheet', help='Perform evaluation on bugs from last week and generate a csv file', action='store_true')
     args = parser.parse_args()
 
     model_file_name = f'{args.goal}model'
 
+    probabilities = True
```
marco-c: Nit: let's call these `probabilities_enabled` and `importances_enabled`, so you don't have to rename the pre-existing `importances` variable

bugbug/model.py
johngian: Can we tackle this in a different PR, since it involved changes to alla the models we currently have.

johngian: Can we tackle this in a different PR since it involves changes in all the rest models as well ?

marco-c: OK!

johngian:
@marco-c PR updated

marco-c:
@johngian I've made a couple refactorings, could you double check this is still working for you?

Issue #51, opened 2019-01-02
Add DB versioning


marco-c:
When a new DB is available, it should automatically be downloaded.

sbeesm:
We also need to add last-modified field to the dropbox link headers, right?

Then maybe we can compare timestamps of the existing DB in the file directory and the one on the server?

marco-c:
We can just add a file containing the version alongside the DB files, that's the easiest solution.

sbeesm:
But this would always require to download the DB (along with the version file) first and then check with the existing version.

marco-c:
> But this would always require to download the DB (along with the version file) first and then check with the existing version.

Not really, it would just require downloading the version file.

Anyway, I've noticed that Dropbox returns an ETag in the headers, so we could just check the ETag before downloading (if it's the same as the previous one, we don't download, if it is different, then we download).

Thinking about it though, I prefer to avoid any kind of connection on every run. So I'd just add a version file for each database in the repository itself. We will have to remember to update them whenever we update the DBs.

sbeesm:
So in `db.download()`, we can add a version file along with the database in the repository and then in `run.py`, we can check the version and download the DB if necessary?

marco-c:
Yes

marco-c:
Basically:
- we'd have a VERSION file for each DB in the repo;
- when downloading a DB, we save the current version in another file locally;
- when we run, we check that the VERSION from the repository is the same as the local version file.

PR #171, opened 2019-02-13
Add DB versioning

sbeesm:
Fixes https://github.com/mozilla/bugbug/issues/51

### DIFF
--- a/DB_VERSION.json
+++ b/DB_VERSION.json
@@ -1,0 +1,1 @@
+{"bugs.json": "v1", "commits.json": "v1"}
--- a/DB_VERSION.json
+++ b/DB_VERSION.json
@@ -1,1 +1,0 @@
-{"bugs.json": "v1", "commits.json": "v1"}
--- a/bugbug/db.py
+++ b/bugbug/db.py
@@ -96,17 +96,13 @@ def is_outdated(path):
 
 
 def update_ver_file(db_path):
-    if not os.path.exists(VER_PATH):
-        ver_dict = {
-            db_path: DATABASES[db_path]['version']
-        }
-        with open(VER_PATH, 'w') as db:
-            json.dump(ver_dict, db)
-        return
-
-    with open(VER_PATH) as db:
-        ver_dict = json.load(db)
+    if os.path.exists(VER_PATH):
+        with open(VER_PATH) as db:
+            ver_dict = json.load(db)
+    else:
+        ver_dict = {}
 
     ver_dict[db_path] = DATABASES[db_path]['version']
+
     with open(VER_PATH, 'w') as db:
         json.dump(ver_dict, db)
--- a/bugbug/db.py
+++ b/bugbug/db.py
@@ -24,28 +24,22 @@ def register(path, url, version):
 # Download and extract databases.
 def download():
     for path, info in DATABASES.items():
-        if os.path.exists(path) and not is_updated_ver(path):
+        if os.path.exists(path) and not is_outdated(path):
             continue
 
         xz_path = f'{path}.xz'
 
-        ver_path = os.path.join(os.path.dirname(path), 'DB_VERSION.json')
+        ver_path = 'data/DB_VERSION.json'
 
         # Only download if the xz file is not there yet.
-        if not os.path.exists(xz_path) or is_updated_ver(path):
+        if not os.path.exists(xz_path) or is_outdated(path):
             urlretrieve(DATABASES[path]['url'], xz_path)
 
         with open(path, 'wb') as output_f:
             with lzma.open(xz_path) as input_f:
                 shutil.copyfileobj(input_f, output_f)
 
-        if not os.path.exists(ver_path):
-            ver_dict = {}
-            with open(ver_path, 'w') as db:
-                ver_dict[path] = info['version']
-                json.dump(ver_dict, db)
-        else:
-            update_ver_file(ver_path, path)
+        update_ver_file(ver_path, path)
 
 
 def read(path):
@@ -92,22 +86,27 @@ def delete(path, match):
     os.rename(f'{path}_new', path)
 
 
-def is_updated_ver(db_path):
-    path = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
-    ver_path = os.path.join(path, 'data', 'DB_VERSION.json')
+def is_outdated(path):
+    ver_path = 'data/DB_VERSION.json'
 
     if not os.path.exists(ver_path):
         return True
     else:
         with open(ver_path) as db:
             ver = json.load(db)
-        if ver.get(db_path):
-            return not DATABASES[db_path]['version'] == ver[db_path]
+        if ver.get(path):
+            return not DATABASES[path]['version'] == ver[path]
         else:
             return True
 
 
 def update_ver_file(ver_path, db_path):
+    if not os.path.exists(ver_path):
+        ver_dict = {}
+        ver_dict[db_path] = DATABASES[db_path]['version']
+        with open(ver_path, 'w') as db:
+            json.dump(ver_dict, db)
+
     with open(ver_path) as db:
         ver_dict = json.load(db)
 
--- a/bugbug/db.py
+++ b/bugbug/db.py
@@ -7,6 +7,7 @@ import json
 import lzma
 import os
 import shutil
+from datetime import datetime
 from urllib.request import urlretrieve
 
 DATABASES = {}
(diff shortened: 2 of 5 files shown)


sbeesm:
In the local version file, for now I've just added the `datetime` when a DB file is downloaded.

And in the repo, should I create a new folder for DB version files?

marco-c:
The local files should contain the same content as the version files of the repo, at the time of the download.

marco-c:
First download:
- There's no local version file, there is a repo version file with v1
- Download
- Store v1 in the local version file

After an update:
- There is a local version file with v1, there is a repo version file with v2
- Download
- Store v2 in the local version file

When there is no update:
- There is a local version file with v2, there is a repo version file with v2
- Don't download

sbeesm:
> First download:
> 
> * There's no local version file, there is a repo version file with v1

But if there is a version file on the repo, won't there be already a local file when a user clones or pull data from upstream?

marco-c:
> But if there is a version file on the repo, won't there be already a local file when a user clones or pull data from upstream?

The version file in the repo and the local version file are two separate files.

sbeesm:
Okay, so we need to add a local version file (ignoring it in git) and check it with the repo version, right?

And just to clarify, we don't need to always download the repo version file from github right? We'd check it from our local clone?

sbeesm:
And instead of creating version files for each DB, can we just create one json file with keys as DB names and their versions as the values?

marco-c:
> Okay, so we need to add a local version file (ignoring it in git) and check it with the repo version, right?

Yes.

> And just to clarify, we don't need to always download the repo version file from github right? We'd check it from our local clone?

Correct!

> And instead of creating version files for each DB, can we just create one json file with keys as DB names and their versions as the values?

Yes, it's the same.

DB_VERSION.json
```diff
@@ -0,0 +1 @@
+{"bugs.json": "v1", "commits.json": "v1"}
```
marco-c: Actually, instead of having a separate file, let's just add a third parameter to the db.register function to specify the version

sbeesm: But then won't we need to change the arguments to db.register in `bugzilla.py` and `repository.py` every time we update DB?

And adding the argument to db.register would then create a new version file or we would add that version to `DATABASES` dictionary?

marco-c: > But then won't we need to change the arguments to db.register in `bugzilla.py` and `repository.py` every time we update DB?

Yes, but this is no different than updating the VERSION file.

> And adding the argument to db.register would then create a new version file or we would add that version to `DATABASES` dictionary?

Everything stays the same, except that the information about the DB version is not going to be in a file in the repo, but is going to be passed with db.register.

sbeesm: Yes, but we need to store the version passed in db.register, right? So that we can check with the local file if our DB need updating?

In db.register, can we do something like:
```
def register(path, url, version):
    DATABASES[path] = [url, version]
```

And in db.download, we can compare the vesions?

marco-c: This is really just an implementation detail, not such a big problem, but I prefer you store a dict in DATABASES[path] (with keys `url` and `version`).

sbeesm:
I've added the parameter `version` to db.register instead of a new file.

But there is a problem, the functions download both the databases even if only one db is changed. Although that can be fixed by passing the particular db as argument but then it would also need adding an argument [here](https://github.com/mozilla/bugbug/pull/171/files#diff-6e38f16215ae91c11fc5c54b74c66d54R62).

Any suggestions how we can solve this?

marco-c:
You could just run db.download() all the time instead of only when args.train is True and completely avoid the db.is_updated_ver() call in run.py.

marco-c:
Or actually, don't change run.py at all. We will download a new DB only when the user wants to retrain.

bugbug/db.py
```diff
@@ -80,3 +90,27 @@ def delete(path, match):
 
     os.unlink(path)
     os.rename(f'{path}_new', path)
+
+
+def is_updated_ver(db_path):
```
marco-c: The naming is a bit confusing, I was assuming `is_updated_ver` meant 'Is the database up to date?', but instead it means 'Is the database out of date?'.
I'd either make this be 'Is the database up to date?' (returning True when the download shouldn't be done and False when it should), or change its name.

bugbug/db.py
```diff
@@ -23,20 +23,30 @@ def register(path, url):
...
 def download():
-    for path, url in DATABASES.items():
-        if os.path.exists(path):
+    for path, info in DATABASES.items():
+        if os.path.exists(path) and not is_updated_ver(path):
             continue
 
         xz_path = f'{path}.xz'
 
+        ver_path = os.path.join(os.path.dirname(path), 'DB_VERSION.json')
```
marco-c: Let's make this a constant 'data/DB_VERSION.json'

bugbug/db.py
```diff
@@ -23,20 +23,30 @@ def register(path, url):
...
-        if not os.path.exists(xz_path):
-            urlretrieve(DATABASES[path], xz_path)
+        if not os.path.exists(xz_path) or is_updated_ver(path):
+            urlretrieve(DATABASES[path]['url'], xz_path)
 
         with open(path, 'wb') as output_f:
             with lzma.open(xz_path) as input_f:
                 shutil.copyfileobj(input_f, output_f)
 
+        if not os.path.exists(ver_path):
```
marco-c: I'd move all of this into `update_ver_file`

bugbug/db.py
```diff
@@ -23,20 +23,24 @@ def register(path, url):
...
 def download():
-    for path, url in DATABASES.items():
-        if os.path.exists(path):
+    for path, info in DATABASES.items():
+        if os.path.exists(path) and not is_outdated(path):
             continue
 
         xz_path = f'{path}.xz'
 
+        ver_path = 'data/DB_VERSION.json'
```
marco-c: Define this as a constant at the top of the file, and use it both in `is_outdated` and `update_ver_file`

bugbug/db.py
```diff
@@ -80,3 +84,32 @@ def delete(path, match):
...
     os.unlink(path)
     os.rename(f'{path}_new', path)
+
+
+def is_outdated(path):
+    ver_path = 'data/DB_VERSION.json'
+
+    if not os.path.exists(ver_path):
+        return True
+    else:
```
marco-c: Nit: no need for the else: after a return

bugbug/db.py
```diff
@@ -80,3 +84,32 @@ def delete(path, match):
...
+        with open(ver_path) as db:
+            ver = json.load(db)
+        if ver.get(path):
```
marco-c: Nit: `if path in ver`

bugbug/db.py
```diff
@@ -80,3 +84,32 @@ def delete(path, match):
...
+            return not DATABASES[path]['version'] == ver[path]
```
marco-c: This is strange, use XXX != YYY instead of not XXX == YYY

bugbug/db.py
```diff
@@ -80,3 +84,32 @@ def delete(path, match):
...
+        else:
+            return True
```
marco-c: You could actually write everything on one line:
```Python
return DATABASES[path]['version'] != ver[path] if path in ver else True
```

bugbug/db.py
```diff
@@ -80,3 +84,32 @@ def delete(path, match):
...
+
+
+def update_ver_file(ver_path, db_path):
+    if not os.path.exists(ver_path):
+        ver_dict = {}
+        ver_dict[db_path] = DATABASES[db_path]['version']
```
marco-c: Create the dict directly with this key, instead of adding it later:
```Python
ver_dict = {
    db_path: DATABASES[db_path]['version']
}
```

bugbug/db.py
```diff
@@ -80,3 +84,32 @@ def delete(path, match):
...
+        with open(ver_path) as db:
+            ver = json.load(db)
+        if ver.get(path):
+            return not DATABASES[path]['version'] == ver[path]
+        else:
+            return True
+
+
+def update_ver_file(ver_path, db_path):
+    if not os.path.exists(ver_path):
```
marco-c: At the end of this branch, you should return, as it's unneeded to reread and rewrite the file.

sbeesm:
The code looks much cleaner now.

Thank You!

bugbug/db.py
```diff
@@ -80,3 +84,29 @@ def delete(path, match):
...
+def is_outdated(path):
+    if not os.path.exists(VER_PATH):
+        return True
+
+    with open(VER_PATH) as db:
+        ver = json.load(db)
+    return DATABASES[path]['version'] != ver[path] if path in ver else True
+
+
+def update_ver_file(db_path):
```
marco-c: Even better, here you could do something like this:
```
def update_ver_file(db_path):
    if os.path.exists(VER_PATH):
        with open(VER_PATH) as db:
            ver_dict = json.load(db)
    else:
        ver_dict = {}

    ver_dict[db_path] = DATABASES[db_path]['version']

    with open(VER_PATH, 'w') as db:
        json.dump(ver_dict, db)
```

marco-c:
I'll merge it later today, after updating the bugs DB, so we don't force everyone to redownload the DB twice.

PR #172, opened 2019-02-14
Add a parameter to only download commits starting from a given date

marco-c:
In the Bugzilla case, we download all bugs from a start date to an end date.
In the repo case, we can ignore all bugs before the start date, but if we want to have all commits for all bugs, we can't limit the end date.

Limiting the start date is enough though to reduce the number of commits from ~500k to ~80k.

### DIFF
--- a/bugbug/repository.py
+++ b/bugbug/repository.py
@@ -11,6 +11,7 @@ import os
 import subprocess
 from collections import namedtuple
 from datetime import datetime
+from datetime import timedelta
 
 import hglib
 from parsepatch.patch import Patch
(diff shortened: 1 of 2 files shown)


bugbug/repository.py
```diff
@@ -166,4 +173,6 @@ def get_commit_map():
     parser.add_argument('repository_dir', help='Path to the repository', action='store')
     args = parser.parse_args()
 
-    download_commits(args.repository_dir)
+    two_years_and_six_months_ago = datetime.utcnow() - timedelta(547)
```
calixteman: nit: datetime.utcnow() - relativedelta(years=2, months=6)

PR #173, opened 2019-02-14
Ignore commits that have been backed out

### LINKED ISSUE #167: In some features related to commits, ignore commits that have been backed out
For example, to count the number of lines modified by the commits of a bug, we only want to sum the lines of non-backed out commits.

ayir:
fix #167


marco-c:
Looks good, you'll just need to do the same to components_touched and components_touched_num too.

PR #174, opened 2019-02-15
'Is reporter a developer' as a feature

### LINKED ISSUE #83: 'Is reporter a developer' as a feature
Similar to #30.

This feature is going to be a boolean saying whether the reporter of the bug is a developer. As a first implementation, we can check if there is any commit where they are authors.

poojan124:
'Is reporter a developer' as a feature

Fixes #83.

### DIFF
--- a/bugbug/bug_features.py
+++ b/bugbug/bug_features.py
@@ -277,6 +277,11 @@ class op_sys(object):
         return bug['op_sys']
 
 
+class is_reporter_a_developer(object):
+    def __call__(self, bug, author_ids, **kwargs):
+        return bug_reporter()(bug).strip() in author_ids
+
+
 def cleanup_url(text):
     text = re.sub(r'http[s]?://(hg.mozilla|searchfox|dxr.mozilla)\S+', '__CODE_REFERENCE_URL__', text)
     return re.sub(r'http\S+', '__URL__', text)
@@ -318,6 +323,13 @@ def cleanup_crash(text):
     return re.sub(r'bp-[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{6}[0-9]{6}\b', '__CRASH_STATS_LINK__', text)
 
 
+def get_author_ids():
+    author_ids = set()
+    for commit in repository.get_commits():
+        author_ids.add(commit['author_email'])
+    return author_ids
+
+
 class BugExtractor(BaseEstimator, TransformerMixin):
     def __init__(self, feature_extractors, cleanup_functions, rollback=False, rollback_when=None, commit_data=False):
         self.feature_extractors = feature_extractors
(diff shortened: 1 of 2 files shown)


bugbug/bug_features.py
```diff
@@ -267,6 +267,11 @@ def __call__(self, bug, **kwargs):
         return len(set(component for commit in bug['commits'] for component in commit['components'] if not commit['ever_backedout']))
 
 
+class is_developer(object):
```
marco-c: `is_reporter_a_developer`

poojan124:
@marco-c how can i update database here? so that it passes the CI checks.

bugbug/bug_features.py
poojan124: Done

bugbug/bug_features.py
```diff
@@ -333,6 +345,7 @@ def transform(self, bugs):
         results = []
 
         reporter_experience_map = defaultdict(int)
+        author_ids = get_author_ids()
```
marco-c: Only call get_author_ids if commit_data is not None (like we do for `self.commit_map`)

marco-c: (this will probably fix CI too)

bugbug/repository.py
```diff
@@ -55,6 +55,7 @@ def _transform(commit):
         'components': list(),
         'author_experience': author_experience[commit],
         'author_experience_90_days': author_experience_90_days[commit],
+        'author_email': commit.author_email.decond('utf-8'),
```
marco-c: Did you try to run the script? `decond` must be wrong :P

bugbug/bug_features.py
poojan124: i dont get it which condition to put here? error saying no author_email in commits.

marco-c: the same we have a few lines earlier for `self.commit_map`.

poojan124: Done

bugbug/repository.py
poojan124: ;_; Done

bugbug/bug_features.py
```diff
@@ -333,6 +345,7 @@ def transform(self, bugs):
         results = []
 
         reporter_experience_map = defaultdict(int)
+        author_ids = None if self.commit_map is None else get_author_ids()
```
marco-c: Do it the other way around, like for commit_map, it is more readable

```Python
author_ids = get_author_ids() if self.commit_map else None
```

poojan124: This is not working if commit_map is empty.
```
cm = {}
>>> 1 if cm else 0
0
```
I can do this instead
```
>>> 1 if cm is not None else 0
```

marco-c: When is commit_map empty?

poojan124: Ok so i mined small data(1000) using new template, and all those commits contains `bug_id = ""` so `get_commit_map` is returning `{} dictionary. `     
So you want to me to continue with this one? 
`author_ids = get_author_ids() if self.commit_map else None`

marco-c: Ah ok, yeah in general it should never be empty, unless you have wrong data in the commits.json DB (you might want to redownload it).

poojan124:
@marco-c I think all good now.

PR #175, opened 2019-02-15
Add total and 90 days reviewer experience 

### LINKED ISSUE #152: Add 'reviewer experience' as feature

ayush1999:
Fixes #152.


bugbug/repository.py
```diff
@@ -130,14 +133,19 @@ def download_commits(repo_dir, date_from):
...
     # Previous commits by the author, in a 90 days window.
     commits_by_author = defaultdict(list)
 
     global author_experience
+    global reviewer_experience
     global author_experience_90_days
+
     for commit in commits:
         author_experience[commit] = total_commits_by_author[commit.author]
+        reviewer_experience[commit] = total_reviews_by_reviewer[commit.author]
```
marco-c: You are using the author here

marco-c:
This doesn't fully fix #152 as it's only adding the total experience, we should add the 90 days experience too.
We can do that in a separate PR though.

bugbug/repository.py
ayush1999: Ah, sorry. Forgot to change that.

ayush1999:
Sure, I'll open a separate PR as soon as this gets merged.

bugbug/repository.py
```diff
@@ -130,14 +133,19 @@ def download_commits(repo_dir, date_from):
...
     # Previous commits by the author, in a 90 days window.
     commits_by_author = defaultdict(list)
 
     global author_experience
+    global reviewer_experience
     global author_experience_90_days
+
     for commit in commits:
         author_experience[commit] = total_commits_by_author[commit.author]
+        reviewer_experience[commit] = total_reviews_by_reviewer[commit.reviewers]
```
marco-c: You are using a list of reviewers here :smile:

ayush1999: `commit.reviewers` returns a byte string, (`b'name'`), not a list.

marco-c: https://github.com/mozilla/version-control-tools/blob/d5db0980ab004b3aada29afdcf798a8a95520240/hgext/mozext/__init__.py#L1062

ayush1999:
Ah, I see. Just adding the `{reviewers}` to template gives us all reviewers as a single string ('AB' instead of 'A', 'B'). I'll see why this is happening (probably due to `rawcommand`).

ayush1999:
Yes, `hg.rawcommand(args)` seems to combine all the elements in the list. Hence we get results like `jdescottesladybenko` instead of `jdescottes, ladybenko`

ayush1999:
Seeing that using the template parameter doesn't return the reviewers in the required format, I've updated it to use re to extract the data from commit description. 

1. In cases where there are more than one reviewers, the average of the reviewer experience is returned. 
2. In cases where there are no reviewers, 0 is returned.

marco-c:
Using the template features, you can likely make reviewers be returned as a list (e.g. by using `join`).

ayush1999:
I'll look into it. Also, I get a `*** failed to import extension hgmo from /home/ayush99/.mozbuild/version-control-tools/hgext/hgmo: No module named utils` when I try to  use hg, why is this happening?

marco-c:
I don't know why it's happening, maybe try updating mercurial to 4.8.

Paste here the entire error message if it still happens.

ayush1999:
I think it's probably because it supports only Python2.7, while my system uses Python3 by default, which may be causing this error.

ayush1999:
@marco-c Anything else I should add in this PR?

ayush1999:
@marco-c Anything else to be added in this PR?

marco-c:
> @marco-c Anything else to be added in this PR?

I haven't reviewed it yet as I thought it was not working for you. Is it working now?

ayush1999:
This is working for me, the error in mercurial I get doesn't really have anything to do with this.

marco-c:
> This is working for me, the error in mercurial I get doesn't really have anything to do with this.

Ah ok, since you mentioned it here I assumed the error was preventing the PR from being usable.

I'm getting this error when trying to run your code:
> hglib.error.CommandError: (255, b'', b'hg: parse error: not displayable without template')

Are you not seeing it? Which version of hg and Python are you using?

ayush1999:
I don't get any such error. I'm on Python 3.7, hg 4.5.3.

bugbug/repository.py
```diff
@@ -130,15 +133,26 @@ def download_commits(repo_dir, date_from):
...
+    global reviewer_experience
     global author_experience_90_days
+
     for commit in commits:
         author_experience[commit] = total_commits_by_author[commit.author]
+
         total_commits_by_author[commit.author] += 1
 
+        if commit.reviewers != ('',):
+            reviewer_experience[commit] = sum(total_reviews_by_reviewer[reviewer] for reviewer in commit.reviewers) / len(commit.reviewers)
```
marco-c: Let's use the sum, not the average

bugbug/repository.py
```diff
@@ -130,15 +133,26 @@ def download_commits(repo_dir, date_from):
...
     global author_experience
+    global reviewer_experience
     global author_experience_90_days
+
     for commit in commits:
         author_experience[commit] = total_commits_by_author[commit.author]
+
         total_commits_by_author[commit.author] += 1
 
+        if commit.reviewers != ('',):
```
marco-c: When there are no reviewers, what is `rev[6]`? Could we make it so that commit.reviewers is empty when there are no reviewers, instead of being a tuple with only the '' element?

marco-c:
> I don't get any such error. I'm on Python 3.7, hg 4.5.3.

It looks like the template doesn't work yet with Mercurial 4.8+. We will have to wait a little to merge the PR, but overall it looks good.

ayush1999:
Oh, I see. Should I open a PR for reviewer experience for 90 days too? We can let that be open along with this. I can also implement the regex if this needs to be merge. @marco-c

bugbug/repository.py
ayush1999: rev[6] is `b' '` when there are no reviewers. Sure, that can be done.

marco-c:
You can implement the 90 days exp in this PR directly, so we don't have two PRs open and rebasing will be easier in case you will have to rebase.

bugbug/repository.py
```diff
@@ -109,6 +111,7 @@ def hg_log(hg, first_rev):
             date=dt,
             bug=rev[4],
             ever_backedout=(rev[5] != b''),
+            reviewers=tuple(rev[6].decode('utf-8').split(',')) if rev[6] != b'' else None
```
marco-c: I don't know how long the Mercurial bug will take to get fixed, so let's resort to parsing the commit message for now (do it here though, so we can easily switch to the template later.

ayush1999: You mean using regex to get the list of reviewers from the commit message, right? (`r=''`)

marco-c: Exactly. You can use this function: https://github.com/mozilla/version-control-tools/blob/2c2812d4a41b690203672a183b1dd85ca8b39e01/pylib/mozautomation/mozautomation/commitparser.py#L129.

marco-c: Just copy the function here, making adjustments if necessary, then add a link to the source of the function and say that we can remove it and use the template when https://bugzilla.mozilla.org/show_bug.cgi?id=1528938 is fixed.

ayush1999:
@marco-c Good to go?

bugbug/repository.py
```diff
@@ -50,6 +53,48 @@
 files_touched_prev = defaultdict(int)
 files_touched_prev_90_days = defaultdict(int)
 
+# This is only a temporary hack: Should be removed after the template issue with reviewers
+# gets fixed. Most of this code is copied from https://github.com/mozilla/version-control-tools/blob/2c2812d4a41b690203672a183b1dd85ca8b39e01/pylib/mozautomation/mozautomation/commitparser.py#L129
```
marco-c: Add a link to the bug I mentioned `gets fixed (BUGZILLA LINK).`

marco-c: Can you move the comment right on top of the function, without the two newlines?

infra/pypi.py
```diff
@@ -14,5 +14,4 @@
 
 os.environ['TWINE_USERNAME'] = data['secret']['pypi']['username']
 os.environ['TWINE_PASSWORD'] = data['secret']['pypi']['password']
-
-subprocess.run(['twine', 'upload', 'dist/*'], check=True)
```
marco-c: Remove the changes from this file

bugbug/repository.py
```diff
@@ -276,17 +323,29 @@ def download_commits(repo_dir, date_from):
 
     # Total previous number of commits by the author.
     total_commits_by_author = defaultdict(int)
+    total_reviews_by_reviewer = defaultdict(int)
     # Previous commits by the author, in a 90 days window.
     commits_by_author = defaultdict(list)
 
     global author_experience
+    global reviewer_experience
     global author_experience_90_days
```
marco-c: Also add reviewer_experience_90_days

ayush1999:
Done.

PR #183, opened 2019-02-19
Add majority voting ensemble in component nn classifier

johngian:


johngian:
@marco-c r?

bugbug/models/component_nn.py
```diff
@@ -178,7 +205,14 @@ def __init__(self, *args, **kwargs):
...
         }
-        self.clf = ComponentNNClassifier(**kwargs)
+
+        estimators = []
+        for i, param in enumerate(self.params):
+            kwargs['params'] = param
+            estimator = ComponentNNClassifier(**kwargs)
+            estimators.append(('model_{}'.format(i), estimator))
+
+        self.clf = VotingClassifier(estimators=estimators, voting='soft', weights=[1, 1, 1])
```
marco-c: If the VotingClassifier is doing the label encoding for us, then maybe we have to remove the encoding from KerasClassifier, or make it optional.

(I wonder why some classifiers do the encoding by themselves and some don't, it's painful!)

bugbug/models/component_nn.py
```diff
@@ -30,59 +31,76 @@
...
         self.long_desc_vocab_size = kwargs.pop('long_desc_vocab_size')
         self.long_desc_emb_sz = kwargs.pop('long_desc_emb_sz')
+        self.params = kwargs.pop('params')
+        return self
 
     def model_creator(self, X, y):
         short_desc_inp = Input(shape=(self.short_desc_maxlen,), name='title_sequence')
         short_desc_emb = Embedding(self.short_desc_vocab_size, self.short_desc_emb_sz)(short_desc_inp)
-        short_desc_emb = SpatialDropout1D(0.2)(short_desc_emb)
+        short_desc_emb = SpatialDropout1D(self.params[0])(short_desc_emb)
```
marco-c: Could we make these params like the others, where you have names?

bugbug/model.py
```diff
@@ -29,6 +29,14 @@ def __init__(self, lemmatization=False):
 
         self.calculate_importance = True
 
+    @property
+    def le(self):
```
marco-c: I hate this, but I guess it's a necessary evil for now until we make the label encoding part better...

johngian: We can always move it to the model definition and define it per model.

marco-c: I mean I don't like the fact that we have to access the internal properties :(
I wish the classifiers didn't encode the labels and the responsibility of encoding were of the users of the classifiers.

johngian:
@marco-c PR updated

bugbug/nn.py
```diff
@@ -36,8 +35,7 @@ def __init__(self, epochs, batch_size):
     def fit(self, X, y):
         X_dict = numpy_to_dict(X)
 
-        self._le = LabelEncoder().fit(y)
```
marco-c: Maybe we have to remove the usage of _le in `predict` too?

johngian:
@marco-c PR updated

Issue #109, opened 2019-01-23
Add number of bug dependencies as a feature


marco-c:

ludoro:
I would like to tackle this! Should I just create another class in the bug_feature.py file? If I manage this I will solve similar problems as well.

marco-c:
Correct!

PR #184, opened 2019-02-19
Added Class for number of bug dependencies, not sure on the return value

ludoro:
I tried to solve issue #109 by adding a class in the file bug_features.py 
However I still do not know what value to return, I found a "depends_on" in other file but I am very unsure about that. Would love some inputs!


marco-c:
You can take a look at the bugs data by loading the JSON file in data/bugs.json.
The field you are looking for is indeed `depends_on`, which is a list field defining which bugs are depending on the bug.

marco-c:
> You can take a look at the bugs data by loading the JSON file in data/bugs.json.

Or simply:
```
from bugbug import bugzilla

example_bug = next(bugzilla.get_bugs())

# Do what you want with example_bug
...
```

marco-c:
Please reopen once you're ready.

PR #188, opened 2019-02-20
Ignore backed out commits when calculating author experience

### LINKED ISSUE #187: Author and reviewer experience should ignore backouts
Author and reviewer experience should not be increased for commits that are backed out.

ayir:
Fixes #187


bugbug/repository.py
```diff
@@ -139,7 +139,8 @@ def download_commits(repo_dir, date_from):
     global author_experience_90_days
     for commit in commits:
         author_experience[commit] = total_commits_by_author[commit.author]
-        total_commits_by_author[commit.author] += 1
+        if not commit.ever_backedout:
+            total_commits_by_author[commit.author] += 1
```
marco-c: You'll need to also ignore it for the 90 days experience below.

The best way to ignore both would probably be to add it at the beginning of the for loop:
```Python
for commit in commits:
    if commit.ever_backedout:
        continue

    ...
```

marco-c: Adding a comment like `# We don't want to consider backed out commits when calculating author/reviewer experience`.

marco-c: Ah, actually, we will have to set author_experience and reviewer_experience anyways, otherwise we fail later on when getting from the map.

bugbug/repository.py
```diff
@@ -139,6 +139,8 @@ def download_commits(repo_dir, date_from):
     global author_experience_90_days
     for commit in commits:
         author_experience[commit] = total_commits_by_author[commit.author]
+        if commit.ever_backedout:
+            continue
```
marco-c: See https://github.com/mozilla/bugbug/pull/188#discussion_r258457469, do as you were doing before, but adding a comment like I said.

And also, you'll need to support ignoring backed out commits for the 90 days experience too.

PR #189, opened 2019-02-20
Add rollback to qaneeded model

ayush1999:
Fixes: #21

### DIFF
--- a/bugbug/models/qaneeded.py
+++ b/bugbug/models/qaneeded.py
@@ -43,7 +43,7 @@ class QANeededModel(Model):
         ]
 
         self.extraction_pipeline = Pipeline([
-            ('bug_extractor', bug_features.BugExtractor(feature_extractors, cleanup_functions)),
+            ('bug_extractor', bug_features.BugExtractor(feature_extractors, cleanup_functions, rollback=True, rollback_when=self.rollback)),
             ('union', ColumnTransformer([
                 ('data', DictVectorizer(), 'data'),
 
@@ -56,6 +56,9 @@ class QANeededModel(Model):
         self.clf = xgboost.XGBClassifier(n_jobs=16)
         self.clf.set_params(predictor='cpu_predictor')
 
+    def rollback(self, change):
+        return change['added'] in ['qawanted', 'qe-verify']
+
     def get_labels(self):
         classes = {}
 
@@ -64,7 +67,7 @@ class QANeededModel(Model):
 
             for entry in bug_data['history']:
                 for change in entry['changes']:
-                    if change['added'].startswith('qawanted'):
+                    if change['added'].startswith('qawanted') or change['added'].startswith('qe-verify'):
                         classes[bug_id] = 1
                     elif 'flags' in entry:
                         for flag in entry['flags']:
--- a/bugbug/models/qaneeded.py
+++ b/bugbug/models/qaneeded.py
@@ -57,7 +57,7 @@ class QANeededModel(Model):
         self.clf.set_params(predictor='cpu_predictor')
 
     def rollback(self, change):
-        return change['added'] in ['qawanted', 'qe-verify']
+        return any(change['added'].startswith(prefix) for prefix in ['qawanted', 'qe-verify'])
 
     def get_labels(self):
         classes = {}
@@ -69,11 +69,6 @@ class QANeededModel(Model):
                 for change in entry['changes']:
                     if change['added'].startswith('qawanted') or change['added'].startswith('qe-verify'):
                         classes[bug_id] = 1
-                    elif 'flags' in entry:
-                        for flag in entry['flags']:
-                            if flag['name'].startswith('qe-verify'):
-                                classes[bug_id] = 1
-
             if bug_id not in classes:
                 classes[bug_id] = 0
 
--- a/tests/test_qaneeded.py
+++ b/tests/test_qaneeded.py
@@ -20,16 +20,19 @@ def test_rollback():
 
     histories = {}
     for bug in bugzilla.get_bugs():
-        histories[int(bug['id'])] = bug
+        histories[int(bug['id'])] = bug['history']
 
-    def contains_field(bug_id):
-        count = False
-        for history in histories[bug_id]['history']:
+    def rollback_point(bug_id):
+        count = 0
+        for history in histories[bug_id]:
             for change in history['changes']:
                 if model.rollback(change):
-                    count = True
-                    break
+                    return count
+                count += 1
         return count
 
-    assert contains_field(1390433), 'A bug field should start with qawanted or qe-verify'
-    assert not contains_field(1389136), 'A bug field should start with qawanted or qe-verify'
+    assert rollback_point(1390433) == 35, 'A bug field should start with qawanted or qe-verify'
+    assert rollback_point(1389136) == 9, 'A bug field should start with qawanted or qe-verify'
+
+    assert rollback_point(1388990) == 29
+    assert rollback_point(1389223) == 8
--- a/tests/test_qaneeded.py
+++ b/tests/test_qaneeded.py
@@ -1,0 +1,35 @@
+# -*- coding: utf-8 -*-
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from bugbug import bugzilla
+from bugbug.models.qaneeded import QANeededModel
+
+
+def test_get_qaneeded_labels():
+    model = QANeededModel()
+    classes = model.get_labels()
+    assert not classes[1389220]
+    assert classes[1389223], 'Bug should contain qawanted in a field'
+    assert classes[1390433], 'Bug should contain qe-verify in a field'
+
+
+def test_rollback():
+    model = QANeededModel()
+
+    histories = {}
+    for bug in bugzilla.get_bugs():
+        histories[int(bug['id'])] = bug
+
+    def contains_field(bug_id):
+        count = False
+        for history in histories[bug_id]['history']:
+            for change in history['changes']:
+                if model.rollback(change):
+                    count = True
+                    break
+        return count
+
+    assert contains_field(1390433), 'A bug field should start with qawanted or qe-verify'
+    assert not contains_field(1389136), 'A bug field should start with qawanted or qe-verify'


bugbug/models/qaneeded.py
```diff
@@ -56,6 +56,9 @@ def __init__(self, lemmatization=False):
         self.clf = xgboost.XGBClassifier(n_jobs=16)
         self.clf.set_params(predictor='cpu_predictor')
 
+    def rollback(self, change):
+        return change['field_name'] in ['qawanted', 'qe-verify']
```
marco-c: Analyze a few bugs manually, IIRC qe-verify is never in 'field_name'.

ayush1999: Yes, neither is `qawanted`. `field_name` contains `qa-contact`, which is the closest it comes.

marco-c: Yep. So you'll need to use some other value from `change`.

ayush1999: We need to rollback to when `qaneeded` or `qe-verify` was added, right? So maybe setting change['added'] in `['qawanted', 'qe-verify']` should work?

marco-c: Have you checked if they can be in change['added']?

ayush1999: Yes, they are present in `change['added']`.

marco-c: Good, then it should work!
Please also add a test with a bug with qawanted and one with qe-verify.

marco-c: Also, if they are present in change['added'], then we should update get_labels to take that into account.

ayush1999: `get_labels` already uses this information to give bugs a target label.

ayush1999:
Adding the rollback function as `change['added'] in ['qawanted', 'qe-verify']` actually reduces the accuracy to ~0.73. Is adding this a good option? @marco-c

ayush1999:
Done. Results: The accuracy increases from 0.81 to 0.93.

bugbug/models/qaneeded.py
```diff
@@ -64,7 +67,7 @@ def get_labels(self):
 
             for entry in bug_data['history']:
                 for change in entry['changes']:
-                    if change['added'].startswith('qawanted'):
+                    if change['added'].startswith('qawanted') or change['added'].startswith('qe-verify'):
                         classes[bug_id] = 1
                     elif 'flags' in entry:
```
marco-c: The second part here might have the be removed if it is unnecessary. Can you check if there is any bug where this happens?

tests/test_qaneeded.py
```diff
@@ -0,0 +1,13 @@
...
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from bugbug.models.qaneeded import QANeededModel
+
+
+def test_get_qaneeded_labels():
+    model = QANeededModel()
+    classes = model.get_labels()
+    assert not classes[1389220]
+    assert classes[1389223]
```
marco-c: Can you also add a test with qawanted?

bugbug/models/qaneeded.py
ayush1999: Yes, there are quite a few bugs with added as qe-verify.

bugbug/models/qaneeded.py
marco-c: Looking at the bug you used for the test, this doesn't work. The bug has added == qe-verify+.

bugbug/models/qaneeded.py
marco-c: I'm not talking about the part you added (we talked about that already, I suggested you to add it :P).
I'm talking about the second part of the pre-existing code, which is analyzing flags.

marco-c:
> Done. Results: The accuracy increases from 0.81 to 0.93.

Mmmh, unexpected! The accuracy in theory should decrease, as we are reducing the information the model can work on.

bugbug/models/qaneeded.py
ayush1999: Oh! I just checked: there are around 1235 bugs with `name` as `qe-verify`. I think we should let this be here. Right?

bugbug/models/qaneeded.py
ayush1999: Fixed that. Added the `startswith` method for this.

tests/test_qaneeded.py
ayush1999: Done.

bugbug/models/qaneeded.py
marco-c: Can you show me an example? I want to check the difference between one of them and 1389223.

tests/test_qaneeded.py
marco-c: And we should test the case where qe-verify is in the flags.

Also, please add a rollback test.

bugbug/models/qaneeded.py
marco-c: (Also, if there are cases like that, then you need to add support for it to rollback)

ayush1999: > Can you show me an example? I want to check the difference between one of them and 1389223.

Sure, for example bug id 1389252.

marco-c: 1389252 doesn't match that: https://bugzilla.mozilla.org/rest/bug/1389252/history.

ayush1999: Oh, sorry. I was looking at the wrong `flags` key.

ayush1999:
@marco-c Sorry, I was looking at the wrong `flags` key. I just checked ; the second condition is redundant, I'll remove it from the loop.

ayush1999:
I'e added rollback tests for two bugs (one True and other False). The tracking model also doesn't contain a rollback test. Should I add a test there too? (in this PR?)

marco-c:
> Should I add a test there too? (in this PR?)

No, let's keep this PR about the qaneeded model.

bugbug/models/qaneeded.py
```diff
@@ -56,6 +56,9 @@ def __init__(self, lemmatization=False):
         self.clf = xgboost.XGBClassifier(n_jobs=16)
         self.clf.set_params(predictor='cpu_predictor')
 
+    def rollback(self, change):
+        return change['added'].startswith('qawanted') or change['added'].startswith('qe-verify')
```
marco-c: It'd be more Pythonic to do something like:
```Python
any(change['added'].startswith(prefix) for prefix in ['qawanted', 'qe-verify']
```

tests/test_qaneeded.py
```diff
@@ -0,0 +1,36 @@
...
+from bugbug import bugzilla
+from bugbug.models.qaneeded import QANeededModel
+
+
+def test_get_qaneeded_labels():
+    model = QANeededModel()
+    classes = model.get_labels()
+    assert not classes[1389220]
+    assert classes[1389223]
+    assert classes[1390433]
```
marco-c: Add a message to the asserts (e.g. Bug should contain qawanted for the qawanted one)

tests/test_qaneeded.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+def test_rollback():
+    model = QANeededModel()
+
+    bugs = {}
```
marco-c: Call this `histories`, since you're only storing the history stuff.

tests/test_qaneeded.py
```diff
@@ -0,0 +1,36 @@
...
+    for bug in bugzilla.get_bugs():
+        bugs[int(bug['id'])] = bug['history']
+
+    def test(bug_id):
```
marco-c: The name of the function is too generic.

BTW, we also want to test that we are rolling back to the right point in the history

ayush1999: To test if rollback is to the right time, should I check if all comments are made before the `rollback_date`?

marco-c: You could simply assert on the number of changes you loop through before finding the right one.

marco-c: Basically make count a integer increasing it for every iteration of the inner loop, then in the caller do:
`assert rollback_point(BUG_ID) == X`

ayush1999: Sorry for the delay, just finished with my exams. Correct me if I'm wrong, but rollback is supposed to take a bug and return the same bug in an initial stage, right? Because whenever I test the function, the value before and after rollback always comes as equal.

marco-c: Rollback is used to go back in the history of a bug, not exclusively to the initial state, but also to a given state (this is what the `rollback_when` function is for, to know at which point in the history to stop).

In your case, if a bug has qawanted in the history, rollback should go back to the point when qawanted was added.

ayush1999:
I've added a count that increments through all changes and returns whenever `rollback(change)`  is True. Please review @marco-c

tests/test_qaneeded.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+def test_rollback():
+    model = QANeededModel()
+
+    histories = {}
+    for bug in bugzilla.get_bugs():
+        histories[int(bug['id'])] = bug
+
+    def contains_field(bug_id):
```
marco-c: This is redundant with the other function, you can remove it and just always use the other function.

tests/test_qaneeded.py
```diff
@@ -0,0 +1,52 @@
...
+            for change in history['changes']:
+                if not model.rollback(change):
+                    count += 1
+                else:
+                    ret = True
+                    break
+        return count
+
+    assert contains_field(1390433), 'A bug field should start with qawanted or qe-verify'
+    assert not contains_field(1389136), 'A bug field should start with qawanted or qe-verify'
```
marco-c: Make these two use rollback_point too

tests/test_qaneeded.py
```diff
@@ -0,0 +1,52 @@
...
+    assert classes[1389223], 'Bug should contain qawanted in a field'
+    assert classes[1390433], 'Bug should contain qe-verify in a field'
+
+
+def test_rollback():
+    model = QANeededModel()
+
+    histories = {}
+    for bug in bugzilla.get_bugs():
+        histories[int(bug['id'])] = bug
```
marco-c: You were storing the history before, so I suggested to change the name to `histories`. Why did you move to storing the whole bug instead?

ayush1999: Removed it. I thought it'd maintain uniformity, but it seems redundant now.

tests/test_qaneeded.py
```diff
@@ -0,0 +1,43 @@
...
+        histories[int(bug['id'])] = bug['history']
+
+    def rollback_point(bug_id):
+        count = 0
+        ret = False
+        for history in histories[bug_id]:
+            if ret:
+                break
+            for change in history['changes']:
+                if not model.rollback(change):
```
marco-c: This can be simplified:
```Python
if model.rollback(change):
    return count
count += 1
```
And you can completely remove the `ret` variable.

ayush1999:
Done. Please review.

PR #190, opened 2019-02-21
Add script to get untriaged bugs

johngian:


scripts/get_untriaged.py
```diff
@@ -0,0 +1,47 @@
...
+    bugzilla.download_bugs_between(three_months_ago, today)
+
+    # Get untriaged bugs
+    bugs = bugzilla.get_bugs()
+    untriaged_bugs = []
+    for bug in bugs:
+        for history in bug['history']:
+            for change in history['changes']:
+                if change['field_name'] == 'component' and change['added'] == 'Untriaged':
+                    untriaged_bugs.append(bug)
```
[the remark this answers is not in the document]
marco-c: We should do the opposite, find bugs for which one of change['removed'] was 'Untriaged', so we have bugs which were Untriaged and then were triaged by someone.

scripts/get_untriaged.py
```diff
@@ -0,0 +1,47 @@
...
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--token',
+                        help='Bugzilla API token')
+    parser.add_argument('--days-back',
+                        type=int,
+                        help='Days since to fetch bugs')
+
+    args = parser.parse_args()
+    print(get_untriaged(args))
```
marco-c: It'd be nice, after getting the untriaged bugs, to apply the models we have for triage and calculate the metrics for them.

scripts/get_untriaged.py
```diff
@@ -0,0 +1,101 @@
...
+
+            if classifiable:
+                print("Classifying bug with ID: {}".format(bug['id']))
+                classification = model.classify(bug)[0]
+                print("Classified bug as: {}".format(classification))
+
+            else:
+                print("Skipping bug with ID: {}".format(bug['id']))
+
+            correct_prediction = full_component == classification
```
marco-c: We should make sure we apply the same rules from https://github.com/mozilla/bugbug/blob/c0373fb6dd1f4d583b865407c8a2389787fdcfce/bugbug/models/component.py#L60, e.g. a `Core::Audio/Video: Playback` issue classified as `Core::Audio/Video` should be OK.

We can probably just extract (part of) the logic from the for loop at https://github.com/mozilla/bugbug/blob/c0373fb6dd1f4d583b865407c8a2389787fdcfce/bugbug/models/component.py#L118 into a function and use it both there and here.

scripts/get_untriaged.py
```diff
@@ -0,0 +1,101 @@
...
+    three_months_ago = today - timedelta(days=args.days_back)
+
+    # Set bugzilla token and download bugs
+    bugzilla.set_token(args.token)
+    bugzilla.download_bugs_between(three_months_ago, today)
+
+    # Get untriaged bugs
+    bugs = bugzilla.get_bugs()
+    untriaged_bugs = []
+    for bug in bugs:
```
marco-c: Can you make sure we only return the bugs we've just downloaded? (bugzilla.get_bugs() will get all bugs from the bugs DB).

You can do that by simply storing the bug IDs when you call download_bugs_between, and add a if bug['id'] not in bug_ids: continue here

johngian:
@marco-c PR updated. I moved the component filtering to a method so we can call it in other parts of the codebase as well.

bugbug/models/component.py
```diff
@@ -57,86 +113,31 @@ def __init__(self, lemmatization=False):
...
-            'WebExtensions'
-        }
-
-        classes = {}
-
-        conflated_components = [
-            'Core::Audio/Video', 'Core::Graphics', 'Core::IPC', 'Core::JavaScript', 'Core::Layout', 'Core::Networking',
-            'Core::Print', 'Core::WebRTC', 'Firefox::Activity Streams', 'Toolkit::Password Manager',
-        ]
+    def filter_components(self, bug_data):
```
marco-c: Nit: `filter_component`, since it operates on a single bug

PR #192, opened 2019-02-25
Ignore missing categorical features

johngian:


bugbug/models/component_nn.py
```diff
@@ -260,7 +261,7 @@ def __init__(self, *args, **kwargs):
             ('union', StructuredColumnTransformer([
                 ('platform', make_pipeline(DictExtractor('platform'), OrdinalEncoder()), 'data'),
                 ('op_sys', make_pipeline(DictExtractor('op_sys'), OrdinalEncoder()), 'data'),
-                ('bug_reporter', make_pipeline(DictExtractor('bug_reporter'), OrdinalEncoder()), 'data'),
+                ('bug_reporter', make_pipeline(DictExtractor('bug_reporter'), MissingOrdinalEncoder()), 'data'),
```
marco-c: It should be feasible without subclassing, see for example the part about `You can set the parameters using the names issued` on https://scikit-learn.org/stable/modules/generated/sklearn.pipeline.Pipeline.html.

johngian: I don't think that it's going to be helpful here. The problem is that even though its internally supported, `OrdinalEncoder` doesn't allow passing this as an argument because it never calls the underlying functions with this argument in the signature.

There is an open bug in scikit for that.

marco-c: Oh, OK. That's a pity :(

bugbug/utils.py
```diff
@@ -34,3 +35,14 @@ def fit(self, x, y=None):
 
     def transform(self, data):
         return np.array([elem[self.key] for elem in data]).reshape(-1, 1)
+
+
+class MissingOrdinalEncoder(OrdinalEncoder):
```
marco-c: Can you add a comment explaining that this is necessary until supported upstream in OrdinalEncoder?

PR #197, opened 2019-02-27
Add more negative examples for DevDocNeeded model

### LINKED ISSUE #101: In the DevDocNeeded model, consider bugs for which the dev-doc-needed keyword was removed as negative examples
See also #79.

ranpox:
Fixes #101

### DIFF
--- a/bugbug/models/devdocneeded.py
+++ b/bugbug/models/devdocneeded.py
@@ -78,7 +78,8 @@ class DevDocNeededModel(Model):
                         classes[bug_id] = 0
                     # Bugs that go from dev-doc-needed to dev-doc-complete are guaranteed to be good
                     # Bugs that go from not having dev-doc-needed to having dev-doc-complete are bugs
-                    #   that were missed by previous scans through content but someone realized it should have been flagged and updated the docs, found the docs already updated.
+                    # that were missed by previous scans through content but someone realized it 
+                    # should have been flagged and updated the docs, found the docs already updated.
                     elif change['field_name'] == 'keywords' and any(keyword in change['added'] for keyword in ['dev-doc-needed', 'dev-doc-complete']):
                         classes[bug_id] = 1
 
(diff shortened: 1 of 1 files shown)


bugbug/models/devdocneeded.py
```diff
@@ -73,6 +73,9 @@ def get_labels(self):
 
             for entry in bug_data['history']:
                 for change in entry['changes']:
+                    if change['field_name'] == 'keywords' and 'dev-doc-needed' in change['removed'] and 'dev-doc-complete' not in change['added']:
```
marco-c: Can you add a comment to explain the rationale? (Basically the suggestion from the issue)

bugbug/models/devdocneeded.py
```diff
@@ -73,6 +73,9 @@ def get_labels(self):
...
+                        classes[bug_id] = 0
+                        continue
```
marco-c: Instead of continue, just use elif in the next line

bugbug/models/devdocneeded.py
```diff
@@ -73,7 +73,13 @@ def get_labels(self):
 
             for entry in bug_data['history']:
                 for change in entry['changes']:
-                    if change['field_name'] == 'keywords' and any(keyword in change['added'] for keyword in ['dev-doc-needed', 'dev-doc-complete']):
+                    # Bugs that get dev-doc-needed removed from them at some point after it's been added (this suggests a false positive among human-analyzed bugs)
+                    if change['field_name'] == 'keywords' and 'dev-doc-needed' in change['removed'] and 'dev-doc-complete' not in change['added']:
+                        classes[bug_id] = 0
+                    # Bugs that go from dev-doc-needed to dev-doc-complete are guaranteed to be good
+                    # Bugs that go from not having dev-doc-needed to having dev-doc-complete are bugs
+                    #   that were missed by previous scans through content but someone realized it should have been flagged and updated the docs, found the docs already updated.
```
marco-c: Remove the two spaces before `that`, so that you align it with the previus line.
Also, the second line is too long, break it more or less at the same point where you broke the previous line (so after `realized it`).

marco-c:
There is a flake8 error in the Travis build:
> ./bugbug/models/devdocneeded.py:81:97: W291 trailing whitespace

PR #199, opened 2019-02-27
Differentiate between Firefox and external DLLs during cleanup

### LINKED ISSUE #88: When cleaning up DLL names, differentiate between third-party DLLs and Firefox ones
We're currently replacing all DLL names with a _DLL_NAME_ token, but we should differentiate between third-party DLLs and Firefox DLLs (XUL.dll, and so on).

LakshyAAAgrawal:
Introduced added regex in "cleanup_dll" function, so that it differentiates between the DLLs used by firefox and external DLLs. The filenames have been added into the regex raw string. Propose to separate it into a list of filenames, and then add into regex.
Closes #88


bugbug/bug_features.py
```diff
@@ -300,7 +300,107 @@ def cleanup_hex(text):
 
 
 def cleanup_dll(text):
-    return re.sub(r'\w+(\.dll|\.so|\.dylib)\b', '__DLL_NAME__', text)
+    firefox_dlls = [
+        'libmozwayland.so',
```
marco-c: Put a few names on the same line, otherwise this gets too long

bugbug/bug_features.py
```diff
@@ -300,7 +300,107 @@ def cleanup_hex(text):
...
+        'msvcp140.dll',
+        'nss3.dll',
+        'nssckbi.dll',
+        'nssdbm3.dll',
+        'qipcap64.dll',
+        'softokn3.dll',
+        'ucrtbase.dll',
+        'vcruntime140.dll',
+        'xul.dll',
+        'xul.dll.sig',
```
marco-c: Remove the files ending with `.sig`.

tests/test_cleanup_functions.py
```diff
@@ -48,8 +48,9 @@ def test_cleanup_hex():
 def test_cleanup_dll():
     tests = [
         ('Crashing thread: 0 scdetour.dll scdetour.dll@0x2dd77', 'Crashing thread: 0 __DLL_NAME__ __DLL_NAME__@0x2dd77'),
-        ('Crash in libxul.so@0x287ad36 | libxul.so@0x270c062', 'Crash in __DLL_NAME__@0x287ad36 | __DLL_NAME__@0x270c062'),
-        ('Crash in libsystem_pthread.dylib@0x14fc', 'Crash in __DLL_NAME__@0x14fc')
+        ('Crash in libxul.so@0x287ad36 | libxul.so@0x270c062', 'Crash in libxul.so@0x287ad36 | libxul.so@0x270c062'),
+        ('Crash in libsystem_pthread.dylib@0x14fc', 'Crash in __DLL_NAME__@0x14fc'),
+        ('Crash in liblgpllibs.so@0x14fc exmpl.so@0xask ', 'Crash in liblgpllibs.so@0x14fc __DLL_NAME__@0xask ')
     ]
```
marco-c: Also add a test with a Firefox DLL and a test with a Firefox dylib, to make sure we don't rewrite them.

bugbug/bug_features.py
```diff
@@ -300,7 +300,29 @@ def cleanup_hex(text):
 
 
 def cleanup_dll(text):
-    return re.sub(r'\w+(\.dll|\.so|\.dylib)\b', '__DLL_NAME__', text)
+    firefox_dlls = [
```
marco-c: Let's also make this a constant, defining it outside of the function (just before the function) and in upper-case (a Python convention for constants).

marco-c: The entire `inv_match` can be the constant (but rename it something like FIREFOX_DLLS_MATCH), this way we only calculate it once and not for every call of the function.

bugbug/bug_features.py
```diff
@@ -299,8 +299,31 @@ def cleanup_hex(text):
...
+    'api-ms-win-crt-time-l1-1-0.dll', 'api-ms-win-crt-utility-l1-1-0.dll', 'd3dcompiler_47.dll', 'freebl3.dll',
+    'mozavcodec.dll', 'mozavutil.dll', 'mozglue.dll', 'msvcp140.dll', 'nss3.dll', 'nssckbi.dll', 'nssdbm3.dll', 'qipcap64.dll',
+    'softokn3.dll', 'ucrtbase.dll', 'vcruntime140.dll', 'xul.dll', 'clearkey.dll', 'libfreebl3.dylib', 'liblgpllibs.dylib', 'libmozavcodec.dylib',
+    'libmozavutil.dylib', 'libmozglue.dylib', 'libnss3.dylib', 'libnssckbi.dylib', 'libnssdbm3.dylib', 'libplugin_child_interpose.dylib', 'libsoftokn3.dylib'
+]).replace('.', r'\.')
+
+
 def cleanup_dll(text):
-    return re.sub(r'\w+(\.dll|\.so|\.dylib)\b', '__DLL_NAME__', text)
+    regex = r'\b(?!' + FIREFOX_DLLS_MATCH + r')\w+(\.dll|\.so|\.dylib)\b'
```
marco-c: One last thing, please use a f-string instead of concatenating strings.
```Python
fr'\b(?!{FIREFOX_DLLS_MATCH})\w+(\.dll|\.so|\.dylib)\b'
```

marco-c: One last thing, please use a f-string instead of concatenating strings.
```Python
fr'\b(?!{FIREFOX_DLLS_MATCH})\w+(\.dll|\.so|\.dylib)\b'
```

PR #204, opened 2019-03-03
Bug assignee

### LINKED ISSUE #61: Create a classifier to suggest an assignee for a bug
A new classifier, similar to the ones already present under bugbug/models/, should be created to predict the assignee of a bug.
This will be similar to the component model, except the label will be the assignee instead of the product/component.

LakshyAAAgrawal:
Created a class for bug assignment.

Fixes #61.


marco-c:
Can you rebase on top of master? There are some unrelated commits.

LakshyAAAgrawal:
The output of AssigneeModel.train() are in [assignee_train_output.txt](https://github.com/mozilla/bugbug/files/2923229/assignee_train_output.txt)

bugbug/models/assignee.py
```diff
@@ -0,0 +1,81 @@
...
+
+                ('comments', self.text_vectorizer(min_df=0.0001), 'comments'),
+            ])),
+        ])
+
+        self.clf = xgboost.XGBClassifier(n_jobs=16)
+        self.clf.set_params(predictor='cpu_predictor')
+
+    def get_labels(self):
+
```
marco-c: Nit: Remove this empty line

bugbug/models/assignee.py
```diff
@@ -0,0 +1,81 @@
...
+        classes = {}
+
+        for bug_data in bugzilla.get_bugs():
+            if bug_data['assigned_to_detail']['email'] == 'nobody@mozilla.org':
```
marco-c: There should be at least another `nobody` that we should exclude

marco-c: The NSS one, but I don't remember the address. Check all addresses in all bugs that contain the string `nobody`

bugbug/models/assignee.py
```diff
@@ -0,0 +1,81 @@
...
+                continue
+
+            bug_id = int(bug_data['id'])
+            classes[bug_id] = bug_data['assigned_to_detail']['email']
+
+        component_counts = Counter(classes.values()).most_common()
```
marco-c: You should not use the `component` word here, as this model is about assignees.

marco-c: I would also filter out assignees that are assigned to very few bugs (say, less than 5).

bugbug/models/assignee.py
```diff
@@ -0,0 +1,81 @@
...
+        ]
+
+        cleanup_functions = [
+            bug_features.cleanup_fileref,
+            bug_features.cleanup_url,
+            bug_features.cleanup_synonyms,
+        ]
+
+        self.extraction_pipeline = Pipeline([
+            ('bug_extractor', bug_features.BugExtractor(feature_extractors, cleanup_functions, rollback=True)),
```
marco-c: We should probably rollback to the point where the assignee was decided, rather than the point where the bug was filed, but we can do this in a follow-up.

LakshyAAAgrawal: What I understand is that, I will need to create a function func_name which will be passed as an argument [rollback_when](https://github.com/mozilla/bugbug/blob/master/bugbug/bug_features.py#L361) to [BugExtractor](https://github.com/mozilla/bugbug/blob/master/bugbug/bug_features.py#L356) and also set rollback=True. 
```python
def func_name(change):
    addresses_to_exclude = [
            'nobody@bugzilla.org',
            'nobody@example.com',
            'nobody@fedoraproject.org',
            'nobody@mozilla.org',
            'nobody@msg1.fake',
            'nobody@nss.bugs',
            'nobody@t4b.me'
    ]
    if change['field_name'] == 'assigned_to':
        if change['removed'] in addresses_to_exclude:
            return(True)
    return(False)

```
func_name() takes an argument [change](https://github.com/mozilla/bugbug/blob/master/bugbug/bug_snapshot.py#L101) from [bug_snapshot.rollback](https://github.com/mozilla/bugbug/blob/master/bugbug/bug_snapshot.py#L96), and returns True if change has "field_name":"assigned_to" and change['removed'] belongs to the addresses associated with "No Assignee".
Is this correct?

marco-c: Yes, correct! But you can simplify by just checking if the assignee was ever changed, you don't need to check that it was changed from a nobody to a non-nobody.

marco-c: As I said though, this can be done in a follow-up! (follow-up means another PR)

bugbug/models/assignee.py
```diff
@@ -0,0 +1,89 @@
...
+        ]
+        for bug_data in bugzilla.get_bugs():
+            if bug_data['assigned_to_detail']['email'] in addresses_to_exclude:
+                continue
+
+            bug_id = int(bug_data['id'])
+            classes[bug_id] = bug_data['assigned_to_detail']['email']
+
+        assignee_counts = Counter(classes.values()).most_common()
+        top_assignees = set(assignee if count > 5 else None for assignee, count in assignee_counts)
```
marco-c: You don't need to assign the None value and then remove it later, you can just use a filter condition like: `set(assignee for assignee, count in assignee_counts if count > 5`

marco-c: Also, make 5 a constant that you define at the top of the file (MINIMUM_ASSIGNMENTS)

bugbug/models/assignee.py
LakshyAAAgrawal: I will create another PR for this. 
1. Could you suggest where to define 'func_name'?
2. Could you also suggest an appropriate name for 'func_name'? is 'is_change_assignee' a good name?
3. I will follow your advise, however, just for my reference, I wanted to know why wouldn't it be better to rollback to the earliest assignee change from nobody to Non-nobody, as that would preserve all further assignee changes. On the contrary, this will lead to loss of that info.

marco-c: 1. You can define it in the model class, look at the other models using rollback.
2. As in the other models, you can just call it `rollback`.
3. What I suggested is just a simplification, the result will be the same (you'll rollback to the first assignee change in both cases)

LakshyAAAgrawal: Thanks for the explanation on point 3. I just realized that it will be the case.
Is there anything else left to do on this issue?
I have added the rollback in [another branch](https://github.com/LakshyAAAgrawal/bugbug/tree/rollback_assignee_change)( branched out from this branch ). Should I create a pull request from that, or wait for this to merge?

marco-c: I've merged the PR now.

In general, feel free to open PRs even if they depend on another PR to land first (you can just rebase it once the first lands).

Issue #210, opened 2019-03-06
Use qaurgent too in the qa model


marco-c:

ayush1999:
This should go in both the `rollback` and `get_labels` functions, right?

ayush1999:
Should I take this up? (Also, this should go in both the `rollback` and `get_labels` function, right?)

marco-c:
As you want :smiley: 

Yep, in both functions, as the others.

PR #211, opened 2019-03-08
Add qaurgent to the qa model

ayush1999:
Fixes: #210


marco-c:
Did you check if the number of labels increased?

ayush1999:
Just checked: the number of labels doesn't change: 113619 before and after the change.

ayush1999:
Just checked: the number of labels increased by a small amount : from 5821 to 5824.

PR #213, opened 2019-03-08
Rollback assignee change

LakshyAAAgrawal:
Added rollback function to AssigneeModel and used it for rollback in Bug Feature Extraction Using BugExtractor.
Refer https://github.com/mozilla/bugbug/pull/204#discussion_r262426790

### DIFF
--- a/bugbug/models/assignee.py
+++ b/bugbug/models/assignee.py
@@ -55,7 +55,7 @@ class AssigneeModel(Model):
         ]
 
         self.extraction_pipeline = Pipeline([
-            ('bug_extractor', bug_features.BugExtractor(feature_extractors, cleanup_functions, rollback=True)),
+            ('bug_extractor', bug_features.BugExtractor(feature_extractors, cleanup_functions, rollback=True, rollback_when=self.rollback)),
             ('union', ColumnTransformer([
                 ('data', DictVectorizer(), 'data'),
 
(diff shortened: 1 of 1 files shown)


LakshyAAAgrawal:
Hey, the output of AssigneeModel.train() is attached in [text_out.txt](https://github.com/mozilla/bugbug/files/2948106/text_out.txt)

marco-c:
Could you post the average precision/recall values before/after this change?

LakshyAAAgrawal:
Hey,
Will something like [this](https://stackoverflow.com/a/31587532) need to be implemented, or is there some in-built way to print the required information?

marco-c:
The required information is already printed, it is a subset of what you attached here.

LakshyAAAgrawal:
After Change

| No Confidence Threshold > than | Precision | Recall |
|-----------------------------------------|-------------|----------|
\- | 0.56 | 0.56 |
0.6 | 0.87 | 0.87 |
0.7 | 0.89 | 0.90 |
0.8 | 0.93 | 0.94 |
0.9 | 0.96 | 0.97 |

LakshyAAAgrawal:
I will post the results without rollback in a few hours.

marco-c:
It's surprising that it got better, we have removed information, I was thinking it would get worse.

LakshyAAAgrawal:
Should I report back with more testing and data? If yes, what?

marco-c:
No this is OK ;)

PR #214, opened 2019-03-09
In the uplift model, rollback bugs to the time the uplift was accepted or rejected

### LINKED ISSUE #60: Use the bug snapshot transform in the "uplift" model
Depends on #5.

chidauri:
@marco-c please review.By the way I am able to run the model after these changes.

Fixes #60.


chidauri:
closes #60

bugbug/models/uplift.py
```diff
@@ -56,6 +56,9 @@ def __init__(self, lemmatization=False):
         self.clf = xgboost.XGBClassifier(n_jobs=16)
         self.clf.set_params(predictor='cpu_predictor')
 
+    def rollback(self, change):
+        return (change['added'].startswith('approval-mozilla-') and (change['added'].endswith('+') or change['added'].endswith('-'))) or (change['removed'].startswith('approval-mozilla-') and (change['removed'].endswith('+') or change['removed'].endswith('-')))
```
marco-c: This is not really readable.
You can make it more readable by either adding temporary variables, or splitting the conditions (checking one first, then another inside it, etc.), or adding newlines between conditions.

chidauri: I think the changes i've made are wrong, and  i'll have to change the whole rollback function.

Because (approval-mozilla-) is found in "attachments" as in line 67(https://github.com/mozilla/bugbug/blob/e1bf7d6fb546b1a48e631e0854d230de669611fe/bugbug/models/uplift.py#L67) and not in bug["history"].
Right?

marco-c: Yes, that's right!

chidauri:
@marco-c please take a look at the conflict.

bugbug/models/uplift.py
chidauri: Done.

marco-c:
> should i resolve the conflict?

Yes

bugbug/bug_snapshot.py
```diff
@@ -242,8 +242,10 @@ def rollback(bug, when, verbose=True, all_inconsistencies=False):
                                 found_flag = f
 
                         # TODO: always assert here, once https://bugzilla.mozilla.org/show_bug.cgi?id=1514415 is fixed.
+
                         if obj['id'] not in [1052536, 1201115, 1213517, 794863] and not (to_remove == 'in-testsuite+' and obj['id'] in [1318438, 1312852, 1332255, 1344690, 1362387, 1380306]) and not (to_remove == 'in-testsuite-' and bug['id'] in [1321444, 1342431, 1370129]) and not (to_remove == 'approval-comm-esr52?' and bug['id'] == 1352850) and not (to_remove == 'checkin+' and bug['id'] in [1308868, 1357808, 1361361, 1365763, 1328454]) and not (to_remove == 'checkin-' and bug['id'] == 1412952) and not (to_remove == 'webcompat?' and obj['id'] in [1360579, 1364598]) and not (to_remove == 'qe-verify-' and bug['id'] in [1322685, 1336510, 1363358, 1370506, 1374024, 1377911, 1393848, 1396334, 1398874, 1419371]):
                             assert found_flag is not None, f'flag {to_remove} not found in {bug["id"]}'
+
```
marco-c: Please remove the whitespace changes in this file

bugbug/models/uplift.py
```diff
@@ -56,6 +56,12 @@ def __init__(self, lemmatization=False):
         self.clf = xgboost.XGBClassifier(n_jobs=16)
         self.clf.set_params(predictor='cpu_predictor')
 
+    def rollback(self, change):
+        if change['field_name'].startswith('flagtypes.name') and change['added'].startswith('approval-mozilla-'):
+            if change['added'].endswith('+') or change['added'].endswith('-'):
```
marco-c: This can be added as another "and" of the previous if.

bugbug/models/uplift.py
```diff
@@ -56,6 +56,9 @@ def __init__(self, lemmatization=False):
         self.clf = xgboost.XGBClassifier(n_jobs=16)
         self.clf.set_params(predictor='cpu_predictor')
 
+    def rollback(self, change):
+        return (change['field_name'].startswith('flagtypes.name') and change['added'].startswith('approval-mozilla-') and (change['added'].endswith('+') or change['added'].endswith('-')))
```
marco-c: Sorry, one last thing.
For `change['field_name'].startswith('flagtypes.name')`, I think it is unnecessary to use `startswith`, it can be just an equality check. Could you confirm? (to confirm, just check that in all histories of all bugs there is no case where change['field_name'].startswith('flagtypes.name') but change['field_name'] != 'flagtypes.name'.

chidauri: checked and there is not such case 👍

PR #215, opened 2019-03-11
Expose n_grams from terminal when using TfIdfVectorizer

ayush1999:
Added option to set n_grams from terminal, example usage: `python run.py --train --goal=bug --ngrams=2`. Currently, I've added this only for the bug, component, regression and tracking models.


ayush1999:
@marco-c Training is taking a lot of time for the component and tracking models, is that reasonable?

marco-c:
> @marco-c Training is taking a lot of time for the component and tracking models, is that reasonable?

Yes, there are a lot of training examples for those two models.

bugbug/model.py
```diff
@@ -18,15 +18,15 @@
 
 
 class Model():
-    def __init__(self, lemmatization=False):
+    def __init__(self, lemmatization=False, ngrams=1):
```
marco-c: Not all models are using ngrams, so this shouldn't be an option of the base model class.

run.py
```diff
@@ -23,6 +23,7 @@
                         help='Goal of the classifier',
                         choices=['bug', 'regression', 'tracking', 'qaneeded', 'uplift', 'component', 'devdocneeded', 'defectfeaturetask'],
                         default='bug')
+    parser.add_argument('--ngrams', default='1')
```
marco-c: Not all classifiers support this, so I'm not a fan of adding the option in run.py.

ayush1999:
Could you tell me what kind of implementation do you have in mind regarding this? I assumed based on our conversation in the issue thread, that it'd be easier to pass this option to run.py.

marco-c:
> Could you tell me what kind of implementation do you have in mind regarding this? I assumed based on our conversation in the issue thread, that it'd be easier to pass this option to run.py.

I don't have an implementation in mind for this, I just know I don't like to add it in run.py.
You can just add the option to the models for now, and we can manually change it when we want to try different ngram numbers.

ayush1999:
Sure, I'll check the cases for these models where accuracy increases and add them.

ayush1999:
Some results:
```
# ngrams=1
>>> python run.py --train --goal=bug
X: (11012, 21998), y: (11012,)
Cross Validation scores:
Accuracy: f0.9528737862686614 (+/- 0.012266587420558554)
Precision: f0.9834396554490479 (+/- 0.004077661158072395)
Recall: f0.962142792039289 (+/- 0.014032244937659843)
X_train: (17276, 21998), y_train: (17276,)
X_test: (1102, 21998), y_test: (1102,)

# ngrams=2
>>> python run.py --train --goal=bug
X: (11012, 95178), y: (11012,)
Cross Validation scores:
Accuracy: f0.9532774191684084 (+/- 0.014276582067371067)
Precision: f0.9828806760304394 (+/- 0.005048220763169109)
Recall: f0.9631846597610929 (+/- 0.016588291686093024)
X_train: (17276, 95178), y_train: (17276,)
X_test: (1102, 95178), y_test: (1102,)
```

ayush1999:
@marco-c The results are just slightly better. Is it a good idea to add this feature?

marco-c:
> @marco-c The results are just slightly better. Is it a good idea to add this feature?

What are the results for the other models? I see you shared the results for the `bug` model and they didn't improve by much, but I'd like to see the other models too (the problems they aim to address are different, so the important features are different).

ayush1999:
Sure, some other results: (Sorry for the delay, took some time to train):
```
# ngrams = 1
>>> python run.py --goal=tracking --train
X: (95354, 60581), y: (60581, )
Cross Validation scores:
Accuracy: f0.926973319943
Precision: f0.329060105975
Recall: f0.928728414442

#ngrams = 2
>>> python run.py --train --goal=tracking
X: (95354, 377763), y: (95354,)
Cross Validation scores:
Accuracy: f0.9474003017905837 (+/- 0.004124596180095733)
Precision: f0.40747622722572946 (+/- 0.01934933264230578)
Recall: f0.915541601255887 (+/- 0.01164661663484547)

**Comment by ayush1999** [comment]
Sure, some other results:
```
# ngrams = 1
>>> python run.py --goal=tracking --train
X: (95354, 60581), y: (60581, )
Cross Validation scores:
Accuracy: f0.926973319943
Precision: f0.329060105975
Recall: f0.928728414442

#ngrams = 2
>>> python run.py --train --goal=tracking
X: (95354, 377763), y: (95354,)
Cross Validation scores:
Accuracy: f0.9474003017905837 (+/- 0.004124596180095733)
Precision: f0.40747622722572946 (+/- 0.01934933264230578)
Recall: f0.915541601255887 (+/- 0.01164661663484547)

marco-c:
That's a good improvement for the tracking model!

Can you try the component model too?

ayush1999:
The component model leads to a `memoryerror` and my terminal hangs. Should I try decreasing the train and test dataset for this?

marco-c:
> The component model leads to a memoryerror and my terminal hangs. Should I try decreasing the size of
train and test dataset for this?

Maybe you could just tweak min_df, or use max_features.
Reducing the datasets will not give useful results here.

ayush1999:
Sure, I tried doing that and it works now, though it take a considerable time. The results were pretty big, so I've uploaded them as two different files on a public gist: [here](https://gist.github.com/ayush1999/85f789e484a302624a3862f919d2bd05).

marco-c:
For the component model I don't see an improvement. We should enable cross-validation though, so we can actually see if there is an improvement or not. It will take 5x more than it took now, brace yourself!

ayush1999:
@marco-c  Should I add cv for the component model in this PR? I'm looking into ways of running these models on google colab, so I can leave them to train for hours without encountering any kind of error.

ayush1999:
BTW, why is cross validation disabled for the component model? (Any specific reason?)

marco-c:
It was disabled because it's really slow, so I couldn't quickly iterate. Now we could probably add it back (but not in this PR).

marco-c:
Note: you can add it to this PR for a while if it makes your life easier, we'll simply remove it before merging.

ayush1999:
Hmm, sure. Is it possible to run this model on the cloud? This will take around 6-7 hours on my system, and I couldn't find a resource for testing this out on colab (I'm still working on this). Maybe mozilla has access to some cloud resources/virtual machine, so I can test this there?

ayush1999:
(The commits made above are only for the purpose of making changes compatible with colab)

ayush1999:
Update: I finally got these models running on colab!

marco-c:
> Hmm, sure. Is it possible to run this model on the cloud? This will take around 6-7 hours on my system, and I couldn't find a resource for testing this out on colab (I'm still working on this). Maybe mozilla has access to some cloud resources/virtual machine, so I can test this there?

Yes, I need to figure out how to get you access. In the meantime, please use Colab.

ayush1999:
I've been trying to run the component model for a few days now, but the colab session always gets interrupted and needs to be reconnected and restarted. I guess it is probably due to the training time of the model exceeding the allowed running time per session on colab.

marco-c:
You could test it by replacing https://github.com/mozilla/bugbug/blob/master/.taskcluster.yml#L71 with `component`.

marco-c:
Now we're running on Taskcluster, so we don't have the limitations of Travis.

ayush1999:
Oh, cool. I'm exploring taskcluster right now. How am I supposed to run these tests using taskcluster?

marco-c:
The tasks defined in that file will run automatically when you update the PR.

ayush1999:
@marco-c Why are we using `"` instead of `'` as quotation for strings now? (It's causing flake8 to fail).

marco-c:
> @marco-c Why are we using " instead of ' as quotation for strings now? (It's causing flake8 to fail).

We've switched to [Black](https://github.com/ambv/black).

Note that the linting failure you have is not caused by " instead of ', see the logs of the linting task: https://tools.taskcluster.net/groups/N02jeGziTpicJLEluhxXKQ/tasks/MHzL3WlPSDaxvsp4Ip8ICQ/runs/0/logs/public%2Flogs%2Flive.log#L2361.

ayush1999:
Oh, I see. Can you tell me why the lint build is failing? Running the subprocess command results in exception. Am I missing something?

marco-c:
You can fix that by running `git fetch -all` and then `git push --tags`. Then it will work on your next push.

ayush1999:
Great, thanks, that test passes now (sorry for the inconvenience, I'm getting used to TaskCluster). Also, if I add the component model to the training test, in order to test the performance, what should I set the `maxruntime` as?

marco-c:
Try changing it to 8 hours, it should be hopefully enough.

ayush-1506:
@marco-c Unfortunately I lost access to my previous github account. (This is my new account). Please close this PR, I'll open a new fresh one.

PR #216, opened 2019-03-12
Fix #126

### LINKED ISSUE #126: Clean the test data for cleanup functions
The test data for cleanup functions (https://github.com/mozilla/bugbug/tree/master/tests/fixtures/bug_features) has a lot of redundant fields. Essentially, we'd need only the field that is being tested. For example, it'd be easier if each test element in `has_str.json` had only the `has_str` key. (And so on for other test files).

HSR05:
I beautified the code and now it makes sense to me. I can clearly see where the dicts. @ayush1999 Can you take up an example file and guide me about cleaning the redundant data. I will clean up the rest. Thankyou.

CHeers 👍


ayush1999:
Also, run flake8 locally before pushing to make sure the code is properly styled.

Issue #111, opened 2019-01-23
Split commit changes-related features into real code changes and test code changes


marco-c:
Right now we have single `added`, `deleted` and `types` values: https://github.com/mozilla/bugbug/blob/1784a10346d9dedcb2e18e076f3d482e39be93a1/bugbug/repository.py#L48.

We should have a `added` and `test_added`, `deleted` and `test_deleted`, and so on.

LakshyAAAgrawal:
Hey,
Could someone please provide a little more info regarding this?

marco-c:
At the moment, the `lines_added` feature is considering both real changes and test-only changes, we have to split the feature in two, `lines_added` and `test_lines_added`.
`lines_added` will be the number of lines added by the commit in Firefox code.
`test_lines_added` will be the number of lines added by the commit in testing code.

LakshyAAAgrawal:
Hey,
Really sorry to have got it wrong.
Should I instead check for the filename in the variable _path_ if it contains 'test' or not using regex, and if it does, then consider them to be test code changes?

PR #218, opened 2019-03-13
added 'test_added' and 'test_deleted' fields

LakshyAAAgrawal:
Added check based on "ever_backedout" to decide if it is a test commit or real.

Fixes https://github.com/mozilla/bugbug/issues/111

### DIFF
--- a/bugbug/repository.py
+++ b/bugbug/repository.py
@@ -8,6 +8,7 @@ import concurrent.futures
 import itertools
 import multiprocessing
 import os
+import re
 from collections import defaultdict
 from collections import namedtuple
 from datetime import datetime
@@ -50,7 +51,9 @@ def _transform(commit):
         'bug_id': int(commit.bug.decode('utf-8')) if commit.bug else None,
         'ever_backedout': commit.backedoutby != b'',
         'added': 0,
+        'test_added': 0,
         'deleted': 0,
+        'test_deleted': 0,
         'files_modified_num': 0,
         'types': set(),
         'components': list(),
@@ -67,8 +70,13 @@ def _transform(commit):
             obj['types'].add('binary')
             continue
 
-        obj['added'] += len(stats['added']) + len(stats['touched'])
-        obj['deleted'] += len(stats['deleted']) + len(stats['touched'])
+        if re.search(r'\/?[^\w](test|tests)\/', path):
+            obj['test_added'] += len(stats['added']) + len(stats['touched'])
+            obj['test_deleted'] += len(stats['deleted']) + len(stats['touched'])
+        else:
+            obj['added'] += len(stats['added']) + len(stats['touched'])
+            obj['deleted'] += len(stats['deleted']) + len(stats['touched'])
+
         ext = os.path.splitext(path)[1]
         if ext in ['.js', '.jsm']:
             type_ = 'JavaScript'
--- a/bugbug/repository.py
+++ b/bugbug/repository.py
@@ -40,6 +40,7 @@ def _init(repo_dir):
     HG = hglib.open(repo_dir)
 
 
+# This code was adapted from https://github.com/mozsearch/mozsearch/blob/2e24a308bf66b4c149683bfeb4ceeea3b250009a/router/router.py#L127
 def is_test(path):
     return ('/test/' in path or '/tests/' in path or '/mochitest/' in path or '/unit/' in path or '/gtest/'
             in path or 'testing/' in path or '/jsapi-tests/' in path or '/reftests/' in path or '/reftest/'
--- a/bugbug/repository.py
+++ b/bugbug/repository.py
@@ -8,7 +8,6 @@ import concurrent.futures
 import itertools
 import multiprocessing
 import os
-import re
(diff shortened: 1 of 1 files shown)


LakshyAAAgrawal:
Is this approach appropriate?

marco-c:
Mmmh, test changes are not changes made in backed out commits. Where did you get this idea? :smile:

marco-c:
Answering to https://github.com/mozilla/bugbug/issues/111#issuecomment-473938303:
> Hey,
> Really sorry to have got it wrong.
> Should I instead check for the filename in the variable path if it contains 'test' or not using regex, and if it does, then consider them to be test code changes?

You can check the test files here: https://searchfox.org/mozilla-central/search?q=&case=false&regexp=false&path=test
Try to make a rule out of this.
The best option might be to check if one of the path segments is 'test' or 'tests', but we have to check if there are many false positives.

LakshyAAAgrawal:
Hey,
I have added the test for 'test' or 'tests' segment in the 'path' variable. Could you tell me how to decide whether a file within these directories is a test or not, so that I could attempt to make a rule based on them?
Aren't all the files within these directories relevant to testing, and thus be counted as an addition to test?

marco-c:
Let's use this: https://github.com/mozsearch/mozsearch/blob/2e24a308bf66b4c149683bfeb4ceeea3b250009a/router/router.py#L127.

bugbug/repository.py
```diff
@@ -41,6 +41,15 @@ def _init(repo_dir):
 
 
 def _transform(commit):
+    def is_test(p):
```
marco-c: Move the function outside of _transform, rename `p` to `path`.
Add a link to the source code where you got the function.

bugbug/repository.py
```diff
@@ -41,6 +41,15 @@ def _init(repo_dir):
...
+        if '/unit/' in p:
+            return True
+        if 'test' not in p:
+            return False
```
marco-c: The first two ifs are probably unnecessary, I can't see when they would help

marco-c: Maybe they are optimizations, but I think we can remove them in our case.

LakshyAAAgrawal:
Hey,
I have added the is_test function from the [source](https://github.com/mozsearch/mozsearch/blob/2e24a308bf66b4c149683bfeb4ceeea3b250009a/router/router.py#L127) mentioned above.

However, I think I found an issue while making the said changes.
flake8 is testing against 'W504 line break after binary operator' as well as 'W503 line break before binary operator'.
According to the [PEP8 documentation](https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator), line break before binary operator should be allowed.

Therefore, I think W503 should be added in ignore, in the file [setup.cfg](https://github.com/mozilla/bugbug/blob/master/setup.cfg).
Should I do it?

marco-c:
I prefer keeping a vanilla configuration (with the line limit being the only exception). Could you fix the problem in the code instead?

LakshyAAAgrawal:
The current code is consistent with the existing flake8 check. But, I think it might lead to issues in the future with other operators like '+' in string concatenation.

bugbug/repository.py
marco-c: The second part of the comment still applies

marco-c:
> The current code is consistent with the existing flake8 check. But, I think it might lead to issues in the future with other operators like '+' in string concatenation.

What issues could it lead to?
The only issue I can think of is failing the linting check, but we can always update the formatting of the code to pass the linting check.

LakshyAAAgrawal:
Yes, I agree.
My only thought was that if there were multiple string concatenations like

```python
return('This_is_a_very_long_string_of_text'+'This_is_a_very_long_string_of_text'+'This_is_a_very_long_string_of_text'+'This_is_a_very_long_string_of_text'+'This_is_a_very_long_string_of_text'+'This_is_a_very_long_string_of_text'+'This_is_a_very_long_string_of_text')
```
It could be better typed as
```python
return('This_is_a_very_long_string_of_text'
       +'This_is_a_very_long_string_of_text'
       +'This_is_a_very_long_string_of_text'
       +'This_is_a_very_long_string_of_text'
       +'This_is_a_very_long_string_of_text'
       +'This_is_a_very_long_string_of_text'
       +'This_is_a_very_long_string_of_text')
```
However, I see the point with maintaining a vanilla configuration.

marco-c:
There is just one thing left to do, https://github.com/mozilla/bugbug/pull/218#discussion_r267319861.

bugbug/repository.py
LakshyAAAgrawal: I have added the link in the commit message:

>  Added function is_test to check whether a path is related to test or not. The conditions for is_test were obtained from https://github.com/mozsearch/mozsearch/blob/2e24a308bf66b4c149683bfeb4ceeea3b250009a/router/router.py#L127

Should I also add it as a comment in the code?

marco-c: Yes, just add a comment saying
# This code was adapted from XXX

Issue #221, opened 2019-03-14
need possible update in requirements.txt


SiluPanda:
while running the run.py after installing all packages in requirements.txt,
i got following error:

1. cannot import name '_validate_lengths' from 'numpy.lib.arraypad' (/home/silupanda/anaconda3/lib/python3.7/site-packages/numpy/lib/arraypad.py)

The is actually is a sklearn and numpy version issue which appears to be satisfied while running the requirements.txt.
reference of the issue: https://github.com/scikit-image/scikit-image/issues/3551

SiluPanda:
added a PR of a small change in requirement.txt https://github.com/mozilla/bugbug/pull/222 . @marco-c  kindly review

sbeesm:
Which version of `scikit-image` are you using? Try updating it to version `0.14.2` and the problem should be solved.

See #120 for reference.

SiluPanda:
i was using everything as it is in requirement.txt.
version is 0.20.3. 
i agree that this can be avoided with multiple ways, i did a PR #222 with one of the solutions.

marco-c:
When you install the requirements from our requirements file, you'll get scikit-image 0.14.2, so this issue doesn't really apply.

PR #222, opened 2019-03-14
small change in requirement.txt

SiluPanda:
added a small change to requitement.txt to avoid https://github.com/mozilla/bugbug/issues/221


marco-c:
See https://github.com/mozilla/bugbug/issues/221#issuecomment-473841869.

PR #229, opened 2019-03-21
Add number of previous commits touching the same modules as a feature

### LINKED ISSUE #113: Add number of previous commits touching the same modules as a feature
This is similar to https://github.com/mozilla/bugbug/commit/a002e0ce53080699a78cada08a95fd2355953c42#diff-fd96da16451dabb38a55be1721e8be9e, but for components.

It's a little harder than author or reviewer experience, as we need to parse the patch in order to get the components, and we're currently parsing the patches in multiple processes.

sbeesm:
Fixes #113.

### DIFF
--- a/bugbug/repository.py
+++ b/bugbug/repository.py
@@ -24,8 +24,9 @@ COMMITS_DB = 'data/commits.json'
 db.register(COMMITS_DB, 'https://www.dropbox.com/s/mz3afgncx0siijc/commits.json.xz?dl=1', 'v1')
 
 COMPONENTS = {}
+COMP_TOUCHED_PREV = {}
 
-Commit = namedtuple('Commit', ['node', 'author', 'desc', 'date', 'bug', 'backedoutby', 'author_email'])
+Commit = namedtuple('Commit', ['node', 'author', 'desc', 'date', 'bug', 'backedoutby', 'author_email', 'files', 'file_copies'])
 
 author_experience = {}
 author_experience_90_days = {}
@@ -66,6 +67,7 @@ def _transform(commit):
         'author_experience': author_experience[commit],
         'author_experience_90_days': author_experience_90_days[commit],
         'author_email': commit.author_email.decode('utf-8'),
+        'comp_touched_prev': 0,
     }
 
     patch = HG.export(revs=[commit.node], git=True)
@@ -105,11 +107,13 @@ def _transform(commit):
 
     obj['components'] = list(set('::'.join(COMPONENTS[fl]) for fl in patch_data.keys() if COMPONENTS.get(fl)))
 
+    obj['comp_touched_prev'] = COMP_TOUCHED_PREV[commit]
+
     return obj
 
 
 def _hg_log(revs):
-    template = '{node}\\0{author}\\0{desc}\\0{date}\\0{bug}\\0{backedoutby}\\0{author|email}\\0'
+    template = '{node}\\0{author}\\0{desc}\\0{date}\\0{bug}\\0{backedoutby}\\0{author|email}\\0{join(files,"|")}\\0{join(file_copies,"|")}\\0'
 
     args = hglib.util.cmdbuilder(b'log', template=template, no_merges=True, rev=revs[0] + b':' + revs[-1], branch='central')
     x = HG.rawcommand(args)
@@ -128,6 +132,8 @@ def _hg_log(revs):
             bug=rev[4],
             backedoutby=rev[5],
             author_email=rev[6],
+            files=rev[7],
+            file_copies=rev[8],
         ))
 
     return revs
@@ -215,6 +221,25 @@ def download_commits(repo_dir, date_from):
     r = requests.get('https://index.taskcluster.net/v1/task/gecko.v2.mozilla-central.latest.source.source-bugzilla-info/artifacts/public/components.json')
     r.raise_for_status()
     COMPONENTS = r.json()
+
+    components_touched = {}
+
+    for commit in commits:
+        comp = set('::'.join(COMPONENTS[fl]) for fl in commit.files.decode('utf-8').split('|') if fl in COMPONENTS)
+        for cp in comp:
+            if cp in components_touched:
+                components_touched[cp] += 1
+            else:
+                components_touched[cp] = 1
+
+        COMP_TOUCHED_PREV[commit] = sum(components_touched[cp]-1 for cp in comp)
+
+        if len(commit.file_copies) > 0:
+            for fl in commit.file_copies.decode('utf-8').split('|'):
+                cpy, src = fl[:fl.find('(')-1], fl[fl.find('(')+1:-1]
+
+                if cpy in COMPONENTS and src in COMPONENTS:
+                    components_touched['::'.join(COMPONENTS[cpy])] = components_touched['::'.join(COMPONENTS[src])]
 
     print(f'Mining commits using {multiprocessing.cpu_count()} processes...')
 
--- a/bugbug/repository.py
+++ b/bugbug/repository.py
@@ -126,13 +126,13 @@ def _hg_log(revs):
         dt = datetime.fromtimestamp(posixtime)
 
         file_copies = {}
-        for file_copy in rev[8].split(b'|'):
+        for file_copy in rev[8].decode('utf-8').split('|'):
             if not file_copy:
                 continue
 
-            parts = file_copy.split(b' (')
-            copied = parts[0].decode('utf-8')
-            orig = parts[1][:-1].decode('utf-8')
+            parts = file_copy.split(' (')
+            copied = parts[0]
+            orig = parts[1][:-1]
             file_copies[orig] = copied
 
         revs.append(Commit(
--- a/bugbug/repository.py
+++ b/bugbug/repository.py
@@ -8,7 +8,6 @@ import concurrent.futures
 import itertools
 import multiprocessing
 import os
-import time
 from collections import defaultdict
 from collections import namedtuple
 from datetime import datetime
@@ -24,14 +23,15 @@ from bugbug import db
 COMMITS_DB = 'data/commits.json'
 db.register(COMMITS_DB, 'https://www.dropbox.com/s/mz3afgncx0siijc/commits.json.xz?dl=1', 'v1')
 
-COMPONENTS = {}
-COMP_TOUCHED_PREV = defaultdict(int)
-COMP_TOUCHED_PREV_90_DAYS = defaultdict(int)
+path_to_component = {}
 
 Commit = namedtuple('Commit', ['node', 'author', 'desc', 'date', 'bug', 'backedoutby', 'author_email', 'files', 'file_copies'])
 
 author_experience = {}
 author_experience_90_days = {}
+
+components_touched_prev = defaultdict(int)
+components_touched_prev_90_days = defaultdict(int)
 
 
 def get_commits():
@@ -69,8 +69,8 @@ def _transform(commit):
         'author_experience': author_experience[commit.node],
         'author_experience_90_days': author_experience_90_days[commit.node],
         'author_email': commit.author_email.decode('utf-8'),
-        'comp_touched_prev': 0,
-        'comp_touched_prev_90_days': 0,
+        'components_touched_prev': components_touched_prev[commit.node],
+        'components_touched_prev_90_days': components_touched_prev_90_days[commit.node],
     }
 
     patch = HG.export(revs=[commit.node], git=True)
@@ -108,11 +108,7 @@ def _transform(commit):
     # Covert to a list, as a set is not JSON-serializable.
     obj['types'] = list(obj['types'])
 
-    obj['components'] = list(set(COMPONENTS[fl] for fl in patch_data.keys() if COMPONENTS.get(fl)))
-
-    obj['comp_touched_prev'] = COMP_TOUCHED_PREV[commit.node]
-
-    obj['comp_touched_prev_90_days'] = COMP_TOUCHED_PREV_90_DAYS[commit.node]
+    obj['components'] = list(set(path_to_component[path] for path in patch_data.keys() if path_to_component.get(path)))
 
     return obj
 
@@ -129,6 +125,16 @@ def _hg_log(revs):
         posixtime = float(rev[3].split(b'.', 1)[0])
         dt = datetime.fromtimestamp(posixtime)
 
+        file_copies = {}
+        for file_copy in rev[8].split(b'|'):
+            if not file_copy:
+                continue
+
+            parts = file_copy.split(b' (')
+            copied = parts[0].decode('utf-8')
+            orig = parts[1][:-1].decode('utf-8')
+            file_copies[orig] = copied
+
         revs.append(Commit(
             node=rev[0],
             author=rev[1],
@@ -138,7 +144,7 @@ def _hg_log(revs):
             backedoutby=rev[5],
             author_email=rev[6],
             files=rev[7].decode('utf-8').split('|'),
-            file_copies={fl[fl.find('(') + 1:-1]: fl[:fl.find('(') - 1] for fl in rev[8].decode('utf-8').split('|')} if len(rev[8]) > 0 else {},
+            file_copies=file_copies,
         ))
 
     return revs
@@ -199,8 +205,6 @@ def download_commits(repo_dir, date_from):
 
     global author_experience
     global author_experience_90_days
-
-    start = time.perf_counter()
     for commit in commits:
         author_experience[commit.node] = total_commits_by_author[commit.author]
         # We don't want to consider backed out commits when calculating author/reviewer experience.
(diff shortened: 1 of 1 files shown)


bugbug/repository.py
```diff
@@ -216,6 +222,25 @@ def download_commits(repo_dir, date_from):
...
+
+    for commit in commits:
+        comp = set('::'.join(COMPONENTS[fl]) for fl in commit.files.decode('utf-8').split('|') if fl in COMPONENTS)
+        for cp in comp:
+            if cp in components_touched:
+                components_touched[cp] += 1
+            else:
+                components_touched[cp] = 1
+
+        COMP_TOUCHED_PREV[commit] = sum(components_touched[cp]-1 for cp in comp)
```
rhcu: Hi. Just noticed failing Travis check. :)
You need to add whitespace around arithmetic operator on this line and on line 239 (in 2 places) to have it passed.

sbeesm: Hi! Thanks for pointing it out.

I have this thing in my notice. Actually I was waiting for @marco-c  to review this PR. So that if there'd be any other changes, I'll fix this missing whitespace along with those so as to avoid some extra commits for minor syntax changes.

bugbug/repository.py
```diff
@@ -128,6 +132,8 @@ def _hg_log(revs):
             bug=rev[4],
             backedoutby=rev[5],
             author_email=rev[6],
+            files=rev[7],
+            file_copies=rev[8],
```
marco-c: Parse them here directly (decode and split).
file_copies can be a map (from original path to copied path).

bugbug/repository.py
```diff
@@ -216,6 +222,25 @@ def download_commits(repo_dir, date_from):
     r.raise_for_status()
     COMPONENTS = r.json()
 
+    components_touched = {}
+
+    for commit in commits:
+        comp = set('::'.join(COMPONENTS[fl]) for fl in commit.files.decode('utf-8').split('|') if fl in COMPONENTS)
+        for cp in comp:
```
marco-c: You can increase COMP_TOUCHED_PREV[commit] in this loop too, so you don't have to add a second loop.

bugbug/repository.py
```diff
@@ -216,6 +222,25 @@ def download_commits(repo_dir, date_from):
     r.raise_for_status()
     COMPONENTS = r.json()
 
+    components_touched = {}
+
+    for commit in commits:
+        comp = set('::'.join(COMPONENTS[fl]) for fl in commit.files.decode('utf-8').split('|') if fl in COMPONENTS)
```
marco-c: Since we are repeating `'::'.join(COMPONENTS[fl])` all the time, maybe we should just do this once at the beginning by transforming COMPONENTS

marco-c:
Looks good, just a few comments.

There is one thing missing though, we want the same value for the past 90 days window too (just like for author and reviewer experience).

bugbug/repository.py
marco-c: When you notice some issue, feel free to fix it right away. It's OK if there are multiple commits, indeed it's better to have one commit for each thing you do.

sbeesm: Okay.

sbeesm:
> There is one thing missing though, we want the same value for the past 90 days window too (just like for author and reviewer experience).

For this, can we have a list of commits for the previous 90 days from a particuar commit?
Then we'd only check for components from these commits:

```
prev_commits_90_days = []

for commit in commits:
    cut = None

    for i, prev_commit in enumerate(prev_commits_90_days):
        if (commit.date - prev_commit.date).days <= 90:
            break

            cut = i

    if cut is not None:
        prev_commits_90_days = prev_commits_90_days[cut + 1:]
```

But then won't we have to repeat the complete process again (in a nested loop) to find the components again and check for file_copies and then populate the list for components touched for previous 90 days?

marco-c:
> For this, can we have a list of commits for the previous 90 days from a particuar commit?

Let's try, we need to check if the memory usage grows too much.

sbeesm:
Right now, to calculate the values for past 90 days, we are just repeating the same steps again. Any suggestions how we might improve on that?

And should we calculate the values in the same loop in which we are calculating the `author_experience`? This way we can avoid looping over commits multiple times.

marco-c:
> Right now, to calculate the values for past 90 days, we are just repeating the same steps again. Any suggestions how we might improve on that?

> And should we calculate the values in the same loop in which we are calculating the author_experience? This way we can avoid looping over commits multiple times.

The part after that which parses all patches takes several hours, so optimizing this is only worth if it takes a long time. How long does it take currently? (if you return early before "Mining commits ...").

sbeesm:
> How long does it take currently? (if you return early before "Mining commits ...").

Should I mine all the commits to get the time?

Here are some results for 1000 commits. The time for execution is less every time I run the script. Is it somehow related to cache memory?

```
========================================
2019-03-29 20:49:36 - Start Program
========================================
Mining 1024 commits using 4 processes...
100%|█████████████████████████████████████████████████████████████████████████████████████| 4/4 [01:28<00:00, 22.16s/it]
Analyzing 957 patches...
========================================
2019-03-29 20:51:44 - End Program
Elapsed time: 0:02:07.184135
========================================


(base) sbeesm@LAPTOP-6V5RSGD0:/mnt/f/COMPILED/Git/bugbug$ python bugbug/repository.py /mnt/f/COMPILED/Mercurial/mozilla-central
========================================
2019-03-29 20:55:04 - Start Program
========================================
Mining 1024 commits using 4 processes...
100%|█████████████████████████████████████████████████████████████████████████████████████| 4/4 [00:16<00:00,  4.06s/it]
Analyzing 957 patches...
========================================
2019-03-29 20:55:58 - End Program
Elapsed time: 0:00:54.060148
========================================


(base) sbeesm@LAPTOP-6V5RSGD0:/mnt/f/COMPILED/Git/bugbug$ python bugbug/repository.py /mnt/f/COMPILED/Mercurial/mozilla-central
========================================
2019-03-29 20:56:55 - Start Program
========================================
Mining 1024 commits using 4 processes...
100%|█████████████████████████████████████████████████████████████████████████████████████| 4/4 [00:15<00:00,  3.91s/it]
Analyzing 957 patches...
========================================
2019-03-29 20:57:43 - End Program
Elapsed time: 0:00:48.016388
========================================

```

marco-c:
> Should I mine all the commits to get the time?

Yes, to have an idea of how much overhead the code you're adding has and if it's worth optimizing it.

Measure the first loop (the block of code which calculates the author/reviewer experience) and your loop (the block of code you added, which calculates the component stuff). You can measure them by using https://docs.python.org/3/library/time.html#time.process_time.

```Python
start = time.process_time()
# AUTHOR/REVIEWER
for commit in commits:
    ....
print('Time: {}'.format(time.process_time() - start))

start = time.process_time()
# COMPONENT STUFF
for commit in commits:
    ....
print('Time: {}'.format(time.process_time() - start))
```

And it'd also be nice if you can measure the last part too:
```Python
start = time.process_time()
with concurrent.futures....:
    ...
print('Time: {}'.format(time.process_time() - start))
```

sbeesm:
The components loop roughly takes double the time taken by the author loop.

And unfortunately while mining commits, this error occurred:

```
Mining 141863 commits using 4 processes...
100%|███████████████████████████████████████████████████████████████████████████████| 555/555 [1:52:41<00:00, 12.18s/it]Analyzing 124586 patches...

Time (author): 0.40625
Time (component): 0.875

Mining commits using 4 processes...
 89%|███████████████████████████████████████████████████████████████▍       | 111312/124586 [3:52:03<3:28:02,  1.06it/s]concurrent.futures.process._RemoteTraceback:
"""
  File "/home/sbeesm/miniconda3/lib/python3.7/concurrent/futures/process.py", line 232, in _process_worker
    r = call_item.fn(*call_item.args, **call_item.kwargs)
  File "/home/sbeesm/miniconda3/lib/python3.7/concurrent/futures/process.py", line 191, in _process_chunk
    return [fn(*args) for args in chunk]
  File "/home/sbeesm/miniconda3/lib/python3.7/concurrent/futures/process.py", line 191, in <listcomp>
    return [fn(*args) for args in chunk]
  File "bugbug/repository.py", line 78, in _transform
    patch = HG.export(revs=[commit.node], git=True)
  File "/home/sbeesm/miniconda3/lib/python3.7/site-packages/hglib/client.py", line 818, in export
    out = self.rawcommand(args)
  File "/home/sbeesm/miniconda3/lib/python3.7/site-packages/hglib/client.py", line 263, in rawcommand
    raise error.CommandError(args, ret, out, err)
hglib.error.CommandError: (255, b"# HG changeset patch\n# User Jan Varga <jan.varga@gmail.com>\n# Date 1543524541 -3600\n#      Thu Nov 29 21:49:01 2018 +0100\n# Node ID eba8447d393532c4f429c84a14c9a3a3acb5c447\n# Parent  1309fe77cfaa7237aeb193db5b3544f7c2a0f6f0\nBug 1286798 - Part 33: Restrict localStorage from being available on some pages; r=asuth\n\nThis matches the old implementation. localStorage shouldn't be available on some pages, for example about:home.", b'abort: Bad address')
"""

The above exception was the direct cause of the following exception:

  File "bugbug/repository.py", line 323, in <module>
    download_commits(args.repository_dir, two_years_and_six_months_ago)
  File "bugbug/repository.py", line 294, in download_commits
    db.write(COMMITS_DB, commits)
  File "/mnt/f/COMPILED/Git/bugbug/bugbug/db.py", line 136, in write
    store.write(elems)
  File "/mnt/f/COMPILED/Git/bugbug/bugbug/db.py", line 58, in write
    for elem in elems:
  File "/home/sbeesm/miniconda3/lib/python3.7/site-packages/tqdm/_tqdm.py", line 1022, in __iter__
    for obj in iterable:
  File "/home/sbeesm/miniconda3/lib/python3.7/concurrent/futures/process.py", line 476, in _chain_from_iterable_of_lists    for element in iterable:
  File "/home/sbeesm/miniconda3/lib/python3.7/concurrent/futures/_base.py", line 586, in result_iterator
    yield fs.pop().result()
  File "/home/sbeesm/miniconda3/lib/python3.7/concurrent/futures/_base.py", line 425, in result
    return self.__get_result()
  File "/home/sbeesm/miniconda3/lib/python3.7/concurrent/futures/_base.py", line 384, in __get_result
    raise self._exception
hglib.error.CommandError: (255, b"# HG changeset patch\n# User Jan Varga <jan.varga@gmail.com>\n# Date 1543524541 -3600\n#      Thu Nov 29 21:49:01 2018 +0100\n# Node ID eba8447d393532c4f429c84a14c9a3a3acb5c447\n# Parent  1309fe77cfaa7237aeb193db5b3544f7c2a0f6f0\nBug 1286798 - Part 33: Restrict localStorage from being available on some pages; r=asuth\n\nThis matches the old implementation. localStorage shouldn't be available on some pages, for example about:home.", b'abort: Bad address')
```
Any idea what might be wrong?

marco-c:
Never seen that error before, try again, maybe it was a temporary issue.

Also, use `time.perf_counter()` instead of `time.process_time()` (so we consider elapsed time during sleep too).

sbeesm:
I tried running it again, but the program stops at same point of time.
But this time it shows a different error, something related to extensions:

```
Mining 141278 commits using 4 processes...
100%|███████████████████████████████████████████████████████████████████████████████| 552/552 [1:54:03<00:00, 12.40s/it]Analyzing 124043 patches...

Time (author): 1.020003600000564
Time (component): 2.846416600000339

Mining commits using 4 processes...
 89%|██████████████████████████████████████████████████████████████▉        | 109953/124043 [4:16:45<2:49:05,  1.39it/s]concurrent.futures.process._RemoteTraceback:
"""
  File "/home/sbeesm/miniconda3/lib/python3.7/concurrent/futures/process.py", line 232, in _process_worker
    r = call_item.fn(*call_item.args, **call_item.kwargs)
  File "/home/sbeesm/miniconda3/lib/python3.7/concurrent/futures/process.py", line 191, in _process_chunk
    return [fn(*args) for args in chunk]
  File "/home/sbeesm/miniconda3/lib/python3.7/concurrent/futures/process.py", line 191, in <listcomp>
    return [fn(*args) for args in chunk]
  File "bugbug/repository.py", line 78, in _transform
    patch = HG.export(revs=[commit.node], git=True)
  File "/home/sbeesm/miniconda3/lib/python3.7/site-packages/hglib/client.py", line 818, in export
    out = self.rawcommand(args)
  File "/home/sbeesm/miniconda3/lib/python3.7/site-packages/hglib/client.py", line 258, in rawcommand
    ret = self.runcommand(args, inchannels, outchannels)
  File "/home/sbeesm/miniconda3/lib/python3.7/site-packages/hglib/client.py", line 192, in runcommand
    channel, data = self._readchannel()
  File "/home/sbeesm/miniconda3/lib/python3.7/site-packages/hglib/client.py", line 153, in _readchannel
    % (ret, serr.strip()))
hglib.error.ServerError: server exited with status -11: b'*** failed to import extension pushlog from /mnt/c/Users/HP/.mozbuild/version-control-tools/hgext/pushlog: No module named utils\n*** failed to import extension hgmo from /mnt/c/Users/HP/.mozbuild/version-control-tools/hgext/hgmo: No module named utils'
"""

The above exception was the direct cause of the following exception:

  File "bugbug/repository.py", line 352, in <module>
    download_commits(args.repository_dir, two_years_and_six_months_ago)
  File "bugbug/repository.py", line 296, in download_commits
    db.write(COMMITS_DB, commits)
  File "/mnt/f/COMPILED/Git/bugbug/bugbug/db.py", line 136, in write
    store.write(elems)
  File "/mnt/f/COMPILED/Git/bugbug/bugbug/db.py", line 58, in write
    for elem in elems:
  File "/home/sbeesm/miniconda3/lib/python3.7/site-packages/tqdm/_tqdm.py", line 1022, in __iter__
    for obj in iterable:
  File "/home/sbeesm/miniconda3/lib/python3.7/concurrent/futures/process.py", line 476, in _chain_from_iterable_of_lists    for element in iterable:
  File "/home/sbeesm/miniconda3/lib/python3.7/concurrent/futures/_base.py", line 586, in result_iterator
    yield fs.pop().result()
  File "/home/sbeesm/miniconda3/lib/python3.7/concurrent/futures/_base.py", line 425, in result
    return self.__get_result()
  File "/home/sbeesm/miniconda3/lib/python3.7/concurrent/futures/_base.py", line 384, in __get_result
    raise self._exception
hglib.error.ServerError: server exited with status -11: b'*** failed to import extension pushlog from /mnt/c/Users/HP/.mozbuild/version-control-tools/hgext/pushlog: No module named utils\n*** failed to import extension hgmo from /mnt/c/Users/HP/.mozbuild/version-control-tools/hgext/hgmo: No module named utils'
```

But I've already added the extensions to the file. And this error occurs only after mining 89% of commits but the previous ones run fine. If the error was really related to extensions, I wouldn't be able to analyse commits in the first place (for _hg_log) right?

marco-c:
Since it ends at 89% (and so it's almost done), you could just add a try except around the "Mining commits" phase and print the time in the except.

Can you commit the WIP code with the timing, I'd like to see it.

marco-c:
OK, looks good, let's see what the numbers are now.

sbeesm:
Time (author): 1.108501200000319

Time (component): 3.6927001000003656

Exception Time: 14591.4676359

marco-c:
OK, so the difference is so big that we don't have to attempt to optimize the component loop!

sbeesm:
Should we try calculating author experience and components in one single loop?

marco-c:
No it's not worth it. The time it takes for them is so little compared to the rest, that it's better to have readable code rather than a minuscule improvement.

sbeesm:
Okay.

So do I need to make any additional changes? Or should I commit the same code as before(without timing)?

bugbug/repository.py
```diff
@@ -24,8 +25,10 @@
 db.register(COMMITS_DB, 'https://www.dropbox.com/s/mz3afgncx0siijc/commits.json.xz?dl=1', 'v1')
 
 COMPONENTS = {}
+COMP_TOUCHED_PREV = defaultdict(int)
```
marco-c: Since these are not constants, make them lowercase (and move them right after the author_experience dicts).

bugbug/repository.py
```diff
@@ -63,9 +66,11 @@ def _transform(commit):
         'files_modified_num': 0,
         'types': set(),
         'components': list(),
-        'author_experience': author_experience[commit],
-        'author_experience_90_days': author_experience_90_days[commit],
+        'author_experience': author_experience[commit.node],
+        'author_experience_90_days': author_experience_90_days[commit.node],
         'author_email': commit.author_email.decode('utf-8'),
+        'comp_touched_prev': 0,
+        'comp_touched_prev_90_days': 0,
```
marco-c: Here you shouldn't use 0, but the actual values :)

bugbug/repository.py
```diff
@@ -103,13 +108,17 @@ def _transform(commit):
     # Covert to a list, as a set is not JSON-serializable.
     obj['types'] = list(obj['types'])
 
-    obj['components'] = list(set('::'.join(COMPONENTS[fl]) for fl in patch_data.keys() if COMPONENTS.get(fl)))
+    obj['components'] = list(set(COMPONENTS[fl] for fl in patch_data.keys() if COMPONENTS.get(fl)))
+
+    obj['comp_touched_prev'] = COMP_TOUCHED_PREV[commit.node]
+
+    obj['comp_touched_prev_90_days'] = COMP_TOUCHED_PREV_90_DAYS[commit.node]
```
marco-c: And you can remove these two, as you put them directly in the code above.

bugbug/repository.py
```diff
@@ -128,6 +137,8 @@ def _hg_log(revs):
             bug=rev[4],
             backedoutby=rev[5],
             author_email=rev[6],
+            files=rev[7].decode('utf-8').split('|'),
+            file_copies={fl[fl.find('(') + 1:-1]: fl[:fl.find('(') - 1] for fl in rev[8].decode('utf-8').split('|')} if len(rev[8]) > 0 else {},
```
marco-c: It'd be better to generate this with a loop earlier, so it is more readable, something along these lines (adding decode/adding b/removing b when necessary):
```Python
        file_copies = {}
        for file_copy in rev[8].split(b'|'):
            if not file_copy:
                continue

            parts = file_copy.split(b' (')
            copied = parts[0]
            orig = parts[1][:-1]
            file_copies[orig] = copied
```

bugbug/repository.py
```diff
@@ -206,22 +219,81 @@ def download_commits(repo_dir, date_from):
...
             commits_by_author[commit.author].append(commit)
 
+    print()
+    print('Time (author): {}'.format(time.perf_counter() - start))
+
     global COMPONENTS
     r = requests.get('https://index.taskcluster.net/v1/task/gecko.v2.mozilla-central.latest.source.source-bugzilla-info/artifacts/public/components.json')
     r.raise_for_status()
     COMPONENTS = r.json()
+    COMPONENTS = {fl: '::'.join(cmp) for fl, cmp in COMPONENTS.items()}
```
marco-c: Nit: rename `cmp` to `component`, as `cmp` is a reserved keyword.

Use `path` instead of `fl`, here and elsewhere.

bugbug/repository.py
```diff
@@ -206,22 +219,81 @@ def download_commits(repo_dir, date_from):
...
+
+    global COMP_TOUCHED_PREV
+    global COMP_TOUCHED_PREV_90_DAYS
+
+    components_touched = defaultdict(int)
+    prev_commits_90_days = []
+
+    start = time.perf_counter()
+    for commit in commits:
+        comp = set(COMPONENTS[fl] for fl in commit.files if fl in COMPONENTS)
```
marco-c: Rename `comp` to `components`

bugbug/repository.py
```diff
@@ -206,22 +219,81 @@ def download_commits(repo_dir, date_from):
...
+
+        for cp in comp:
```
marco-c: Rename `cp` to `components`.

bugbug/repository.py
```diff
@@ -24,8 +25,10 @@
 db.register(COMMITS_DB, 'https://www.dropbox.com/s/mz3afgncx0siijc/commits.json.xz?dl=1', 'v1')
 
 COMPONENTS = {}
```
marco-c: Rename this to `path_to_component`, so the rest of the code gets more readable (since you are adding many variables with similar names).

marco-c:
> So do I need to make any additional changes? Or should I commit the same code as before(without timing)?

I added a few comments, and yes please also remove the timing stuff.

bugbug/repository.py
```diff
@@ -129,6 +125,16 @@ def _hg_log(revs):
         posixtime = float(rev[3].split(b'.', 1)[0])
         dt = datetime.fromtimestamp(posixtime)
 
+        file_copies = {}
+        for file_copy in rev[8].split(b'|'):
```
marco-c: decode here, so you can avoid decoding twice later (and remove all the `b`)

PR #233, opened 2019-03-25
Add feature importance visualization for multiclass problems

### LINKED ISSUE #162: Make get_feature_importances work for multiclass problems
Right now `get_feature_importances` fail for multiclass problems (e.g. component).

yatinmaan:
Fixes #162

### DIFF
--- a/bugbug/model.py
+++ b/bugbug/model.py
@@ -41,26 +41,56 @@ class Model():
         return []
 
     def get_important_features(self, cutoff, shap_values):
-        # Calculate the values that represent the fraction of the model output variability attributable
-        # to each feature across the whole dataset.
-        shap_sums = shap_values.sum(0)
-        abs_shap_sums = np.abs(shap_values).sum(0)
-        rel_shap_sums = abs_shap_sums / abs_shap_sums.sum()
+        # shap_values is a list for models with vector output
+        if isinstance(shap_values, list):
+            shap_values = np.array(shap_values)
+            class_shap_values = []
 
-        cut_off_value = cutoff * np.amax(rel_shap_sums)
+            for matrix in shap_values:
+                shap_sum = (np.abs(matrix)).sum(0)
+                class_shap_values.append(shap_sum)
 
-        # Get indices of features that pass the cut off value
-        top_feature_indices = np.where(rel_shap_sums >= cut_off_value)[0]
-        # Get the importance values of the top features from their indices
-        top_features = np.take(rel_shap_sums, top_feature_indices)
-        # Gets the sign of the importance from shap_sums as boolean
-        is_positive = (np.take(shap_sums, top_feature_indices)) >= 0
-        # Stack the importance, indices and shap_sums in a 2D array
-        top_features = np.column_stack((top_features, top_feature_indices, is_positive))
-        # Sort the array (in decreasing order of importance values)
-        top_features = top_features[top_features[:, 0].argsort()][::-1]
+            # Importance of features (sum of all classes)
+            feature_shap_vals = np.array(class_shap_values).sum(0)
+            sorted_feature_indices = np.argsort(feature_shap_vals)[::-1]
+            # features: matrix of [features x classes] sorted by importance of features
+            features = (np.array(class_shap_values).T)[sorted_feature_indices]
 
-        return top_features
+            cut_off_value = cutoff * np.amax(feature_shap_vals)
+            # Gets the number of features that pass the cut off value
+            cut_off_len = len(np.where(feature_shap_vals >= cut_off_value)[0])
+
+            # Stack the original indices and feature importance along with the features
+            # above the cut off length
+            top_features = np.column_stack((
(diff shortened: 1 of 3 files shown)


yatinmaan:
@marco-c [self.clf.fit(X_train, y_train)](https://github.com/mozilla/bugbug/blob/947b695fb477bca7a5924b27beeb187658ca1044/bugbug/model.py#L138) crashes for me with a `std:bad_alloc` error I set [calculate_importance](https://github.com/mozilla/bugbug/blob/947b695fb477bca7a5924b27beeb187658ca1044/bugbug/models/component.py#L45) to `True`

Any idea what might be the cause?

marco-c:
This is an out of memory error, to test your patch I suggest reducing the size of the dataset (just modify https://github.com/mozilla/bugbug/blob/937e889c517b040158ae2de0385db672114d2126/bugbug/bugzilla.py#L42 to only return a subset of bugs by using itertools.islice).

yatinmaan:
@marco-c Updated PR, its ready for review now.

> This is an out of memory error, to test your patch I suggest reducing the size of the dataset (just modify
> 
> [bugbug/bugbug/bugzilla.py](https://github.com/mozilla/bugbug/blob/937e889c517b040158ae2de0385db672114d2126/bugbug/bugzilla.py#L42)
> 
> Line 42 in [937e889](/mozilla/bugbug/commit/937e889c517b040158ae2de0385db672114d2126)
>  return db.read(BUGS_DB) 
> to only return a subset of bugs by using itertools.islice).

I tried that, it made the training faster, but still ended up with the same error.

Using smaller `test_size` and `train_size` in [train_test_split](https://github.com/mozilla/bugbug/blob/19ab979206ead0d19b7053968fc2a48a4e119e70/bugbug/model.py#L112) is what finally (sort of) fixed it.

bugbug/model.py
```diff
@@ -113,12 +143,26 @@ def bugs():
             explainer = shap.TreeExplainer(self.clf)
             shap_values = explainer.shap_values(X_train)
 
-            # TODO: Actually implement feature importance visualization for multiclass problems.
-            if isinstance(shap_values, list):
-                shap_values = np.sum(np.abs(shap_values), axis=0)
-
             important_features = self.get_important_features(importance_cutoff, shap_values)
-
+            if isinstance(shap_values, list):
```
marco-c: Let's move the printing into a function, so we can use it here and in run.py

bugbug/model.py
```diff
@@ -41,26 +41,56 @@ def get_feature_names(self):
...
+            # above the cut off length
+            top_features = np.column_stack((
+                sorted_feature_indices[:cut_off_len],
+                feature_shap_vals[sorted_feature_indices][:cut_off_len],
+                features[:cut_off_len]
+            ))
+
+            return top_features
+
+        else:
```
marco-c: Couldn't the one-level case be handled by doing something like `shap_values = [shap_values]`, so that we share the same code for single-class and multiclass problems?

yatinmaan: - Single class importances are sorted according to relative shap sums, multi-class uses just shap sums

 - In single class case, we can effectively determine the sign of the importance value whereas in multi-class we use sum of absolute values

So I thought it would be better to leave single-class as it is, but I can merge them if you want.

marco-c: Couldn't we determine the sign in the multiclass case?

We can sort by absolute value to have the most important features to show, but then we can show the sign.

run.py
```diff
@@ -76,10 +76,30 @@
...
-
                 feature_names = model.get_feature_names()
-                for i, (importance, index, is_positive) in enumerate(importances):
-                    print(f'{i + 1}. \'{feature_names[int(index)]}\' ({"+" if (is_positive) else "-"}{importance})')
+
+                if len(importances.shape) < 2:
+                    for i, (importance, index, is_positive) in enumerate(importances):
+                        print(f'{i + 1}. \'{feature_names[int(index)]}\' ({"+" if (is_positive) else "-"}{importance})')
+                else:
+                    # Number of classes to be printed
```
marco-c: Maybe it would be best here to only print the features which contributed to the selection of the predicted class.

marco-c: What an user wants to know here is "why did the classifier select this class?"

bugbug/model.py
marco-c: |  | Feature 1 (most important between all classes) | Feature 2 (second most important between all classes) |
| ------------- |:-------------:|:-------------:|
|  **Class1** | +1,3 | -1,2 |
| **Class2** | +0,01 | +7,9 |
|  **Class3** | +99,3 | -11,2 |

marco-c:
@yatinmaan are you interested in finishing this up? :)

marco-c:
@yatinmaan are you interested in finishing this up?
I'm going to close the PR otherwise.

marco-c:
Closing because the author is not responsive.

Issue #26, opened 2018-12-06
Create a "similar bug" recommender


marco-c:
"similar" not in terms of duplicate bug, but in terms of features. E.g. two bugs that are completely unrelated from each other, but that are of the same type and have similarly sized patches.

spectre1811:
It means we have to build a kind of recommender system based upon the features and the size of the patches extracted from the bug info.?

marco-c:
> It means we have to build a kind of recommender system based upon the features and the size of the patches extracted from the bug info.?

Yes (note that the size of the patches is just one of the many features that we can consider).

spectre1811:
Yeah i figured that thing. So can i work upon this thing as my gsoc project because i think this would take some time to be completed

marco-c:
Yes, this is a big enough project for a GSoC.

spectre1811:
I want to work on this. Can i be assigned this?

marco-c:
@spectre1811 see the contributing rules in README, an issue is assigned only when there is a PR to fix it.

marco-c:
See also #517.

PR #238, opened 2019-03-31
Applying document similarity to bug summaries

chidauri:


marco-c:
We need a way to evaluate the results, or we can't compare different solutions.

One possibility is to get groups of bugs (each group would be composed of a bug and all bugs that are duplicates of it), and check that the tool would find, for each bug in the group, all other members of the group.

This is not a great measure though because, for a given bug, there could be bugs that were missed by humans but that are actually duplicates of that bug. So, both the recall and precision can't be properly calculated.

Try to do some research and find how other people have approached this, by reading some publications about bug duplicate detection.

If we don't find a better way to measure, then we'll do what I proposed and just assume that the human duplicate detection was perfect.

marco-c:
As discussed on IRC, we'll consider the dataset as being clean (so all duplicates as being really duplicates, and no duplicates outside of those detected by humans).

marco-c:
@chidauri are you still interested in working on this?
I'm going to close the PR otherwise.

chidauri:
@marco-c  i was a little busy, i will continue this.

chidauri:
@marco-c seems like duplicate classifier is already being worked upon , so i can help in that or pick up any other issue, whatever you say, you can close this PR.

marco-c:
@chidauri you could either work on #26, which is similar (but not based on textual features but on other features), or help with #517 (which is the work you started to do in this bug).

chidauri:
@marco-c can't this be used as it is in #517 ? I'll make it return a list of bug ids of top x bugs.

marco-c:
@chidauri yep! We'll also need some code to evaluate the results, as discussed earlier.

marco-c:
TBC in #547.

PR #242, opened 2019-04-03
Add assignee model to run.py

### LINKED ISSUE #241: Add assignee model to run.py
See also #61.

rhcu:
Fixes #241


marco-c:
Looks good, but you also need to add it to the arg choices.

PR #246, opened 2019-04-05
Use bug type as a label, when possible

### LINKED ISSUE #240: Use task/enhancement/defect data from Bugzilla as labels
We have recently (today) introduced a new field on Bugzilla that specifies whether a bug is a task, a enhancement, or a defect.
We can use this data as labels for the bug and the defect/feature/task models.

marco-c:
Fixes #240


marco-c:
Merged manually.

PR #254, opened 2019-04-09
Import Trainer class from release-services repository

Lothiraldan:
This basically import the `trainer.py` file from the `release-services`
repository at hash 77cdddd. I removed imports and reference to cli-common
helpers that will likely needs to be reimplemented, like the raven support.

Also defines 4 docker images, one per model to train.

### DIFF
--- a/scripts/trainer.py
+++ b/scripts/trainer.py
@@ -4,7 +4,6 @@ import argparse
 import lzma
 import os
 import shutil
-from datetime import datetime, timedelta
 from logging import INFO, basicConfig, getLogger
 from urllib.request import urlretrieve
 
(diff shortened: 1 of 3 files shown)


scripts/trainer.py
```diff
@@ -0,0 +1,107 @@
...
+
+from bugbug.models.component import ComponentModel
+from bugbug.models.defect_enhancement_task import DefectEnhancementTaskModel
+from bugbug.models.regression import RegressionModel
+from bugbug.models.tracking import TrackingModel
+
+basicConfig(level=INFO)
+logger = getLogger(__name__)
+
+BASE_URL = "https://index.taskcluster.net/v1/task/project.releng.services.project.testing.bugbug_data.latest/artifacts/public"
```
marco-c: We should make this configurable, I'll file an issue

marco-c: Filed #256.

PR #261, opened 2019-04-10
Add option to analyze old bugs

### LINKED ISSUE #248: Create an option in bug model for analyzing old bugs
Details are here: https://github.com/mozilla/bugbug/issues/244#issuecomment-480284787 and https://github.com/mozilla/bugbug/pull/247#issuecomment-480756156

rhcu:
Fixes #248. I have started adding the option. Should I add it to all models or only to Bug model? 
I have not finished yet the patch, so it is failing. :)

### DIFF
--- a/bugbug/model.py
+++ b/bugbug/model.py
@@ -17,7 +17,7 @@ from bugbug.nlp import SpacyVectorizer
 
 
 class Model:
-    def __init__(self, lemmatization=False, old=False):
+    def __init__(self, lemmatization=False, historical=False):
         if lemmatization:
             self.text_vectorizer = SpacyVectorizer
         else:
--- a/bugbug/model.py
+++ b/bugbug/model.py
@@ -17,7 +17,7 @@ from bugbug.nlp import SpacyVectorizer
 
 
 class Model:
-    def __init__(self, lemmatization=False, historical=False):
+    def __init__(self, lemmatization=False):
         if lemmatization:
             self.text_vectorizer = SpacyVectorizer
         else:
--- a/bugbug/model.py
+++ b/bugbug/model.py
@@ -17,7 +17,7 @@ from bugbug.nlp import SpacyVectorizer
 
 
 class Model:
-    def __init__(self, lemmatization=False):
+    def __init__(self, lemmatization=False, old=False):
         if lemmatization:
             self.text_vectorizer = SpacyVectorizer
         else:
--- a/bugbug/models/bug.py
+++ b/bugbug/models/bug.py
@@ -42,8 +42,10 @@ class BugModel(Model):
             bug_features.product(),
             bug_features.component(),
         ]
-        if (historical is True):
+
+        if (historical):
             feature_extractors.append(bug_features.had_severity_enhancement())
+
         cleanup_functions = [
             bug_features.cleanup_url,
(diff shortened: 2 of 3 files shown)


marco-c:
Yeah, let's add it only to the Bug model for now.
Call it `historical` though, nicer than `old` :P

rhcu:
Sorry, I don't understand why linting fails. I checked the code locally with flake8, and there were no errors. Can you help to find to a problem?

marco-c:
@rhcu sorry I was messing with CI, you'll need to rebase on top of current master.

rhcu:
I rebased but there is still an error, and I cannot find it in Taskcluster's log.

marco-c:
Could you try rebasing on top of master again?

rhcu:
Same after rebasing. The following error is shown:
Screen Shot 2019-04-11 at 1 37 38 PM

marco-c:
This is strange... Could you try removing your fork and forking again?

marco-c:
Or, try running `git fetch --all` and then `git push --tags`.

rhcu:
> Or, try running `git fetch --all` and then `git push --tags`.

Done. Also, updated branch to solve the merge conflicts.

marco-c:
OK, it's green now ;)

rhcu:
> OK, it's green now ;)

Thanks for helping with tags! :)

rhcu:
Should I add anything to the code?

bugbug/model.py
```diff
@@ -17,7 +17,7 @@
 
 
 class Model:
-    def __init__(self, lemmatization=False):
+    def __init__(self, lemmatization=False, historical=False):
```
marco-c: We are not using the parameter here, so you can remove it.

bugbug/models/bug.py
```diff
@@ -42,7 +42,8 @@ def __init__(self, lemmatization=False):
             bug_features.product(),
             bug_features.component(),
         ]
-
+        if (historical is True):
```
marco-c: Nit: keep a space before and after the if block

```
]

if ...:
    ...

cleanup_functions = [
```

marco-c: Nit: since `historical` is a boolean, you can just do `if historical:`

run.py
```diff
@@ -100,8 +101,10 @@
 
     if args.train:
         db.download()
-
```
marco-c: Nit: don't remove the newline here

run.py
```diff
@@ -100,8 +101,10 @@
...
-        model = model_class(args.lemmatization)
+        if args.historical and model_class == BugModel:
```
marco-c: Remove the second check here, if somebody passes `--historical` for an unsupported model, the script will just fail.

bugbug/model.py
rhcu: If I remove it, I get the following error:
` Model.__init__(self, lemmatization, historical)`

`TypeError: __init__() takes from 1 to 2 positional arguments but 3 were given`

since the constructor from this class is inherited by `bug.py`

marco-c: You'll need to remove the argument here, and stop passing it to Model.__init__ in bug.py

marco-c:
Thanks ;)

PR #262, opened 2019-04-10
Add basic HTTP service

Lothiraldan:
It doesn't support batch requests or authentication yet


http_service/models.py
```diff
@@ -0,0 +1,83 @@
...
+        with open(f"{file_path}.etag", "w") as f:  # noqa
+            f.write(new_etag)
+    else:
+        print(f"ETAG for {model_url} is ok")
+
+    return file_path
+
+
+def get_model_path(name):
+    file_name = f"{name}model"  # noqa: E999
```
marco-c: Nit: can you make this black-compliant and remove noqa?

http_service/models.py
```diff
@@ -0,0 +1,83 @@
...
+    file_path = os.path.join("models", file_name)
+
+    return file_path
+
+
+MODELS = {
```
marco-c: Defined twice

http_service/app.py
```diff
@@ -0,0 +1,44 @@
...
+        setattr(g, attribute, model)
+
+    return getattr(g, attribute, model)
+
+
+@application.route("/<model>/predict/<bug_id>")
+def hello(model, bug_id):
+    bugs = bugzilla._download([bug_id])
+    model = get_model(model)
+    probs = model.classify(copy.deepcopy(list(bugs.values())), True)
```
marco-c: Since we are not using the bugs data later on, we can also avoid copying

http_service/app.py
```diff
@@ -0,0 +1,44 @@
...
+
+    if not hasattr(g, attribute):
+        model = load_model(model)
+        setattr(g, attribute, model)
+
+    return getattr(g, attribute, model)
+
+
+@application.route("/<model>/predict/<bug_id>")
+def hello(model, bug_id):
```
marco-c: Nit: nicer name for this func

http_service/app.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def get_model(model):
+    attribute = f"bugbug_model_{model}"
+
+    if not hasattr(g, attribute):
+        model = load_model(model)
+        setattr(g, attribute, model)
+
+    return getattr(g, attribute, model)
```
marco-c: I'm not familiar with `g`. Will this cache the model between requests?
According to http://flask.pocoo.org/docs/1.0/appcontext/#storing-data it doesn't, so maybe we should use something else so we avoid loading the model on each request?

Lothiraldan: Indeed, I was remembering incorrectly that it was surviving requests, I will store the models somewhere else.

http_service/models.py
Lothiraldan: It should be black-compliant, I removed the noqa

PR #265, opened 2019-04-10
Add Docker build task

La0:


.taskcluster.yml
```diff
@@ -146,6 +146,38 @@ tasks:
...
+      - taskId: {$eval: as_slugid("docker_build")}
+        created: {$fromNow: ''}
+        deadline: {$fromNow: '1 hour'}
+        provisionerId: aws-provisioner-v1
+        workerType: releng-svc
+        payload:
+          capabilities:
+            privileged: true
+          maxRunTime: 3600
+          image: babadie/taskboot:latest
```
marco-c: Could you file an issue about switching to mozilla/ once we can?

.taskcluster.yml
```diff
@@ -146,6 +146,38 @@ tasks:
...
+            GIT_REVISION: ${head_rev}
+          command:
+            - taskboot
+            - build
+            - infra/dockerfile.base
+            - --write
+            - /base.tar
+          artifacts:
+            public/bugbug-base.tar:
+              expires: {$fromNow: '2 weeks'}
```
marco-c: Maybe we could set a lower expiration, so that we don't upload a lot of data to S3? (if we have many PRs, with many iterations)

.taskcluster.yml
```diff
@@ -146,6 +146,38 @@ tasks:
...
+              path: /base.tar
+              type: file
+        scopes:
+          - docker-worker:capability:privileged
+        metadata:
+          name: bugbug docker build
+          description: bugbug docker build
```
marco-c: Nit: maybe specify 'bugbug base docker build'

PR #266, opened 2019-04-10
Update root dockerignore

Lothiraldan:
The sooner we get this right, the less time we will waste invalidating Docker cache both locally and on the CI.


.dockerignore
```diff
@@ -1,4 +1,100 @@
-venv
-data/
+# Git
+.git
+.gitignore
+
+# CI
+.codeclimate.yml
+.travis.yml
```
marco-c: .taskcluster.yml too

Issue #258, opened 2019-04-09
Add a cache for pip packages in Taskcluster builds


marco-c:

La0:
Or use the docker image produced by the docker build task and run inside

marco-c:
Filed #272 for that.

La0:
I made some tests, it will be more complicated than i thought: setting up the Taskcluster cache is easy, but using it while building a Docker image is complex.

The protocol for mounting an host folder during an image build [is still an experimental feature](https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/experimental.md) and has several limitations:

- we can only use `mode=bind` for our use case
- the source must be in the build context, so in the midlle of our repo...
- but even with `readwrite` enabled, the doc says : **Written data will be discarded.**
- which means we cannot really cache anything...

I tested that with img and this Dockerfile:

```
#syntax=docker/dockerfile:experimental
FROM alpine

RUN echo 'taskboot !' > /etc/taskboot
RUN --mount=type=bind,source=/.cache,target=/cache,rw \
  echo 'plop' > /cache/plop
```

The build goes OK, but i don't have the `plop` file on my host cache... 

The *old* way to do that is to `COPY` a cache in the image, then use a staging image... that's dirty too.

marco-c:
OK, yeah, sounds like building all the images in the same task is the best option.

marco-c:
What if we build the base image in a task, and have another task to build the derived images.
In both tasks, we'd pull the previous base image first, so that we can benefit from the docker caching.

marco-c:
I'm not sure if it still makes sense to have a pip cache, now that we are benefiting from Docker layer caching.

marco-c:
> I'm not sure if it still makes sense to have a pip cache, now that we are benefiting from Docker layer caching.

@La0 what do you think? I think we can close this.

La0:
Yes, we can.

PR #274, opened 2019-04-12
Docker full build using compose

La0:
Use internal cache for pip calls
Build all images in one step

Refs #258


.gitignore
```diff
@@ -1,4 +1,4 @@
-*model
```
marco-c: Need to keep this too for local dev

La0: All right

PR #277, opened 2019-04-12
Make infra/version_check.py only run on branches and tags, not PRs

### LINKED ISSUE #271: Make infra/version_check.py only run on branches and tags, not pull requests
It is painful for contributors otherwise.

sbeesm:
Fixes #271.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -70,8 +70,8 @@ tasks:
                  git checkout ${head_rev} &&
                  python infra/version_check.py"
           metadata:
-            name: bugbug version_check
-            description: bugbug version_check
+            name: bugbug tag version check
+            description: bugbug tag version check
             owner: ${user}@users.noreply.github.com
             source: ${repository}/raw/${head_rev}/.taskcluster.yml
 
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -39,16 +39,16 @@ tasks:
           maxRunTime: 3600
           image: python
           $if: 'tasks_for == "github-push"'
-          then: 
-          command:
-            - "/bin/bash"
-            - "-lcx"
-            - "git clone ${repository} &&
-               cd bugbug &&
-               git checkout ${head_rev} &&
-               pip install -r test-requirements.txt &&
-               flake8 &&
-               python infra/version_check.py"
+          then:
+              command:
+                - "/bin/bash"
+                - "-lcx"
+                - "git clone ${repository} &&
+                   cd bugbug &&
+                   git checkout ${head_rev} &&
+                   pip install -r test-requirements.txt &&
+                   flake8 &&
+                   python infra/version_check.py"
           else:
               command:
                 - "/bin/bash"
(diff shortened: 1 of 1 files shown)


sbeesm:
> Submitting the task to Taskcluster failed. Details
> TemplateError at template.tasks[0].payload: $if has undefined properties: image maxRunTime

@marco-c sorry, I'm not familiar with `yml`.

Could you please tell me what's wrong here and how to fix this?

marco-c:
I'm not so familiar with the format either, you'll have to use the docs https://taskcluster-web.netlify.com/docs/reference/integrations/github/taskcluster-yml-v1 and a bit of trial and error.

sbeesm:
I've defined a separate task that runs the script when task is github-push.

Although the checks have passed, but how can I verify that it does the right thing?

.taskcluster.yml
```diff
@@ -45,14 +45,38 @@ tasks:
...
+            maxRunTime: 3600
+            image: python
+            command:
+              - "/bin/bash"
+              - "-lcx"
+              - "git clone ${repository} &&
+                 cd bugbug &&
+                 git checkout ${head_rev} &&
+                 pip install -r test-requirements.txt &&
+                 flake8 &&
```
marco-c: This task won't need to run flake8, given that we already run it in the linting task.

.taskcluster.yml
```diff
@@ -45,14 +45,38 @@ tasks:
...
+               flake8"
         metadata:
           name: bugbug lint
           description: bugbug lint
           owner: ${user}@users.noreply.github.com
           source: ${repository}/raw/${head_rev}/.taskcluster.yml
 
+      - $if: 'tasks_for == "github-push"'
+        then:
+          taskId: {$eval: as_slugid("version_check_task")}
```
marco-c: This task should be added as a dependency of the release task.

marco-c:
> Although the checks have passed, but how can I verify that it does the right thing?

After merging, we'll verify by looking at the Taskcluster logs of the tasks on master.

.taskcluster.yml
sbeesm: So do you mean to define this task in the release task itself or just adding the dependency like we've done for other tasks?

marco-c: Just the dependency

.taskcluster.yml
```diff
@@ -45,14 +45,37 @@ tasks:
...
+          payload:
+            maxRunTime: 3600
+            image: python
+            command:
+              - "/bin/bash"
+              - "-lcx"
+              - "git clone ${repository} &&
+                 cd bugbug &&
+                 git checkout ${head_rev} &&
+                 pip install -r test-requirements.txt &&
```
marco-c: We don't need to install the test-requirements in this case

.taskcluster.yml
```diff
@@ -45,14 +45,37 @@ tasks:
...
+                 python infra/version_check.py"
+          metadata:
+            name: bugbug version_check
+            description: bugbug version_check
```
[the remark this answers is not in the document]
sbeesm: Should we do this for `taskId` as well?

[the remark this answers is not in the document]
marco-c: No the taskId is OK

PR #289, opened 2019-04-18
Add a taskcluster hook update task

### LINKED ISSUE #283: Define a Taskcluster hook to start the pipeline
The hook will be triggered either on cron, or on receival of a pulse message (which will be sent after we deploy a new version).

This is an example hook which is triggered on cron: https://tools.taskcluster.net/hooks/project-releng/services-production-codecoverage%2Fbot.

This is an example hook which is triggered by a pulse message: https://tools.taskcluster.net/hooks/project-releng/services-testing-staticanalysis%2Fbot.

At first, the hook will do nothing, then we can implement the actual behavior separately.

Lothiraldan:
Fixes #283.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -237,3 +237,36 @@ tasks:
             description: bugbug docker push
             owner: ${user}@users.noreply.github.com
             source: ${repository}/raw/${head_rev}/.taskcluster.yml
+
+      - $if: 'tasks_for == "github-push" && head_branch[:10] == "refs/tags/"'
+        then:
+          taskId: {$eval: as_slugid("update_hook")}
+          dependencies:
+            - {$eval: as_slugid("docker_push")}
+          scopes:
+            - hooks:modify-hook:project-relman/bugbug
+            - assume:hook-id:project-relman/bugbug
+          created: {$fromNow: ''}
+          deadline: {$fromNow: '1 hour'}
+          provisionerId: aws-provisioner-v1
+          workerType: github-worker
+          payload:
+            features:
+              taskclusterProxy:
+                true
+            maxRunTime: 3600
+            image: mozilla/taskboot:latest
+            env:
+              GIT_REPOSITORY: ${repository}
+              GIT_REVISION: ${head_rev}
+            command:
+              - taskboot
+              - build-hook
+              - infra/taskcluster-hook-pipeline-start.json
+              - project-relman
+              - bugbug
+          metadata:
+            name: bugbug update hooks
+            description: bugbug update hooks
+            owner: ${user}@users.noreply.github.com
+            source: ${repository}/raw/${head_rev}/.taskcluster.yml
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ share/python-wheels/
 .installed.cfg
 *.egg
 MANIFEST
+cache/
--- a/infra/taskcluster-hook-pipeline-start.json
+++ b/infra/taskcluster-hook-pipeline-start.json
@@ -1,0 +1,59 @@
+{
+    "schedule": [
+        "0 0 * * * *"
+    ],
+    "metadata": {
+        "description": "",
+        "name": "BugBug data pipeline",
+        "owner": "mcastelluccio@mozilla.com"
+    },
+    "task": {
+        "created": {
+            "$fromNow": "0 seconds"
+        },
+        "deadline": {
+            "$fromNow": "2 hours"
+        },
+        "expires": {
+            "$fromNow": "1 month"
+        },
+        "extra": {},
+        "metadata": {
+            "description": "",
+            "name": "BugBug data pipeline",
+            "owner": "mcastelluccio@mozilla.com",
+            "source": "https://github.com/mozilla/bugbug"
+        },
+        "payload": {
+            "artifacts": {},
+            "cache": {},
+            "capabilities": {},
+            "command": [
+                "/hello"
+            ],
+            "env": {},
+            "features": {
+                "taskclusterProxy": true
+            },
+            "image": "hello-world",
+            "maxRunTime": 7200
+        },
+        "priority": "normal",
+        "provisionerId": "aws-provisioner-v1",
+        "retries": 5,
+        "routes": [
+            "notify.email.release-mgmt-analysis@mozilla.com.on-failed",
+            "notify.irc-channel.#bugbug.on-failed"
+        ],
+        "schedulerId": "-",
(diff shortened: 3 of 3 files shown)


taskcluster-hook-pipeline-start.json
```diff
@@ -0,0 +1,51 @@
...
+  },
+  "priority": "normal",
+  "provisionerId": "aws-provisioner-v1",
+  "retries": 5,
+  "routes": [
+    "index.project.relman.services.project.bugbug.data_pipeline.latest"
+  ],
+  "schedulerId": "-",
+  "scopes": [
+    "secrets:get:project/relman/bugbug/deploy",
```
marco-c: The needed secret will be project/relman/bugbug/production (project/relman/bugbug/deploy is for deployments, project/relman/bugbug/production is for operation)

requirements.txt
```diff
@@ -17,3 +17,4 @@ keras==2.2.4
 tqdm==4.31.1
 python-dateutil==2.8.0
 zstandard==0.11.0
+taskcluster==7.0.1
```
marco-c: Since this is only needed for deployment, we could avoid adding it to the requirements and just install it in .taskcluster.yml directly.

taskcluster-hook-pipeline-start.json
```diff
@@ -0,0 +1,62 @@
...
+        "provisionerId": "aws-provisioner-v1",
+        "retries": 5,
+        "routes": [
+            "index.project.relman.services.project.bugbug.data_pipeline.latest"
+        ],
+        "schedulerId": "-",
+        "scopes": [
+            "secrets:get:project/relman/bugbug/production",
+            "notify:email:*",
+            "docker-worker:cache:services-testing-static-analysis-bot",
```
marco-c: You can remove this one for now, we'll introduce a cache later.

taskcluster-hook-pipeline-start.json
```diff
@@ -0,0 +1,62 @@
...
+            "maxRunTime": 7200
+        },
+        "priority": "normal",
+        "provisionerId": "aws-provisioner-v1",
+        "retries": 5,
+        "routes": [
+            "index.project.relman.services.project.bugbug.data_pipeline.latest"
+        ],
+        "schedulerId": "-",
+        "scopes": [
```
marco-c: Instead of defining all scopes here, we can use `assume:hook-id:project-relman/bugbug` and define all the required scopes in the hook-id:project-relman/bugbug role.

taskcluster-hook-pipeline-start.json
```diff
@@ -0,0 +1,62 @@
...
+            "secrets:get:project/relman/bugbug/production",
+            "notify:email:*",
```
marco-c: For notifications, they could be routes so we don't have to implement anything in our task as they are handled automatically by Taskcluster.
Here's an example: https://github.com/marco-c/breakpad-win-update-symbols/blob/master/upload-task.json#L9 (but we'd notify the #bugbug channel and release-mgmt-analysis@mozilla.com).

.taskcluster.yml
```diff
@@ -177,6 +177,34 @@ tasks:
           owner: ${user}@users.noreply.github.com
           source: ${repository}/raw/${head_rev}/.taskcluster.yml
 
+      - taskId: {$eval: as_slugid("update_hook")}
+        scopes:
+          - hooks:modify-hook:garbage/bugbug-test
+          - assume:hook-id:garbage/bugbug-test
```
marco-c: You probably don't need this one

Lothiraldan:
Once https://github.com/mozilla/task-boot/pull/15 will be merged, we will be able to move to taskboot

.taskcluster.yml
```diff
@@ -177,6 +177,35 @@ tasks:
           owner: ${user}@users.noreply.github.com
           source: ${repository}/raw/${head_rev}/.taskcluster.yml
 
+      - taskId: {$eval: as_slugid("update_hook")}
+        scopes:
+          - hooks:modify-hook:project-relman/bugbug
+          - assume:hook-id:project-relman/bugbug
```
marco-c: We should check if it works after removing this (maybe not, but just want to be sure)

.taskcluster.yml
```diff
@@ -177,6 +177,35 @@ tasks:
           owner: ${user}@users.noreply.github.com
           source: ${repository}/raw/${head_rev}/.taskcluster.yml
 
+      - taskId: {$eval: as_slugid("update_hook")}
```
marco-c: Before merging, we should only run this task on tags, and only after the deployment of the docker images is done (so you can add the docker push task as a dependency of this task).

.taskcluster.yml
Lothiraldan: I had to re-add it, it was failing without https://tools.taskcluster.net/groups/NJps5ECFTciS6xf0sfSEqw/tasks/fbJ6T3rfTG-oq2x9h1NtfQ/runs/0/logs/public%2Flogs%2Flive.log, maybe you dit changed something since?

marco-c: OK, so yes, let's keep it.

.taskcluster.yml
Lothiraldan: Do we want also to move to taskboot new command before merging?

marco-c: I think it's basically ready, right? If so, yes.

marco-c: Note: to deploy only on tags, you'll need to add an if like the one we have for the PyPI release task.

Lothiraldan: Got it, I'm waiting for the taskboot PR to be merged and to have validated the dependencies before squashing and adding the condition

.taskcluster.yml
```diff
@@ -175,6 +175,37 @@ tasks:
...
+              taskclusterProxy:
+                true
+            maxRunTime: 3600
+            image: mozilla/taskboot:latest
+            command:
+              - "/bin/sh"
+              - "-lcxe"
+              - "git clone ${repository} /code &&
+                 cd /code &&
+                 git checkout ${head_rev} &&
```
La0: It would be better to use the clone feature of taskboot.
The payload would be:
```yaml
command:
  - taskboot
  - build-hook
  - taskcluster-hook-pipeline-start.json
  - project-relman
  - bugbug
env:
  GIT_REPOSITORY: ${repository}
  GIT_REVISION: ${head_rev}
```

marco-c: I've quickly made this change before merging

PR #290, opened 2019-04-18
Use Taskboot Push artifact

### LINKED ISSUE #288: Use taskboot to push images to Docker Hub
The credentials are in the project/relman/bugbug/deploy secret.

La0:
Fixes #288

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -161,7 +161,7 @@ tasks:
                cd /code &&
                git checkout ${head_rev} &&
                python http_service/download_models.py &&
-               taskboot --target /code build-compose --write /images"
+               taskboot --target /code build-compose --registry=registry.hub.docker.com --write /images"
           artifacts:
             public/bugbug:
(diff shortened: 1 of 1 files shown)


.taskcluster.yml
```diff
@@ -161,7 +161,7 @@ tasks:
                cd /code &&
                git checkout ${head_rev} &&
                python http_service/download_models.py &&
-               taskboot --target /code build-compose --write /images"
+               taskboot --target /code build-compose --registry=registry.hub.docker.com --write /images"
```
marco-c: A thought for the future: it'd be nicer if the registry was specified when pushing (in theory you could push the same image to different registries).

La0: The thing is that `push-artifact` reads the full `RepoTags` written in the image, and use that for pushing (so you can push to multiple registries already).
So I set the registry during the build steps, instead of relying on multiple arguments at push to specify where to push what

PR #293, opened 2019-04-19
Start working on the data pipeline task definitio

Lothiraldan:


data-pipeline.yml
```diff
@@ -0,0 +1,222 @@
+version: 1
+tasks:
+  - ID: commit-retrieval
+    created: {$fromNow: ''}
+    deadline: {$fromNow: '1 hour'}
```
marco-c: This task will very likely take longer than 1 hour, I think we should be safe with 12 hours.

data-pipeline.yml
```diff
@@ -0,0 +1,222 @@
...
+    provisionerId: aws-provisioner-v1
+    workerType: tutorial
```
marco-c: workerType for this should be release-svc-compute

data-pipeline.yml
```diff
@@ -0,0 +1,222 @@
...
+    scopes:
+      - docker-worker:cache:services-testing-bugbug-data
```
marco-c: We can remove the cache for now, and add it later

data-pipeline.yml
```diff
@@ -0,0 +1,222 @@
...
+        public/commits.json.xz:
+          path: /data/commits.json.xz
+          type: file
+
+      cache:
+        services-testing-bugbug-data: /cache
+
+      features:
+        taskclusterProxy:
+          true
```
marco-c: We might not need the proxy for this one (as it probably doesn't need to access any secret)

data-pipeline.yml
```diff
@@ -0,0 +1,222 @@
...
+      name: bugbug commit retrieval
+      description: bugbug commit retrieval
+      owner: release-mgmt-analysis@mozilla.com
+      source: https://github.com/mozilla/bugbug/raw/master/data-pipeline.yml
+
+  - ID: bugs-retrieval
+    created: {$fromNow: ''}
+    deadline: {$fromNow: '1 hour'}
+    provisionerId: aws-provisioner-v1
+    workerType: tutorial
```
marco-c: This one can be a smaller workerType, like releng-svc

data-pipeline.yml
```diff
@@ -0,0 +1,222 @@
...
+      name: bugbug bugs retrieval
+      description: bugbug bugs retrieval
+      owner: release-mgmt-analysis@mozilla.com
+      source: https://github.com/mozilla/bugbug/raw/master/data-pipeline.yml
+
+  - ID: train-component
+    created: {$fromNow: ''}
+    deadline: {$fromNow: '1 hour'}
+    provisionerId: aws-provisioner-v1
+    workerType: tutorial
```
marco-c: The train task will need a beefier worker type, so releng-svc-compute

data-pipeline.yml
```diff
@@ -0,0 +1,222 @@
...
+           taskboot --target /code build-compose --registry=registry.hub.docker.com --write /images --service bugbug-http-service"
+      artifacts:
+        public/bugbug:
+          expires: {$fromNow: '2 weeks'}
+          path: /images
+          type: directory
+    scopes:
+      - docker-worker:capability:privileged
+    metadata:
+      name: bugbug docker http build
```
marco-c: Will we push the image to AWS/Heroku in the same task, or in a task dependent on this one?

infra/spawn_data_pipeline.py
```diff
@@ -0,0 +1,111 @@
...
+        options["rootUrl"] = proxy_url
+
+    if "rootUrl" not in options:
+        # Always have a value in root url
+        options["rootUrl"] = TASKCLUSTER_DEFAULT_URL
+
+    return options
+
+
+def main():
```
La0: @marco-c This is basically https://github.com/mozilla/task-boot/issues/14 without adding a json-e function.

data-pipeline.yml
Lothiraldan: A dependent one

data-pipeline.yml
```diff
@@ -0,0 +1,211 @@
+version: 1
+tasks:
+  - ID: commit-retrieval
+    created: {$fromNow: ''}
+    deadline: {$fromNow: '12 hours'}
+    provisionerId: aws-provisioner-v1
+    workerType: release-svc-compute
+    payload:
+      maxRunTime: 86400
```
marco-c: The maxRunTime should match the deadline

data-pipeline.yml
```diff
@@ -0,0 +1,211 @@
...
+
+    metadata:
+      name: bugbug commit retrieval
+      description: bugbug commit retrieval
+      owner: release-mgmt-analysis@mozilla.com
+      source: https://github.com/mozilla/bugbug/raw/master/data-pipeline.yml
+
+  - ID: bugs-retrieval
+    created: {$fromNow: ''}
+    deadline: {$fromNow: '1 hour'}
```
marco-c: This one also will take a very long time

data-pipeline.yml
```diff
@@ -0,0 +1,211 @@
...
+
+    metadata:
+      name: bugbug bugs retrieval
+      description: bugbug bugs retrieval
+      owner: release-mgmt-analysis@mozilla.com
+      source: https://github.com/mozilla/bugbug/raw/master/data-pipeline.yml
+
+  - ID: train-component
+    created: {$fromNow: ''}
+    deadline: {$fromNow: '1 hour'}
```
marco-c: Some of the train task will take longer than 1 hour (this one probably around 3?). We could use 7 for all of them, and then update it when we have more data on how long they take.

data-pipeline.yml
```diff
@@ -0,0 +1,211 @@
...
+      image: mozilla/bugbug-train-defect
+
+      artifacts:
+        public/defectenhancementtaskmodel.xz:
+          path: /defectenhancementtaskmodel.xz
+          type: file
+
+      features:
+        taskclusterProxy:
+          true
```
marco-c: I think all train task won't need access to secrets

data-pipeline.yml
```diff
@@ -0,0 +1,211 @@
+version: 1
+tasks:
+  - ID: commit-retrieval
+    created: {$fromNow: ''}
+    deadline: {$fromNow: '12 hours'}
+    provisionerId: aws-provisioner-v1
+    workerType: release-svc-compute
+    payload:
+      maxRunTime: 43299
```
marco-c: Nit: why 12 hours + 99 seconds? :P

marco-c:
We should make all tasks report to #bugbug channel and release-mgmt-analysis@mozilla.com in case of failure (with a route, like the hook).

data-pipeline.yml
Lothiraldan: Is it only for accessing secrets? Would they need it for pushing / getting artifacts?

data-pipeline.yml
Lothiraldan: Nice catch, just a typo, I meant 43200

data-pipeline.yml
marco-c: It is needed for a few things that require auth, but artifacts do not require it IIRC.

infra/taskcluster-hook-pipeline-start.json
```diff
@@ -29,13 +29,15 @@
             "cache": {},
             "capabilities": {},
             "command": [
-                "/hello"
+                "/bin/bash",
+                "-lcx",
+                "git clone https://github.com/mozilla/bugbug && cd bugbug && git checkout master && pip install taskcluster json-e requests pyyaml && python infra/spawn_data_pipeline.py data-pipeline.yml"
```
marco-c: We should have an image for the hook task, so that we don't have to clone and pip install whenever the hook is triggered.

data-pipeline.yml
```diff
@@ -0,0 +1,229 @@
...
+      - notify.irc-channel.#bugbug.on-failed
+    metadata:
+      name: bugbug bugs retrieval
+      description: bugbug bugs retrieval
+      owner: release-mgmt-analysis@mozilla.com
+      source: https://github.com/mozilla/bugbug/raw/master/data-pipeline.yml
+
+  - ID: train-component
+    created: {$fromNow: ''}
+    deadline: {$fromNow: '7 hour'}
```
marco-c: I think the deadline should be the maximum runtime of the task + the maximum runtime of the tasks it depends on.

infra/taskcluster-hook-pipeline-start.json
```diff
@@ -29,25 +29,27 @@
             "cache": {},
             "capabilities": {},
             "command": [
-                "/hello"
+                "/bin/bash",
+                "-lcx",
+                "git clone https://github.com/mozilla/bugbug && cd bugbug && git checkout data-pipeline-definition && pip install taskcluster json-e requests pyyaml && python infra/spawn_data_pipeline.py data-pipeline.yml"
```
marco-c: This should be an image, instead of doing the git clone + dependency installation at runtime.

We can do it as a follow-up (just remember to file an issue if you want to do it in a follow-up)

Lothiraldan: I think @La0 suggested that this should be a task-boot commando I guess we will use the task-boot image which will have all the required command

marco-c: It might be a while until we implement it in taskboot, as we haven't settled on a decision about this yet. We might also keep task-boot only about building/deploying.

Anyway, let's file an issue so we don't forget to make this avoid the git clone + dependency installation at runtime.

Lothiraldan:
Closing the PRs to stop having notifications

PR #294, opened 2019-04-19
Add support to get secret from taskcluster

Lothiraldan:


scripts/bug_retriever.py
```diff
@@ -2,29 +2,19 @@
...
 import lzma
-import os
 import shutil
 from datetime import datetime
 from logging import INFO, basicConfig, getLogger
 
 from dateutil.relativedelta import relativedelta
 
 from bugbug import bug_snapshot, bugzilla, labels
+from bugbug.scripts import get_secret
```
marco-c: This should be bugbug.utils

bugbug/utils.py
```diff
@@ -50,3 +55,39 @@ def fit(self, X, y=None):
...
+        # Always have a value in root url
+        options["rootUrl"] = TASKCLUSTER_DEFAULT_URL
+
+    return options
+
+
+def get_secret(secret_id):
+    """ Return the secret value
+
+    TODO: Support task-cluster secret API
```
marco-c: Nit: remove TODO now

bugbug/utils.py
```diff
@@ -3,11 +3,16 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
+import os
+
 import numpy as np
+import taskcluster
```
marco-c: You'll need to add taskcluster to the requirements.txt now, since we need it to run bugbug

PR #301, opened 2019-04-21
Add number of previous commits touching the same directories as a feature

### LINKED ISSUE #278: Add number of previous commits touching the same directories as a feature
Similar to #112 and #113.

sbeesm:
Fixes #278.


sbeesm:
@marco-c should we consider migrating commits for directories of files in file_copies? 

I've checked some `file_copies` and for most of them, the change was only the file name and not any directory.

marco-c:
> @marco-c should we consider migrating commits for directories of files in file_copies?

> I've checked some file_copies and for most of them, the change was only the file name and not any directory.

Yes we should do it, as otherwise we wouldn't be precise. Even if in most cases just the file name was changed, we can't just assume it's always the case.

sbeesm:
@marco-c I think we need to take care of some edge cases as well.

The cases where a combination of same directories occur in a new commit that were touched by a previous commit as well.

For instance:
Lets say, commit A touches multiple directories as: ``` {'b2g', 'installer', 'dom', 'apps', 'wappush'} ```.

So for each directory, we have a count of one commit touching them.
But as we traverse down through the list of commits, now we only know that these directories were touched by some previous commit but we don't know that probably same commit touched them. 

That would be important for a case when for a new commit X, we again see a combination of these directories, ```{'b2g', 'installer', 'dom', 'apps', 'layout', ...}```

(For the sake of simplicity, lets assume we don't see any of these directories again before commit X)
So for the similar directories, commit A should be considered as one commit touching them but with this:
```
for directory in directories:
    directories_touched_prev[commit.node] += directories_touched[directory]

    directories_touched[directory] += 1
```

We always assume that the directories are always touched by different commits and add 1 for every directory match, in this case storing 4 as a number of previous of commits, which is wrong.

This could also be the case with components and files.

Any suggestions how we might fix this?

bugbug/repository.py
```diff
@@ -456,6 +500,15 @@ def download_commits(repo_dir, date_from):
             for path_prev in prev_commit.files:
                 files_touched_90_days[path_prev] += 1
 
+            directories_prev = set(
+                directory
+                for path in prev_commit.files
+                for directory in os.path.dirname(path).split("/")[:2]
+                if directory != ""
+            )
```
Lothiraldan: This will extract each directory of level 0 and level 1 independently. So a modification on a file `docshell/build/nsDocShellModule.cpp` and a modification on a file `image/build/moz.build` will both haves the same `build` feature. Is it expected? I don't know about the mozilla-central architecture to know if it makes sense or not so please forgive me if my question is silly.

bugbug/repository.py
```diff
@@ -465,12 +518,36 @@ def download_commits(repo_dir, date_from):
...
                     files_touched_90_days[copied] = files_touched_90_days[orig]
 
+                    orig_directories = [
+                        directory
+                        for directory in os.path.dirname(orig).split("/")[:2]
+                        if directory != ""
+                    ]
+                    copied_directories = [
+                        directory
+                        for directory in os.path.dirname(copied).split("/")[:2]
```
Lothiraldan: Maybe the `os.path.dirname(path).split("/")[:2] if directory != ""` could be extracted somewhere?

Lothiraldan: Also this will be at most a list of size 2, right?

bugbug/repository.py
marco-c: No it's not expected, e.g. for `docshell/build/nsDocShellModule.cpp` we should have `docshell` ad `docshell/build`.

bugbug/repository.py
sbeesm: > Maybe the `os.path.dirname(path).split("/")[:2] if directory != ""` could be extracted somewhere?

But for the original and copied file paths, we'd still have to do it twice as the path may change as well. 

> Also this will be at most a list of size 2, right?

Yes.

sbeesm:
> There is a bit too much code duplication for my own personal taste

To avoid duplication to some extent, we may define functions to extract components and directories. These functions can extract data based on what we pass to them (list of files for commit.files and dictionary for file_copies).

And then we can use these functions for commits and previous 90 days commits.

bugbug/repository.py
```diff
@@ -434,12 +434,14 @@ def download_commits(repo_dir, date_from):
...
-            for directory in os.path.dirname(path).split("/")[:2]
-            if directory != ""
-        )
+        directories = set()
+        for path in commit.files:
+            path_dirs = (
+                os.path.dirname(path).split("/")[:2] if os.path.dirname(path) else []
+            )
+            if path_dirs:
+                directories.update([path_dirs[0], "/".join(path_dirs)])
```
Lothiraldan: I think we should definitely extract this part in a separate function that takes a path and return a list of directories or an empty list

bugbug/repository.py
```diff
@@ -275,6 +282,20 @@ def get_revs(hg, date):
...
 
 
+def get_directories(files):
+    if isinstance(files, str):
+        files = [files]
+
+    directories = set()
+    for path in files:
+        path_dirs = (
+            os.path.dirname(path).split("/")[:2] if os.path.dirname(path) else []
```
Lothiraldan: Small improvement, you can limit the split works by passing maxsplit.
```suggestion
            os.path.dirname(path).split("/", 2)[:2] if os.path.dirname(path) else []
```

Do you have an example where `os.path.dirname(path)` returns a Falsy value?

sbeesm: It returns false when just a file name is encountered in `commit.files`.

Example for this file `old-configure.in`, it returns an empty string.

sbeesm:
Any suggestions on https://github.com/mozilla/bugbug/pull/301#issuecomment-486346195?

marco-c:
I thought I already answered, but apparently I didn't :/

If I got your comment right, the problem you were thinking of is something like this:
Commit A modifies 'dom' and 'apps'
Commit B modifies 'dom'
Commit C modifies 'dom' and 'apps'

With the current implementation, we lose the information about the fact that commit A and commit C were both modifying those two directories, as commit A will have dom=0 and apps=0, commit B will have dom=1, commit C will have dom=2 and apps=1.

If this is the problem you were thinking of, yes, we lose some information, but I think this information might be too detailed to be really useful. We can consider adding combinations of directories too in the future, but I wouldn't expect them to bring much.

sbeesm:
> Commit A modifies 'dom' and 'apps'
> Commit B modifies 'dom'
> Commit C modifies 'dom' and 'apps'
> 
> With the current implementation, we lose the information about the fact that commit A and commit C were both modifying those two directories, as commit A will have dom=0 and apps=0, commit B will have dom=1, commit C will have dom=2 and apps=1.

Yes, and because we're adding these directories to calculate number of previous commits, the # prev commits for Commit C becomes `2 (dom) + 1 (apps) = 3`. But there are only two commits.

So I guess this could only be solved by adding combinations of directories as you suggested.

marco-c:
> Yes, and because we're adding these directories to calculate number of previous commits, the # prev commits for Commit C becomes 2 (dom) + 1 (apps) = 3. But there are only two commits.

Are you sure about this? The # prev commits for Commit C should be 2, it's a bug if it isn't.

sbeesm:
> Are you sure about this? The # prev commits for Commit C should be 2, it's a bug if it isn't.

Yes. With the current implementation: 

```
for directory in directories:
    directories_touched_prev[commit.node] += directories_touched[directory]

    directories_touched[directory] += 1
```


For a commit, as we're iterating over its directories one at a time, we add the # previous commits for each directory individually without considering the fact that multiple directories might be touched by one commit.

So if a commit C modifies 'dom' and 'apps', it adds Commit A two times, once for each 'dom' and 'apps'.

sbeesm:
And unfortunately, this also seems to be the case with # previous commits touching components and files.

sbeesm:
This actually seems to be a problem with how we're managing count of previous directories. 

Instead of having a count like this:
`directories_touched[directory] += 1`

Can we instead maintain a set of commit nodes for each file (having commit nodes touching each directory)? This way we won't have duplicate nodes for different directories of a file.

marco-c:
OK, I see your point. Yeah, using the commit nodes sounds like a good idea. Can you file a follow-up issue?

PR #308, opened 2019-04-25
Build base image with python slim

### LINKED ISSUE #300: Try switching to Python slim for the base image
Might be a first (big) step towards #292.

La0:
Fixes #300

### DIFF
--- a/infra/dockerfile.base
+++ b/infra/dockerfile.base
@@ -1,12 +1,14 @@
 #syntax=docker/dockerfile:experimental
 FROM python:3.7.3-slim
 
-COPY requirements.txt /code/bugbug/
+# Setup dependencies in a cachable step
+COPY requirements.txt /requirements.txt
+RUN apt-get update && \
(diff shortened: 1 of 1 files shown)


infra/dockerfile.base
```diff
@@ -1,12 +1,14 @@
...
-RUN --mount=type=cache,target=/cache/pip \
-	pip install --cache-dir=/cache/pip /code/bugbug/
+FROM python:3.7.3-slim
+
+# Setup dependencies in a cachable step
+COPY requirements.txt /requirements.txt
+RUN apt-get update && \
+		apt-get install -y gcc g++ libgomp1 && \
+		pip install --disable-pip-version-check -r /requirements.txt && \
+		apt-get remove -y gcc g++ && \
```
marco-c: Should we remove libgomp1 too?

La0: It's needed to run `bugbug-train`

La0: It's needed to run `bugbug-train`

PR #310, opened 2019-04-25
Make Spacy and Gensim dependencies optional

### LINKED ISSUE #307: Make NLP dependencies optional
Some dependencies are only needed in some images, we should make some dependencies optional so that they are only installed when necessary.

For example, Spacy is the biggest dependency, but it is currently not really used by any image.

Lothiraldan:
Fixes #307.

### DIFF
--- a/bugbug/nlp.py
+++ b/bugbug/nlp.py
@@ -7,7 +7,6 @@ from collections import defaultdict
 from functools import lru_cache
 
 import numpy as np
-from gensim.models import KeyedVectors
 from sklearn.base import BaseEstimator, TransformerMixin
 from sklearn.feature_extraction.text import TfidfVectorizer
 
@@ -16,10 +15,15 @@ try:
     import spacy
 
     from spacy.tokenizer import Tokenizer
+    from gensim.models import KeyedVectors
 
     nlp = spacy.load("en_core_web_sm")
 except ImportError:
     spacy = None
+
+OPT_MSG_MISSING = (
+    "Optional dependencies are missing, install them with `pip install bugbug[nlp]`"
+)
 
 
 def spacy_token_lemmatizer(text):
@@ -34,8 +38,7 @@ class SpacyVectorizer(TfidfVectorizer):
 
         # Detect when the optional dependency is missing
         if spacy is None:
-            msg = "Optional dependency Spacy is missing"
-            raise NotImplementedError(msg)
+            raise NotImplementedError(OPT_MSG_MISSING)
 
         super().__init__(tokenizer=spacy_token_lemmatizer, *args, **kwargs)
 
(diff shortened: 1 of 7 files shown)


bugbug/nlp.py
```diff
@@ -3,17 +3,38 @@
...
+try:
+    import spacy
+
+    from spacy.tokenizer import Tokenizer
+    from gensim.models import KeyedVectors
+except ImportError:
+    spacy = None
+
+try:
+    nlp = spacy.load("en_core_web_sm")
```
marco-c: We should only do this when spacy is available

bugbug/nlp.py
```diff
@@ -3,17 +3,38 @@
...
 from sklearn.feature_extraction.text import TfidfVectorizer
-from spacy.tokenizer import Tokenizer
 
-nlp = spacy.load("en_core_web_sm")
+# Spacy is an optional dependency
+try:
+    import spacy
+
+    from spacy.tokenizer import Tokenizer
+    from gensim.models import KeyedVectors
```
marco-c: We should also detect when gensim is not available.

bugbug/nlp.py
Lothiraldan: Indeed, it should be fixed now

bugbug/nlp.py
Lothiraldan: Indeed, I introduced a new atomic boolean for this effect

bugbug/nlp.py
```diff
@@ -3,17 +3,40 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
+from __future__ import print_function
```
marco-c: This should not be needed, right?

bugbug/nlp.py
```diff
@@ -3,17 +3,40 @@
...
+
+    HAS_OPTIONAL_DEPENDENCIES = True
+except ImportError:
+    pass
+
+try:
+    if HAS_OPTIONAL_DEPENDENCIES:
+        nlp = spacy.load("en_core_web_sm")
+except OSError:
+    msg = "Spacy model is missing, install it with: %s -m spacy download en_core_web_sm"
```
marco-c: Nit: use f-string here and in OPT_MSG_MISSING

bugbug/nlp.py
```diff
@@ -25,6 +48,11 @@ def spacy_token_lemmatizer(text):
 
 class SpacyVectorizer(TfidfVectorizer):
     def __init__(self, *args, **kwargs):
+
+        # Detect when the optional dependency is missing
```
marco-c: Nit: `# Detect when the spacy optional dependency is missing` (at some point we might split Spacy and Gensim into separate optional deps)

bugbug/nlp.py
```diff
@@ -37,6 +65,10 @@ def get_word_embeddings():
 
 class MeanEmbeddingTransformer(BaseEstimator, TransformerMixin):
     def __init__(self):
+        # Detect when the optional dependency is missing
```
marco-c: Nit: `# Detect when the gensim optional dependency is missing` (at some point we might split Spacy and Gensim into separate optional deps)

PR #311, opened 2019-04-25
Use Taskcluster cache during docker build

La0:
Needs https://github.com/mozilla/task-boot/pull/17 to be merged & deployed


La0:
@marco-c The build worked with the cache, it seems a bit faster.

We need to merge taskboot PR, then remove 965df55d38ce2c023024bade5a956c2b8ed1ca01, and push again...

PR #314, opened 2019-04-26
Don't use pip cache directory

marco-c:
This slightly reduces the size of the images.


marco-c:
This reduces the base image size from ~700 to ~400 MB!

PR #319, opened 2019-04-26
Also moves keras as an optional dependency - fix #317

### LINKED ISSUE #317: Make Keras/Tensorflow dependencies optional
See also #307.

Lothiraldan:


bugbug/models/component_nn.py
```diff
@@ -29,10 +17,39 @@
...
+        Embedding,
+        Flatten,
+        GlobalMaxPooling1D,
+        SpatialDropout1D,
+    )
+    from keras.models import Model as KerasModel
+
+    HAS_OPTIONAL_DEPENDENCIES = True
+except ImportError:
+    pass
```
marco-c: The whole file can't be used without Keras, so we might as well throw here

bugbug/nn.py
```diff
@@ -3,16 +3,34 @@
...
+HAS_OPTIONAL_DEPENDENCIES = False
+
+try:
+    from keras.preprocessing.sequence import pad_sequences
+    from keras.preprocessing.text import Tokenizer
+    from keras.utils import to_categorical
+
+    HAS_OPTIONAL_DEPENDENCIES = True
+except ImportError:
+    pass
```
marco-c: Same here

PR #321, opened 2019-04-28
Http service batch

Lothiraldan:


Lothiraldan:
@marco-c I have pushed a first version of the http service using a background worker. The "database" is currently a dict in memory, we will likely wants to move a real database and a real caching service like Redis.

We will need to think about how many web workers we will have and how we route the requests for a given bug id. With the current design, if you ask two different web workers about the same bug, they will both schedule the classification for the given bug id.

marco-c:
(We could use https://python-rq.org/ for Redis, that's what I used for code coverage for background jobs)

http_service/app.py
```diff
@@ -3,37 +3,80 @@
...
+POOL = Pool(4)
+
+
+class Database(UserDict):
+    # We might want to customize the database with cache invalidation or plug
+    # it into a real database / caching service
+    pass
+
+
+MODELS_DB = defaultdict(Database)
```
marco-c: Nit: RESULTS_DB sounds like a better name

http_service/app.py
```diff
@@ -3,37 +3,80 @@
...
+
+
+def schedule_bug_classification(model_name, bug_id):
+    """ Schedule the classification of a single bug_id
+    """
+    res = POOL.apply_async(classify_bug, (model_name, bug_id, BUGZILLA_TOKEN))
```
marco-c: For performance reasons, it'd be better to do the classification by batches, but this would increase the complexity of the architecture.
We can do it as a follow-up.

http_service/app.py
```diff
@@ -43,16 +86,18 @@ def model_prediction(model, bug_id):
...
+    for bug in bugs:
+        if bug not in MODELS_DB[model_name]:
+            schedule_bug_classification(model_name, bug)
+
+    data = {}
+
+    for bug in bugs:
+        data[bug] = get_bug_classification(model_name, bug)
 
     return jsonify(**data)
```
marco-c: We could return 202 when the results are not ready yet

http_service/app.py
```diff
@@ -3,37 +3,80 @@
...
+class Database(UserDict):
+    # We might want to customize the database with cache invalidation or plug
+    # it into a real database / caching service
+    pass
+
+
+MODELS_DB = defaultdict(Database)
+
+
+def schedule_bug_classification(model_name, bug_id):
```
marco-c: Shouldn't we check if a result is already available, before scheduling a new classification?

marco-c:
> We will need to think about how many web workers we will have and how we route the requests for a given bug id. With the current design, if you ask two different web workers about the same bug, they will both schedule the classification for the given bug id.

I was thinking we could have X web workers (1 at first, but configurable) and Y background workers (1 at first, but configurable), with a queue of jobs between them.

The web workers put jobs in the queue or get results from the queue. The background workers get jobs from the queue and put results in the queue.

Lothiraldan:
I've used https://python-rq.org in the past and it did the job.

I agree with the design, I would make just one change, instead of returning the result, the background workers would insert their results in Redis directly. They need a Redis connection for RQ anyway. This way we don't need to consume the queue on the web worker side, what do you think?

http_service/app.py
Lothiraldan: I tried to do the classifications for all given bugs but the return structure of `classify_bug` is not very row-oriented friendly. How much performance gain could we expect doing batches?

http_service/app.py
Lothiraldan: I'm hesitant, I would say it's the responsibility of the caller to choose if there is a result already available (and if they want to schedule it anyway). Maybe we can introduce a new higher-level function that would decides if it should schedule the classification or not.

Lothiraldan:
One thing I was not sure if rq supports was batching jobs, it seems it does not, so we will have to do the batching ourselves, which might be cumbersome to do with overlapping batches.

http_service/app.py
marco-c: I'm not sure, maybe we can try taking a model and 100 bugs and test how long it takes in batches of 1, 5, 10, 100.

http_service/app.py
marco-c: It works for me if we put it in the caller, or a higher level function.

marco-c:
> I agree with the design, I would make just one change, instead of returning the result, the background workers would insert their results in Redis directly. They need a Redis connection for RQ anyway. This way we don't need to consume the queue on the web worker side, what do you think?

Agreed!

> One thing I was not sure if rq supports was batching jobs, it seems it does not, so we will have to do the batching ourselves, which might be cumbersome to do with overlapping batches.

Let's do the quick performance test, if it doesn't buy us much we can avoid it.

Lothiraldan:
> > I agree with the design, I would make just one change, instead of returning the result, the background workers would insert their results in Redis directly. They need a Redis connection for RQ anyway. This way we don't need to consume the queue on the web worker side, what do you think?
> 
> Agreed!
> 
> > One thing I was not sure if rq supports was batching jobs, it seems it does not, so we will have to do the batching ourselves, which might be cumbersome to do with overlapping batches.
> 
> Let's do the quick performance test, if it doesn't buy us much we can avoid it.

Theoretically it should be much faster but if we need to reconstruct 1000s of dicts, if the performance gain is not big, it might be slower

Lothiraldan:
I have pushed a first version using RQ that seems to works locally, I have a few things I want to check and fix but the happy path should be covered

http_service/app.py
Lothiraldan: I checked quickly what the overhead was and it was non-negligeable, it took 9s to process the 100 bugs in one batch (with a preloaded model) and 29s to process them in batchs of size 10. I tried with a batch of size 1 but the benchmark was still running after 22 minutes.

I checked the RQ Worker implementation and the worker is forking right before handling a job and is designed to handle only one job. That means that we cannot cache the model in memory at least without some customization of the Worker. It should also be possible to have a custom Worker able to process several RQ tasks at once instead of one by one, so we could still schedule them one by one, but I'm not sure how much work is it needed.

@marco-c Do you think we will have lot of requests in the first time? It takes about 2 seconds to classify a bug right now with the model loading.

marco-c: I tried with the scripts we are running in relman-auto-nag. The first needed to classify 443 bugs, the second 453 bugs, the third 1380 bugs.
Usually the numbers will be smaller though (e.g. the third is run every hour, but it only needs to classify bugs that it hasn't classified yet, so the first run will classify 1380 bugs, the second run only bugs that weren't classified before, so < 10).

Lothiraldan:
@marco-c Code is ready, but we need to setup the separate dyno type and Redis on Heroku before releasing it

http_service/Dockerfile
```diff
@@ -11,8 +11,6 @@ COPY . /code/bugbug_http_service/
 # Load the models
 WORKDIR /code/
 
-COPY ./models /code/models
-
 ARG CHECK_MODELS
```
marco-c: We don't need CHECK_MODELS and ensure_models.sh anymore here, right? Given that we move the classification in the bg part.

http_service/Dockerfile.bg_worker
```diff
@@ -0,0 +1,17 @@
...
+
+COPY . /code/bugbug_http_service/
+
+# Load the models
+WORKDIR /code/
+
+ARG CHECK_MODELS
+ENV CHECK_MODELS="${CHECK_MODELS}"
+
+RUN python /code/bugbug_http_service/check_models.py
```
marco-c: Should we use ensure_models.sh?

http_service/app.py
```diff
@@ -3,38 +3,92 @@
...
-bugzilla.set_token(os.environ.get("BUGBUG_BUGZILLA_TOKEN"))
+BUGZILLA_TOKEN = os.environ.get("BUGBUG_BUGZILLA_TOKEN")
 
 logging.basicConfig(level=logging.INFO)
 LOGGER = logging.getLogger()
 
 
-def get_model(model):
-    attribute = f"bugbug_model_{model}"
+def get_job_id(model_name, bug_id):
```
marco-c: Nit: I guess you can remove the two unused parameters

http_service/app.py
```diff
@@ -3,38 +3,92 @@
...
+
+    try:
+        job = Job.fetch(job_id.decode("utf-8"), connection=redis_conn)
+    except NoSuchJobError:
+        # The job might have expired from redis
+        return False
+
+    job_status = job.get_status()
+    if job_status in ("running", "started", "queued"):
+        print("Bug already running", model_name, bug_id)
```
marco-c: Nit: Bug classification already running or Classification already running

http_service/app.py
```diff
@@ -43,16 +97,50 @@ def model_prediction(model, bug_id):
...
-    bugs = bugzilla._download([bug_id])
-    model = get_model(model)
-    probs = model.classify(list(bugs.values()), True)
-    indexes = probs.argmax(axis=-1)
-    suggestions = model.clf._le.inverse_transform(indexes)
+    status_code = 200
+    data = get_bug_classification(model_name, bug_id)
+
+    if not data:
+        schedule_bug_classification(model_name, [bug_id])
```
marco-c: Shouldn't we check if it's already running before scheduling?

http_service/app.py
```diff
@@ -43,16 +97,50 @@ def model_prediction(model, bug_id):
...
+
+    bugs = batch_body["bugs"]
+
+    status_code = 200
+    data = {}
+    missing_bugs = []
+
+    for bug in bugs:
+        data[str(bug)] = get_bug_classification(model_name, bug)
+        if not data[str(bug)]:
```
marco-c: What happens if we have a security bug? Will we always return 202?

http_service/models.py
```diff
@@ -60,3 +65,60 @@ def retrieve_model(name):
...
+
+
+def classify_bug(
+    model_name, bug_ids, bugzilla_token, expiration=DEFAULT_EXPIRATION_TTL
+):
+    # This should be called in a process worker so it should be safe to set
+    # the token here
+    bug_ids_set = set(map(int, bug_ids))
+    bugzilla.set_token(bugzilla_token)
+    bugs = bugzilla._download(bug_ids)
```
marco-c: Maybe we could use download, so we store the bug data locally and don't have to redownload it if somebody asks for the classification of the same bug again with a different model (since the cache only works for same model / same bug).
We'd need a way to avoid the DB growing forever though.

http_service/models.py
```diff
@@ -60,3 +65,60 @@ def retrieve_model(name):
...
+        # TODO: Find a better error format
+        encoded_data = json.dumps({"available": False})
+
+        redis.set(redis_key, encoded_data)
+        redis.expire(redis_key, expiration)
+
+    if not bugs:
+        return "NOK"
+
+    # TODO: Cache the model in the process memory, it's quite hard as the RQ
```
marco-c: Ah, this is painful, IIRC loading a model is not so cheap

http_service/models.py
```diff
@@ -60,3 +65,60 @@ def retrieve_model(name):
...
+    # worker is forking before starting
+    model = load_model(model_name)
+
+    # TODO: Classify could choke on a single bug which could make the whole
+    # job to fails. What should we do here?
```
marco-c: Maybe we could do the same as for missing bugs earlier?

http_service/docker-compose.yml
```diff
@@ -0,0 +1,53 @@
+version: '3.2'
```
marco-c: Can we remove the http service stuff from the root docker-compose and only leave it here?

http_service/app.py
Lothiraldan: No because we add an error placeholder for missing bugs here: https://github.com/mozilla/bugbug/pull/321/files/225b0a03969aeb5c63c064dec0215e18a632b59f#diff-7908c821177e4af7e08af4a9ca055d60R88

It can be security bugs or non-existing bug ids

http_service/docker-compose.yml
Lothiraldan: If we do that, we won't build the Docker http image during release or we would have to also run `taskboot build-compose` against this docker-compose.

We still need to build the bg worker image

http_service/models.py
Lothiraldan: I'm not sure what you mean, do you want to use the bug database from the index to avoid getting data from bugzilla directly? Or introduce a cache layer, like Redis?

http_service/models.py
Lothiraldan: We might be able to preload the models when the worker starts with: https://python-rq.org/docs/workers/#performance-notes

http_service/models.py
Lothiraldan: The issue is that we are giving X bug ids to `classify` and get an Exception in return so we don't know which one was failing

http_service/models.py
marco-c: I mean using `download_bugs` instead of `_download`, which will store the bug IDs locally. But we can consider this as a future enhancement.

http_service/models.py
marco-c: Ah right, I remember we talked about that.

http_service/models.py
Lothiraldan: Once there are in the DB, I don't see an easy way to get individuals bugs from the DB. Moreover it we start storing things on disk, we need to be extra careful about cache invalidation. I would prefer to put it as a future enhancement

http_service/models.py
Lothiraldan: @marco-c Do you want to try doing that now or later?

marco-c: Let's file an issue for doing that later

http_service/models.py
marco-c: Let's file an issue for doing that later

marco-c:
We can land after https://bugzilla.mozilla.org/show_bug.cgi?id=1557809 is fixed.

PR #328, opened 2019-05-02
Update the index URLs in bugbug

### LINKED ISSUE #325: Store pipeline tasks in the index, and update Taskcluster URLs in scripts
The scripts are still using the old URLs from releng-services.

Lothiraldan:
Fixes #325.

### DIFF
--- a/http_service/download_models.py
+++ b/http_service/download_models.py
@@ -14,7 +14,7 @@ import requests
 logging.basicConfig(level=logging.INFO)
 LOGGER = logging.getLogger()
 
-BASE_URL = "https://index.taskcluster.net/v1/task/project.relman.project.bugbug.train_{}.latest/artifacts/public"
+BASE_URL = "https://index.taskcluster.net/v1/task/project.relman.bugbug.train_{}.latest/artifacts/public"
 
 MODELS_NAMES = ("defectenhancementtask", "component", "regression")
 MODELS_DIR = os.path.join(os.path.dirname(__file__), "models")
--- a/http_service/download_models.py
+++ b/http_service/download_models.py
@@ -14,7 +14,7 @@ import requests
 logging.basicConfig(level=logging.INFO)
 LOGGER = logging.getLogger()
 
-BASE_URL = "https://index.taskcluster.net/v1/task/project.relman.project.bugbug.production.train_{}.latest/artifacts/public"
+BASE_URL = "https://index.taskcluster.net/v1/task/project.relman.project.bugbug.train_{}.latest/artifacts/public"
 
 MODELS_NAMES = ("defectenhancementtask", "component", "regression")
 MODELS_DIR = os.path.join(os.path.dirname(__file__), "models")
--- a/http_service/download_models.py
+++ b/http_service/download_models.py
@@ -14,7 +14,7 @@ import requests
 logging.basicConfig(level=logging.INFO)
 LOGGER = logging.getLogger()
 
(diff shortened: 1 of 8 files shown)


infra/taskcluster-hook-pipeline-start.json
```diff
@@ -51,7 +51,9 @@
         "retries": 5,
         "routes": [
             "notify.email.release-mgmt-analysis@mozilla.com.on-failed",
-            "notify.irc-channel.#bugbug.on-failed"
+            "notify.irc-channel.#bugbug.on-failed",
+            "project.relman.services.project.bugbug.production.bugbug_data.latest",
```
marco-c: This way the task triggered by the hook (the data pipeline start task) will be indexed here, instead we have to index the actual tasks that do these things.

marco-c: We can just use project.relman.bugbug instead of project.relman.services.project.bugbug

Also, remove 'production' as we currently we only have one channel

data-pipeline.yml
```diff
@@ -48,6 +49,7 @@ tasks:
     routes:
       - notify.email.release-mgmt-analysis@mozilla.com.on-failed"
       - notify.irc-channel.#bugbug.on-failed
+      - index.project.relman.bugbug.bugbug_data.latest
```
marco-c: Two tasks can't have the same index path, or the one finishing last will overwrite the first.

We should update the trainer script to use two different URLs.

data-pipeline.yml
```diff
@@ -152,6 +157,7 @@ tasks:
     routes:
       - notify.email.release-mgmt-analysis@mozilla.com.on-failed"
       - notify.irc-channel.#bugbug.on-failed
+      - index.project.relman.bugbug.bugbug_train.latest
```
marco-c: Same here, we can't have the same index path for multiple tasks.

data-pipeline.yml
```diff
@@ -17,6 +17,7 @@ tasks:
     routes:
       - notify.email.release-mgmt-analysis@mozilla.com.on-failed"
       - notify.irc-channel.#bugbug.on-failed
+      - index.project.relman.bugbug.bugbug_data.latest
```
marco-c: Nit: given that we have project.relman.bugbug.XXX, we can avoid rewriting bugbug again. So this could be something like: index.project.relman.bugbug.data_bugzilla.latest

marco-c: Similarly with the other index names

http_service/download_models.py
```diff
@@ -14,7 +14,7 @@
 logging.basicConfig(level=logging.INFO)
 LOGGER = logging.getLogger()
 
-BASE_URL = "https://index.taskcluster.net/v1/task/project.releng.services.project.testing.bugbug_train.latest/artifacts/public"
+BASE_URL = "https://index.taskcluster.net/v1/task/project.relman.project.bugbug.production.train_{}.latest/artifacts/public"
```
marco-c: You'll need to remove the 'production.' here

http_service/download_models.py
```diff
@@ -14,7 +14,7 @@
 logging.basicConfig(level=logging.INFO)
 LOGGER = logging.getLogger()
 
-BASE_URL = "https://index.taskcluster.net/v1/task/project.releng.services.project.testing.bugbug_train.latest/artifacts/public"
+BASE_URL = "https://index.taskcluster.net/v1/task/project.relman.project.bugbug.train_{}.latest/artifacts/public"
```
marco-c: there is an additional 'project' here that should be removed

PR #335, opened 2019-05-02
Make training code more generic to make it possible to train on other kinds of objects (e.g. commits)

marco-c:
This also adds a model to predict if a commit will be backed out (not greatly performing of course, as it is quite hard to predict).


bugbug/feature_cleanup.py
```diff
@@ -0,0 +1,177 @@
...
+
+    return text
+
+
+def crash(text):
+    return re.sub(
+        r"bp-[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{6}[0-9]{6}\b",
+        "__CRASH_STATS_LINK__",
+        text,
+    )
```
Lothiraldan: You might wants to pre-compile the regex if those functions are gonna be called a lot

run.py
```diff
@@ -100,6 +103,10 @@
         from bugbug.models.assignee import AssigneeModel
 
         model_class = AssigneeModel
+    elif args.goal == "backout":
+        from bugbug.models.backout import BackoutModel
+
+        model_class = BackoutModel
 
     if args.train:
         db.download()
```
Lothiraldan: There is a bit of duplication between this script and `scripts/trainer.py`, do we want to reduce the duplication somehow?

bugbug/feature_cleanup.py
marco-c: Good idea, I'll file a follow-up for this though (as in this PR I've just moved them from one file to another).

marco-c: Filed https://github.com/mozilla/bugbug/issues/338.

run.py
marco-c: Yeah, I was thinking about the same!
Filed #339.

PR #337, opened 2019-05-03
 Add the rollback test in the data pipeline - fix #275 

### LINKED ISSUE #275: Bug snapshot test should not be a CI test, but a test run after training
It is more of an integration test.

We should run it on the DB fixtures on CI, and run it on the entire DB only after training.

Lothiraldan:
Fixes #275.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -115,7 +115,7 @@ tasks:
                pip install -r requirements.txt &&
                mkdir data/ &&
                cp tests/fixtures/bugs.json data/bugs.json &&
-               python -m bugbug.bug_snapshot --verbose"
+               python -m bugbug.bug_snapshot"
         metadata:
           name: bugbug rollback test
           description: bugbug rollback test
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -115,7 +115,7 @@ tasks:
                pip install -r requirements.txt &&
                mkdir data/ &&
                cp tests/fixtures/bugs.json data/bugs.json &&
-               python -m bugbug.bug_snapshot"
+               python -m bugbug.bug_snapshot --verbose"
         metadata:
           name: bugbug rollback test
           description: bugbug rollback test
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -113,6 +113,8 @@ tasks:
                cd bugbug &&
                git checkout ${head_rev} &&
                pip install -r requirements.txt &&
+               mkdir data/ &&
+               cp tests/fixtures/bugs.json data/bugs.json &&
                python -m bugbug.bug_snapshot"
         metadata:
           name: bugbug rollback test
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -113,7 +113,6 @@ tasks:
                cd bugbug &&
                git checkout ${head_rev} &&
                pip install -r requirements.txt &&
-               python -c 'from bugbug import bugzilla, db; db.download()' &&
                python -m bugbug.bug_snapshot"
         metadata:
           name: bugbug rollback test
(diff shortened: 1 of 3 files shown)


.taskcluster.yml
```diff
@@ -113,7 +113,8 @@ tasks:
                cd bugbug &&
                git checkout ${head_rev} &&
                pip install -r requirements.txt &&
-               python -c 'from bugbug import bugzilla, db; db.download()' &&
+               mkdir data/ &&
+               cp tests/fixtures/bugs.json data/bugs.json &&
                python -m bugbug.bug_snapshot"
```
marco-c: We could turn it into a pytest, that'd be nicer

marco-c: So we can completely remove this task

infra/data-pipeline.yml
```diff
@@ -164,6 +164,37 @@ tasks:
       owner: release-mgmt-analysis@mozilla.com
       source: https://github.com/mozilla/bugbug/raw/master/data-pipeline.yml
 
+  - ID: rollback-test-task
+    created: {$fromNow: ''}
+    deadline: {$fromNow: '49 hours'}
+    provisionerId: aws-provisioner-v1
+    workerType: releng-svc
+    dependencies:
```
marco-c: The rollback test should be able to run right after the download bugzilla task.

tests/test_bug_snapshot.py
```diff
@@ -0,0 +1,23 @@
...
+from bugbug.bug_snapshot import rollback
+
+
+def test_bug_snapshot(get_fixture_path):
+    mock_bugs_db_path = get_fixture_path("bugs.json")
+
+    with open(mock_bugs_db_path, "r") as mock_bug_db:
+
+        for i, line in enumerate(mock_bug_db):
+            bug = json.loads(line)
```
marco-c: I think you could replace all this (from line 13 to 18) with `for bug in bugzilla.get_bugs():` (it should work as the mock bug DB is stored in data/)

tests/test_bug_snapshot.py
```diff
@@ -0,0 +1,23 @@
...
+
+            print(bug["id"])
+            print(i)
```
marco-c: Nit: remove the printing

infra/data-pipeline.yml
```diff
@@ -56,6 +56,34 @@ tasks:
...
+  - ID: rollback-test-task
+    created: {$fromNow: ''}
+    deadline: {$fromNow: '25 hours'}
+    provisionerId: aws-provisioner-v1
+    workerType: releng-svc
+    dependencies:
+      - bugs-retrieval
+    payload:
+      maxRunTime: 3600
+      image: python
```
marco-c: Probably better to use the bugbug base image, so we don't have to install the requirements (and we don't have to define yet another image)

infra/data-pipeline.yml
```diff
@@ -56,6 +56,34 @@ tasks:
...
+      command:
+        - "/bin/bash"
+        - "-lcx"
+        - "git clone ${repository} &&
+           cd bugbug &&
+           git checkout ${head_rev} &&
```
marco-c: Related to the prev comment, head_rev is undefined here, so it wouldn't know what to checkout to. If we use the base image, this should be fixed automatically

tests/test_bug_snapshot.py
Lothiraldan: As we are in a loop, I think the prints will be helpful to quickly identify which bug is triggering an error. It's also helpful to check that the mock DB is correctly read by running `pytest -s`

marco-c: OK, sounds good!

PR #344, opened 2019-05-03
Use commit nodes to calculate number of previous commits touching the same directories, components and files

### LINKED ISSUE #336: Number of previous commits touching the same files/directories/components can be overcounted
See https://github.com/mozilla/bugbug/pull/301#issuecomment-488806551 and https://github.com/mozilla/bugbug/pull/301#issuecomment-488826806

sbeesm:
Fixes #336.

### DIFF
--- a/bugbug/repository.py
+++ b/bugbug/repository.py
@@ -48,11 +48,11 @@ reviewer_experience = {}
 reviewer_experience_90_days = {}
 author_experience_90_days = {}
 
-components_touched_prev = defaultdict(int)
-components_touched_prev_90_days = defaultdict(int)
+components_touched_prev = defaultdict(set)
+components_touched_prev_90_days = defaultdict(set)
 
-files_touched_prev = defaultdict(int)
-files_touched_prev_90_days = defaultdict(int)
+files_touched_prev = defaultdict(set)
+files_touched_prev_90_days = defaultdict(set)
 
 directories_touched_prev = defaultdict(set)
 directories_touched_prev_90_days = defaultdict(set)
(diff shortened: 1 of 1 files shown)


marco-c:
We should do the same for files and components, no?

Can you check how slower it is?

bugbug/repository.py
```diff
@@ -452,9 +452,9 @@ def download_commits(repo_dir, date_from):
         directories = get_directories(commit.files)
 
         for directory in directories:
-            directories_touched_prev[commit.node] += directories_touched[directory]
+            directories_touched_prev[commit.node].update(directories_touched[directory])
```
marco-c: If the memory usage is too high after this change, we could still keep directories_touched_prev as defaultdict(int) making this code a little more complex.
If the memory usage is reasonable, we can avoid it.

sbeesm: I think the memory usage would be high as we're storing a set of commit nodes for each commit in directories_touched_prev. And the size of these sets would increase as we traverse more commits.

And how can we check if the memory usage is high or not?

sbeesm:
> We should do the same for files and components, no?

Yes. Should we do it in this PR itself?

> Can you check how slower it is?

Is it somehow possible to run `repository.py` on AWS? Otherwise it takes a lot of time on my laptop (6-7 hours).

marco-c:
> Yes. Should we do it in this PR itself?

Yes.

> Is it somehow possible to run repository.py on AWS? Otherwise it takes a lot of time on my laptop (6-7 hours).

You can reduce the number of commits just for testing purposes (by simply changing the two_years_and_six_months_ago variable to something more recent).

bugbug/repository.py
marco-c: Assuming you have less than 32 GB on your laptop, if the script still works after your changes without crashing because of running out of memory, then the memory usage is still reasonable.

marco-c: Basically, the memory usage is too high if you can't run the script on your laptopt.

sbeesm: Okay.

sbeesm:
I tested the script with 4 months of recent data (10537 commits) and the time it takes is:

Time (author): 0.09422849999998562

Time (component, file, dir): 3042.9868890000002

Time (mining commits): 1407.9845606999997

marco-c:
> I tested the script with 4 months of recent data (10537 commits) and the time it takes is:

What is the time before your changes?

marco-c:
I might have found a better implementation, so let's put this on hold for a while.

sbeesm:
> What is the time before your changes?

Time (author): 0.04589599999997063

Time (component, file, dir): 2030.1190452

Time (mining commits): 1194.5243415

> I might have found a better implementation, so let's put this on hold for a while.

Okay.

marco-c:
The issue was fixed as part of #354.

Issue #338, opened 2019-05-03
Feature cleanup regexes could be compiled


marco-c:
Since they are being used thousand of times (https://github.com/mozilla/bugbug/pull/335#discussion_r280687823).

PR #351, opened 2019-05-07
Use re.compile to speedup feature cleanups

ayush-1506:
Fixes : #338 .

Each cleanup function first compiles the regex to a regex object, and then performs the `sub` on the text using this compiled regex.


ayush-1506:
Yes, I realised this later. I think making them classes is better, as it'll maintain uniformity throughout cleanup functions and feature extractors.

Issue #333, opened 2019-05-02
To calculate history features (e.g. experience), should analyze more commits


marco-c:
We will only store the newer commit data like the current state, but in order to calculate history stuff we need to start from the beginning.

PR #354, opened 2019-05-08
Optimize history features calculation for commits

marco-c:
This also fixes a couple of issues, #333 and #336.

I suggest reviewing the first two commits in isolation, and the third commit by looking at the actual code rather than the diff (since the algorithm is changing completely, looking at the diff will not make much sense).

### DIFF
--- a/bugbug/repository.py
+++ b/bugbug/repository.py
@@ -35,6 +35,7 @@ Commit = namedtuple(
         "author",
         "desc",
         "date",
+        "pushdate",
         "bug",
         "backedoutby",
         "author_email",
@@ -220,7 +221,7 @@ def _transform(commit):
 
 
 def _hg_log(revs):
-    template = '{node}\\0{author}\\0{desc}\\0{date}\\0{bug}\\0{backedoutby}\\0{author|email}\\0{join(files,"|")}\\0{join(file_copies,"|")}\\0'
+    template = '{node}\\0{author}\\0{desc}\\0{date}\\0{bug}\\0{backedoutby}\\0{author|email}\\0{join(files,"|")}\\0{join(file_copies,"|")}\\0{pushdate}\\0'
 
     args = hglib.util.cmdbuilder(
         b"log",
@@ -234,8 +235,9 @@ def _hg_log(revs):
 
     revs = []
     for rev in hglib.util.grouper(template.count("\\0"), out):
-        posixtime = float(rev[3].split(b".", 1)[0])
-        dt = datetime.fromtimestamp(posixtime)
+        date = datetime.fromtimestamp(float(rev[3].split(b".", 1)[0]))
+
+        pushdate = datetime.utcfromtimestamp(float(rev[9].split(b"-", 1)[0]))
 
         file_copies = {}
         for file_copy in rev[8].decode("utf-8").split("|"):
@@ -252,7 +254,8 @@ def _hg_log(revs):
                 node=rev[0],
                 author=rev[1],
                 desc=rev[2],
-                date=dt,
+                date=date,
+                pushdate=pushdate,
                 bug=rev[4],
                 backedoutby=rev[5],
                 author_email=rev[6],
--- a/bugbug/repository.py
+++ b/bugbug/repository.py
@@ -5,6 +5,7 @@
 
 import argparse
 import concurrent.futures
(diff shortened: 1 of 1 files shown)


bugbug/repository.py
```diff
@@ -233,8 +234,9 @@ def _hg_log(revs):
 
     revs = []
     for rev in hglib.util.grouper(template.count("\\0"), out):
-        posixtime = float(rev[3].split(b".", 1)[0])
-        dt = datetime.fromtimestamp(posixtime)
+        date = datetime.fromtimestamp(float(rev[3].split(b".", 1)[0]))
```
calixteman: You should probably use utcfromtimestamp too here else you'll have a local time here.

marco-c: I copied this from hglib, so I assumed that's the local timestamp

marco-c: Fixed with 3fd81a928eebe8c622784c9742e942a0444a8c35.

bugbug/repository.py
```diff
@@ -411,131 +352,99 @@ def download_commits(repo_dir, date_from):
...
-        )
+            experiences_by_commit["total"][experience_type][commit.node] += exp
+            experiences_by_commit["90_days"][experience_type][commit.node] += (
+                exp - experiences[day - 90][experience_type][item]
+            )
 
-        for component in components:
-            components_touched_prev[commit.node] += components_touched[component]
+            # We don't want to consider backed out commits when calculating experiences.
+            if not commit.backedoutby:
```
calixteman: I'd say that even backed out patches contribute to your experience and maybe your learn more in such a case...
Maybe we could create a special experience for backed out patches: it could help to measure "seniority" (even if I'm not sure of the definition of this word).

bugbug/repository.py
```diff
@@ -411,131 +352,99 @@ def download_commits(repo_dir, date_from):
...
     }
 
-    global components_touched_prev
-    global components_touched_prev_90_days
+    first_pushdate = commits[0].pushdate
 
-    global files_touched_prev
-    global files_touched_prev_90_days
+    experiences = defaultdict(lambda: defaultdict(lambda: defaultdict(int)))
+    complex_experiences = defaultdict(lambda: defaultdict(lambda: defaultdict(set)))
```
calixteman: nit: doc what complex experiences are... it's unclear for me.

bugbug/repository.py
```diff
@@ -411,131 +352,99 @@ def download_commits(repo_dir, date_from):
...
-        directories_touched_90_days = defaultdict(int)
-        for prev_commit in prev_commits_90_days:
-            components_prev = set(
+                if orig in path_to_component and copied in path_to_component:
+                    orig_component = path_to_component[orig]
+                    copied_component = path_to_component[copied]
+                else:
+                    orig_component = copied_component = None
+
+                for prev_day in complex_experiences.keys():
```
calixteman: nit: move the "for" in the previous "if" to avoid the "else" and check for None in the for loop

bugbug/repository.py
marco-c: It's for the case where you need to have a set of commits.
E.g. say commit A is modifying "dom" and "apps", commit B is modifying "apps", commit C is modifying "dom" and "apps".

For commit C, you should have that the number of previous commits touching the same dirs is 2 (A and B), but if you just use the sum you will have 3 (A twice + B)

bugbug/repository.py
marco-c: I mainly didn't want to overcount them, as most of the cases a patch is backed out and then relanded after a fix.
I'll file a follow-up issue to have a special experience for backed out patches.

marco-c: Filed #373.

bugbug/repository.py
marco-c: I'd need to do two for loops this way

bugbug/repository.py
marco-c: Fixed with 90bf7cf006ec218059e1a298f452c4314c5246aa.

marco-c:
> nit: I'd put 90 in a global variable or in a conf file.

Fixed with fe1cda4fbc7555a0e03fecee3bca0c8e767c8a2d.

PR #362, opened 2019-05-09
Try updating the hook before trigerring the route. ref #345

### LINKED ISSUE #345: On a new release, the hook is triggered in its old version rather than the new
It looks like the hook is triggered before being updated. Maybe we have to stop using the route and use the insertTask API here. We need to check with Taskcluster.

Lothiraldan:
Fixes #345.


marco-c:
Oh, the route was on the docker push task, that's why!

Issue #372, opened 2019-05-09
Add some complexity measures of modified files


marco-c:
- [x] Size of modified file (before? after? both?) - Done in 97d514e1dbb24a1faa649feda5cc746ed7053fef.
- [x] Size of modified function (before? after? both?) - Done in 66d17b05d6fa624d9ed80c9202849ba61494f3ce.
- [x] Cyclomatic complexity (of the modified functions?)
- [ ] Number of tokens (whole file? modified functions? both?)
- [ ] Parameter count of functions (whole file? modified functions? both?)
- [ ] Number of exit points for functions
- [ ] Dependency count (e.g. number of #include, number of imports, etc.).
- [ ] What else?

There is a tool that does this (https://terryyin.github.io/lizard/), but it doesn't support all languages.

The solution for this would probably have something in common with the solution for #371.

marco-c:
I've closed this as we're done with the first iteration, and filed a few issues to track further additions.

marco-c:
I've closed this as we're done with the first iteration, and filed a few issues to track further additions.

Issue #371, opened 2019-05-09
Differentiate between code lines added and comment lines added


marco-c:
A patch that adds one line of code and ten lines of comments is likely different than a patch which adds eleven lines of comments!

marco-c:
One way to do this is:
- Get the files before/after the commit with hg cat
- Remove comments from the files before/after using some language agnostic tool
- Calculate the diff of the two

Another way would be to just remove lines starting with // or between /* and */ in a diff, but this wouldn't be precise (in the case of multiline commits for example, if the diff is in the middle of the multiline commit).

The first option is likely way slower than the second, but it is more precise and would probably allow us to also do #372 more easily.

marco-c:
For removing comments, we could use a generic parser like pygments or tree-sitter.

sbeesm:
> * Get the files before/after the commit with hg cat

With hg cat, we also need to add the revision right? So how would we find the revision number for a particular commit?


> * Remove comments from the files before/after using some language agnostic tool

And we only want to know code lines added and not comment lines?

marco-c:
> With hg cat, we also need to add the revision right? So how would we find the revision number for a particular commit?

We can use the node of the commit (see in repository.py, we are already using hg cat now).

> And we only want to know code lines added and not comment li

Ideally both.

sbeesm:
Okay, so with hg cat, we get the after file and for the before file, we'd use the previous commit from the list, right?

But why do we need to remove comment lines if we want to calculate both?

marco-c:
> Okay, so with hg cat, we get the after file and for the before file, we'd use the previous commit from the list, right?

Or we store the parent of the commit for each commit, getting it in hg_log.

> But why do we need to remove comment lines if we want to calculate both?

We don't *have to*, it's just a possible way to implement this.

sbeesm:
> Or we store the parent of the commit for each commit, getting it in hg_log.

Oh yes, that would be a better option.

And for parsing files, I was thinking of first reading the whole file into a string then splitting into list of lines. For single line comments, we can just check the first char so parsing them should be easy. But for multi-line comments we can have a loop starting when we encounter ```/*``` or ```'''```, and keep counting the lines until we find the end.

marco-c:
[...]

PR #394, opened 2019-05-10
Collect size of files modified by the patch

marco-c:
This can be considered as one of the complexity measures (see #372).

(Depends on #377, I'll rebase this after it's merged)

### DIFF
--- a/bugbug/repository.py
+++ b/bugbug/repository.py
@@ -101,7 +101,8 @@ def get_commits():
 
 def _init(repo_dir):
     global HG
-    HG = hglib.open(repo_dir)
+    os.chdir(repo_dir)
+    HG = hglib.open(".")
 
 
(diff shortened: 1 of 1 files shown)


bugbug/repository.py
```diff
@@ -211,6 +214,23 @@ def _transform(commit):
             type_ = ext
         obj["types"].add(type_)
 
+        try:
```
calixteman: Maybe it'd be faster to keep # of added/removed lines for each revision and after getting all the log, computing the size of the file for each revision.

marco-c: I've thought about that, but then we'd need to analyze all the revisions in sequence and we couldn't use multiple processes to analyze them.

calixteman: I'd guess that it's probably faster to have two runs even if the second is single threaded.
Maybe we could write an extension to just get the '\n' count without retrieving all the file.

marco-c: Yeah, but in the end we'll probably need to hg.cat the file anyway for #371 and other parts of #372, so it might not be worth optimizing this now if we have to do this anyway later.

Issue #20, opened 2018-12-05
Create a classifier to detect the type of bug


marco-c:
Needed for https://github.com/mozilla/relman-auto-nag/issues/223.

To do this, we'll need to collect some labels.
We can automatically create some by using the `perf` and `memory` keywords.

There are other types that we could try to classify other than perf and memory:
- Security - Security vulnerability.
- Crash - Program unexpectedly stops running.
- Hang - Program keeps running but without response.
- Incorrect rendering - Components or video cannot be correctly rendered.
- Wrong functionality - Incorrect functionalities besides rendering issues.
- Web incompatibility - Program does not work correctly for a major website or many websites due to incompatible APIs or libraries, or a functionality, which was removed on purpose, but is still used in the wild.
- Add-on or plug-in incompatibility - Program does not work correctly for a major add-on/plug-in or many add-ons/plug-ins due to incompatible APIs or libraries, or a functionality, which was removed on purpose, but is still used in the wild.
- Compile - Compilation errors.
- Feature - Introduce or remove features.
- Improvement - Minor functional or aesthetical improvement.
- Test-only problem - Errors that only break tests.
- Other - Other types that don't belong to the previous categories, e.g., data corruption and license incompatibility.

marco-c:
We can automatically create some also by using the `crash` keyword, and some keywords for the security bugs (listed in https://wiki.mozilla.org/Security_Severity_Ratings).

ayush-1506:
Should this can be closed now?

marco-c:
Not yet, as we don't have any labels for many of the types in the first comment.

PR #395, opened 2019-05-10
Multilabel classifier for detecting type of bug

ayush-1506:
Initial fix for #20

A few questions regarding this classifier:

1. [class_names](https://github.com/mozilla/bugbug/blob/master/bugbug/model.py#L66) would need to be redefined, as it isn't valid for this classifier. (get_labels returns a dict where values are a list, not a string). I think the best option would be to remove it from the `train` method and make it an attribute, initializing it in the constructor. However, I'd need a green light before changing this.

2. Similar to the presence of various tags for security bugs (https://github.com/mozilla/bugbug/issues/20#issuecomment-491064287), where can I find other tags that actually refer to the same bug type?

@marco-c

### DIFF
--- a/bugbug/models/assignee.py
+++ b/bugbug/models/assignee.py
@@ -105,11 +105,13 @@ class AssigneeModel(BugModel):
         for assignee, count in assignee_counts:
             print(f"{assignee}: {count}")
 
-        return {
+        classes = {
             bug_id: assignee
             for bug_id, assignee in classes.items()
             if assignee in top_assignees
         }
+
+        return classes, set(classes.values())
 
     def get_feature_names(self):
         return self.extraction_pipeline.named_steps["union"].get_feature_names()
--- a/bugbug/models/backout.py
+++ b/bugbug/models/backout.py
@@ -75,7 +75,7 @@ class BackoutModel(CommitModel):
         for commit_data in repository.get_commits():
             classes[commit_data["node"]] = 1 if commit_data["ever_backedout"] else 0
 
-        return classes
+        return classes, [0, 1]
 
     def get_feature_names(self):
         return self.extraction_pipeline.named_steps["union"].get_feature_names()
--- a/bugbug/models/bugtype.py
+++ b/bugbug/models/bugtype.py
@@ -3,18 +3,32 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
-import itertools
-
+import numpy as np
 import xgboost
 from imblearn.over_sampling import BorderlineSMOTE
 from sklearn.compose import ColumnTransformer
 from sklearn.feature_extraction import DictVectorizer
-from sklearn.pipeline import Pipeline
 from sklearn.multiclass import OneVsRestClassifier
+from sklearn.pipeline import Pipeline
 
-from bugbug import bug_features, bugzilla, feature_cleanup, labels
+from bugbug import bug_features, bugzilla, feature_cleanup
 from bugbug.model import BugModel
-import numpy as np
+
+keyword_dict = {
+    "sec-critical": "security",
+    "sec-high": "security",
+    "sec-moderate": "security",
+    "sec-low": "security",
+    "sec-other": "security",
+    "sec-audit": "security",
+    "sec-vector": "security",
+    "sec-want": "security",
+    "memory-footprint": "memory",
+    "memory-leak": "memory",
+    "crash": "crash",
+    "crashreportid": "crash",
+    "perf": "performance",
+}
 
 
 class BugTypeModel(BugModel):
@@ -28,7 +42,7 @@ class BugTypeModel(BugModel):
             bug_features.severity(),
             # Ignore keywords that would make the ML completely skewed
             # (we are going to use them as 100% rules in the evaluation phase).
-            bug_features.keywords({"regression", "talos-regression", "feature"}),
+            bug_features.keywords(set(keyword_dict.keys())),
             bug_features.is_coverity_issue(),
             bug_features.has_crash_signature(),
             bug_features.has_url(),
@@ -83,20 +97,16 @@ class BugTypeModel(BugModel):
 
     def get_labels(self):
         classes = {}
-        keyword_list = []
-        for bug_data in bugzilla.get_bugs():
-            for keyword in bug_data["keywords"]:
-                if keyword not in keyword_list:
-                    keyword_list.append(keyword)
+        keyword_list = list(set(keyword_dict.values()))
 
         for bug_data in bugzilla.get_bugs():
             target = np.zeros(len(keyword_list))
             for keyword in bug_data["keywords"]:
-                target[keyword_list.index(keyword)] = 1
+                target[keyword_list.index(keyword_dict[keyword])] = 1
 
             classes[int(bug_data["id"])] = target
 
-        return classes
+        return classes, keyword_list
 
     def get_feature_names(self):
         return self.extraction_pipeline.named_steps["union"].get_feature_names()
--- a/bugbug/models/bugtype.py
+++ b/bugbug/models/bugtype.py
@@ -1,0 +1,102 @@
+# -*- coding: utf-8 -*-
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import itertools
+
+import xgboost
+from imblearn.over_sampling import BorderlineSMOTE
+from sklearn.compose import ColumnTransformer
+from sklearn.feature_extraction import DictVectorizer
+from sklearn.pipeline import Pipeline
+from sklearn.multiclass import OneVsRestClassifier
+
+from bugbug import bug_features, bugzilla, feature_cleanup, labels
+from bugbug.model import BugModel
+import numpy as np
+
+
+class BugTypeModel(BugModel):
+    def __init__(self, lemmatization=False, historical=False):
+        BugModel.__init__(self, lemmatization)
+
+        self.sampler = BorderlineSMOTE(random_state=0)
+
+        feature_extractors = [
+            bug_features.has_str(),
+            bug_features.severity(),
+            # Ignore keywords that would make the ML completely skewed
+            # (we are going to use them as 100% rules in the evaluation phase).
+            bug_features.keywords({"regression", "talos-regression", "feature"}),
+            bug_features.is_coverity_issue(),
+            bug_features.has_crash_signature(),
+            bug_features.has_url(),
+            bug_features.has_w3c_url(),
+            bug_features.has_github_url(),
+            bug_features.whiteboard(),
+            bug_features.patches(),
+            bug_features.landings(),
+            bug_features.title(),
+            bug_features.blocked_bugs_number(),
+            bug_features.ever_affected(),
+            bug_features.affected_then_unaffected(),
+            bug_features.product(),
+            bug_features.component(),
+        ]
+
+        cleanup_functions = [
+            feature_cleanup.url(),
+            feature_cleanup.fileref(),
+            feature_cleanup.synonyms(),
+        ]
+
+        self.extraction_pipeline = Pipeline(
+            [
+                (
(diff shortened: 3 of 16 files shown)


bugbug/models/bugtype.py
```diff
@@ -0,0 +1,102 @@
...
+        BugModel.__init__(self, lemmatization)
+
+        self.sampler = BorderlineSMOTE(random_state=0)
+
+        feature_extractors = [
+            bug_features.has_str(),
+            bug_features.severity(),
+            # Ignore keywords that would make the ML completely skewed
+            # (we are going to use them as 100% rules in the evaluation phase).
+            bug_features.keywords({"regression", "talos-regression", "feature"}),
```
marco-c: Here we don't need to remove these keywords, but we need to remove the keywords we are using in get_labels

bugbug/models/bugtype.py
```diff
@@ -0,0 +1,102 @@
...
+                    ),
+                ),
+            ]
+        )
+
+        self.clf = OneVsRestClassifier(xgboost.XGBClassifier(n_jobs=16))
+
+    def get_labels(self):
+        classes = {}
+        keyword_list = []
```
marco-c: Why are you using all keywords as labels?

ayush-1506: Any value which comes in keywords would be a part of the label, right? (I need to group these according to their type, as multiple values can be of a single bug type.)

If not, is there a list of all available keywords that I should be using?

bugbug/models/bugtype.py
ayush-1506: Ah, right. Will fix that.

bugbug/models/bugtype.py
```diff
@@ -0,0 +1,102 @@
...
+                                self.text_vectorizer(min_df=0.001),
+                                "comments",
+                            ),
+                        ]
+                    ),
+                ),
+            ]
+        )
+
+        self.clf = OneVsRestClassifier(xgboost.XGBClassifier(n_jobs=16))
```
marco-c: We could also try using xgboost support for multilabel classification (as a follow-up, not in this PR).

bugbug/models/bugtype.py
marco-c: There are keywords for a lot of disparate things, they don't only specify type. Also, some keywords can be grouped together to mean a single type.

There is an initial list of keywords in #20 (perf, memory, crash, keywords from https://wiki.mozilla.org/Security_Severity_Ratings).
We need to discover new ones we can use too, but that can be done later (iterate all bugs, put all keywords in a set, print them and see which ones we could use).

ayush-1506: For a start, should I implement only `crash` vs `perf` vs `memory` vs `security` in this PR? And we can keep adding others as we find them.

marco-c:
> class_names would need to be redefined, as it isn't valid for this classifier. (get_labels returns a dict where values are a list, not a string). I think the best option would be to remove it from the train method and make it an attribute, initializing it in the constructor. However, I'd need a green light before changing this.

We have three options:
1) Make get_labels return two values, the classes dict + a list of class names, and remove the class_names = sorted(list(set(classes.values())), reverse=True) line from model.py
2) Add a new function 'def get_class_names(classes)' to model.py and make it `class_names = sorted(list(set(classes.values())), reverse=True)` by default. In your model, override it and return a hardcoded list of class names.

1 should be faster at runtime, as we can avoid iterating the dict, but the difference is negligible.

I'd prefer not to make it an attribute, as some models calculate it in get_labels.

bugbug/models/bugtype.py
marco-c: Yes!

bugbug/models/assignee.py
```diff
@@ -105,11 +105,19 @@ def get_labels(self):
...
-            for bug_id, assignee in classes.items()
-            if assignee in top_assignees
-        }
+        return (
+            {
+                bug_id: assignee
+                for bug_id, assignee in classes.items()
+                if assignee in top_assignees
+            },
+            sorted(
```
marco-c: Let's move the sorting in model.py, so we don't have to replicate the code in each model.

bugbug/models/assignee.py
```diff
@@ -105,11 +105,19 @@ def get_labels(self):
...
+                list(
+                    set(value for value in classes.values() if value in top_assignees)
+                ),
+                reverse=True,
+            ),
+        )
```
marco-c: Nit: you can do:
```
classes = XXX
return classes, list(set(classes.values()))
```

bugbug/models/bugtype.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def get_labels(self):
+        classes = {}
+        keyword_list = ["performance", "memory", "crash", "security"]
+
+        def bug_type(keyword):
+            keyword_dict = {
```
marco-c: Define this keyword_dict as a global variable at the beginning of the file, so you can also use it instead of keywords_to_ignore (and avoid defining the same keywords multiple times).

bugbug/models/bugtype.py
```diff
@@ -0,0 +1,133 @@
...
+                    "sec-vector",
+                    "sec-want",
+                ],
+                "memory": ["memory-footprint", "memory-leak"],
+                "crash": ["crash", "crashreportid"],
+                "performance": ["perf"],
+            }
+            for key, val in keyword_dict.items():
+                if keyword in val:
+                    return key
```
marco-c: It would be better if the dict was:
KEYWORD -> TYPE

This way you don't have to iterate, but you can just access the dict.

bugbug/models/bugtype.py
```diff
@@ -0,0 +1,133 @@
...
+                    ),
+                ),
+            ]
+        )
+
+        self.clf = OneVsRestClassifier(xgboost.XGBClassifier(n_jobs=16))
+
+    def get_labels(self):
+        classes = {}
+        keyword_list = ["performance", "memory", "crash", "security"]
```
marco-c: This too can be generated from the new global variable you will add

bugbug/models/component.py
```diff
@@ -181,11 +181,19 @@ def get_labels(self):
...
+                for bug_id, component in classes.items()
+                if component in top_components
+            },
+            sorted(
+                list(
+                    set(value for value in classes.values() if value in top_components)
+                ),
+                reverse=True,
+            ),
+        )
```
marco-c: Same comments as `assignee`

bugbug/models/defect.py
```diff
@@ -243,7 +243,10 @@ def get_bugbug_labels(self, kind="bug"):
         return {bug_id: label for bug_id, label in classes.items() if bug_id in bug_ids}
 
     def get_labels(self):
-        return self.get_bugbug_labels("bug")
+        return (
+            self.get_bugbug_labels("bug"),
+            sorted(list(set(self.get_bugbug_labels("bug").values())), reverse=True),
```
marco-c: We have a predefined set of classes here, you can just return them directly (1 and 0).

bugbug/models/defect_enhancement_task.py
```diff
@@ -27,7 +27,7 @@ def get_labels(self):
             "{} tasks".format(sum(1 for label in classes.values() if label == "task"))
         )
 
-        return classes
+        return classes, sorted(list(set(classes.values())), reverse=True)
```
marco-c: Same here, we have a defined set of labels (`task`, `enhancement` and `defect`).

bugbug/models/devdocneeded.py
```diff
@@ -108,7 +108,7 @@ def get_labels(self):
             if bug_id not in classes:
                 classes[bug_id] = 0
 
-        return classes
+        return classes, sorted(list(set(classes.values())), reverse=True)
```
marco-c: Same here, 1 and 0.

bugbug/models/qaneeded.py
```diff
@@ -90,7 +90,7 @@ def get_labels(self):
             if bug_id not in classes:
                 classes[bug_id] = 0
 
-        return classes
+        return classes, sorted(list(set(classes.values())), reverse=True)
```
marco-c: Same here, 1 and 0.

bugbug/models/regression.py
```diff
@@ -11,4 +11,9 @@ def __init__(self, lemmatization=False):
         DefectModel.__init__(self, lemmatization)
 
     def get_labels(self):
-        return self.get_bugbug_labels("regression")
+        return (
+            self.get_bugbug_labels("regression"),
+            sorted(
+                list(set(self.get_bugbug_labels("regression").values())), reverse=True
+            ),
+        )
```
marco-c: Same here, 1 and 0.

bugbug/models/tracking.py
```diff
@@ -113,7 +113,7 @@ def get_labels(self):
             if bug_id not in classes:
                 classes[bug_id] = 0
 
-        return classes
+        return classes, sorted(list(set(classes.values())), reverse=True)
```
marco-c: Same here, 1 and 0.

bugbug/models/uplift.py
```diff
@@ -93,7 +93,7 @@ def get_labels(self):
                     elif flag["status"] == "-":
                         classes[bug_id] = 0
 
-        return classes
+        return classes, sorted(list(set(classes.values())), reverse=True)
```
marco-c: Same here, 1 and 0.

run.py
```diff
@@ -37,6 +37,7 @@
             "assignee",
             # commit classifiers
             "backout",
+            "bugtype",
```
marco-c: This should be in the `# bug classifiers` section (that is, after `assignee`).

tests/test_bug.py
```diff
@@ -8,7 +8,7 @@
 
 def test_get_bug_labels():
     model = DefectModel()
-    classes = model.get_labels()
+    classes = model.get_labels()[0]
```
marco-c: Nit: classes, _ = model.get_labels() is nicer

ayush-1506:
Done, please review.

bugbug/models/backout.py
```diff
@@ -75,7 +75,7 @@ def get_labels(self):
         for commit_data in repository.get_commits():
             classes[commit_data["node"]] = 1 if commit_data["ever_backedout"] else 0
 
-        return classes
+        return classes, set(classes.values())
```
marco-c: This can just be 1 and 0

bugbug/models/defect.py
```diff
@@ -243,7 +243,7 @@ def get_bugbug_labels(self, kind="bug"):
         return {bug_id: label for bug_id, label in classes.items() if bug_id in bug_ids}
 
     def get_labels(self):
-        return self.get_bugbug_labels("bug")
+        return (self.get_bugbug_labels("bug"), {0, 1})
```
marco-c: Nit: the 0 and 1 can be in a list instead of a set

marco-c: Since you are transforming to a list in model.py, you might as well have a list directly here (as you don't need the set to eliminate duplicate values like in other cases)

marco-c: (Similarly for the other models where you have a predefined list of classes)

bugbug/models/bugtype.py
```diff
@@ -0,0 +1,112 @@
...
+            target = np.zeros(len(keyword_list))
+            for keyword in bug_data["keywords"]:
+                target[keyword_list.index(keyword_dict[keyword])] = 1
+
+            classes[int(bug_data["id"])] = target
+
+        return classes, keyword_list
+
+    def get_feature_names(self):
+        return self.extraction_pipeline.named_steps["union"].get_feature_names()
```
marco-c: As a follow-up, we should also add a `overwrite_classes` method (as when we have the keywords, we can be 100% sure of the type).

ayush-1506:
Sure, I'll add this in a fresh PR.

bugbug/models/bugtype.py
marco-c: Filed https://github.com/mozilla/bugbug/issues/403.

bugbug/models/bugtype.py
marco-c: Filed #404 for this.

PR #398, opened 2019-05-12
Add a central place where the models are defined - fix #367

### LINKED ISSUE #367: Create a module listing all models
See https://github.com/mozilla/bugbug/pull/341#discussion_r282297696.

Lothiraldan:
Fixes #367.

Also add some helpers to load a model. I'm wondering if we should also put the
model retrieval code inside the same module? I'm thinking about this code
https://github.com/mozilla/bugbug/pull/321/files#diff-7908c821177e4af7e08af4a9ca055d60R41

bugbug/run.py is doing also this kind of dispatch, we might want to merge the
two implementations later but bugbug/run.py is much more complex.


bugbug/models/__init__.py
```diff
@@ -1 +1,28 @@
...
+    "component": ComponentModel,
+    "regression": RegressionModel,
+    "bug": BugModel,
+}
+
+
+def load_model(model_name, model_dir=None):
+    if model_dir is None:
+        model_dir = "."
+
```
marco-c: Could you add an assertion that `model_name` is in MODELS?

The code would fail anyway in line 27 if that's not the case, but it'd be nice to have a better error message.

comment_level_labeler.py
```diff
@@ -20,13 +21,9 @@
...
 if args.goal == "str":
-    from bugbug.models.bug import BugModel
-
-    model = BugModel.load("bugmodel")
+    model = load_model("bug")
 elif args.goal == "regressionrange":
-    from bugbug.models.regression import RegressionModel
-
-    model = RegressionModel.load("regressionmodel")
+    model = load_model("regression")
```
marco-c: Could you apply the same treatment to run.py?

marco-c:
> bugbug/run.py is doing also this kind of dispatch, we might want to merge the
two implementations later but bugbug/run.py is much more complex.

We could partially do this already

marco-c:
> We could partially do this already

I mean, a least using this newly introduced method

marco-c:
> I'm wondering if we should also put the
model retrieval code inside the same module? I'm thinking about this code
https://github.com/mozilla/bugbug/pull/321/files#diff-7908c821177e4af7e08af4a9ca055d60R41

I'd like to keep the infra code as separated as possible from the core code, so that the core code could be used by others with other infra.

bugbug/models/component_nn.py
```diff
@@ -21,22 +21,6 @@
     "Optional dependencies are missing, install them with: pip install bugbug[nn]\n"
 )
 
-try:
```
marco-c: Could we import the model modules only when load_model is called, so we avoid these changes?

Lothiraldan: I will try to see how we could do it in a proper way

marco-c:
Just noticed we can use this module in scripts/trainer.py too.

Lothiraldan:
@marco-c Should be ready for merging now

bugbug/models/component_nn.py
```diff
@@ -21,6 +21,7 @@
     "Optional dependencies are missing, install them with: pip install bugbug[nn]\n"
 )
 
+
```
marco-c: Nit: extraneous whitespace change

scripts/trainer.py
```diff
@@ -73,21 +37,22 @@ def go(self, model):
...
-            raise Exception("valid_models is likely not up-to-date anymore")
+        log_messages = {
+            "component": "Training *component* model",
+            "defectenhancementtask": "Training *defect vs enhancement vs task* model",
+            "regression": "Training *regression vs non-regression* model",
+            "training": "Training *tracking* model",
+        }
+
+        if log_messages.get(model_name):
+            logger.info(log_messages[model_name])
```
marco-c: We could simply do `Training *{model}* name` (it's not exactly the same, but it works anyway and will require less code).

scripts/trainer.py
```diff
@@ -37,15 +37,7 @@ def go(self, model_name):
...
-        log_messages = {
-            "component": "Training *component* model",
-            "defectenhancementtask": "Training *defect vs enhancement vs task* model",
-            "regression": "Training *regression vs non-regression* model",
-            "training": "Training *tracking* model",
-        }
-
-        if log_messages.get(model_name):
-            logger.info(log_messages[model_name])
+        logger.info(f"Training *{model_name}* name")
```
marco-c: s/name/model :P

marco-c: (my fault, I see I used the wrong string in my comment earlier)

marco-c: So this should be logger.info(f"Training *{model_name}* model")

tests/test_models.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+def test_import_all_models():
+    """ Try loading all defined models to ensure that their full qualified
+    names are still good
+    """
+
+    for model_name in MODELS.keys():
+        print("Try loading model", model_name)
+        get_model_class(model_name)
```
marco-c: The test is failing because the optional dependencies are not installed.
We could either catch the optional dependencies exception and ignore it, or just make sure the models exist without importing them, or actually install the optional dependencies.

Lothiraldan: Let's install the optional dependencies, if we start testing the model we would need them no matter what

scripts/trainer.py
Lothiraldan: It should be fixed now.

I should have checked it before pasting.

PR #411, opened 2019-05-14
Add a pre-commit hook using codespell

### LINKED ISSUE #410: Use codespell in precommit hook

marco-c:
Fixes #410


Lothiraldan:
Why not defining a mirror repository as codespell is written in Python? It should be pretty easy bit requires maintenance for each new version

Here is for example the one for isort: https://github.com/pre-commit/mirrors-isort

marco-c:
Good idea, done!

PR #416, opened 2019-05-15
Replace parsepatch with rs_parsepatch

calixteman:


bugbug/repository.py
```diff
@@ -17,7 +17,7 @@
 import hglib
 import requests
 from dateutil.relativedelta import relativedelta
-from parsepatch.patch import Patch
+import rs_parsepatch as rs_pp
```
marco-c: Nit: I'd prefer a more explanatory name, e.g. leave it as `rs_parsepatch`, or `parsepatch`

bugbug/repository.py
```diff
@@ -216,15 +219,14 @@ def _transform(commit):
...
-            else:
-                raise
-
-        sizes.append(after.count(b"\n"))
+        if not stats["deleted"] and not stats["binary"]:
+            try:
+                after = HG.cat([path.encode("utf-8")], rev=commit.node)
+                sizes.append(after.count(b"\n"))
+            except hglib.error.CommandError as e:
+                print((e, stats))
```
marco-c: What error can we have here?

bugbug/repository.py
```diff
@@ -216,15 +219,14 @@ def _transform(commit):
...
-        except hglib.error.CommandError as e:
-            if b"no such file in rev" in e.err:
-                after = b""
-            else:
-                raise
-
-        sizes.append(after.count(b"\n"))
+        if not stats["deleted"] and not stats["binary"]:
+            try:
+                after = HG.cat([path.encode("utf-8")], rev=commit.node)
```
marco-c: Doesn't the rev= need to be bytes?

bugbug/repository.py
calixteman: If a file has been deleted but rs_pp didn't see it then we'll catch it here

marco-c: OKOK, can we remove the print statement?

PR #430, opened 2019-05-15
overwrite_classes for stepstoreproduce model

ayush-1506:
Fixes : #424

### DIFF
--- a/bugbug/models/stepstoreproduce.py
+++ b/bugbug/models/stepstoreproduce.py
@@ -92,5 +92,16 @@ class StepsToReproduceModel(BugModel):
 
         return classes, [0, 1]
 
+    def overwrite_classes(self, bugs, classes, probabilities):
+        for i, bug in enumerate(bugs):
+            if bug["cf_has_str"] == "no":
+                classes[i] = 0 if not probabilities else [1.0, 0.0]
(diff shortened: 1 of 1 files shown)


bugbug/models/stepstoreproduce.py
```diff
@@ -92,5 +92,20 @@ def get_labels(self):
...
 
+    def overwrite_classes(self, bugs, classes, probabilities):
+        for i, bug in enumerate(bugs):
+            if bug["cf_has_str"] == "no":
+                classes[i] = 0 if not probabilities else [1.0, 0.0]
+            elif bug["cf_has_str"] == "yes":
+                classes[i] = 1 if not probabilities else [0.0, 1.0]
+            elif "stepswanted" in bug["keywords"]:
+                classes[i] = 0 if not probabilities else [1.0, 0.0]
+            else:
```
marco-c: Let's remove this part about the history, as it's too slow compared to the other, quicker, checks.

ayush-1506: Now that I think of it, wouldn't bugs that don't fall into the first three categories obviously fall into this category? So we can just replace it with an else statement.

ayush-1506: Oh, maybe not. My bad.

Issue #427, opened 2019-05-15
Add stepstoreproduce training task


marco-c:

ayush-1506:
@marco-c What is a training task in this context? Is there an example of this for some other model?

marco-c:
[Taskcluster](https://docs.taskcluster.net/docs) runs our pipeline, which is defined in infra/data-pipeline.yml. Each item of that YAML file defines a task that is run on Taskcluster. A training task is a task that performs the training of a model.

Here is an example for the component model: https://github.com/mozilla/bugbug/blob/73a5192073ed6e805afc7cadc8143d68db5c38ef/infra/data-pipeline.yml#L88, with the Docker image defined at https://github.com/mozilla/bugbug/blob/73a5192073ed6e805afc7cadc8143d68db5c38ef/infra/dockerfile.train_component#L1.

PR #441, opened 2019-05-15
Add stepstoreproduce training task

ayush-1506:
Fixes : #427


infra/dockerfile.train_stepstoreproduce
```diff
@@ -0,0 +1,3 @@
+FROM mozilla/bugbug-base:latest
+
+CMD bugbug-train stepstoreproduce
```
marco-c: Can you try if this command works for you locally? You'll probably need to update some other file in order for it to run successfully.

ayush-1506: How do I run these tasks locally?

marco-c: For the purpose of this PR, you can just run this command locally. bugbug-train is defined in setup.py, and it corresponds to scripts.trainer:main.
So you can run `python3 -c "from scripts import trainer; trainer.main()" stepstoreproduce`.

ayush-1506: Yes, this needed to be added to the trainer.py script too. I've added this now.

marco-c: Thanks. We'll need to wait for an image on Docker Hub to be created for this.

marco-c:
@ayush-1506 could you resolve the conflicts in the PR?

infra/dockerfile.train_stepstoreproduce
marco-c: The image has been added, so we can now merge this.

PR #445, opened 2019-05-16
[WIP] Calculate experiences in parallel

marco-c:
Need to check how much memory is used. If it is too much, I could reduce the number of processes to 2/3 instead of len(tasks).


marco-c:
I need to try rebasing this on top of current master and see if it makes a difference.

marco-c:
This is not worth it anymore, as experience calculation is pretty fast now.
We can re-evaluate it in case we need to add more experience types in the future.

PR #458, opened 2019-05-17
Prepare heroku deployment.

### LINKED ISSUE #376: Deploy http service somewhere
Either Heroku, or GCP, or AWS.
Depends on the task-boot issue to support deploying to one of those services.

Lothiraldan:
Fixes #376.


.taskcluster.yml
```diff
@@ -224,6 +224,42 @@ tasks:
             owner: ${user}@users.noreply.github.com
             source: ${repository}/raw/${head_rev}/.taskcluster.yml
 
+      - $if: 'tasks_for == "github-push" && head_branch[:10] == "refs/tags/"'
```
marco-c: We should deploy at the end of the data pipeline

http_service/Dockerfile
```diff
@@ -18,4 +18,4 @@ ENV CHECK_MODELS="${CHECK_MODELS}"
 
 RUN python /code/bugbug_http_service/check_models.py
 
-CMD ["gunicorn", "-b", "0.0.0.0:8000", "bugbug_http_service.app", "--preload", "--timeout", "30", "-w", "3"]
+CMD gunicorn -b 0.0.0.0:$PORT bugbug_http_service.app --preload --timeout 30 -w 3
```
marco-c: Is the other format not OK on HEroku?

http_service/models.py
```diff
@@ -0,0 +1,70 @@
...
+LOGGER = logging.getLogger()
+
+MODELS = {
+    "defectenhancementtask": DefectEnhancementTaskModel,
+    "component": ComponentModel,
+    "regression": RegressionModel,
+}
+MODELS_NAMES = MODELS.keys()
+MODELS_DIR = os.path.join(os.path.dirname(__file__), "models")
+BASE_URL = "https://index.taskcluster.net/v1/task/project.releng.services.project.testing.bugbug_train.latest/artifacts/public"
```
marco-c: We should use the new URL

http_service/models.py
```diff
@@ -0,0 +1,70 @@
...
+import requests
+
+from bugbug.models.component import ComponentModel
+from bugbug.models.defect_enhancement_task import DefectEnhancementTaskModel
+from bugbug.models.regression import RegressionModel
+
+logging.basicConfig(level=logging.INFO)
+LOGGER = logging.getLogger()
+
+MODELS = {
```
marco-c: I wonder if we couldn't reuse at least part of the central place where we define models

.taskcluster.yml
Lothiraldan: If we do that, that means we need to wait for a full data-pipeline in order to deploy any code change, including a hotfix, that's sub-obtimal. But we are generating an HTTP Service Docker image without the models so it's effectively broken.

I'm not fan of either solutions, @marco-c what do you think?

http_service/Dockerfile
Lothiraldan: No the other format couldn't be updated as the list-based format is directly executed by Docker. This format means that it's wrapped by bash for argument parsing and the nice side-effect is that bash is doing the variable interpolation for us.

http_service/models.py
Lothiraldan: Yes now that we have merged the centralization PR, we should, let me update this PR

.taskcluster.yml
marco-c: For now we have to do the non-broken thing, we can figure out how to deploy fixes later (file an issue so we don't forget).

Lothiraldan: Done https://github.com/mozilla/bugbug/issues/486

http_service/models.py
```diff
@@ -0,0 +1,62 @@
...
+from urllib.request import urlretrieve
+
+import requests
+
+from bugbug.models import load_model as bugbug_load_model
+
+logging.basicConfig(level=logging.INFO)
+LOGGER = logging.getLogger()
+
+MODELS_NAMES = ["defectenhancementtask", "component", "regression"]
```
marco-c: We need a better way here, so we can avoid hardcoding the models. Maybe we could parse the yml file to check what we are training? Let's file an issue so we don't forget.

PR #470, opened 2019-05-20
Add "presence of image" between attachments as a feature

### LINKED ISSUE #454: Add "presence of image" in attachments as a feature
When users file a bug uploading an image, it's very likely they want to show that there is something wrong with Firefox.
So, the presence of an image between the attachments can be an indication of the bug being a defect.

marco-c:
Fixes #454

### DIFF
--- a/bugbug/bug_features.py
+++ b/bugbug/bug_features.py
@@ -7,6 +7,7 @@ import re
 from collections import defaultdict
 from datetime import datetime, timezone
 
+import dateutil.parser
 import pandas as pd
 from libmozdata import versions
 from sklearn.base import BaseEstimator, TransformerMixin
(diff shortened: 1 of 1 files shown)


bugbug/bug_features.py
```diff
@@ -258,6 +259,23 @@ def __call__(self, bug, **kwargs):
         )
 
 
+class has_image_attachment_at_bug_creation(object):
+    def __call__(self, bug, **kwargs):
+        bug_creation_time = dateutil.parser.parse(bug["creation_time"])
+        return any(
+            "image" in attachment["content_type"]
+            and dateutil.parser.parse(attachment["creation_time"]) == bug_creation_time
```
calixteman: I'd say that attachment["creation_time"] == bug_creation_time is enough

marco-c: Done with 2cecd1d.

PR #471, opened 2019-05-20
commit author seniority as a feature

### LINKED ISSUE #355: Add "seniority" feature to commit data
Right now we are calculating the experience in terms of number of commits. We should also add a "seniority" feature, which would be the time since the first commit in the repo.

ayush-1506:
Fixes #355 

Seniority is equal to standardized `datetime.now() - commit_date` only when the author isn't in the `seniority` dict. Else, we won't count the commit since it isn't the author's first commit.


bugbug/repository.py
```diff
@@ -331,6 +335,17 @@ def calculate_experiences(commits):
...
+
+    for commit in commits:
+        if commit.author not in seniority:
+            time_lapse = datetime.now() - commit.date
+            if hasattr(time_lapse, "years"):
+                seniority[commit.author] += time_lapse.years
+            if hasattr(time_lapse, "months"):
+                seniority[commit.author] += time_lapse.months / 12
+            if hasattr(time_lapse, "days"):
+                seniority[commit.author] += time_lapse.days / 365
```
marco-c: Simply store the seniority as elapsed days instead of elapsed years.

ayush-1506:
@marco-c I've replaced it with difference between current commit and first commit date for the author for each commit. Is this what was needed?

marco-c:
Looks good, but let's use `commit.pushdate` instead of `commit.date`, as it's often more precise.

Issue #25, opened 2018-12-06
Create a DUPLICATE detector


marco-c:

marco-c:
This is the current code on Bugzilla: https://github.com/mozilla-bteam/bmo/blob/88101dcba7f4e60c862c14b6ba9c0861c62d3395/Bugzilla/Bug.pm#L766-L847.

sylvestre:
We have a keyword _dupeme_ which might be useful to help

sylvestre:
I improve the data with this in autonag:
https://github.com/mozilla/relman-auto-nag/pull/548

PR #484, opened 2019-05-23
Duplicate classifier

ayush-1506:
Fixes : #25 

This is WIP and needs substantial revisions.


bugbug/models/duplicate.py
```diff
@@ -0,0 +1,129 @@
...
+            if val in bugs:
+                text1.append(self.get_text(bugs[ind]))
+                text2.append(self.get_text(bugs[val]))
+
+        l = len(text1)
+        count = 500
+        for ind, val in bugs.items():
+            if not count:
+                break
+            if "duplicate" not in val["resolution"]:
```
marco-c: The bug to which all bugs are duplicated is still part of the duplicate group, even if it is not resolved as "duplicate".
Here it would be better to just skip bugs which are in the duplicate_map

bugbug/models/duplicate.py
```diff
@@ -0,0 +1,129 @@
...
+            feature_cleanup.url(),
+            feature_cleanup.synonyms(),
+            feature_cleanup.crash(),
+        ]
+        for cleanup_function in cleanup_functions:
+            X = cleanup_function(X)
+        return X
+
+    def get_text(self, bug):
+        return bug["summary"] + " ".join(comment["text"] for comment in bug["comments"])
```
marco-c: Try what happens to the accuracy if you don't consider all comments, but just the first (replace `bug["comments"]` with `bug["comments"][0:1]`).

bugbug/models/duplicate.py
```diff
@@ -0,0 +1,129 @@
...
+    def get_duplicate_map(self):
+        
+        bugs = bugzilla.get_bugs()
+        bug = {}
+        duplicate_bug_map = {}
+        count = 500
+        for bug_data in bugs:
+            bug_id = int(bug_data["id"])
+
+            if bug_data["resolution"] == "DUPLICATE":
```
marco-c: Instead of using the resolution, use the values of dupe_of and duplicates. This way you put in the map all bugs that are part of the duplicate group (both the bugs that are resolved as duplicate of a bug, and the bug which is the "main" one).

bugbug/models/duplicate.py
ayush-1506: If we consider the entire comment, accuracy is 0.9278897136797455.
If we consider only the first comment, accuracy is 0.9734888653234358
(Both were obtained using count=2500)

bugbug/models/duplicate.py
marco-c: Note: say we have bug 1, bug 2, bug 3, bug 4, bug 5.
bug 1 and bug 2 are duplicate.
bug 3 and bug 4 are duplicate.
bug 5 is not a duplicate of anything else.

When training, we need the following labels:
1 & 2 -> 1
1 & 3 -> 0
1 & 4 -> 0
1 & 5 -> 0
2 & 3 -> 0
2 & 4 -> 0
2 & 5 -> 0
3 & 4 -> 1
3 & 5 -> 0
4 & 5 -> 0

marco-c:
As discussed on IRC, the model's accuracy seems high enough, so we should finish the PR by integrating the model better in the current bugbug architecture (modifying/extending the current architecture when necessary).

Once we merge the PR, we will create a training task, and we will try to use it in practice to see if the model is actually working.

Then, we will have to find a way to make the application of the model faster. Right now this requires comparing each bug with every other bug, which is pretty slow. One option is to use some similarity technique (e.g. https://radimrehurek.com/gensim/tut3.html) to reduce the number of bugs to compare to (e.g. given a bug, find the top X most similar, then apply the model only on those instead of the whole DB).

ayush-1506:
@marco-c review? (I've also added a new `super_classifier` class, since we needed the LabelEncoder, that doesn't come with LinearSVM.) To get the bugs, I'm storing the bug data as an attribute, and retrieving the required id.

bugbug/models/duplicate.py
```diff
@@ -0,0 +1,172 @@
...
+        self._le = LabelEncoder()
+
+    def fit(self, X, y):
+        super().fit(X, y)
+        self._le.fit(y)
+
+
+class DuplicateModel(BugCoupleModel):
+    def __init__(self, lemmatization=False):
+
```
marco-c: Nit: no newline between function definition and first line of the function

bugbug/models/duplicate.py
```diff
@@ -0,0 +1,172 @@
...
+        BugCoupleModel.__init__(self, lemmatization)
+
+        self.cross_validation_enabled = True
```
marco-c: This is true by default, so you can avoid setting it.

bugbug/models/duplicate.py
```diff
@@ -0,0 +1,172 @@
...
+        self.sampler = BorderlineSMOTE(random_state=0)
+        self.calculate_importance = False
+
+        self.bugs = {}
```
marco-c: We can't store the bugs as a property of the model, or the model will be huge.
Also, the dataset of bugs will be different in production.

bugbug/models/duplicate.py
```diff
@@ -0,0 +1,172 @@
...
+        count = 2500
+        for bug_data in bugzilla.get_bugs():
+
+            if bug_data["duplicates"] != []:
+                duplicate_ids.append(bug_data["id"])
+                for duplicate_bug in bug_data["duplicates"]:
+                    count -= 1
+                    duplicate_ids.append(duplicate_bug)
+                    if bug_data["id"] in self.bugs and int(duplicate_bug) in self.bugs:
+                        classes[str(bug_data["id"]) + "|" + str(duplicate_bug)] = 1
```
marco-c: Use tuples instead of strings, `classes[(bug_data["id"], duplicate_bug)] = 1`

bugbug/models/duplicate.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def get_text(self, bug):
+        return bug["summary"] + " ".join(
+            comment["text"] for comment in bug["comments"][0:1]
+        )
+
+    def get_feature_names(self):
+        return self.extraction_pipeline.named_steps["tfidf"].get_feature_names()
+
+    def classify(
```
marco-c: You shouldn't need to overwrite `classify` once the model is properly integrated with the current architecture, you'll just need to change the function in model.py a bit to support tuples of bugs instead of only single bugs.

ayush-1506: Ah, sure. I was trying to make sure I don't touch model.py. Will do :+1:

marco-c: We will definitely need to make a few changes to model.py and bug_features.py in order to support a model like this, which is quite different than the other models.

bugbug/models/duplicate.py
```diff
@@ -0,0 +1,172 @@
...
+from sklearn.feature_extraction.text import TfidfVectorizer
+from sklearn.pipeline import Pipeline
+from sklearn.preprocessing import LabelEncoder
+from sklearn.svm import LinearSVC
+
+from bugbug import bugzilla, feature_cleanup
+from bugbug.model import BugCoupleModel
+
+
+class super_classifier(CalibratedClassifierCV):
```
marco-c: Nit: rename `super_classifier` to `LinearSVCWithLabelEncoding`

marco-c: We should likely refactor the label encoding stuff in general, but in the future.

bugbug/model.py
```diff
@@ -65,12 +65,25 @@ def train(self, importance_cutoff=0.15):
         classes, class_names = self.get_labels()
         class_names = sorted(list(class_names), reverse=True)
 
+        # flag is true only for the duplicate model
+        flag = False
```
marco-c: I'm not sure you need this, if you make get_id of the BugCoupleModel return the tuple of bugs.

bugbug/models/duplicate.py
```diff
@@ -0,0 +1,145 @@
...
+                    ColumnTransformer([("text", self.text_vectorizer(), "text")]),
+                ),
+            ]
+        )
+
+        self.clf = LinearSVCWithLabelEncoding(LinearSVC())
+
+    def get_labels(self):
+        bugs = {}
+        count = 20000
```
marco-c: You can use https://docs.python.org/3/library/itertools.html#itertools.islice instead of implementing it yourself.
Something like `for bug_data in itertools.islice(bugzilla.get_bugs(), 0, 20000):` should work.

bugbug/models/duplicate.py
```diff
@@ -0,0 +1,145 @@
...
+        for bug_data in bugzilla.get_bugs():
+            if not count:
+                break
+            count -= 1
+            bugs[bug_data["id"]] = bug_data
+
+        classes = {}
+        duplicate_ids = (
+            []
+        )  # Only store id's of bugs that have duplicate or are duplicate
```
marco-c: Nit: put the comment on the line before the variable (here and in the other cases).

bugbug/models/duplicate.py
```diff
@@ -0,0 +1,145 @@
...
+            if not count:
+                break
+        print("Number of hybrid labels are ", len(classes) - l2)
+
+        return classes, [0, 1]
+
+    def get_feature_names(self):
+        return self.extraction_pipeline.named_steps["union"].get_feature_names()
+
+    def overwrite_classes(self, bug_id, classes, probabilities):
```
marco-c: You can remove overwrite_classes, in this case we won't need it

bugbug/models/duplicate.py
```diff
@@ -0,0 +1,145 @@
...
+
+            if bug_data["duplicates"] != []:
+                duplicate_ids.append(bug_data["id"])
+                for duplicate_bug in bug_data["duplicates"]:
+                    if duplicate_bug in bugs:
+                        duplicate_ids.append(duplicate_bug)
+                    if duplicate_bug in bugs:
+                        classes[(bug_data["id"], duplicate_bug)] = {
+                            "label": 1,
+                            "data": (bugs[bug_data["id"]], bugs[duplicate_bug]),
```
marco-c: We shouldn't store the data as part of the labels

bugbug/model.py
```diff
@@ -221,13 +223,29 @@ class BugModel(Model):
...
+# for the duplicate model class
+class BugCoupleModel(Model):
+    def get_id(self, bug):
+        return bug[0]["id"], bug[1]["id"]
+
+    def items_gen(self, classes):
+        bugs = {}
+        for bug in bugzilla.get_bugs():
+            bugs[bug["id"]] = bug
+        res = []
```
marco-c: No need to create a big list here that you will throw away rigth after iterating through it.

You can return a generator instead, like the other items_gen:
```
return ((bugs[bug_id1], bugs[bug_id2]) for bug_id1, bug_id2 in classes)
```

bugbug/models/duplicate.py
```diff
@@ -0,0 +1,135 @@
...
+                ),
+            ]
+        )
+
+        self.clf = LinearSVCWithLabelEncoding(LinearSVC())
+
+    def get_labels(self):
+        bugs = []
+
+        for bug_data in islice(bugzilla.get_bugs(), 0, 20000):
```
marco-c: Define the counts as constants at the top of the file.

bugbug/models/duplicate.py
```diff
@@ -0,0 +1,135 @@
...
+            bugs.append(bug_data["id"])
+
+        classes = {}
+        # Only store id's of bugs that have duplicate or are duplicate
```
marco-c: Nit: `IDs` or `ids`

`that have duplicates or are duplicates` or `that are in a duplicate group`

bugbug/models/duplicate.py
```diff
@@ -0,0 +1,135 @@
...
+        duplicate_ids = []
+
+        # Store all remaining ids
+        all_ids = []
+
+        count = 4500
+        for bug_data in bugzilla.get_bugs():
+
+            if bug_data["duplicates"] != []:
```
marco-c: Nit: you could add a continue here so that you don't have to have an increased indentation for the rest of the code.
```
    if len(bug_data['duplicates']) == 0:
        continue

   ...
```

marco-c: This way, the code will be more readable

bugbug/models/duplicate.py
```diff
@@ -0,0 +1,135 @@
...
+                duplicate_ids.append(bug_data["id"])
+                for duplicate_bug in bug_data["duplicates"]:
+                    if duplicate_bug in bugs:
+                        duplicate_ids.append(duplicate_bug)
+                        classes[(bug_data["id"], duplicate_bug)] = 1
+                        # classes[(bug_data["id"], duplicate_bug)] = {
```
marco-c: Remove the commented out part here

bugbug/models/duplicate.py
```diff
@@ -0,0 +1,135 @@
...
+                    if not count:
+                        break
+            if not count:
+                break
+
+        for bug in bugs:
+            if bug not in duplicate_ids:
+                all_ids.append(bug)
+
+        le = len(classes)
```
marco-c: Use a better name for this, like `duplicates_num`

bugbug/models/duplicate.py
```diff
@@ -0,0 +1,135 @@
...
+        print("Number of purely duplicate labels are: ", le)
```
marco-c: Nit: use an f-string: `f"Number of duplicate bugs: {duplicates_num}`

marco-c: Here and in the other places

bugbug/models/duplicate.py
marco-c: And actually, you could use `duplicates_num` in place of `count` earlier, and break from the loop when duplicates_num reaches the wanted number

bugbug/models/duplicate.py
marco-c: Here and in other places

bugbug/models/duplicate.py
```diff
@@ -0,0 +1,135 @@
...
+
+        for bug_data in islice(bugzilla.get_bugs(), 0, 20000):
+            bugs.append(bug_data["id"])
+
+        classes = {}
+        # Only store id's of bugs that have duplicate or are duplicate
+        duplicate_ids = []
+
+        # Store all remaining ids
+        all_ids = []
```
marco-c: `all_ids` is misleading, better something like `non_duplicate_ids`.

bugbug/models/duplicate.py
```diff
@@ -0,0 +1,135 @@
...
+                        #    "data": (bugs[bug_data["id"]], bugs[duplicate_bug]),
+                        # }
+                        count -= 1
+                    if not count:
+                        break
+            if not count:
+                break
+
+        for bug in bugs:
+            if bug not in duplicate_ids:
```
marco-c: Remember that checking if an element is in a list is way slower than checking if an element is in a set, so you should use a set here.

bugbug/models/duplicate.py
```diff
@@ -0,0 +1,135 @@
...
+        print("Number of purely duplicate labels are: ", le)
+
+        all_ids = set(all_ids)
+
+        duplicate_ids = set(duplicate_ids)
+
+        # When the bug has no duplicates, we create dup-nondup labels = 0
+        count = 2500
+        for key in duplicate_ids:
+            for key2 in all_ids:
```
marco-c: Remember that iterating sets is really slow compared to iterating lists, so you should use lists here

bugbug/models/duplicate.py
```diff
@@ -0,0 +1,135 @@
...
+
+        count = 4500
+        for bug_data in bugzilla.get_bugs():
+
+            if bug_data["duplicates"] != []:
+                duplicate_ids.append(bug_data["id"])
+                for duplicate_bug in bug_data["duplicates"]:
+                    if duplicate_bug in bugs:
+                        duplicate_ids.append(duplicate_bug)
+                        classes[(bug_data["id"], duplicate_bug)] = 1
```
marco-c: Every bug in a group of duplicates is a duplicate of every other bug in the same group. Here, you are only setting the label to 1 for the main bug and all other bugs of the group, we could instead set the label to 1 for all possible combinations of the bugs in the group.

You can use itertools for generating combinations.

marco-c: Actually, since we are limiting the number to 4500, let's not do this or we risk overfitting. We need to file an issue about this though, so that we do it in the future.

bugbug/bug_features.py
```diff
@@ -434,63 +434,87 @@ def __init__(
     def fit(self, x, y=None):
         return self
 
+    # in case of duplicate model, bugs is a vector of tuples
```
marco-c: The changes here are too specific to the duplicate mode, we should make them more generic for any future model that might want to use a couple of bugs as a trainable item.
I'll see if I can come up with some idea tomorrow.

ayush-1506: One way can be to use a `flag` variable as an attribute (set to False initially) in the base Model. The models that use tuples of bugs at a time can set flag to True in their constructors. We can then pass this as an argument to BugExtractor. This way we can be sure that models with tuples always have self.flag= False and have a common set of extractors for them.

marco-c: `flag` is definitely not a great name for an option, as it could mean basically anything :P

Anyway, this doesn't change the current situation much, as we could already automatically set the flag in BugExtractor simply by checking whether the elements of bugs are dicts or tuples.

marco-c: Here's what we could do:
```
    def transform(self, elems):
        results = []

        def apply_transform(bug):
            # Do all we need to do here

            return {
                "data": data,
                "title": bug["summary"],
                "first_comment": bug["comments"][0]["text"],
                "comments": " ".join([c["text"] for c in bug["comments"]]),
            }

        for elem in elems:
            if isinstance(elem, dict):
                results.append(apply_transform(elem))
            elif isinstance(elem, tuple):
                bug1, bug2 = elem
                result1 = apply_transform(bug1)
                result2 = apply_transform(bug2)
                results.append({
                    "data1": result1["data"],
                    "data2": result2["data"],
                    "title1": result1["title"],
                    "title2": result2["title"],
                    ...
                })
            else:
                assert False, 'Unexpected element type'

        return pd.DataFrame(results)
```

marco-c: Then, in the model, you somehow merge title and first_comment of the two bugs, either with a built-in transformer if you find one that would work here, or with a custom transformer.

marco-c: > Then, in the model, you somehow merge title and first_comment of the two bugs, either with a built-in transformer if you find one that would work here, or with a custom transformer.

Or, for the time being, you can add a `merge_data` option to BugExtractor do:
```
if merge_data:
                results.append({
                    "text": f"{result1['title']} {result1['first_comment']} {result2['title']} {result2['first_comment']}",
                })
else:
                results.append({
                    "data1": result1["data"],
                    "data2": result2["data"],
                    "title1": result1["title"],
                    "title2": result2["title"],
                    ...
                })
```
instead of:
```
                results.append({
                    "data1": result1["data"],
                    "data2": result2["data"],
                    "title1": result1["title"],
                    "title2": result2["title"],
                    ...
                })
```

bugbug/models/duplicate.py
```diff
@@ -0,0 +1,129 @@
...
+from sklearn.calibration import CalibratedClassifierCV
+from sklearn.compose import ColumnTransformer
+from sklearn.pipeline import Pipeline
+from sklearn.preprocessing import LabelEncoder
+from sklearn.svm import LinearSVC
+
+from bugbug import bug_features, bugzilla, feature_cleanup
+from bugbug.model import BugCoupleModel
+
+count = 6000
```
marco-c: It isn't a rule, but usually in Python constants are uppercase.
Please also make this a bit more informative, `count` is not a great name.
I would define three different constants, one for duplicate, one for non-duplicate, and one for hybrid.

ayush-1506: Ideally, we'd need to set a limit only for the third case : (nondup - non-dup). I have set `count` now just for running it on my system. Should I remove it for the first two cases?

marco-c: Are you able to run the model on your system if you set the limit only for the third case?

ayush-1506: No, it runs out of memory.

marco-c: OK, for now use a limit that allows you to run the model, then we'll make this configurable with options.

bugbug/bug_features.py
```diff
@@ -423,24 +423,27 @@ def __init__(
...
         self.rollback = rollback
         self.rollback_when = rollback_when
         self.commit_map = repository.get_commit_map() if commit_data else None
+        self.merge_data = merge_data
         assert self.commit_map is None or len(self.commit_map) > 0
 
     def fit(self, x, y=None):
         return self
 
+    # in case of duplicate model, bugs is a vector of tuples
```
marco-c: You can remove this comment now, as this is generic

bugbug/model.py
```diff
@@ -221,13 +223,26 @@ class BugModel(Model):
...
 class CommitModel(Model):
     def get_id(self, commit):
         return commit["node"]
 
-    def items_gen(self):
+    def items_gen(self, classes):
         return (commit for commit in repository.get_commits())
+
+
+# for the duplicate model class
```
marco-c: Here too, as this could be used by other models in the future.

bugbug/models/duplicate.py
```diff
@@ -0,0 +1,131 @@
...
+                    duplicate_ids.append(duplicate_bug)
+                    classes[(bug_data["id"], duplicate_bug)] = 1
+                    duplicates_num += 1
+                if duplicates_num == NUM_DUPLICATES:
+                    break
+            if duplicates_num == NUM_DUPLICATES:
+                break
+
+        for bug in bugs:
+            if bug not in set(duplicate_ids):
```
marco-c: Unfortunately I think Python is not smart enough to calculate this just once, you have to do something like:
```
duplicate_ids_set = set(duplicate_ids)
for bug in bugs:
    if bug not in duplicate_ids_set:
...
```

bugbug/models/duplicate.py
```diff
@@ -0,0 +1,131 @@
...
+                classes[(key, key2)] = 0
+                dup_nondup_num += 1
+                if dup_nondup_num == NUM_DUP_NONDUPS:
+                    break
+            if dup_nondup_num == NUM_DUP_NONDUPS:
+                break
+
+        print(f"Number of purely non-duplicate labels are {dup_nondup_num}")
+        # Non we map non-dup to non-dup bug. However, these are really
+        # large in number, so I'm restricting the number to 2.5K
```
marco-c: Remove the second part of this comment (from However to 2.5K)

bugbug/models/duplicate.py
marco-c: Filed https://github.com/mozilla/bugbug/issues/521 about this.

bugbug/models/duplicate.py
marco-c: Filed https://github.com/mozilla/bugbug/issues/522 about this.

PR #488, opened 2019-05-24
Optimize experience calculation by having lists specific to items instead of carrying over the entire history all the time

marco-c:


bugbug/repository.py
```diff
@@ -310,27 +310,53 @@ def _hg_log(revs):
...
+            return self.list[-1]
+
+        return self.list[day - self.start_day]
+
+    def __setitem__(self, day, value):
+        if day == self.last_day:
+            self.list[day - self.start_day] = value
+        elif day > self.last_day:
+            last_val = self.list[-1]
+            self.list.extend(last_val for _ in range(self.last_day, day - 1))
```
calixteman: I'd say that adding last_val (N % self.maxlen) times instead of N times is better.
I'm not sure of that, just need to be checked but there is probably a modulo to put somewhere to improve speed.

bugbug/repository.py
```diff
@@ -310,27 +310,53 @@ def _hg_log(revs):
...
 
 
+class exp_queue:
+    def __init__(self, start_day, maxlen, default):
+        self.list = deque([default] * maxlen, maxlen=maxlen)
+        self.start_day = start_day - (maxlen - 1)
+        self.default = default
+
+    @property
+    def last_day(self):
```
calixteman: nit: you could replace with self.last_day = start_day (in ctor) or = day (in __setitem__)

bugbug/repository.py
marco-c: I think with 7740622 this should be good.

bugbug/repository.py
marco-c: I did that at first, but then I wanted to avoid having an additional variable to think about.

bugbug/repository.py
```diff
@@ -436,58 +474,42 @@ def update_complex_experiences(experience_type, day, items):
...
 
         # When a file is moved/copied, copy original experience values to the copied path.
         if len(commit.file_copies) > 0:
             for orig, copied in commit.file_copies.items():
                 orig_directories = get_directories(orig)
                 copied_directories = get_directories(copied)
+                for orig_directory, copied_directory in zip(
+                    orig_directories, copied_directories
+                ):
+                    experiences["directory"][copied_directory] = copy.deepcopy(
```
calixteman: nit: move xp["directory"] outside of the loop to avoid extra lookup

bugbug/repository.py
```diff
@@ -436,58 +474,42 @@ def update_complex_experiences(experience_type, day, items):
 
         # We don't want to consider backed out commits when calculating experiences.
         if not commit.backedoutby:
-            for item in items:
-                complex_experiences[day][experience_type][item].append(commit.node)
-
-    prev_days = 0
+            for i, item in enumerate(items):
+                experiences[experience_type][item][day] = all_commit_lists[i] + [
```
calixteman: nit: move xp[experience_type] outside of the loop

bugbug/repository.py
```diff
@@ -380,15 +418,15 @@ def update_experiences(experience_type, day, items):
 
         # We don't want to consider backed out commits when calculating experiences.
         if not commit.backedoutby:
-            for item in items:
-                experiences[day][experience_type][item] += 1
+            for i, item in enumerate(items):
+                experiences[experience_type][item][day] = total_exps[i] + 1
```
calixteman: nit: move xp[...] outside of the loop

bugbug/repository.py
marco-c: ea29fe3 is improving even more, with a comment added in f6d77e1, please double check them!

bugbug/repository.py
```diff
@@ -310,27 +310,57 @@ def _hg_log(revs):
...
-        rev=f"{rev_start}:tip",
+        b"log", template="{node}\n", no_merges=True, branch="central", rev=f"0:tip"
     )
     x = hg.rawcommand(args)
     return x.splitlines()
 
 
+class exp_queue:
+    def __init__(self, start_day, maxlen, default):
+        self.list = deque([default] * maxlen, maxlen=maxlen)
```
calixteman: when init with default=[] then it's the same instance of default everywhere in queue.
So for example q[0].append(1) will give deque([1], [1], ...).
Is it expected ?

bugbug/repository.py
```diff
@@ -310,27 +310,57 @@ def _hg_log(revs):
...
+
+    def __setitem__(self, day, value):
+        if day == self.last_day:
+            self.list[day - self.start_day] = value
+        elif day > self.last_day:
+            last_val = self.list[-1]
+            # We need to extend the list except for 2 elements (the last, which
+            # is going to be the same, and the one we are adding now).
+            self.list.extend(
+                last_val for _ in range(min(day - self.last_day, self.list.maxlen) - 2)
```
calixteman: If last_val is a list then we've the same instance everywhere. Is it expected ?

bugbug/repository.py
```diff
@@ -310,27 +310,57 @@ def _hg_log(revs):
...
+        return self.list[day - self.start_day]
+
+    def __setitem__(self, day, value):
+        if day == self.last_day:
+            self.list[day - self.start_day] = value
+        elif day > self.last_day:
+            last_val = self.list[-1]
+            # We need to extend the list except for 2 elements (the last, which
+            # is going to be the same, and the one we are adding now).
+            self.list.extend(
```
calixteman: The value ((min(day - self.last_day, self.list.maxlen) - 2) >= -1 so in case it's 0 or -1 there is no need to call range + extend, so just add a test before.
Anyway it looks pretty fine.

bugbug/repository.py
marco-c: Elements of the queue are never modified. I've used a tuple now, so this is basically asserted.

bugbug/repository.py
marco-c: Yes, as the values in the queue are the cumulative values for a specific day (so that we can subtract now - 90 days ago for example).

bugbug/repository.py
marco-c: Done with 4c24c50.

Issue #340, opened 2019-05-03
Pipeline should use tagged images, so we can rollback when things are broken


marco-c:

marco-c:
It should probably be a JSON-e variable in the data-pipeline.yml file, and the hook will take care of filling it

Lothiraldan:
I can imagine how we could pass the version number with the pulse message but I'm not sure how that would works in the cron configuration

marco-c:
The version number could be a config of the hook itself.
The task triggered by the hook could either parse its own "image" and read the tag from there, or we could add the tag to "env".

Lothiraldan:
There is two part for this issue:
- Tagging the image with versions. I have opened a PR in task-boot to allow to pass the Git tag when building: https://github.com/mozilla/task-boot/pull/29
- Render the hook with the right version number.

I think we will need to "save" the current version somewhere anyway as the data pipeline will need to access the version in order to build an updated version of the http service image. What about taskcluster index, could we store the latest version there?

marco-c:
The task building the http service image will have to use the same tag as the tasks in the data pipeline, so the tag could be filled by the infra/spawn_data_pipeline.py script itself.

The infra/spawn_data_pipeline.py script could read the tag from its env (or its image, if it can access that), and trigger tasks from data-pipeline.yml with the same tag.

marco-c:
To support this, we need to reland #489 and #491 (and #494), fixing the issue from #501.

marco-c:
Another prerequisite is uploading the image both with 'latest' and version tag when we deploy. Basically making use of https://github.com/mozilla/task-boot/commit/3374b5c288d548d0598255950f5eb57c5fe7c356.

marco-c:
Another thing we need to do is to make the index URLs use the tag. E.g. the trainer script is always downloading from "latest", it should instead download from the task with the same tag as itself.

Lothiraldan:
I'm summing up the work to do and already done:
- We need to build Docker images with tagged images: #554 
- We need to spawn the data pipeline with tagged images: #507 + #553 
- We need to push artifacts in Taskcluster index with tags
- We need to retrieve artifacts from Taskcluster index with tags
- We need to build the HTTP images with tags at the end of the data pipeline

marco-c:
This should be done, we just need to confirm everything is working.

marco-c:
To verify, we can try an actual rollback and see if everything works.

PR #489, opened 2019-05-24
Add support for specific Docker tag in spawn_data_pipeline.py

Lothiraldan:
This is needed in #340. We can either compute the Docker image like I do in this PR or pass the Docker Tag as context as the JSON file is rendered as JSON-e. I have a small preference for the way of this PR as we don't need to put the logic `logic or env.DOCKER_TAG` in all of the pipelines files.

### DIFF
--- a/infra/spawn_data_pipeline.py
+++ b/infra/spawn_data_pipeline.py
@@ -69,7 +69,7 @@ def main():
 
     id_mapping = {}
 
-    docker_tag = os.getenv("DOCKER_TAG", None)
+    docker_tag = os.getenv("TAG", None)
 
     # First pass, do the template rendering and dependencies resolution
     tasks = []
--- a/infra/spawn_data_pipeline.py
+++ b/infra/spawn_data_pipeline.py
@@ -69,6 +69,8 @@ def main():
 
     id_mapping = {}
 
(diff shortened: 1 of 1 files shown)


infra/spawn_data_pipeline.py
```diff
@@ -69,6 +69,8 @@ def main():
 
     id_mapping = {}
 
+    docker_tag = os.getenv("DOCKER_TAG", None)
```
marco-c: Nit: Can we call it just "TAG"?

infra/spawn_data_pipeline.py
```diff
@@ -105,6 +107,14 @@ def main():
 
         payload["dependencies"] = new_dependencies
 
+        # Override the Docker image tag if needed
+        if docker_tag:
+            base_image = payload["payload"]["image"]
+            tagless_image = base_image.rsplit(":", 1)[0]
```
marco-c: Let's actually assert the tag doesn't exist.

Lothiraldan: Done, I have allowed the tag `latest` as it's more or less the equivalent of no tag

marco-c: :+1:

PR #498, opened 2019-05-24
Ignore changesets with 'ignore-this-changeset' in the desc

### LINKED ISSUE #496: Ignore "# ignore-this-changeset" commits
When "# ignore-this-changeset" is in the desc, we can ignore the commit.

marco-c:
Fixes #496

This slows down a bit `get_revs` (30 seconds instead of 15 seconds), but it is minimal compared to the savings in the rest of the process.


marco-c:
We can't do this because of #502.

Issue #428, opened 2019-05-15
Add some basic tests for the steps to reproduce model


marco-c:

sladyn98:
@marco-c I would like to work on this could you guide me a bit. Thanks

marco-c:
Hello @sladyn98, this will be something similar to https://github.com/mozilla/bugbug/blob/1bae5834abd49f70c69da84d36b66f5f5cf503b0/tests/test_tracking.py, for the StepsToReproduce model (https://github.com/mozilla/bugbug/blob/master/bugbug/models/stepstoreproduce.py).

sladyn98:
@marco-c Kindly review :)

marco-c:
Fixed by #503.

PR #503, opened 2019-05-25
test_stepstoreproduce.py: Add basic tests

sladyn98:
Added basic tests with respec to the
stepstoreproduce model.

Reference: https://github.com/mozilla/bugbug/issues/428


sladyn98:
@marco-c Im sorry this is the first time reading the log of TaskCluster. I am not able to read where exaclty is the log. Could you help me a bit

marco-c:
On this page, click on Details. You will see a list of tasks, the failing ones are in red and say "failed". Click on the name of one of them. Click on "Run Logs", select one of the two options.

Note: you should be able to reproduce the errors locally too.

sladyn98:
Got it :+1: And to run this repo I just need to execute `run.py` as mentioned in the readme.Is it necessary to run the mining scripts?

marco-c:
@sladyn98 so you didn't try to run the test that you wrote?

You don't need to run the training nor the mining scripts, you just need to run the test that you added.

sladyn98:
@marco-c I did run the test and I got a key error. I suppose thats not what the classes dictionary expects.Thats why the tests are failing

tests/test_stepstoreproduce.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+from bugbug.models.stepstoreproduce import StepsToReproduceModel
+
+
+def test_get_labels():
+    model = StepsToReproduceModel()
+    classes, _ = model.get_labels()
+    # assert not classes[1101825]
+    assert classes[1042096]
```
ayush-1506: @sladyn98 The stepstoreproduce model doesn't have this key in it's labels.

ayush-1506:
@sladyn98 Both these models are independent of each other. You should look at a bug that satisfies the `stepstoreproduce` criteria (found in the get_labels method) and add a check to see if the result is consistent.

sladyn98:
Is there a way to find a key ,such that the key exists?I mean would I have to run the script...
cc @ayush-1506 , @marco-c

ayush-1506:
No. All you need is the bugs.json file. You don't need to run the repository mining script.

sladyn98:
@marco-c @ayush-1506 Done Thanks and kindly review.

sladyn98:
I did try adding a negative test but there were no bugs for which the label was 0. This was the only bug.

ayush-1506:
That shouldn't happen. Can you tell me what calling the `get_labels` method for this classifier returns for you? How large is your bugs DB file?

sladyn98:
I just ran pytest and printed the classes.
```
0 bugs have no steps to reproduce
1 bugs have steps to reproduce
{1488310: 1}
```
Doesnt the bugs db for tests get populated from the json?

ayush-1506:
Yes, I'm guessing you don't have the complete bugs.json file for some reason. What is the size of data/bugs.json ?

sladyn98:
Its showing me `385.1 kB (3,85,144 bytes) ` on my linux system

ayush-1506:
Yes, it should be around 1.4 GBs. Delete the data folder and run `python run.py --goal=component --train` (You can use and possible value for goal, this will just download the bugs and commits DB).

ayush-1506:
Or you can directly download it from https://github.com/mozilla/bugbug/blob/master/bugbug/bugzilla.py#L21 and paste it in the data folder.

sladyn98:
Yeah I did keep it for download :)

sladyn98:
@ayush-1506 The tests pass on my local machine.But they are failing on task cluster

sladyn98:
I'm guessing it's because the tests are being picked up from the mock dB and the key is not present there.

marco-c:
> I'm guessing it's because the tests are being picked up from the mock dB and the key is not present there.

Yes, you need to add it to the mock DB.

sladyn98:
And the bugs are not properly formatted in the json file which is making it difficult to read. Could you tell me the fields in a bug along with the start and end field. Thanks

marco-c:
> And the bugs are not properly formatted in the json file which is making it difficult to read. Could you tell me the fields in a bug along with the start and end field. Thanks

You'll need to make a bit of effort yourself to solve the issue, otherwise it's useless for you (you don't learn much) and it is useless for me (as I spend more time explaining every small detail to you than to just fix the issue myself).

sladyn98:
Yeah agreed. Will come up with a solution soon. Thanks

sladyn98:
@marco-c Kindly have a look I added some mock data.The tests are passing too.

tests/test_stepstoreproduce.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+from bugbug.models.stepstoreproduce import StepsToReproduceModel
+
+
+def test_get_labels():
+    model = StepsToReproduceModel()
+    classes, _ = model.get_labels()
+    assert classes[1488310]
+    assert not classes[1372243]
```
marco-c: Could you also add a bug which is not in classes (you can use one of the bugs which is already in the mock DB and has `cf_has_str` not set to `yes` and not set to `no`).

marco-c: assert XXX not in classes

sladyn98: You mean choose a bug that has `cf_has_str : ---- ` and assert that it is not in classes?

marco-c: Yes!

sladyn98: Done

sladyn98: Done kindly review.This should suffice IMO :)

Issue #501, opened 2019-05-24
Update hook tasks fail


marco-c:
There are a few issues:
1) ${tag} is not available in the JSON-e context, it has to be head_branch[10:] (or head_branch[11:], need to check);
2) the taskboot image doesn't contain the bugbug files, we probably need to clone the repository first
3) the taskboot image doesn't contain /bin/bash (but /bin/sh).

PR #507, opened 2019-05-26
Reimport #491 and fix issues lited in #501

Lothiraldan:
This should fix issues listed in #501 but I'm not sure how to test it properly without creating a tag


marco-c:
You could create the tag, just make sure you only push to the tag and not to master, and remove the deployment tasks from .taskcluster.yml.
After testing, remove the tag.

PR #508, opened 2019-05-27
Swap xp & transform to be able to move file_stuff in _transform

calixteman:


bugbug/repository.py
```diff
@@ -546,22 +570,28 @@ def download_commits(repo_dir, date_from):
 
     print(f"Mining {len(revs)} commits using {processes} processes...")
 
-    CHUNK_SIZE = 256
-    revs_groups = [revs[i : (i + CHUNK_SIZE)] for i in range(0, len(revs), CHUNK_SIZE)]
+    CHUNK_SIZE = int(math.ceil(len(revs) / processes))
```
marco-c: I'd like to do these changes separately, so we can separately evaluate their effect

bugbug/repository.py
```diff
@@ -575,24 +605,26 @@ def download_commits(repo_dir, date_from):
...
-
     commits_num = len(commits)
 
     print(f"Mining {commits_num} commits using {processes} processes...")
 
     global rs_parsepatch
     import rs_parsepatch
 
+    # Exclude commits outside the range we care about.
+    recent_commits = [commit for commit in commits if commit.pushdate > date_from]
```
marco-c: We need to calculate the experiences on the whole history

marco-c: So we can't exclude recent commits

bugbug/repository.py
```diff
@@ -32,6 +33,24 @@
 EXPERIENCE_TIMESPAN = 90
 EXPERIENCE_TIMESPAN_TEXT = f"{EXPERIENCE_TIMESPAN}_days"
 
+FILE_TYPES = {
```
marco-c: Nit: the mapping could be inverted to have a smaller and more readable dict (FileType -> Extension). Then you can define another constant which inverts the inverted mapping

bugbug/repository.py
```diff
@@ -41,31 +60,85 @@ def __init__(
...
+            "seniority_author": self.seniority_author,
+            "total_file_size": self.total_file_size,
+            "average_file_size": self.average_file_size,
+            "maximum_file_size": self.maximum_file_size,
+            "minimum_file_size": self.minimum_file_size,
+            "files_modified_num": self.files_modified_num,
+            "components": self.components,
+            "directories": self.directories,
+        }
+        for attr in self.touched_prev:
```
marco-c: No need to store the properties here, you can just iterate through the properties of the object and only consider those which start with "touched_prev".

bugbug/repository.py
marco-c: Actually, since it's a small part of the runtime, we can keep it.

bugbug/repository.py
```diff
@@ -41,31 +60,85 @@ def __init__(
...
+        self.bug_id = bug_id
         self.backedoutby = backedoutby
+        self.ever_backedout = backedoutby != ""
         self.author_email = author_email
-        self.files = files
-        self.file_copies = file_copies
+        self.files = []
+        self.file_copies = {}
         self.reviewers = reviewers
+        self.added = 0
```
marco-c: Can we avoid the default values? So we break loudly if we don't set them anymore.

bugbug/repository.py
```diff
@@ -546,22 +570,28 @@ def download_commits(repo_dir, date_from):
...
+    revs_groups = [
+        (revs[i], revs[min(i + CHUNK_SIZE, REVS_COUNT) - 1])
+        for i in range(0, REVS_COUNT, CHUNK_SIZE)
+    ]
 
     with concurrent.futures.ProcessPoolExecutor(
         initializer=_init, initargs=(repo_dir,)
     ) as executor:
-        commits = executor.map(_hg_log, revs_groups, chunksize=20)
+        commits = executor.map(_hg_log, revs_groups, chunksize=1)
```
marco-c: Nit: chunksize=1 is the default IIRC, so we could remove it

bugbug/repository.py
calixteman: It was to avoid to call __dict__.
But probably we can call __dict__ and then remove what's useless and change type for required keys.

bugbug/repository.py
```diff
@@ -231,39 +257,51 @@ def _transform(commit):
...
+    return commit
 
-    obj["components"] = list(
-        set(
-            path_to_component[path]
-            for path in commit.files
-            if path in path_to_component
-        )
+
+def hg_modified_files(hg, commit):
```
marco-c: Nit: could you move this *before* _transform?

bugbug/repository.py
calixteman: Since we remove file_** from hg_log, then all the chunks contain almost the same amount of work.
So I think it makes sense to do it here because it's a direct consequence of file_** move and it increases the performance improvement (fyi: 1m39 on my computer with 28 workers for all the log).

marco-c:
Moved to #820.

PR #516, opened 2019-05-29
Fix Docker build by downloading the models inside the image. Fix #504

### LINKED ISSUE #504: Docker HTTP build task is failing
Regression from #458.

```
+ python http_service/download_models.py
  File "http_service/download_models.py", line 7, in <module>
    from models import MODELS_NAMES, retrieve_model
  File "/code/http_service/models.py", line 14, in <module>
    from bugbug.models import load_model as bugbug_load_model
ModuleNotFoundError: No module named 'bugbug'
```

Lothiraldan:
The data pipeline failed before because it tried downloading the model from
outside the Docker image and didn't had bugbug installed.

The clean way of solving this would be to build a base http service image on
release and build another one where we simply download the models but let's
fix it this way for now.


marco-c:
> The clean way of solving this would be to build a base http service image on
release and build another one where we simply download the models but let's
fix it this way for now.

I like this idea, can you file an issue so we don't forget?

Issue #461, opened 2019-05-18
Use zstandard instead of xz


marco-c:
We should compare the compression ratios and, if the difference is small, switch to zstandard.

marco-c:
This is pretty simple.
1. Download the bugs.json.xz file (from https://index.taskcluster.net/v1/task/project.relman.bugbug.data_bugs.latest/artifacts/public/bugs.json.xz), extract it, compress it with zstandard. Compare the sizes of the extracted file vs the xz file vs the zstandard file.
2. Perform the same steps with one of the model files (e.g. https://index.taskcluster.net/v1/task/project.relman.bugbug.train_defectenhancementtask.latest/artifacts/public/defectenhancementtaskmodel.xz).

x249wang:
Here's what I got for the relative file sizes:

| (in KB) | Original  | .xz     | .zst    |
|---------|-----------|---------|---------|
| bugs    | 1,370,732 | 152,193 | 154,090 |
| model   | 67,906    | 39,432  | 41,567  |

I'm using the Windows release of Zstandard v1.4.0, using compression level of 19 (results in smallest file sizes). 

Is the difference considered small enough?

marco-c:
Could you try with the default compression level too?

x249wang:
With default compression level (i.e. 3):

| (in KB) | Original  | .xz     | .zst    |
|---------|-----------|---------|---------|
| bugs    | 1,370,732 | 152,193 | 152,196 |
| model   | 67,906    | 39,432  | 39,433  |

Much closer than previously.

marco-c:
OK, thanks! So we can definitely switch to zstandard.

x249wang:
Sounds good. I can get started with that.

x249wang:
@marco-c I found two instances of the `compress_file()` function being called (`commit_retriever.py`, `bug_retriever.py`), which I can update, but it seems that all the .xz-format model files are downloaded from taskcluster?

marco-c:
> @marco-c I found two instances of the compress_file() function being called (commit_retriever.py, bug_retriever.py), which I can update, but it seems that all the .xz-format model files are downloaded from taskcluster?

Yes, they are downloaded from Taskcluster, but the code to upload to Taskcluster is also in this repository.
You'll need to update both the upload and the download code.
To catch all uses, you can simply `grep "xz" -r scripts/ bugbug/ infra/`.

x249wang:
[...]

PR #524, opened 2019-05-31
Use zstandard instead of xz (#461)

x249wang:


Lothiraldan:
Could you solves the merge conflicts?

marco-c:
You might need to update some additional files too where 'xz' is referenced

marco-c:
And in the db download methods, we'll have to support both, at least for a while.

x249wang:
@marco-c By supporting both methods, do you mean giving user the option to choose which file type to download?

marco-c:
> @marco-c By supporting both methods, do you mean giving user the option to choose which file type to download?

I meant trying to download zstandard first and, if it fails, trying to download xz.

x249wang:
Please let me know if any changes are needed. Thanks in advance.

README.md
```diff
@@ -43,6 +43,9 @@ https://hacks.mozilla.org/2019/04/teaching-machines-to-triage-firefox-bugs/
 
 Run `pip install -r requirements.txt` and `pip install -r test-requirements.txt`
 
+If you update the bugs database, run `zstd data/bugs.json`.
+If you update the commits database, run `zstd data/commits.json`.
+
```
marco-c: These lines have been removed recently, don't add them back.

x249wang: Fixed. Thanks for checking!

marco-c: There is still an addition of a newline here :)

bugbug/db.py
```diff
@@ -55,10 +63,22 @@ def download_support_file(path, file_name):
...
-        if path.endswith(".xz"):
-            extract_file(path[:-3])
+        if path.endswith(".zst"):
+            extract_file(path[:-4], "zst")
+
     except requests.exceptions.HTTPError:
-        print(f"{file_name} is not yet available to download for {path}")
+        try:
+            url = re.sub(".zst$", ".xz", url)
+            path = re.sub(".zst$", ".xz", path)
```
marco-c: You need to escape the `.`, or it will match any character.

You could also implement this with os.path.splitext, something like:
```Python
url = f"{os.path.splitext(url)[0]}.xz"
path = f"{os.path.splitext(path)[0]}.xz"
```

bugbug/db.py
```diff
@@ -55,10 +63,22 @@ def download_support_file(path, file_name):
...
+
+            print(f"Downloading {url} to {path}")
+            utils.download_check_etab(url, path)
```
marco-c: There's a typo here, it should be `etag`, not `etab`.

marco-c:
We should add some tests for the db methods you are touching to make sure they are working as expected.

marco-c:
To write the tests, you can use the `responses` module (look at https://github.com/mozilla/bugbug/blob/f4145b4eca80b0cc854459d239db01dde0790afa/tests/test_repository.py#L244 for an example usage of `responses` for another file that is downloaded with download_check_etag).

bugbug/db.py
```diff
@@ -41,10 +42,17 @@ def is_old_version(path):
...
-            shutil.copyfileobj(input_f, output_f)
+
+        if compression_type == "zst":
+            dctx = zstandard.ZstdDecompressor()
+            with open(f"{path}.zst", "rb") as input_f:
+                dctx.copy_stream(input_f, output_f)
+
+        elif compression_type == "xz":
+            with lzma.open(f"{path}.xz") as input_f:
+                shutil.copyfileobj(input_f, output_f)
```
marco-c: Add an else where you do `assert False, f'Unexpected compression type: {compression_type}'`

bugbug/db.py
```diff
@@ -41,10 +42,17 @@ def is_old_version(path):
     return DATABASES[path]["version"] > prev_version
 
 
-def extract_file(path):
+def extract_file(path, compression_type):
     with open(path, "wb") as output_f:
```
marco-c: I'd do: compression_type = os.path.splitext(path)[1], so you don't need to pass it as a parameter.

marco-c:
The tests should be:
1. Test extract_file with a "xz" file (you can create a xz file in a temporary directory and then try to extract it with extract_file).
2. Test extract_file with a "zst" file
3. Test extract_file with a file which is not "xz" and not "zst" and check that the assertion is thrown
4. Test download_support_file, for a zst URL that exists
5. Test download_support_file, for a zst URL that doesn't exist (and the xz URL exists)
6. Test download_support_file, for a zst URL that doesn't exist (and the xz URL doesn't exist)
7. Test download, for a zst URL that exists
8. Test download, for a zst URL that doesn't exist (and the xz URL exists)
9. Test download, for a zst URL that doesn't exist (and the xz URL doesn't exist)

x249wang:
@marco-c Where can I find valid zst URLs for testing?

marco-c:
> @marco-c Where can I find valid zst URLs for testing?

There aren't valid zst URLs, as this PR isn't merged yet, but for testing you don't need them. They should be mocked out, like the example I showed you, as we don't want to really hit the network during tests.

tests/test_db.py
```diff
@@ -97,3 +106,318 @@ def test_register_db(tmp_path):
...
+
+    db.download(db_path)
+
+    assert os.path.exists(db_path)
+    assert os.path.exists(db_path.with_suffix(db_path.suffix + ".xz"))
+    assert os.path.exists(db_path.with_suffix(db_path.suffix + ".xz.etag"))
+
+
+@responses.activate
+def test_download_bad_format(tmp_path):
```
marco-c: Nit: `test_download_no_zst_no_xz`

marco-c: Or: `test_download_missing`

tests/test_db.py
```diff
@@ -97,3 +106,318 @@ def test_register_db(tmp_path):
...
+        status=200,
+        headers={"ETag": "123", "Accept-Encoding": "zstd"},
+    )
+
+    mock_zst(db_path.with_suffix(db_path.suffix + ".zst"))
+
+    with open(db_path.with_suffix(db_path.suffix + ".zst"), "rb") as content:
+        responses.add(responses.GET, url, status=200, body=content.read())
+
+    os.remove(db_path.with_suffix(db_path.suffix + ".zst"))
```
marco-c: Why are you manually removing the DB here?

tests/test_db.py
```diff
@@ -97,3 +106,318 @@ def test_register_db(tmp_path):
...
+        status=200,
+        headers={"ETag": "123", "Accept-Encoding": "xz"},
+    )
+
+    mock_xz(db_path.with_suffix(db_path.suffix + ".xz"))
+
+    with open(db_path.with_suffix(db_path.suffix + ".xz"), "rb") as content:
+        responses.add(responses.GET, url_xz_support, status=200, body=content.read())
+
+    os.remove(db_path.with_suffix(db_path.suffix + ".xz"))
```
marco-c: Same here

tests/test_db.py
```diff
@@ -97,3 +106,318 @@ def test_register_db(tmp_path):
...
+    )
+    assert os.path.exists(
+        os.path.join(
+            os.path.dirname(db_path), os.path.splitext(support_filename)[0] + ".xz.etag"
+        )
+    )
+
+
+@responses.activate
+def test_download_support_file_bad_format(tmp_path, capfd):
```
marco-c: Same nit here (or _no_zst_no_xz, or _missing)

marco-c:
The tests look great! Just a few minor comments and then we can land this.
Hopefully we haven't missed anything.

tests/test_db.py
x249wang: > Why are you manually removing the DB here?

I created that mock .zst file just for the body of the mock request, and removed it because the next step is to test the ability to download the compressed file.

marco-c: You can create the mock file in another path, so you don't have to remove it.

x249wang: Just updated the tests to save those mock files to another temporary path, and changed the test names to __missing.

PR #526, opened 2019-05-31
Ignore some reporters

ayush-1506:


marco-c:
We should only ignore them in the Duplicate model

ayush-1506:
@marco-c I could add a condition to check if the child class is `DuplicateModel`, if that's what you mean?

marco-c:
> @marco-c I could add a condition to check if the child class is DuplicateModel, if that's what you mean?

No, that would be pretty ugly! The base class should be generic and leave this kind of things to child classes.
So you should do this directly in DuplicateModel, in the get_labels function.

bugbug/models/duplicate.py
```diff
@@ -61,9 +61,11 @@ def __init__(self, lemmatization=False):
 
     def get_labels(self):
         bugs = []
+        reporters = ("intermittent-bug-filer@mozilla.bugs", "wptsync@mozilla.bugs")
```
marco-c: Put this at the top of the file, as a constant set.

```REPORTERS_TO_IGNORE = {"intermittent-bug-filer@mozilla.bugs", "wptsync@mozilla.bugs"}```

File: bugbug/models/duplicate.py [hunk 2]
```diff
@@ -61,9 +61,11 @@ def __init__(self, lemmatization=False):
...
 
         for bug_data in islice(bugzilla.get_bugs(), 0, NUM_DUPLICATES):
-            bugs.append(bug_data["id"])
+            if bug_data["creator"] not in reporters:
```
marco-c: There is another place in get_labels where you are iterating through bugs, we should ignore these reporters there too

PR #532, opened 2019-06-02
Move common dictionary access outside of loop when calculating exp

### LINKED ISSUE #493: When calculating experiences, avoid some unnecessary dict accesses
We can introduce some temporary variables in some cases to avoid accessing the same elements multiple times.
E.g. https://github.com/mozilla/bugbug/pull/488#discussion_r287302102.

ibraheem-moosa:
Fixes #493.


bugbug/repository.py
```diff
@@ -487,10 +487,9 @@ def update_experiences(experience_type, day, items):
                 or commit_type == "backout"
                 and commit.backedoutby
             ):
+                exps = experiences[experience_type][commit_type]
```
calixteman: before: experiences was accessed only when items is not empty.
now: experiences is accessed whatever items are that could lead to an error when items is empty.

bugbug/repository.py
```diff
@@ -560,10 +559,9 @@ def update_complex_experiences(experience_type, day, items):
...
```
calixteman: same as above

marco-c:
@ibraheem-moosa are you still interested in working on this? If yes, feel free to reopen the PR with your updates.

PR #546, opened 2019-06-04
Download previous commits DB and experiences, and only mine data for new commits landed since then

### LINKED ISSUE #537: Commit retrieval task should only download new commits
Unless the format of the DB changed.

marco-c:
Fixes #537

### DIFF
--- a/scripts/commit_retriever.py
+++ b/scripts/commit_retriever.py
@@ -8,7 +8,7 @@ from logging import INFO, basicConfig, getLogger
 
 import hglib
 
-from bugbug import repository
+from bugbug import db, repository
 
 basicConfig(level=INFO)
 logger = getLogger(__name__)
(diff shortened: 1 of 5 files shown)


scripts/commit_retriever.py
```diff
@@ -52,7 +52,18 @@ def retrieve_commits(self):
         hg.pull(update=True)
         hg.close()
 
-        repository.download_commits(self.repo_dir)
+        db.download_version(repository.COMMITS_DB)
+        if not db.is_old_version(repository.COMMITS_DB):
+            db.download(repository.COMMITS_DB, support_files_too=True)
+
+            for commit in repository.get_commits():
```
calixteman: nit: this line looks pretty strange... maybe you could add a method in repo to just get last commit

PR #547, opened 2019-06-05
Add evaluation criteria to latent semantic indexing

### LINKED ISSUE #517: Use a quick way to get similarities between bugs to reduce the set of bugs to compare to
We have to find a way to make the application of the Duplicate model faster. Right now it requires comparing each bug with every other bug, which is pretty slow.

One option is to use some similarity technique (e.g. https://radimrehurek.com/gensim/tut3.html) to reduce the number of bugs to compare to (e.g. given a bug, find the top X most similar, then apply the model only on those instead of the whole DB).

We could also use the similarity implemented in Bugzilla (when you file a new bug, it will suggest you some similar bugs), if there is some kind of API to access that.

chidauri:
Part of #517

Recall: 16 %.
Precision: 2 %  currently.

### DIFF
--- a/duplicate.py
+++ b/duplicate.py
@@ -1,0 +1,116 @@
+# -*- coding: utf-8 -*-
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# Using latent semantic indexing and similarity matrix
+
+import logging
+
+import nltk
+from gensim import models, similarities
+from gensim.corpora import Dictionary
+from nltk.corpus import stopwords
+from nltk.stem.porter import PorterStemmer
+
+from bugbug import bugzilla
+
+logging.basicConfig(
+    format="%(asctime)s : %(levelname)s : %(message)s", level=logging.INFO
+)
+
+nltk.download("stopwords")
+
+
+class SimilarityLsi:
+    def __init__(self):
+
+        self.corpus = []
+
+        for bug in bugzilla.get_bugs():
+            self.corpus.append([bug["id"], bug["summary"]])
+
+        # cleaning the text
+        self.ps = PorterStemmer()
+        texts = [
+            [
+                self.ps.stem(word)
+                for word in summary.lower().split()
+                if word not in set(stopwords.words("english"))
+            ]
+            for bug_id, summary in self.corpus
+        ]
+
+        # Assigning unique integer ids to all words
+        self.dictionary = Dictionary(texts)
+
+        # conversion to bow
+        corpus_final = [self.dictionary.doc2bow(text) for text in texts]
+
+        # initializing and applying the tfidf transformation model on same corpus,resultant corpus is of same dimensions
+        tfidf = models.TfidfModel(corpus_final)
+        corpus_tfidf = tfidf[corpus_final]
+
+        # transform tfidf corpus to latent 300-D space via LATENT SEMANTIC INDEXING
+        self.lsi = models.LsiModel(
+            corpus_tfidf, id2word=self.dictionary, num_topics=300
+        )
+        corpus_lsi = self.lsi[corpus_tfidf]
+
+        # indexing the corpus
+        self.index = similarities.Similarity(
+            output_prefix="simdata.shdat", corpus=corpus_lsi, num_features=300
+        )
+
+    def get_similar_bugs(self, query, duplicate_of, default=10):
+
+        # transforming the query to latent 300-D space
+        for bug_id, summary in self.corpus:
+            if bug_id == query:
+                query_summary = [
+                    self.ps.stem(word)
+                    for word in summary.lower().split()
+                    if word not in set(stopwords.words("english"))
+                ]
+                break
+        vec_bow = self.dictionary.doc2bow(query_summary)
+        vec_lsi = self.lsi[vec_bow]
+
+        # perform a similarity query against the corpus
+        sims = self.index[vec_lsi]
+        sims = sorted(enumerate(sims), key=lambda item: -item[1])
+
+        # bug_id of the k most similar summaries
+        sim_bug_ids = []
+        master_present = False
+        for i, j in enumerate(sims):
+            if duplicate_of == self.corpus[j[0]][0]:
+                master_present = True
+            if i >= 1 and i <= default:  # since i = 0 returns the query itself
+                sim_bug_ids.append(self.corpus[j[0]][0])
+            if i == 2000:
+                break
+
+        if master_present:
+            return sim_bug_ids
+
+        return False
+
+
+def recall_rate():
+    similarity = SimilarityLsi()
+    total_bugs = 0
+    hits = 0
+    for bug in bugzilla.get_bugs():
+        if bug["dupe_of"]:
+            similar_bugs = similarity.get_similar_bugs(bug["id"], bug["dupe_of"])
+            if similar_bugs:
+                total_bugs += 1
+                if bug["dupe_of"] in similar_bugs:
+                    hits += 1
+
+    print(f"The recall rate is {hits/total_bugs * 100} %")
+
+
+if __name__ == "__main__":
+    recall_rate()
--- a/duplicate.py
+++ b/duplicate.py
@@ -109,20 +109,32 @@ class SimilarityLsi:
         return sim_bug_ids
 
 
-def recall_rate():
+def evaluation():
     similarity = SimilarityLsi()
-    total_bugs = 0
-    hits = 0
+    total_r = 0
+    hits_r = 0
+    total_p = 0
+    hits_p = 0
+
     for bug in bugzilla.get_bugs():
         if duplicates[bug["id"]]:
             similar_bugs = similarity.get_similar_bugs(bug["id"])
-            for item in duplicates[bug["id"]]:
-                total_bugs += 1
-                if item in similar_bugs:
-                    hits += 1
 
-    print(f"The recall rate is {hits/total_bugs * 100} %")
+            # Recall
+            for item in duplicates[bug["id"]]:
+                total_r += 1
+                if item in similar_bugs:
+                    hits_r += 1
+
+            # Precision
+            for element in similar_bugs:
+                total_p += 1
+                if element in duplicates[bug["id"]]:
+                    hits_p += 1
+
+    print(f"Recall: {hits_r/total_r * 100} %")
+    print(f"Precision: {hits_p/total_p * 100} %")
 
 
 if __name__ == "__main__":
-    recall_rate()
+    evaluation()
--- a/duplicate.py
+++ b/duplicate.py
@@ -6,6 +6,7 @@
 # Using latent semantic indexing and similarity matrix
 
 import logging
+import re
 from collections import defaultdict
 
 import nltk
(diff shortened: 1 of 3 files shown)


duplicate.py
```diff
@@ -0,0 +1,116 @@
...
+            if similar_bugs:
+                total_bugs += 1
+                if bug["dupe_of"] in similar_bugs:
+                    hits += 1
+
+    print(f"The recall rate is {hits/total_bugs * 100} %")
+
+
+if __name__ == "__main__":
+    recall_rate()
```
marco-c: We need to compute the precision too (for a given bug, how many of the bugs we suggest are actually duplicates of this bug?)

duplicate.py
```diff
@@ -0,0 +1,116 @@
...
+
+        return False
+
+
+def recall_rate():
+    similarity = SimilarityLsi()
+    total_bugs = 0
+    hits = 0
+    for bug in bugzilla.get_bugs():
+        if bug["dupe_of"]:
```
marco-c: Each bug can have many duplicates.
You have a main bug where all bugs are duplicated to (which has "duplicates" as a list of bugs), and you have many duplicate bugs which are duplicated to the main bug (which have "dupe_of" as a single bug). In the end, they are all duplicates of each other.

duplicate.py
```diff
@@ -0,0 +1,116 @@
...
+nltk.download("stopwords")
+
+
+class SimilarityLsi:
+    def __init__(self):
+
+        self.corpus = []
+
+        for bug in bugzilla.get_bugs():
+            self.corpus.append([bug["id"], bug["summary"]])
```
marco-c: We probably want to use the first comment too.

marco-c:
How long does it take to build the Similarity model?
How long does it take for a get_similar_bugs call?

duplicate.py
chidauri: That is why i chose one of the two ways cause then we'll have false accuracy maybe, where we judge one pair of duplicates two times.

marco-c: If we have multiple duplicate bugs, we should measure the recall for all of them, it's OK if we are judging pairs again, as the main bug will be different in each query.

E.g. say you have bug A, B and C that are duplicates of each other. You should search all duplicates of A and see if B and C are caught. You should search all duplicates of B and see if A and C are caught. Same with C.
It's possible, depending on the implementation, that when you search for duplicates of A you find B and C and when you search for duplicates of B you find only A. So we need to measure all possibilities.

chidauri:
For similarity model : 10 mins approx
For get_similar_bugs : 20-25 mins

marco-c:
>  For get_similar_bugs : 20-25 mins

Each call takes 20 minutes? Or all the calls in total?

chidauri:
> > For get_similar_bugs : 20-25 mins
> 
> Each call takes 20 minutes? Or all the calls in total?

in total for every bug.

marco-c:
> in total for all duplicate bugs

Ah OK, I was really afraid :)
How long does a single get_similar_bugs call take?

chidauri:
> > in total for all duplicate bugs
> 
> Ah OK, I was really afraid :)
> How long does a single get_similar_bugs call take?

appox 1 sec.

marco-c:
> appox 1 sec.

OK, that's pretty good!

duplicate.py
chidauri: how do we calculate performance in the case it doesn't find all the duplicates?

marco-c: That's the recall :)
We just need to compute recall considering all bugs in a duplicate group and not only one.

marco-c: duplicate_group = A + duplicates of A
for bug in duplicate_group:
    found_bugs = find_duplicates(bug)
    # Count how many bugs in duplicate_group are in found_bugs (recall)
    # Count how many bugs in found_bugs are not in duplicate_group (precision)

chidauri: what if all the duplicates are not in the range we're training? currently i am using the similarity method to check that where 2000 is the threshold(it doesn't check after that).

marco-c: The method we need to implement is supposed to search for similar bugs in a very big database. If this one you implemented takes too long for 2000 bugs, then we need to switch to something else.

chidauri: No the model works fine and takes fraction of seconds to find similar bugs from the whole database but a lot of time was spent on checking if that duplicate bug is actually present in the database used,so instead i search for them in similar bugs list returned by the model, only in the top 2000 similars. we need to check that, right?

marco-c: The way to solve this is to get all duplicates in a first pass over the whole DB.
Something like:
```
# A map from bug ID to its duplicate IDs.
duplicates = defaultdict(set)
    dupes = bug["duplicates"]
    dupes.append(bug["dupe_of"])

    duplicates[bug['id']] |= dupes
    for dupe in dupes:
        duplicates[dupe].add(bug['id'])
```

So when evaluating get_similar_bugs for a given bug, you don't need to iterate through the bugs again, but you can just compare the results with `duplicates[bug_id]`.

Limiting to 2000 is bound to be imprecise anyways.

chidauri: The recall rate is now 9.6%.

Should we first improve this number or get precision numbers too ? Anyways precision is going to be too low for current model which can really make us afraid ;)

marco-c: Let's get the precision number first. Then, after merging this PR, we will work on trying out additional techniques and comparing the results.

duplicate.py
chidauri: Precision 1.2%

duplicate.py
```diff
@@ -0,0 +1,140 @@
...
+from bugbug import bugzilla
+
+logging.basicConfig(
+    format="%(asctime)s : %(levelname)s : %(message)s", level=logging.INFO
+)
+
+nltk.download("stopwords")
+
+# A map from bug id to its duplicate ids
+all_ids = defaultdict(int)
```
marco-c: You can use a set instead, since you just want to check for the existene of a given ID.

duplicate.py
```diff
@@ -0,0 +1,140 @@
...
+duplicates = defaultdict(set)
+
+for bug in bugzilla.get_bugs():
+    all_ids[bug["id"]] = 1
```
marco-c: Instead of this, you can just do:
```Python
all_ids = set(bug for bug in bugzilla.get_bugs())
```

But, given that we want to ignore bugs from some specific reporters, you can do (like in bugbug/models/duplicate.py):
```Python
        all_ids = set(
            bug["id"]
            for bug in bugzilla.get_bugs()
            if bug["creator"] not in REPORTERS_TO_IGNORE
            and "dupeme" not in bug["keywords"]
        )
```

duplicate.py
```diff
@@ -0,0 +1,140 @@
...
+
+for bug in bugzilla.get_bugs():
+    dupes = set([bugs for bugs in bug["duplicates"] if all_ids[bugs] != 0])
```
marco-c: Don't use the plural form for a variable that contains a single bug data.
`bug for bug in bug['duplicates'] ...`

marco-c: It should be better in terms of performance to have `dupes` be a list instead of a set.
Instead of `duplicates[bug["id"]] |= dupes`, you can do `duplicates[bug["id"]].update(dupes)`

duplicate.py
```diff
@@ -0,0 +1,140 @@
...
+
+        # indexing the corpus
+        self.index = similarities.Similarity(
+            output_prefix="simdata.shdat", corpus=corpus_lsi, num_features=300
+        )
+
+    def get_similar_bugs(self, query, default=10):
+
+        # transforming the query to latent 300-D space
+        for bug_id, summary in self.corpus:
```
marco-c: I think we don't need to iterate through the corpus, if you make query be a bug instead of a bug ID you can access the summary from there.

duplicate.py
marco-c: Check if precision/recall improve if you consider the first comment too.

duplicate.py
chidauri: isn't dictionary faster to check elements?

marco-c: A dictionary and a set are implemented almost in the same way, they are both really quick for checking the presence of an element. See https://wiki.python.org/moin/TimeComplexity.
The main difference is that you can store a value with the key in the case of a dict, and you can only store the key in the case of a set.

duplicate.py
```diff
@@ -0,0 +1,140 @@
...
+
+        # transform tfidf corpus to latent 300-D space via LATENT SEMANTIC INDEXING
+        self.lsi = models.LsiModel(
+            corpus_tfidf, id2word=self.dictionary, num_topics=300
+        )
+        corpus_lsi = self.lsi[corpus_tfidf]
+
+        # indexing the corpus
+        self.index = similarities.Similarity(
+            output_prefix="simdata.shdat", corpus=corpus_lsi, num_features=300
```
chidauri: @marco-c when this runs it creates 4 files named simdatax.shdat in the same destination, do we have problem with that?

marco-c: Yeah, that's fine!

duplicate.py
chidauri: Done all the changes , recall and precision at 8.6% AND 0.9% respectively.

chidauri:
@marco-c after adding first comment , 
recall : 16 % approx
precision : 2 % approx

duplicate.py
```diff
@@ -0,0 +1,165 @@
+# -*- coding: utf-8 -*-
```
marco-c: Let's call the file "similarity.py" instead.

We'll have to refactor it a bit to make it more generic and make it easy to test other algorithms, but we can do it later.

duplicate.py
```diff
@@ -0,0 +1,165 @@
...
+        corpus_lsi = self.lsi[corpus_tfidf]
+
+        # indexing the corpus
+        self.index = similarities.Similarity(
+            output_prefix="simdata.shdat", corpus=corpus_lsi, num_features=300
+        )
+
+    def get_similar_bugs(self, query, default=10):
+
+        # consider first comment too
```
marco-c: Let's rewrite this as `# Consider both summary and first comment`

duplicate.py
```diff
@@ -0,0 +1,165 @@
...
+        for func in cleanup_functions:
+            query_summary = func(query_summary)
+
+        # cleaning the texts
+        query_summary = re.sub("[^a-zA-Z0-9]", " ", query_summary)
+        query_summary = [
+            self.ps.stem(word)
+            for word in query_summary.lower().split()
+            if word not in set(stopwords.words("english")) and len(word) > 1
+        ]
```
marco-c: There is some common code here and in __init__ that you could refactor in a separate function.

marco-c:
> @marco-c after adding first comment ,
recall : 16 % approx
precision : 2 % approx

Still pretty bad, hopefully with some other algorithm it will improve.

We should also try to debug a little, and see if there is a pattern of bugs we fail to find.

similarity.py
```diff
@@ -0,0 +1,160 @@
...
+
+        # indexing the corpus
+        self.index = similarities.Similarity(
+            output_prefix="simdata.shdat", corpus=corpus_lsi, num_features=300
+        )
+
+    def get_similar_bugs(self, query, default=10):
+
+        # consider both summary and first comment
+        query_summary = query["summary"] + " " + query["comments"][0]["text"]
```
chidauri: @marco-c Here we preprocess text again for every bug, instead we can make use of corpus(line 80) which already has preprocessed texts, in that case we'll have to go through a list if that saves time.

marco-c: No that's fine. In production our corpus will be all previous bugs and we will query with a new bug that is not in the corpus.

marco-c:
Need to rebase on top of master

marco-c:
Merged in 8cf9d43aaf8e21fbd1e16cc3c211106829817eff.

marco-c:
Thanks!

PR #548, opened 2019-06-05
script to generate duplicate classifier results

ayush-1506:
The following script:

1. Trains a duplicate model
2. Classifies the bugs from the previous 3 weeks
3. Writes out the result to duplicate_predictions.csv file.

### DIFF
--- a/scripts/generate_duplicate_sheet.py
+++ b/scripts/generate_duplicate_sheet.py
@@ -2,6 +2,7 @@
 
 import csv
 import itertools
+import json
 from datetime import datetime, timedelta
 
 from bugbug import bugzilla
@@ -10,16 +11,23 @@ from bugbug.models.duplicate import DuplicateModel
 m = DuplicateModel.load("duplicatemodel")
 
 REPORTERS_TO_IGNORE = {"intermittent-bug-filer@mozilla.bugs", "wptsync@mozilla.bugs"}
-test_bugs = bugzilla.download_bugs_between(
-    datetime.now() - timedelta(days=21), datetime.now(), store=False
-)
 
-test_bugs = [bug for bug in test_bugs if not bug["creator"] in REPORTERS_TO_IGNORE]
 
-bug_tuples = itertools.combinations(test_bugs, 2)
+try:
+    with open("duplicate_test_bugs.json", "r") as f:
+        test_bugs = json.load(f)
+except FileNotFoundError:
+    test_bugs = bugzilla.download_bugs_between(
+        datetime.now() - timedelta(days=1), datetime.now(), store=False
+    )
+    test_bugs = [bug for bug in test_bugs if not bug["creator"] in REPORTERS_TO_IGNORE]
+    with open("duplicate_test_bugs.json", "w") as f:
+        json.dump(test_bugs, f)
+
+bug_tuples = list(itertools.combinations(test_bugs, 2))
 
 # Warning: Classifying all the test bugs takes a while
-probs = [m.classify(bug_tuples)]
+probs = m.classify(bug_tuples, probabilities=True)
 
 with open("duplicate_predictions.csv", "w") as csvfile:
     spamwriter = csv.writer(csvfile)
--- a/scripts/generate_duplicate_sheet.py
+++ b/scripts/generate_duplicate_sheet.py
@@ -1,0 +1,42 @@
+# -*- coding: utf-8 -*-
+
+import csv
+import itertools
+from datetime import datetime, timedelta
+
+from bugbug import bugzilla
+from bugbug.models.duplicate import DuplicateModel
+
+m = DuplicateModel.load("duplicatemodel")
+
+REPORTERS_TO_IGNORE = {"intermittent-bug-filer@mozilla.bugs", "wptsync@mozilla.bugs"}
+test_bugs = bugzilla.download_bugs_between(
+    datetime.now() - timedelta(days=21), datetime.now(), store=False
+)
+
+test_bugs = [bug for bug in test_bugs if not bug["creator"] in REPORTERS_TO_IGNORE]
+
+bug_tuples = itertools.combinations(test_bugs, 2)
+
+# Warning: Classifying all the test bugs takes a while
+probs = [m.classify(bug_tuples)]
+
+with open("duplicate_predictions.csv", "w") as csvfile:
+    spamwriter = csv.writer(csvfile)
+
+    spamwriter.writerow(
+        ["bug 1 ID", "bug 1 summary", "bug 2 ID", "bug 2 summary", "prediction"]
+    )
+
+    for bug_tuple, prob in zip(bug_tuples, probs):
+
+        if prob[1] > 0.8:
+            spamwriter.writerow(
+                [
+                    f'https://bugzilla.mozilla.org/show_bug.cgi?id={bug_tuple[0]["id"]}',
+                    bug_tuple[0]["summary"],
+                    f'https://bugzilla.mozilla.org/show_bug.cgi?id={bug_tuple[1]["id"]}',
+                    bug_tuple[1]["summary"],
+                    prob[1],
+                ]
+            )
--- a/bugbug/bugzilla.py
+++ b/bugbug/bugzilla.py
@@ -107,7 +107,7 @@ def _download(ids_or_query):
     return new_bugs
 
 
-def download_bugs_between(date_from, date_to, security=False):
+def download_bugs_between(date_from, date_to, security=False, store=True):
     products = {
         "Add-on SDK",
         "Android Background Services",
@@ -169,10 +169,15 @@ def download_bugs_between(date_from, date_to, security=False):
 
             all_bugs += [bug for bug in new_bugs.values()]
 
-            db.append(
-                BUGS_DB,
-                (bug for bug_id, bug in new_bugs.items() if bug_id not in old_bug_ids),
-            )
+            if store:
+                db.append(
+                    BUGS_DB,
+                    (
+                        bug
+                        for bug_id, bug in new_bugs.items()
+                        if bug_id not in old_bug_ids
+                    ),
+                )
 
     return all_bugs
 


scripts/generate_duplicate_sheet.py
```diff
@@ -0,0 +1,47 @@
+# -*- coding: utf-8 -*-
+
+import csv
+import itertools
+from datetime import datetime
+
+from bugbug import bugzilla
+from bugbug.models.duplicate import DuplicateModel
+
+start = datetime(2019, 5, 15)
```
marco-c: Use relativedelta, or you would have to change this all the time

scripts/generate_duplicate_sheet.py
```diff
@@ -0,0 +1,47 @@
...
+test_bugs = bugzilla.download_bugs_between(start, datetime.now())
+
+m = DuplicateModel()
+m.train()
```
marco-c: You are downloading the bugs to evaluate, and then training the model on all bugs. This means the model will be trained on the bugs to evaluate too.

scripts/generate_duplicate_sheet.py
```diff
@@ -0,0 +1,47 @@
...
+
+csvfile = open("duplicate_predictions.csv", "w", newline="\n")
```
marco-c: Use `with` to avoid having to close the file at the end.

marco-c: I think you can avoid `newline`

scripts/generate_duplicate_sheet.py
```diff
@@ -0,0 +1,47 @@
...
+
+spamwriter = csv.writer(csvfile, delimiter=",")
```
marco-c: I think you can avoid `delimiter` and just use the default

scripts/generate_duplicate_sheet.py
```diff
@@ -0,0 +1,47 @@
...
+
+REPORTERS_TO_IGNORE = {"intermittent-bug-filer@mozilla.bugs", "wptsync@mozilla.bugs"}
+
+# Warning: Classifying all the test bugs takes a while
+for bug_tuple in itertools.combinations(test_bugs, 2):
+    if (
+        bug_tuple[0]["creator"] in REPORTERS_TO_IGNORE
+        or bug_tuple[1]["creator"] in REPORTERS_TO_IGNORE
+        or "dupeme" in bug_tuple[0]["keywords"]
+        or "dupeme" in bug_tuple[1]["keywords"]
```
marco-c: We don't have to ignore "dupeme" bugs in this case

marco-c: Indeed, it's an indication that they might be duplicate

scripts/generate_duplicate_sheet.py
```diff
@@ -0,0 +1,47 @@
...
+    ):
+        continue
+    if (
+        max(m.classify((bug_tuple[0], bug_tuple[1]), probabilities=True)[0]) > 0.8
+        and m.classify(bug_tuple)[0] == 1
+    ):
```
marco-c: It would be much faster if you classified all couples together, instead of in a for loop (similar to https://github.com/mozilla/relman-auto-nag/blob/master/auto_nag/bugbug_utils.py#L100). Then you can iterate on the results like in https://github.com/mozilla/relman-auto-nag/blob/master/auto_nag/scripts/regression.py#L68.

scripts/generate_duplicate_sheet.py
```diff
@@ -0,0 +1,47 @@
...
+    if (
+        bug_tuple[0]["creator"] in REPORTERS_TO_IGNORE
+        or bug_tuple[1]["creator"] in REPORTERS_TO_IGNORE
+        or "dupeme" in bug_tuple[0]["keywords"]
+        or "dupeme" in bug_tuple[1]["keywords"]
+    ):
+        continue
+    if (
+        max(m.classify((bug_tuple[0], bug_tuple[1]), probabilities=True)[0]) > 0.8
+        and m.classify(bug_tuple)[0] == 1
```
marco-c: No need to classify things twice.
```
probs = m.classify(bug_tuple, True)
if probs[1] > 0.8:
```

scripts/generate_duplicate_sheet.py
```diff
@@ -0,0 +1,47 @@
...
+spamwriter.writerow(
+    ["bug 1 ID", "bug 1 summary", "bug 2 ID", "bug 2 summary", "prediction"]
+)
+
+REPORTERS_TO_IGNORE = {"intermittent-bug-filer@mozilla.bugs", "wptsync@mozilla.bugs"}
+
+# Warning: Classifying all the test bugs takes a while
+for bug_tuple in itertools.combinations(test_bugs, 2):
+    if (
+        bug_tuple[0]["creator"] in REPORTERS_TO_IGNORE
```
marco-c: You can skip these directly in test_bugs, so you avoid making the combinations altogether

scripts/generate_duplicate_sheet.py
```diff
@@ -0,0 +1,47 @@
...
+        max(m.classify((bug_tuple[0], bug_tuple[1]), probabilities=True)[0]) > 0.8
+        and m.classify(bug_tuple)[0] == 1
+    ):
+        spamwriter.writerow(
+            [
+                f'https://bugzilla.mozilla.org/show_bug.cgi?id={bug_tuple[0]["id"]}',
+                bug_tuple[0]["summary"],
+                f'https://bugzilla.mozilla.org/show_bug.cgi?id={bug_tuple[1]["id"]}',
+                bug_tuple[1]["summary"],
+                m.classify(bug_tuple),
```
marco-c: And you're doing a third classification here

scripts/generate_duplicate_sheet.py
```diff
@@ -0,0 +1,42 @@
...
+
+from bugbug import bugzilla
+from bugbug.models.duplicate import DuplicateModel
+
+m = DuplicateModel()
+m.train()
+
+
+REPORTERS_TO_IGNORE = {"intermittent-bug-filer@mozilla.bugs", "wptsync@mozilla.bugs"}
+delta = timedelta(days=21)
```
marco-c: Nit: just use `datetime.now() - timedelta(days=21)` and remove the delta variable

scripts/generate_duplicate_sheet.py
```diff
@@ -0,0 +1,42 @@
...
+probs = [m.classify(bug_tuple)[0] for bug_tuple in bug_tuples]
+
+with open("duplicate_predictions.csv", "w") as csvfile:
+    spamwriter = csv.writer(csvfile)
+
+    spamwriter.writerow(
+        ["bug 1 ID", "bug 1 summary", "bug 2 ID", "bug 2 summary", "prediction"]
+    )
+
+    # Warning: Classifying all the test bugs takes a while
```
marco-c: This comment should be moved to where the classification is happening

scripts/generate_duplicate_sheet.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+REPORTERS_TO_IGNORE = {"intermittent-bug-filer@mozilla.bugs", "wptsync@mozilla.bugs"}
+delta = timedelta(days=21)
+test_bugs = bugzilla.download_bugs_between(datetime.now() - delta, datetime.now())
+
+test_bugs = [bug for bug in test_bugs if not bug["creator"] in REPORTERS_TO_IGNORE]
+
+bug_tuples = itertools.combinations(test_bugs, 2)
+probs = [m.classify(bug_tuple)[0] for bug_tuple in bug_tuples]
```
marco-c: You are still doing the classification bug couple per bug couple. It makes way more sense to classify them all together, as it's way faster.

marco-c: https://github.com/mozilla/bugbug/pull/548#discussion_r290684822

scripts/generate_duplicate_sheet.py
```diff
@@ -0,0 +1,42 @@
...
+from bugbug import bugzilla
+from bugbug.models.duplicate import DuplicateModel
+
+m = DuplicateModel()
+m.train()
+
+
+REPORTERS_TO_IGNORE = {"intermittent-bug-filer@mozilla.bugs", "wptsync@mozilla.bugs"}
+delta = timedelta(days=21)
+test_bugs = bugzilla.download_bugs_between(datetime.now() - delta, datetime.now())
```
marco-c: https://github.com/mozilla/bugbug/pull/548#discussion_r290682355 still kind of applies.
This will add the bugs to the DB, so when you run the script again you will train the model on these bugs too.

ayush-1506:
@marco-c r?

scripts/generate_duplicate_sheet.py
```diff
@@ -0,0 +1,42 @@
...
+test_bugs = bugzilla.download_bugs_between(
+    datetime.now() - timedelta(days=21), datetime.now(), store=False
+)
+
+test_bugs = [bug for bug in test_bugs if not bug["creator"] in REPORTERS_TO_IGNORE]
+
+bug_tuples = itertools.combinations(test_bugs, 2)
+
+# Warning: Classifying all the test bugs takes a while
+probs = [m.classify(bug_tuples)]
```
marco-c: m.classify should already be returning a list, so if you do this you'll get a list with a single element being a list of predictions.
Also, you should use probabilities=True, or you won't get probabilities.

scripts/generate_duplicate_sheet.py
```diff
@@ -0,0 +1,42 @@
...
+
+from bugbug import bugzilla
+from bugbug.models.duplicate import DuplicateModel
+
+m = DuplicateModel.load("duplicatemodel")
+
+REPORTERS_TO_IGNORE = {"intermittent-bug-filer@mozilla.bugs", "wptsync@mozilla.bugs"}
+test_bugs = bugzilla.download_bugs_between(
+    datetime.now() - timedelta(days=21), datetime.now(), store=False
+)
```
marco-c: How long does the download take? If it takes a long time, you could store the data in a json file

ayush-1506: It takes about 6-7 minutes.

marco-c: OK, so maybe it's worth doing that so that you can iterate faster.
Just do something like:
```
try:
    with open('duplicate_test_bugs.json', 'r') as f:
        test_bugs = json.load(f)
except FileNotFoundError:
        test_bugs = DOWNLOAD
        with open('duplicate_test_bugs.json', 'w') as f:
            json.dump(test_bugs, f)
```

ayush-1506: Sure, also, I get an unexpected error while classifying the model :
```
from bugbug import bugzilla
from bugbug.models.duplicate import DuplicateModel as dm
bugs = bugzilla.get_bugs()
b=  (next(bugs), next(bugs))
m = dm.load("duplicatemodel")

m.classify(b)
```

Error : 
```
  File "<stdin>", line 1, in <module>
  File "/home/ayush99/GitHub/bugbug/bugbug/model.py", line 191, in classify
    X = self.extraction_pipeline.transform(items)
  File "/home/ayush99/GitHub/bbug/lib/python3.7/site-packages/sklearn/pipeline.py", line 540, in _transform
    Xt = transform.transform(Xt)
  File "/home/ayush99/GitHub/bugbug/bugbug/bug_features.py", line 444, in transform
    author_ids = get_author_ids() if self.commit_map else None
AttributeError: 'BugExtractor' object has no attribute 'commit_map'
```
I wasn't getting this earlier today,  not sure what's causing this. @marco can you confirm this?

ayush-1506:
@marco-c r?

PR #553, opened 2019-06-06
Add support for specific Docker tag in spawn_data_pipeline.py

Lothiraldan:
Reverts mozilla/bugbug#499

### DIFF
--- a/infra/spawn_data_pipeline.py
+++ b/infra/spawn_data_pipeline.py
@@ -110,20 +110,26 @@ def main():
         # Override the Docker image tag if needed
         if docker_tag:
             base_image = payload["payload"]["image"]
-            splitted_image = base_image.rsplit(":", 1)
 
-            # If we have already a Docker tag
-            if len(splitted_image) > 1:
-                docker_tag = splitted_image
+            if base_image.startswith("mozilla/bugbug"):
+                splitted_image = base_image.rsplit(":", 1)
 
-                if docker_tag != "latest":
-                    msg = "Don't update image {} for task {} as its already has a tag"
-                    print(msg.format(base_image, task_internal_id))
+                # If we have already a Docker tag
+                if len(splitted_image) > 1:
+                    docker_tag = splitted_image
 
-            tagless_image = splitted_image[0]
+                    if docker_tag != "latest":
+                        msg = (
+                            "Don't update image {} for task {} as it already has a tag"
+                        )
+                        print(msg.format(base_image, task_internal_id))
 
-            new_image = "{}:{}".format(tagless_image, docker_tag)
-            payload["payload"]["image"] = new_image
+                tagless_image = splitted_image[0]
+
+                new_image = "{}:{}".format(tagless_image, docker_tag)
+                msg = "Updating image for task {} to {}"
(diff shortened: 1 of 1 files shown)


Lothiraldan:
I have fixed the `spawn_data_pipeline.py` by updating the Docker tag only for task which uses `latest` or no tag at all. This way we don't touch the bugbug docker tag.

Lothiraldan:
Example output with data-pipeline:

```
Don't update image mozilla/taskboot:0.1.3 for task docker-build as its already has a tag
Don't update image mozilla/taskboot:0.1.3 for task docker-push as its already has a tag
Don't update image mozilla/taskboot:0.1.3 for task http-deploy as its already has a tag
```

marco-c:
Could we restrict the change of the tag to images starting with `mozilla/bugbug`?

Maybe we should go for the JSON-e solution in the future, so that things are more clearly defined and nothing unexpected can happen.

infra/spawn_data_pipeline.py
```diff
@@ -105,6 +107,24 @@ def main():
...
+        if docker_tag:
+            base_image = payload["payload"]["image"]
+            splitted_image = base_image.rsplit(":", 1)
+
+            # If we have already a Docker tag
+            if len(splitted_image) > 1:
+                docker_tag = splitted_image
+
+                if docker_tag != "latest":
+                    msg = "Don't update image {} for task {} as its already has a tag"
```
marco-c: Typo: `as it already has a tag`

marco-c:
> Maybe we should go for the JSON-e solution in the future, so that things are more clearly defined and nothing unexpected can happen.

Let's file an issue about this so we don't forget.

Lothiraldan:
> Could we restrict the change of the tag to images starting with `mozilla/bugbug`?

Yes we could, I will update this PR
> 
> Maybe we should go for the JSON-e solution in the future, so that things are more clearly defined and nothing unexpected can happen.

If I remember correctly we cannot as it is already a JSON-E template given to the hook definition

marco-c:
> If I remember correctly we cannot as it is already a JSON-E template given to the hook definition

Right :(

infra/spawn_data_pipeline.py
```diff
@@ -105,6 +107,30 @@ def main():
...
+
+                # If we have already a Docker tag
+                if len(splitted_image) > 1:
+                    docker_tag = splitted_image
+
+                    if docker_tag != "latest":
+                        msg = (
+                            "Don't update image {} for task {} as it already has a tag"
+                        )
+                        print(msg.format(base_image, task_internal_id))
```
marco-c: Nit: you could use a f-string here...

infra/spawn_data_pipeline.py
```diff
@@ -105,6 +107,30 @@ def main():
...
+
+                tagless_image = splitted_image[0]
+
+                new_image = "{}:{}".format(tagless_image, docker_tag)
+                msg = "Updating image for task {} to {}"
+                print(msg.format(task_internal_id, new_image))
```
marco-c: ...and here.

PR #554, opened 2019-06-06
Build tagged Docker images - #340

Lothiraldan:


marco-c:
We should do the same in the data-pipeline for the http images, right?

marco-c:
Ah I see you put it at the end of the TODO list at https://github.com/mozilla/bugbug/issues/340#issuecomment-499526768.

.taskcluster.yml
```diff
@@ -138,13 +138,19 @@ tasks:
...
+              then: {$eval: 'head_branch[10:]' }
+              else: "latest"
           command:
             - "/bin/sh"
             - "-lcxe"
             - "git clone ${repository} /code &&
                cd /code &&
                git checkout ${head_rev} &&
-               taskboot --cache /cache --target /code build-compose --registry=registry.hub.docker.com --write /images --build-arg CHECK_MODELS=0"
+               taskboot --cache /cache --target /code build-compose --write /images --build-arg CHECK_MODELS=0 --tag $VERSION --tag latest"
```
marco-c: When $VERSION is latest, we'd build the image twice.

marco-c: Filed https://github.com/mozilla/bugbug/issues/558 for this.

PR #561, opened 2019-06-07
Use tagged index urls for pushing artifacts

Lothiraldan:


marco-c:
Instead of adding the code to the scripts (thus making them less generic), we could make the decision task (spawn pipeline) inject "index.project.relman.bugbug.train_duplicate.VERSION" in the routes of the tasks. What do you think?

Lothiraldan:
We can indeed check if each task has an index route starting with `index.project.relman.bugbug` and replace latest by the tag. Few comments:

- We will need to access the version of bugbug to retrieve the artifacts anyway in training images
- It is not explicit that data-pipeline will be dynamically updated before spawning and we cannot use JSON-e as we would need to double-escape part of the templating.
- It would makes it very difficult to makes it generic when moving from `spawn_data_pipeline.py` to a task-boot util.

On the plus side:
- It's quite easy to do
- It doesn't require to add taskcluster code in bugbug itself.

marco-c:
> We will need to access the version of bugbug to retrieve the artifacts anyway in training images

This is fine, as it's still generic.

> It is not explicit that data-pipeline will be dynamically updated before spawning and we cannot use JSON-e as we would need to double-escape part of the templating.

Actually, I'm thinking that in this case we can use JSON-e, as we are rendering JSON-e already in the decision task: https://github.com/mozilla/bugbug/blob/6e39b0a5a5cf772621510632596b7d9ec8ae1774/infra/spawn_data_pipeline.py#L83.

infra/spawn_data_pipeline.py
```diff
@@ -79,7 +80,7 @@ def main():
 
     for task in raw_tasks["tasks"]:
         # Try render the task template
-        context = {}
+        context = {"docker_tag": docker_tag}
```
marco-c: Nit: can we name it "version" instead?

Lothiraldan: Done

infra/data-pipeline.yml
```diff
@@ -27,7 +27,7 @@ tasks:
     routes:
       - notify.email.release-mgmt-analysis@mozilla.com.on-failed"
       - notify.irc-channel.#bugbug.on-failed
-      - index.project.relman.bugbug.data_commits.latest
+      - index.project.relman.bugbug.data_commits.${version}
```
marco-c: We should index both on latest and on ${version}, so for bugbug we use versioned indexes, and other users might want to always use latest.

infra/spawn_data_pipeline.py
```diff
@@ -79,7 +79,7 @@ def main():
 
     for task in raw_tasks["tasks"]:
         # Try render the task template
-        context = {}
+        context = {"version": docker_tag}
```
marco-c: Nit: since you use it only here, you might as well do `"version": os.getenv("TAG", "latest")`

marco-c: Perhaps we could even remove TAG from the env and use get_bugbug_version you just introduced, but that's not a big difference

PR #567, opened 2019-06-07
 Add .taskcluster.yml validator as a pre-commit check 

marco-c:


marco-c:
The lint task expectedly fails with:
```
jsone.shared.InterpreterError: InterpreterError at template.tasks[8].payload.command[2]: unknown context value TAG
```

I'll remove the bad commit now so we can merge this.

PR #584, opened 2019-06-13
Fix default Python version declaration in pre-commit config

Lothiraldan:
The new way of of declaring default Python version was introduced in
pre-commit 1.14.0 and warning about unknown config keys was introduced in
pre-commit 1.17.0.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -36,4 +36,4 @@ repos:
     -   id: check-hooks-apply
     -   id: check-useless-excludes
 default_language_version:
-    python: python3.6
+    python: python3.7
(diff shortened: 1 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -35,4 +35,5 @@ repos:
     hooks:
     -   id: check-hooks-apply
     -   id: check-useless-excludes
-python_version: python3.6
+default_language_version:
+    python: python3.6
```
marco-c: We should use 3.7

Lothiraldan: I would prefer to stay on Python 3.6, as Bugbug targets Python 3.6, we know that all developers have at least Python 3.6 installed when working on bubug.

marco-c: I guess we should switch to targeting 3.7, as I'm already using some features that were added in 3.7.

PR #588, opened 2019-06-13
For bugs where the regression keyword was removed, we can overwrite the assigned class with 'non-regression'

marco-c:


bugbug/models/regression.py
```diff
@@ -12,3 +12,25 @@ def __init__(self, lemmatization=False, historical=False):
...
+                    if change["field_name"] == "keywords":
+                        if "regression" in [
+                            k.strip() for k in change["removed"].split(",")
+                        ]:
+                            regression_keyword_removed = True
+                        elif "regression" in [
+                            k.strip() for k in change["added"].split(",")
+                        ]:
+                            regression_keyword_removed = False
+
```
Lothiraldan: This will detect if the latest change to the bug keywords was removing the `regression` keyword or adding the `regression` keyword, right?

marco-c: It will set the class to 0 when the keyword was removed. I reset regression_keyword_removed to False when the keyword is added so that if it's removed and then added back, the class is not 0.

Lothiraldan: Code is slightly different than https://github.com/mozilla/relman-auto-nag/blob/master/auto_nag/scripts/regression.py#L61 which detects if the `regression` keyword was ever removed: https://github.com/mozilla/relman-auto-nag/blob/master/auto_nag/scripts/regression.py#L61

marco-c: The difference should only be that in that case we didn't check if the 'regression' keyword was re-added, but simply because the check was done automatically by the Bugzilla query: https://github.com/mozilla/relman-auto-nag/blob/master/auto_nag/scripts/regression.py#L37.

Lothiraldan: Got it, sounds good to me

PR #594, opened 2019-06-14
Use a single spawn image for both hooks. Fix #365

### LINKED ISSUE #365: Use generic "spawn" image and stop using "spawn-data" and "spawn-check"
Follow-up to #341.

Lothiraldan:
This removes the need to keep two different Docker images for spawning the
data pipeline and the check pipeline.

I realized that the Docker image in the hook definition are **not** versioned so we will need to find a solution for that before being able to rollback.


marco-c:
> I realized that the Docker image in the hook definition are not versioned so we will need to find a solution for that before being able to rollback.

We can edit the hooks manually.

PR #595, opened 2019-06-16
It Don't try to find inconsistencies in all bugs multiple times

### LINKED ISSUE #593: Don't try to find inconsistencies in all bugs multiple times
We are re-downloading inconsistent bugs at https://github.com/mozilla/bugbug/blob/cc9c5e8c5d69e2dcabf35741a3f1d9b9bb797b39/scripts/bug_retriever.py#L61.

To find them, we are iterating through all bugs.

It makes sense to iterate through all bugs the first time, but the second time we should only iterate through the bugs we found inconsistent the first time (indeed, we can be sure that all the other bugs were not inconsistent).

To fix this, we simply need to modify https://github.com/mozilla/bugbug/blob/cc9c5e8c5d69e2dcabf35741a3f1d9b9bb797b39/bugbug/bug_snapshot.py#L755 adding a "bugs" parameter, and call it in the right way in bug_retriever.py (the first time bugs=bugzilla.get_bugs(), the second and following times bugs=previously_inconsistent_bugs).

Abhi-khandelwal:
Fixes #593

### DIFF
--- a/scripts/bug_retriever.py
+++ b/scripts/bug_retriever.py
@@ -63,20 +63,19 @@ class Retriever(object):
             # We look for inconsistencies in all bugs first, then, on following passes,
             # we only look for inconsistencies in bugs that were found to be inconsistent in the first pass
             if i == 0:
-                inconsistent_bugs_set = bugzilla.get_bugs()
+                inconsistent_bugs = bugzilla.get_bugs()
             else:
-                inconsistent_bugs_set = inconsistent_bugs
-            inconsistent_bugs = bug_snapshot.get_inconsistencies(inconsistent_bugs_set)
-            bug_ids = set(inconsistent_bugs)
+                inconsistent_bugs = inconsistent_bugs
+            inconsistent_bugs_set = set(bug_snapshot.get_inconsistencies(inconsistent_bugs))
 
-            if len(bug_ids) == 0:
+            if len(inconsistent_bugs_set) == 0:
                 break
 
             logger.info(
-                f"Re-downloading {len(bug_ids)} bugs, as they were inconsistent"
+                f"Re-downloading {len(inconsistent_bugs_set)} bugs, as they were inconsistent"
             )
-            bugzilla.delete_bugs(lambda bug: bug["id"] in bug_ids)
-            bugzilla.download_bugs(bug_ids)
+            bugzilla.delete_bugs(lambda bug: bug["id"] in inconsistent_bugs_set)
+            bugzilla.download_bugs(inconsistent_bugs_set)
 
         self.compress_file("data/bugs.json")
 
--- a/scripts/bug_retriever.py
+++ b/scripts/bug_retriever.py
@@ -63,19 +63,19 @@ class Retriever(object):
             # We look for inconsistencies in all bugs first, then, on following passes,
             # we only look for inconsistencies in bugs that were found to be inconsistent in the first pass
             if i == 0:
-                inconsistent_bugs = bugzilla.get_bugs()
+                inconsistent_bugs = bug_snapshot.get_inconsistencies(bugzilla.get_bugs())
             else:
-                inconsistent_bugs = inconsistent_bugs
-            inconsistent_bugs_set = set(bug_snapshot.get_inconsistencies(inconsistent_bugs))
+                inconsistent_bugs = bug_snapshot.get_inconsistencies(inconsistent_bugs)
+            inconsistent_bugs_ids = set(list(map(lambda bug: bug["id"], inconsistent_bugs)))
 
-            if len(inconsistent_bugs_set) == 0:
+            if len(inconsistent_bugs_ids) == 0:
                 break
 
             logger.info(
-                f"Re-downloading {len(inconsistent_bugs_set)} bugs, as they were inconsistent"
+                f"Re-downloading {len(inconsistent_bugs_ids)} bugs, as they were inconsistent"
             )
-            bugzilla.delete_bugs(lambda bug: bug["id"] in inconsistent_bugs_set)
-            bugzilla.download_bugs(inconsistent_bugs_set)
+            bugzilla.delete_bugs(lambda bug: bug["id"] in inconsistent_bugs_ids)
+            bugzilla.download_bugs(inconsistent_bugs_ids)
 
         self.compress_file("data/bugs.json")
 
--- a/scripts/bug_retriever.py
+++ b/scripts/bug_retriever.py
@@ -67,7 +67,7 @@ class Retriever(object):
             else:
                 inconsistent_bugs_set = inconsistent_bugs
             inconsistent_bugs = bug_snapshot.get_inconsistencies(inconsistent_bugs_set)
-            bug_ids = set(inconsistent_bugs_set)
+            bug_ids = set(inconsistent_bugs)
 
             if len(bug_ids) == 0:
                 break
--- a/scripts/bug_retriever.py
+++ b/scripts/bug_retriever.py
@@ -59,9 +59,16 @@ class Retriever(object):
         bugzilla.download_bugs(timespan_ids + labelled_bug_ids)
 
         # Try to re-download inconsistent bugs, up to three times.
+        previously_inconsistent_bugs = []
         for i in range(3):
-            bugs = bugzilla.get_bugs()
-            bug_ids = set(bug_snapshot.get_inconsistencies(bugs))
+            if i == 0:
+                bugs = bugzilla.get_bugs()
+                previously_inconsistent_bugs = bug_snapshot.get_inconsistencies(bugs)
+            else:
+                bugs = previously_inconsistent_bugs
+                previously_inconsistent_bugs = bug_snapshot.get_inconsistencies(bugs)
+            bug_ids = set(previously_inconsistent_bugs)
+
             if len(bug_ids) == 0:
                 break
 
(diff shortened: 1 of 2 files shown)


scripts/bug_retriever.py
```diff
@@ -60,7 +60,8 @@ def retrieve_bugs(self):
 
         # Try to re-download inconsistent bugs, up to three times.
         for i in range(3):
-            bug_ids = set(bug_snapshot.get_inconsistencies())
+            bugs = bugzilla.get_bugs()
+            bug_ids = set(bug_snapshot.get_inconsistencies(bugs))
```
marco-c: This is still downloading all bugs

Abhi-khandelwal: Can you please help me with some extra details ?

marco-c: I've explained what needs to be done in https://github.com/mozilla/bugbug/issues/593#issue-456315165, feel free to ask in that issue about any doubts you might have.

scripts/bug_retriever.py
```diff
@@ -59,9 +59,16 @@ def retrieve_bugs(self):
         bugzilla.download_bugs(timespan_ids + labelled_bug_ids)
 
         # Try to re-download inconsistent bugs, up to three times.
+        previously_inconsistent_bugs = []
         for i in range(3):
-            bugs = bugzilla.get_bugs()
-            bug_ids = set(bug_snapshot.get_inconsistencies(bugs))
+            if i == 0:
+                bugs = bugzilla.get_bugs()
+                previously_inconsistent_bugs = bug_snapshot.get_inconsistencies(bugs)
```
marco-c: This line is the same between the two branches, you can remove it from the two branches and call it after the if-else.

scripts/bug_retriever.py
```diff
@@ -59,9 +59,16 @@ def retrieve_bugs(self):
         bugzilla.download_bugs(timespan_ids + labelled_bug_ids)
 
         # Try to re-download inconsistent bugs, up to three times.
+        previously_inconsistent_bugs = []
         for i in range(3):
-            bugs = bugzilla.get_bugs()
-            bug_ids = set(bug_snapshot.get_inconsistencies(bugs))
+            if i == 0:
```
marco-c: Add a comment before the if saying: `We look for inconsistencies in all bugs first, then, on following passes, we only look for inconsistencies in bugs that were found to be inconsistent in the first pass`.

scripts/bug_retriever.py
```diff
@@ -59,9 +59,16 @@ def retrieve_bugs(self):
         bugzilla.download_bugs(timespan_ids + labelled_bug_ids)
 
         # Try to re-download inconsistent bugs, up to three times.
+        previously_inconsistent_bugs = []
```
marco-c: No need to define this variable here

scripts/bug_retriever.py
marco-c: Nit: you can change the variable name to `inconsistent_bugs`

scripts/bug_retriever.py
```diff
@@ -59,9 +59,16 @@ def retrieve_bugs(self):
...
         for i in range(3):
-            bugs = bugzilla.get_bugs()
-            bug_ids = set(bug_snapshot.get_inconsistencies(bugs))
+            if i == 0:
+                bugs = bugzilla.get_bugs()
+                previously_inconsistent_bugs = bug_snapshot.get_inconsistencies(bugs)
+            else:
+                bugs = previously_inconsistent_bugs
+                previously_inconsistent_bugs = bug_snapshot.get_inconsistencies(bugs)
+            bug_ids = set(previously_inconsistent_bugs)
```
marco-c: And you can change this one to `inconsistent_bugs_set`

scripts/bug_retriever.py
```diff
@@ -60,7 +60,15 @@ def retrieve_bugs(self):
 
         # Try to re-download inconsistent bugs, up to three times.
         for i in range(3):
-            bug_ids = set(bug_snapshot.get_inconsistencies())
+            # We look for inconsistencies in all bugs first, then, on following passes,
+            # we only look for inconsistencies in bugs that were found to be inconsistent in the first pass
+            if i == 0:
+                inconsistent_bugs_set = bugzilla.get_bugs()
```
marco-c: This is not a set, rename it to `inconsistent_bugs`

scripts/bug_retriever.py
```diff
@@ -60,7 +60,15 @@ def retrieve_bugs(self):
...
+            else:
+                inconsistent_bugs_set = inconsistent_bugs
+            inconsistent_bugs = bug_snapshot.get_inconsistencies(inconsistent_bugs_set)
+            bug_ids = set(inconsistent_bugs)
```
marco-c: This one is the variable to rename to `inconsistent_bugs_set` (bug_ids -> inconsistent_bugs_set)

scripts/bug_retriever.py
```diff
@@ -60,7 +60,15 @@ def retrieve_bugs(self):
...
         # Try to re-download inconsistent bugs, up to three times.
         for i in range(3):
-            bug_ids = set(bug_snapshot.get_inconsistencies())
+            # We look for inconsistencies in all bugs first, then, on following passes,
+            # we only look for inconsistencies in bugs that were found to be inconsistent in the first pass
+            if i == 0:
+                inconsistent_bugs_set = bugzilla.get_bugs()
+            else:
+                inconsistent_bugs_set = inconsistent_bugs
+            inconsistent_bugs = bug_snapshot.get_inconsistencies(inconsistent_bugs_set)
```
marco-c: I don't think this is going to work. On the first pass, you are passing an iterator on bugs, on the second you are passing a list of bug IDs (instead, it should be a list of bugs).

You need to modify `get_inconsistencies` to return bugs instead of just their IDs.

scripts/bug_retriever.py
```diff
@@ -60,7 +60,15 @@ def retrieve_bugs(self):
 
         # Try to re-download inconsistent bugs, up to three times.
         for i in range(3):
-            bug_ids = set(bug_snapshot.get_inconsistencies())
+            # We look for inconsistencies in all bugs first, then, on following passes,
+            # we only look for inconsistencies in bugs that were found to be inconsistent in the first pass
+            if i == 0:
+                inconsistent_bugs_set = bugzilla.get_bugs()
+            else:
+                inconsistent_bugs_set = inconsistent_bugs
```
marco-c: Same here, `inconsistent_bugs`

marco-c: Well, you can remove this else completely :)

scripts/bug_retriever.py
marco-c: Or even better, `inconsistent_bug_ids`

scripts/bug_retriever.py
```diff
@@ -60,15 +60,22 @@ def retrieve_bugs(self):
...
         # Try to re-download inconsistent bugs, up to three times.
         for i in range(3):
-            bug_ids = set(bug_snapshot.get_inconsistencies())
-            if len(bug_ids) == 0:
+            # We look for inconsistencies in all bugs first, then, on following passes,
+            # we only look for inconsistencies in bugs that were found to be inconsistent in the first pass
+            if i == 0:
+                inconsistent_bugs = bugzilla.get_bugs()
+            else:
+                inconsistent_bugs = inconsistent_bugs
```
marco-c: You can definitely remove this line, assigning a variable to itself doesn't have any effect :P

Abhi-khandelwal: Yeah , But again our main agenda won't work that in second loop we should pass parameters which are inconsistent at first time.

marco-c: I'm not sure what you mean.

Abhi-khandelwal: Sorry, I got your point.

scripts/bug_retriever.py
```diff
@@ -60,15 +60,22 @@ def retrieve_bugs(self):
...
         # Try to re-download inconsistent bugs, up to three times.
         for i in range(3):
-            bug_ids = set(bug_snapshot.get_inconsistencies())
-            if len(bug_ids) == 0:
+            # We look for inconsistencies in all bugs first, then, on following passes,
+            # we only look for inconsistencies in bugs that were found to be inconsistent in the first pass
+            if i == 0:
+                inconsistent_bugs = bug_snapshot.get_inconsistencies(bugzilla.get_bugs())
+            else:
+                inconsistent_bugs = bug_snapshot.get_inconsistencies(inconsistent_bugs)
```
marco-c: You can actually move it after, like I told you before.
```Python
if i == 0:
    inconsistent_bugs = bugzilla.get_bugs()

inconsistent_bugs = bug_snapshot.get_inconsistencies(inconsistent_bugs)
```

Or even better:
```Python
inconsistent_bugs = bugzilla.get_bugs()
for i in range(3):
    inconsistent_bugs = bug_snapshot.get_inconsistencies(inconsistent_bugs)
```

scripts/bug_retriever.py
```diff
@@ -60,15 +60,22 @@ def retrieve_bugs(self):
...
+            inconsistent_bugs_ids = set(list(map(lambda bug: bug["id"], inconsistent_bugs)))
```
marco-c: Rename the variable as `inconsistent_bug_ids`

marco-c: No need to convert to a list first, you can build the set directly:
```Python
inconsistent_bug_ids = set(bug["id"] for bug in inconsistent_bugs)
```

Abhi-khandelwal: Ohk.

Abhi-khandelwal:
> Thanks!

Thank you so much for helping me , in my first contribution :)

PR #603, opened 2019-06-20
Pre-load bugbug models in the RQ worker. Fix #557

### LINKED ISSUE #557: Pre-load model in HTTP service background worker
Due to RQ worker design, we cannot cache models without implementing our own worker script.

https://github.com/mozilla/bugbug/pull/321/files/225b0a03969aeb5c63c064dec0215e18a632b59f#r291106686

Lothiraldan:
Implements our own background worker script in order to be able to load the
bugbug models before the RQ worker is forking. This way we load them once at
startup and each job will be able to reuse it.

### DIFF
--- a/http_service/worker.py
+++ b/http_service/worker.py
@@ -6,16 +6,21 @@
 
 import os
 import sys
-from os.path import dirname, join
+from os.path import abspath, dirname, join
 
 from redis import Redis
 from rq import Connection, Worker
 
-import http_service.models  # Need to import at with the same name
+sys.path.insert(0, abspath(join(dirname(__file__), "..")))
+
+# We need to make sure that the models.py file is imported with the same name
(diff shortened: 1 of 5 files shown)


http_service/worker.py
```diff
@@ -0,0 +1,29 @@
...
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import os
+import sys
+from os.path import dirname, join
+
+from redis import Redis
+from rq import Connection, Worker
+
+import http_service.models  # Need to import at with the same name
```
marco-c: Can you clarify the comment here?

marco-c:
How does it work exactly? Who's loading worker.py?

marco-c:
> How does it work exactly? Who's loading worker.py?

Maybe you need to change the CMD of the bg_worker?

PR #604, opened 2019-06-20
Start tracking training metrics in taskcluster index. Fix #342

### LINKED ISSUE #342: Store training metrics in the Taskcluster index after training

Lothiraldan:
We will likely wants to track more metrics in the future, it should be as easy
as adding them to `tracking_metrics` in `model.train`.

I think it mights fail when training the component model as it set
`cross_validation_enabled` to False.

### DIFF
--- a/bugbug/model.py
+++ b/bugbug/model.py
@@ -98,6 +98,7 @@ class Model:
 
             scores = cross_validate(pipeline, X_train, y_train, scoring=scorings, cv=5)
 
+            print("Cross Validation scores:")
             for scoring in scorings:
                 score = scores[f"test_{scoring}"]
                 tracking_metrics[f"test_{scoring}"] = {
--- a/bugbug/model.py
+++ b/bugbug/model.py
@@ -138,6 +138,7 @@ class Model:
 
         # Evaluate results on the test set.
         y_pred = self.clf.predict(X_test)
+
         print(f"No confidence threshold - {len(y_test)} classified")
         print(metrics.confusion_matrix(y_test, y_pred, labels=class_names))
         print(classification_report_imbalanced(y_test, y_pred, labels=class_names))
--- a/bugbug/model.py
+++ b/bugbug/model.py
@@ -140,7 +140,10 @@ class Model:
         y_pred = self.clf.predict(X_test)
 
         print(f"No confidence threshold - {len(y_test)} classified")
-        print(metrics.confusion_matrix(y_test, y_pred, labels=class_names))
+        confusion_matrix = metrics.confusion_matrix(y_test, y_pred, labels=class_names)
+        print(confusion_matrix)
+        tracking_metrics["confusion_matrix"] = confusion_matrix.tolist()
+
         print(classification_report_imbalanced(y_test, y_pred, labels=class_names))
 
         # Evaluate results on the test set for some confidence thresholds.
--- a/bugbug/model.py
+++ b/bugbug/model.py
@@ -88,6 +88,8 @@ class Model:
         else:
             pipeline = self.clf
 
+        tracking_metrics = {}
+
         # Use k-fold cross validation to evaluate results.
         if self.cross_validation_enabled:
             scorings = ["accuracy"]
@@ -96,9 +98,12 @@ class Model:
 
             scores = cross_validate(pipeline, X_train, y_train, scoring=scorings, cv=5)
 
-            print("Cross Validation scores:")
             for scoring in scorings:
                 score = scores[f"test_{scoring}"]
+                tracking_metrics[f"test_{scoring}"] = {
+                    "mean": score.mean(),
+                    "std": score.std() * 2,
+                }
                 print(
                     f"{scoring.capitalize()}: f{score.mean()} (+/- {score.std() * 2})"
                 )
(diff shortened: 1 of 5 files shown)


marco-c:
Maybe we could store them in a file instead of the index directly, then they would be accessible by a URL like https://index.taskcluster.net/v1/task/project.relman.bugbug.train_component.latest/artifacts/public/metrics.json and we wouldn't need Taskcluster specific code in the script.

marco-c:
Could you also add a script to retrieve the training metrics for a given model and from a given date?

marco-c:
> Maybe we could store them in a file instead of the index directly, then they would be accessible by a URL like https://index.taskcluster.net/v1/task/project.relman.bugbug.train_component.latest/artifacts/public/metrics.json and we wouldn't need Taskcluster specific code in the script.

For the date part, it could be implemented like we implement the version index (with a JSON-e context variable).

Lothiraldan:
@marco-c I have added a script to retrieve the metrics, do you want me to add it bugbug `console_scripts`?

bugbug/model.py
```diff
@@ -133,7 +138,6 @@ def train(self, importance_cutoff=0.15):
 
         # Evaluate results on the test set.
         y_pred = self.clf.predict(X_test)
-
```
marco-c: Nit: don't remove newline here

infra/data-pipeline.yml
```diff
@@ -1,455 +1,494 @@
...
-    routes:
-      - notify.email.release-mgmt-analysis@mozilla.com.on-failed"
-      - notify.irc-channel.#bugbug.on-failed
-    metadata:
-      name: bugbug docker http service deploy
-      description: bugbug docker http service deploy
-      owner: release-mgmt-analysis@mozilla.com
-      source: https://github.com/mozilla/bugbug/raw/master/data-pipeline.yml
+  $let:
+    now: {$eval: 'now'}
```
marco-c: What if you define year, month, etc. too here?

infra/spawn_data_pipeline.py
```diff
@@ -75,37 +75,36 @@ def main():
     with open(args.data_pipeline_json) as pipeline_file:
         raw_tasks = yaml.safe_load(pipeline_file.read())
 
-    for task in raw_tasks["tasks"]:
```
marco-c: Unrelated to this PR, but we should really add some tests for this module (mocking Taskcluster), as otherwise we notice breakages really late

marco-c: Could you file an issue for this?

scripts/retrieve_training_metrics.py
```diff
@@ -0,0 +1,69 @@
+# -*- coding: utf-8 -*-
+
+import argparse
+import sys
+from os.path import abspath
+
+import requests
+
+LATEST_URI = "train_{}.latest"
+VERSIONNED_URI = "train_{}.{}"
```
marco-c: `VERSIONNED` -> `VERSIONED`

scripts/retrieve_training_metrics.py
```diff
@@ -0,0 +1,69 @@
+# -*- coding: utf-8 -*-
```
marco-c: When you pass a date, the script should get all tasks from that date forward (so we can later generate a graph).

marco-c: Maybe we can implement that later, when we have at least a few tasks indexed by date. Could you file an issue?

bugbug/model.py
```diff
@@ -133,7 +138,6 @@ def train(self, importance_cutoff=0.15):
...
         print(f"No confidence threshold - {len(y_test)} classified")
         print(metrics.confusion_matrix(y_test, y_pred, labels=class_names))
         print(classification_report_imbalanced(y_test, y_pred, labels=class_names))
```
marco-c: You can add metrics from this too (which will be the only metrics when CV is disabled).

infra/data-pipeline.yml
Lothiraldan: I'm not sure if I define them as:

```diff
   $let:
-    now: {$eval: 'now'}
+    year: {$eval: 'now[0:4]'}
+    month: {$eval: 'now[5:7]'}
```

that `now` will be the same datetime object. For year and month it shouldn't be a problem but for minute and seconds it might be problematic.

marco-c: Maybe you can do another let:
```
$let:
    now: {$eval: 'now'}
in:
  $let:
    year: $now[0:4]
...
```

Lothiraldan: I checked locally and it seems fine, I hope json-e implementation used by the taskcluster instance behaves the same

scripts/retrieve_training_metrics.py
Lothiraldan: Sure

infra/spawn_data_pipeline.py
Lothiraldan: Done https://github.com/mozilla/bugbug/issues/613

scripts/retrieve_training_metrics.py
Lothiraldan: https://github.com/mozilla/bugbug/issues/614

bugbug/model.py
Lothiraldan: I've added the confusion matrix to the tracking metrics. Do you want to add also the metrics from  `classification_report_imbalanced`? If yes, I imagine it would be the raw metrics, not the formatted report, right?

marco-c: Yep!

marco-c: Filed #619 for this.

PR #617, opened 2019-06-22
Start working on OpenApi definition and docs

Lothiraldan:


Lothiraldan:
One thing I'm having trouble documenting is when the bug itself is not available as we are currently returning a 200 status code. It might be better to return a 40x status code instead.

Also the answer schema for the batch API is very hard to define because we might have successfully predicted bug in the answer, in-processing bug and unavailable bugs in the same answer payload.

I see three possibilities:
- I can try to find a way to define the answer payload to handle all the cases
- We can move the retrieval of the result to the non-batch API, so we would batch X bugs and for each bug the consumer will ask the "normal" endpoint about them.
- We only document the answer through examples.

marco-c:
> We can move the retrieval of the result to the non-batch API, so we would batch X bugs and for each bug the consumer will ask the "normal" endpoint about them.

This might be too slow if you have a lot of bugs to retrieve.

> We only document the answer through examples.

This might be the easiest way. Maybe a short explanation of how it works followed by examples.

http_service/app.py
```diff
@@ -8,16 +8,42 @@
...
 
-from .models import classify_bug
+from bugbug import get_bugbug_version
+
+from .models import MODELS_NAMES, classify_bug
 
 API_TOKEN = "X-Api-Key"
 
+API_DESCRIPTION = """
+This is the documentation for the BubBug http service, the platform for Bugzilla Machine Learning project.
```
marco-c:
```suggestion
This is the documentation for the BubBug http service, the platform for Bugzilla Machine Learning projects.
```

http_service/app.py
```diff
@@ -8,16 +8,42 @@
...
+
+# Introduction
+
+This service can be used to classify a given bug using one of pre-trained model.
```
marco-c:
```suggestion
This service can be used to classify a given bug using a pre-trained model.
```

http_service/app.py
```diff
@@ -8,16 +8,42 @@
...
+You can classify a single bug or a batch of bugs.
+The classification happens on background so you need to call back the service for getting the results.
```
marco-c:
```suggestion
The classification happens in the background so you need to call back the service for getting the results.
```

http_service/app.py
```diff
@@ -85,15 +142,52 @@ def get_bug_classification(model_name, bug_id):
...
 
 
-@application.route("/<model_name>/predict/<bug_id>")
+@application.route("/<model_name>/predict/<int:bug_id>")
+@cross_origin()
 def model_prediction(model_name, bug_id):
+    """
+    ---
+    get:
+      description: Classify a single bug using given model, answer either 200 if the bug is processed or 202 if at least the bug is being processed
```
marco-c:
```suggestion
      description: Classify a single bug using given model, answer either 200 if the bug is processed or 202 if the bug is being processed
```

http_service/app.py
```diff
@@ -85,15 +142,52 @@ def get_bug_classification(model_name, bug_id):
...
+      summary: Classify a single bug id
```
marco-c:
```suggestion
      summary: Classify a single bug
```

http_service/app.py
```diff
@@ -85,15 +142,52 @@ def get_bug_classification(model_name, bug_id):
...
+          type: integer
+          example: 123456
+      responses:
+        200:
+          description: A single bug prediction
+          content:
+            application/json:
+              schema: BugPrediction
+        202:
+          description: A temporary answer for bug being processed
```
marco-c:
```suggestion
          description: A temporary answer for the bug being processed
```

http_service/app.py
```diff
@@ -110,13 +204,124 @@ def model_prediction(model_name, bug_id):
...
+        ```<br/><br/>
+
+        And you will get the following output once the bugs are available:
+        <br/>
+        ```
+        {"bugs": {"456": {"extra_data": {}, "index": 0, "prob": [0], "suggestion": ""}}}
+        ```<br/><br/>
+
+        Please be aware that the results might be interleaved, so the
+        following output is valid:
```
marco-c: What did you mean here?

http_service/app.py
```diff
@@ -110,13 +204,124 @@ def model_prediction(model_name, bug_id):
...
+        <br/>
+        ```
+        {"bugs": {"123": {"available": False}, "456": {"extra_data": {}, "index": 0, "prob": [0], "suggestion": ""}}}
+        ```
+      summary: Classify a batch of bugs id
```
marco-c:
```suggestion
      summary: Classify a batch of bugs
```

http_service/app.py
```diff
@@ -110,13 +204,124 @@ def model_prediction(model_name, bug_id):
...
+                      index: 0
+                      prob: [0]
+                      suggestion: string
+                    789012:
+                      extra_data: {}
+                      index: 0
+                      prob: [0]
+                      suggestion: string
+        202:
+          description: A temporary answer for bug being processed
```
marco-c:
```suggestion
          description: A temporary answer for bugs being processed
```

http_service/models.py
```diff
@@ -109,7 +109,8 @@ def classify_bug(
 
     model = get_model(model_name)
 
-    model_extra_data = model.get_extra_data()
+    # model_extra_data = model.get_extra_data()
+    model_extra_data = {}
```
marco-c: Why this change?

http_service/templates/doc.html
```diff
@@ -0,0 +1,24 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>BugBug doc</title>
+    <!-- needed for adaptive design -->
```
marco-c: Nit: remove this comment

http_service/templates/doc.html
```diff
@@ -0,0 +1,24 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>BugBug doc</title>
```
marco-c:
```suggestion
    <title>BugBug Documentation</title>
```

http_service/app.py
Lothiraldan: I meant that for a batch, some of the bugs might be available, some of the bugs might still be processing and some of the bugs results might be ready

marco-c: Oh, OK, can you specify that?
Something like:
```
Please be aware that the results might be interleaved, so the following output, where a bug is returned and another one is still being processed, is valid:
```

Lothiraldan:
@marco-c Did you take a look at the generated documentation ? It's on http://localhost:5000/doc when running locally

http_service/models.py
Lothiraldan: Unrelated change, checking that everything is working as it should without it

http_service/app.py
Lothiraldan: Ok done

http_service/models.py
Lothiraldan: Everything is good

PR #622, opened 2019-06-23
option to select size of training set for duplicate classifier

### LINKED ISSUE #521: Add options to the Duplicate classifier to select number of bugs in the training set

chidauri:
closes #521

### DIFF
--- a/bugbug/models/duplicate.py
+++ b/bugbug/models/duplicate.py
@@ -28,7 +28,7 @@ class LinearSVCWithLabelEncoding(CalibratedClassifierCV):
 
 
 class DuplicateModel(BugCoupleModel):
-    def __init__(self, training_size, lemmatization=False):
+    def __init__(self, training_size=14000, lemmatization=False):
         self.num_duplicates = training_size // 2
         self.num_nondups_nondups = self.num_dup_nondups = training_size // 4
 
--- a/bugbug/models/duplicate.py
+++ b/bugbug/models/duplicate.py
@@ -14,10 +14,6 @@ from sklearn.svm import LinearSVC
 from bugbug import bug_features, bugzilla, feature_cleanup
 from bugbug.model import BugCoupleModel
 
-NUM_DUPLICATES = 7000
-NUM_DUP_NONDUPS = 3500
-NUM_NONDUPS_NONDUPS = 3500
-
 REPORTERS_TO_IGNORE = {"intermittent-bug-filer@mozilla.bugs", "wptsync@mozilla.bugs"}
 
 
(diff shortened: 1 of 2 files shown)


run.py
```diff
@@ -45,6 +52,7 @@ def parse_args(args):
                 defectenhancementtask and regression tasks.""",
         action="store_true",
     )
+
```
marco-c: Nit: remove this empty line here (change unrelated to your PR).

run.py
```diff
@@ -22,6 +22,13 @@ def parse_args(args):
         help="Perform lemmatization (using spaCy)",
         action="store_true",
     )
+    parser.add_argument(
+        "--trainsize",
+        nargs="?",
+        default=14000,
+        type=int,
+        help="size of training set of duplicate model",
```
marco-c: Nit: "The size of the training set for the duplicate model"

run.py
```diff
@@ -22,6 +22,13 @@ def parse_args(args):
         help="Perform lemmatization (using spaCy)",
         action="store_true",
     )
+    parser.add_argument(
+        "--trainsize",
```
marco-c: --training-set-size

bugbug/models/duplicate.py
```diff
@@ -32,7 +28,10 @@ def fit(self, X, y):
 
 
 class DuplicateModel(BugCoupleModel):
-    def __init__(self, lemmatization=False):
+    def __init__(self, training_size, lemmatization=False):
+        self.NUM_DUPLICATES = training_size // 2
+        self.NUM_NONDUPS_NONDUPS = self.NUM_DUP_NONDUPS = training_size // 4
```
marco-c: In Python, the convention is to use uppercase for constants. Since these are not constants anymore, make them lower case.

run.py
chidauri: can we use --training_set_size? cause hyphen gives error.

marco-c: Yes, that's fine too

marco-c: A question, what error does it give?

chidauri: args.training-set-size considers it three different words training ,set and size .

marco-c: Who considers it three different words? What error do you see exactly?

chidauri: Traceback (most recent call last):
  File "run.py", line 156, in <module>
    main(parse_args(sys.argv[1:]))
  File "run.py", line 87, in main
    model = model_class(args.training-set-size, args.lemmatization)
AttributeError: 'Namespace' object has no attribute 'training'

chidauri: because it contains the word 'set' .

marco-c: OK, so the "who" is Python :P 
The issue is that in Python you can't have variables with dashes in them!
argparse will automatically take care of that, so if you have an argument like "--training-set-size" you can access it as "args.training_set_size". You shouldn't need to change the argument to "--training_set_size" for that to happen.

chidauri: ohh, so python does have some rules :P

chidauri: Done.

bugbug/models/duplicate.py
```diff
@@ -32,7 +28,10 @@ def fit(self, X, y):
 
 
 class DuplicateModel(BugCoupleModel):
-    def __init__(self, lemmatization=False):
+    def __init__(self, training_size, lemmatization=False):
```
marco-c: We'll have to set a default here, otherwise scripts/trainer.py will fail. You can use the same default you set in run.py.

chidauri: Done.

Issue #535, opened 2019-06-03
Add tests for repository.download_component_mapping


marco-c:
This test will have to use `responses`, similarly to the test for repository.download_commits.

codingisbliss:
Hello macro-c I want to contribute to this bug. I am a newbie to this, how can I get some reading material/ what this code does inforamtion ?

marco-c:
Hello @codingisbliss, this method is pretty much self-contained. It is here: https://github.com/mozilla/bugbug/blob/ff5eba38322593f3a9c186b6da64e704ea0ab2a4/bugbug/repository.py#L645.
Here's an example of another test for another method which had to mock the download of the component mapping: https://github.com/mozilla/bugbug/blob/master/tests/test_repository.py#L244.

You will have to do something similar, but mocking the download in different ways to test all possible cases.

PR #623, opened 2019-06-24
add test for repository.download_component_mapping

cklyyung:
Pull Request for issue #535.
Test cases:
1. Downloading 0 components
2. Downloading 2 components
3. Downloading 2 component with same e-tag as previous download

### DIFF
--- a/tests/test_repository.py
+++ b/tests/test_repository.py
@@ -284,26 +284,7 @@ def test_download_component_mapping():
         responses.HEAD,
         "https://index.taskcluster.net/v1/task/gecko.v2.mozilla-central.latest.source.source-bugzilla-info/artifacts/public/components.json",
         status=200,
-        headers={"ETag": "102"},
-    )
-
-    responses.add(
-        responses.GET,
-        "https://index.taskcluster.net/v1/task/gecko.v2.mozilla-central.latest.source.source-bugzilla-info/artifacts/public/components.json",
-        status=200,
-        json={"README.txt": ["Core", "General"]},
-    )
-
-    repository.download_component_mapping()
-    assert len(repository.path_to_component) == 1
-    assert repository.path_to_component["README.txt"] == "Core::General"
-
-    responses.reset()
-    responses.add(
-        responses.HEAD,
-        "https://index.taskcluster.net/v1/task/gecko.v2.mozilla-central.latest.source.source-bugzilla-info/artifacts/public/components.json",
-        status=200,
-        headers={"ETag": "102"},
+        headers={"ETag": "101"},
     )
 
     responses.add(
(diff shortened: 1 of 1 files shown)


tests/test_repository.py
```diff
@@ -237,6 +237,87 @@ def test_hg_log(fake_hg_repo):
...
+    responses.add(
+        responses.GET,
+        "https://index.taskcluster.net/v1/task/gecko.v2.mozilla-central.latest.source.source-bugzilla-info/artifacts/public/components.json",
+        status=200,
+        json={"README.txt": ["Core", "General"]},
+    )
+
+    repository.download_component_mapping()
+    assert len(repository.path_to_component) == 1
+    assert repository.path_to_component["README.txt"] == "Core::General"
```
marco-c: This case (1 component) is redundant with the previous one, so we can remove it.

tests/test_repository.py
```diff
@@ -237,6 +237,87 @@ def test_hg_log(fake_hg_repo):
...
+        status=200,
+        headers={"ETag": "102"},
+    )
+
+    responses.add(
+        responses.GET,
+        "https://index.taskcluster.net/v1/task/gecko.v2.mozilla-central.latest.source.source-bugzilla-info/artifacts/public/components.json",
+        status=200,
+        json={"AUTHORS": ["mozilla.org", "Licensing"]},
+    )
```
marco-c: And of course when you remove that, you'll have to update this one to match Etag 101

cklyyung: Noted; I removed the test case of 1, and adjusted the e-tag.

tests/test_repository.py
```diff
@@ -237,6 +237,69 @@ def test_hg_log(fake_hg_repo):
...
+
+    responses.reset()
+    responses.add(
+        responses.HEAD,
+        "https://index.taskcluster.net/v1/task/gecko.v2.mozilla-central.latest.source.source-bugzilla-info/artifacts/public/components.json",
+        status=200,
+        headers={"ETag": "101"},
+    )
+
+    responses.add(
```
marco-c: This one could be removed (the code will check the etag and if it's the same it won't download the file).

cklyyung: Ah, that's what my test case is, asserting that the response with the same etag won't be downloaded. If you think it's unnecessary, I will remove it when I have time later today.

marco-c: Exactly, so you just need to mock the HEAD request where the Etag is retrieved, this second request will never happen and so should not be mocked (this way we ensure it doesn't happen, which is what we want)

cklyyung: Ah, I understand, I misunderstood and thought you meant the whole test case.

Issue #611, opened 2019-06-21
Make similarity script more generic so different similarity algorithms can be evaluated


marco-c:
Right now, LSI is backed in the similarity script. We need to modify the similarity script to make it more generic so we can compare different algorithms.

ayush-1506:
I'll add a cosine similarity model using tf-idf vectors and refactor the file a bit. @marco-c Shouldn't similarity.py be in scripts/ ?

marco-c:
> I'll add a cosine similarity model using tf-idf vectors and refactor the file a bit

What are you going to use for the search?

> Shouldn't similarity.py be in scripts/ ?

similarity.py is a bit like model.py. The evaluation part should stay there, the "__main__" part should go in a script. But this is less important than making it generic.

ayush-1506:
For the most similar bugs, I was thinking of a threshold value, (say 0.8), above which all pairs are classified as duplicate.

marco-c:
@ayush-1506 we also have `k`, which is the maximum number of bugs returned. I'm not asking about that though, I'm wondering: once you have a vector representing a bug, how will you search for similar vectors? You can't iterate through all bugs, it will be *extremely* slow.

marco-c:
You will need to use some library like nmslib, or https://scikit-learn.org/stable/modules/neighbors.html#finding-the-nearest-neighbors.

marco-c:
Or even BM25 (https://github.com/mozilla/bugbug/issues/612).

ayush-1506:
Once we have the word vectors, it can be treated as a clustering preblem. If this isn't a good idea, I don't see a way other than iterating the entire buds db.

marco-c:
@ayush-1506 see my comments above. I think you don't really need clustering here, you just need a fast way to find similar vectors.

ayush-1506:
Yep, I'll add this. Thanks!

PR #628, opened 2019-06-24
Make similarity script more generic, add NearestNeighbors similarity

ayush-1506:
Fixes : #611

### DIFF
--- a/bugbug/similarity.py
+++ b/bugbug/similarity.py
@@ -6,6 +6,9 @@
 import re
 from collections import defaultdict
 
+from sklearn.feature_extraction.text import TfidfVectorizer
+from sklearn.neighbors import NearestNeighbors
+
 from bugbug import bugzilla, feature_cleanup
 
 OPT_MSG_MISSING = (
@@ -53,7 +56,11 @@ for bug in bugzilla.get_bugs():
         duplicates[dupe].add(bug["id"])
 
 
-def text_preprocess(text):
+def get_text(bug):
+    return "{} {}".format(bug["summary"], bug["comments"][0]["text"])
+
+
+def text_preprocess(text, join=False):
     for func in cleanup_functions:
         text = func(text)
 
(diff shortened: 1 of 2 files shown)


ayush-1506:
@marco-c Should I add another file to call this script? (from scripts/)

bugbug/similarity.py
```diff
@@ -115,33 +123,92 @@ def get_similar_bugs(self, query, k=10):
...
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--algorithm", choices=["lsi", "neighbors"], help="Algorithm to use"
+    )
+    return parser
 
-    print(f"Recall: {hits_r/total_r * 100}%")
-    print(f"Precision: {hits_p/total_p * 100}%")
 
+def main(args):
```
marco-c: We should move this part in a script under scripts/

bugbug/similarity.py
```diff
@@ -115,33 +123,92 @@ def get_similar_bugs(self, query, k=10):
         # Get IDs of the k most similar bugs
         return [self.corpus[j[0]][0] for j in sims[:k]]
 
+    def evaluation(self):
```
marco-c: This is the same between the two classes.
You should either have a base Similarity class with this function, and make the other classes extend Similarity, or you can move this function out of the classes and make it have a single parameter "similarity_obj" instead of "self"

bugbug/similarity.py
```diff
@@ -116,32 +153,23 @@ def get_similar_bugs(self, query, k=10):
...
-                    hits_p += 1
+        self.vectorizer.fit(text)
+        self.similarity_calculator.fit(self.vectorizer.transform(text))
 
-    print(f"Recall: {hits_r/total_r * 100}%")
-    print(f"Precision: {hits_p/total_p * 100}%")
+    def get_similar_bugs(self, query):
+        bugs = []
+        for bug_data in bugzilla.get_bugs():
+            bugs.append(bug_data["id"])
```
marco-c: Why are you storing data in `bugs` here? It seems like you don't need it?

marco-c: Ah you are using it to get the IDs. You should refactor this, otherwise get_similar_bugs will be extremely slow.

marco-c: E.g. you can make the list of IDs be an attribute of the class, similarly to LSISimilarity (the IDs in that case are stored in the corpus)

bugbug/similarity.py
```diff
@@ -116,32 +153,23 @@ def get_similar_bugs(self, query, k=10):
...
-
-    for bug in bugzilla.get_bugs():
-        if duplicates[bug["id"]]:
-            similar_bugs = similarity.get_similar_bugs(bug)
+class NeighborsSimilarity(BaseSimilarity):
+    def __init__(self, k=10, vectorizer=TfidfVectorizer):
+        self.vectorizer = TfidfVectorizer()
+        self.similarity_calculator = NearestNeighbors(n_neighbors=k)
+        text = []
+        self.bugs = []
```
marco-c: Nit: please rename this bug_ids, or it can be confusing

marco-c:
Could you run the two and compare their results?

ayush-1506:
@marco-c The results seem pretty close. Testing on ~8k bugs:
```
algorithm=lsi
Recall: 45.78313253012048%
Precision: 5.480769230769231%

algorithm=neighbors
Recall: 46.787148594377506%
Precision: 5.600961538461538%
```

scripts/compare_similarity.py
```diff
@@ -0,0 +1,30 @@
+# -*- coding: utf-8 -*-
```
marco-c: Rename the script `evaluate_similarity`, as it's not doing comparison but just running a model to evaluate it.

PR #636, opened 2019-06-27
Merge common parts of .taskcluster.yml.

### LINKED ISSUE #268: Merge common parts of .taskcluster.yml
This will make the file more readable.

Abhi-khandelwal:
Fix #268


Abhi-khandelwal:
Hello @marco-c , am I doing right ?

marco-c:
@Abhi-khandelwal that's mostly a good start, but there are some indentation issues (see the comment from the Taskcluster bot).

Abhi-khandelwal:
@marco-c Can you please guide me that How can I locally test this file.

marco-c:
@Abhi-khandelwal the best way is to keep making changes and pushing them to see if they work. Testing them locally is not so easy (you could test a part of it by running "pre-commit run --all-files", but this won't catch all possible issues).

Abhi-khandelwal:
@marco-c  do we need more changes?

marco-c:
It looks good to me so far, can you rebase on top of master?

Abhi-khandelwal:
Your welcome @marco-c  :)

marco-c:
@imbstack any idea why Taskcluster isn't doing anything here?

imbstack:
@marco-c ah, so it seems like this config is not generating any tasks for this event. We should raise that to you as the user but instead we have a bug where we just crash with ` "Cannot read property 'task' of undefined"` on [this line](https://github.com/taskcluster/taskcluster/blob/e61d672b6a76dfce9a2fcc87f57c8b601d886444/services/github/src/handlers.js#L642). We will fix our stuff so that it reports the error better. It is also likely that this config is doing something unexpected. Does that make sense?

imbstack:
@marco-c ah, so it seems like this config is not generating any tasks for this event. We should raise that to you as the user but instead we have a bug where we just crash with ` "Cannot read property 'task' of undefined"` on [this line](https://github.com/taskcluster/taskcluster/blob/e61d672b6a76dfce9a2fcc87f57c8b601d886444/services/github/src/handlers.js#L642). We will fix our stuff so that it reports the error better. It is also likely that this config is doing something unexpected. Does that make sense?

marco-c:
@imbstack thanks!
So, @Abhi-khandelwal it looks like there is still something to fix :)

Abhi-khandelwal:
@marco-c let me check , @imbstack could you please tell me how can I locally test this?, I already tried with pre-commit but it won't work.

imbstack:
@Abhi-khandelwal, the `.taskcluster.yml` file is a json-e template that gets rendered with context from github events. I've had success debugging things like this in the past by using the [json-e playground](https://taskcluster.github.io/json-e/#Playground). If that doesn't end up being helpful, let me know and I can try to look into this further with you!

imbstack:
@Abhi-khandelwal aha! I was looking at your .taskcluster.yml to try to see how to raise this issue best on our end and I think I see the issue. The `tasks` field you're generating here is not an array of tasks but rather a task object. Perhaps making the `mergeDeep` on line 39 inside an array will fix it?

Abhi-khandelwal:
@imbstack , I got your point , that tasks field should be array of tasks instead of object but I implement mergeDeep function by using this link
https://github.com/mozilla-mobile/android-components/blob/9ddbaaf4084dc2e18a547de9f525ebdc528e1361/.taskcluster.yml

and still I don't able to understand how should I make this an array of tasks.

.taskcluster.yml
```diff
@@ -26,351 +26,343 @@ tasks:
...
-               git -c advice.detachedHead=false checkout ${head_rev} &&
-               pip install --quiet -r test-requirements.txt &&
-               pre-commit run -a" # TODO Be smarter and run only on modified files
         metadata:
-          name: bugbug lint
-          description: bugbug lint
           owner: ${user}@users.noreply.github.com
           source: ${repository}/raw/${head_rev}/.taskcluster.yml
+    in:
+      $if: 'tasks_for in ["github-pull-request", "github-push"]'
```
marco-c: probably the array should start from here.

An array in yaml is something like:
```
- element1
- element2
- element3
```

Or:
```
- taskId: val1
  created: val1
  ...
- taskId: val2
  created: val2
  ...
```

Abhi-khandelwal: @marco-c  you mean I should use `tasks` variable after Line-36 instead of Line-4 ?

marco-c: No, I mean you should do:
```
- $if: 'tasks_for in 
```
instead of
```
$if: 'tasks_for in 
```

Abhi-khandelwal:
@marco-c , I think we should change the path at L-168, under docker_build task, we should define some json or  log file I guess

marco-c:
> @marco-c , I think we should change the path at L-168, under docker_build task, we should define some json or log file I guess

No, that's the correct path. The task is building images and storing them in /images.
Note currently all tasks are working. The goal of this PR is to clean up .taskcluster.yml, but we shouldn't change anything in the current behavior.

Abhi-khandelwal:
So, do I need to fix something else?

marco-c:
> So, do I need to fix something else?

Yeah, as you can see the tasks are failing (https://tools.taskcluster.net/task-group-inspector/#/dfQ8CV4FRFKjy_k0TLj_fw) and there are less tasks than normal (https://tools.taskcluster.net/task-group-inspector/#/SSr8IaHCTuaUP9veKbfdxw).

The best way to debug this would be to get the current .taskcluster.yml and your changed .taskcluster.yml, run it through https://github.com/marco-c/taskcluster_yml_validator/blob/master/taskcluster_yml_validator/__init__.py#L14 and comparing the output of https://github.com/marco-c/taskcluster_yml_validator/blob/master/taskcluster_yml_validator/__init__.py#L24 with the two .taskcluster.yml files.

marco-c:
Closing for inactivity, but feel free to reopen if you still intend to work on this.

PR #639, opened 2019-06-27
Track also imbalance report metrics. Fix #619

### LINKED ISSUE #619: Store imbalanced metrics too

Lothiraldan:


bugbug/_classification.py
```diff
@@ -0,0 +1,95 @@
...
+        geometric_mean_score
+    )
+    iba = iba_gmean(
+        y_true, y_pred, labels=labels, average=None, sample_weight=sample_weight
+    )
+
+    result = {"targets": {}}
+
+    for i, label in enumerate(labels):
+        result["targets"][target_names[i]] = {
```
Lothiraldan: @marco-c Do you prefer the label or the index as the dict keys?

bugbug/_classification.py
```diff
@@ -0,0 +1,95 @@
...
+def classification_report_imbalanced_values(
+    y_true,
+    y_pred,
+    labels=None,
+    target_names=None,
+    sample_weight=None,
+    digits=2,
+    alpha=0.1,
+):
+    """Reimplementation of imblearn.metrics.classification_report_imbalanced
```
marco-c: Whoah, I thought we could access the raw data somehow :(

How bad would it be to parse the output table?

The other option is to just store precision and recall, so we don't have to reimplement all. In the end, that's what I mostly use.

bugbug/_classification.py
marco-c: The label, so it's more readable

bugbug/model.py
```diff
@@ -98,7 +99,6 @@ def train(self, importance_cutoff=0.15):
 
             scores = cross_validate(pipeline, X_train, y_train, scoring=scorings, cv=5)
 
-            print("Cross Validation scores:")
```
marco-c: Can you put this back? And add a print at line 140 to say "Test Set scores:"

bugbug/_classification.py
Lothiraldan: So it should be good, I need to check with another model than defectenhancementtask

bugbug/_classification.py
Lothiraldan: I've removed some extra code in `classification_report_imbalanced_values` that were not needed. I think the amount of code duplication is acceptable now.

I'm more than hesitant to parse the output table, we will likely hit some corner-cases which will makes the parsing code non trivial, we might parse the output wrongly and any dependency update can breaks the parsing either loudly or worse silently.

bugbug/_classification.py
```diff
@@ -0,0 +1,79 @@
...
+    to have access to the raw values. Copied from version 0.4.3"""
+    if labels is None:
+        labels = unique_labels(y_true, y_pred)
+    else:
+        labels = np.asarray(labels)
+
+    if target_names is None:
```
marco-c: You can remove this, since we are always passing the class names

bugbug/_classification.py
```diff
@@ -0,0 +1,79 @@
...
+from imblearn.metrics import (
+    geometric_mean_score,
+    make_index_balanced_accuracy,
+    specificity_score,
+)
+from sklearn.metrics.classification import precision_recall_fscore_support
+from sklearn.utils.multiclass import unique_labels
+
+
+def classification_report_imbalanced_values(
```
marco-c: Move this in model.py

marco-c: So you can remove this file (or we need to find a nicer name for it)

bugbug/_classification.py
```diff
@@ -0,0 +1,79 @@
...
+    )
+    # Index balanced accuracy
+    iba_gmean = make_index_balanced_accuracy(alpha=alpha, squared=True)(
+        geometric_mean_score
+    )
+    iba = iba_gmean(
+        y_true, y_pred, labels=labels, average=None, sample_weight=sample_weight
+    )
+
+    result = {"targets": {}}
```
marco-c: We can probably remove "targets" and just store things in result

marco-c: Ah you are storing the averages too, right, forget it.

bugbug/_classification.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def classification_report_imbalanced_values(
+    y_true,
+    y_pred,
+    labels=None,
+    target_names=None,
+    sample_weight=None,
+    digits=2,
+    alpha=0.1,
```
marco-c: If you have made other modifications to the function, you can also remove the parameters we don't use.

If you have kept it basically 1:1, leave it as is

Lothiraldan: I have only removed the formatting supporting code and changed the content of the loop to create the final dict instead of populating the final format.

For reference the initial code is located here: https://github.com/scikit-learn-contrib/imbalanced-learn/blob/master/imblearn/metrics/_classification.py#L750

marco-c: OK, sounds good. Can you add the link (a perma one) to the comment?

Lothiraldan: Sure

bugbug/utils.py
```diff
@@ -155,3 +156,16 @@ def retry(operation, retries=5, wait_between_retries=30):
                 raise
 
             time.sleep(wait_between_retries)
+
+
+class CustomJsonEncoder(json.JSONEncoder):
+    """ A custom Json Encoder to supports Numpy types
```
marco-c:
```suggestion
    """ A custom Json Encoder to support Numpy types
```

PR #640, opened 2019-06-27
Use tagged version in bugbug hooks. Fix #634

### LINKED ISSUE #634: Add tag in the hook images
Currently we use mozilla/bugbug-spawn-pipeline with no tag.

Lothiraldan:
I introduced a new script because updating the existing script
`set_docker_image_version.py` seems to contradicts its name.

Another solution would be to rename the script to something like
`prepare_hook.py` and do everything we need there.

I also checked if we could use jq directly to makes the changes and it seems a
bit too complex to be maintainable: https://jqplay.org/s/CgYOI1ZXxj. But it's
also a possibility.


marco-c:
We could have a single script, `set_hook_version.py` which sets both TAG and image version.

Yeah using jq is too much :P

marco-c:
> We could have a single script, set_hook_version.py which sets both TAG and image version.

Just like you said as a second option, I missed that.

Yeah, I'd prefer this

PR #644, opened 2019-06-28
Remove fallback on downloading xz files (#624)

### LINKED ISSUE #624: Remove fallback on downloading xz files
It's added in #524 because the current task have xz as artifacts. Once we make a successful deployment, we can switch to zst fully and remove the xz fallback.

x249wang:
Fixes #624.

### DIFF
--- a/bugbug/db.py
+++ b/bugbug/db.py
@@ -99,10 +99,6 @@ def last_modified(path):
     last_modified = utils.get_last_modified(url)
 
     if last_modified is None:
-        base_url = os.path.splitext(url)[0]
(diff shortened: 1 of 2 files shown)


bugbug/db.py
```diff
@@ -132,7 +100,7 @@ def last_modified(path):
 
     if last_modified is None:
         base_url = os.path.splitext(url)[0]
-        last_modified = utils.get_last_modified(f"{base_url}.xz")
+        last_modified = utils.get_last_modified(f"{base_url}.zst")
```
marco-c: You're trying with zst twice here (the first time with the unmodified URL and the second time here). You can instead completely remove this branch

Issue #648, opened 2019-06-28
Write a script for rollbacks


marco-c:
This should help when we want to rollback.

It should:
1. Remind the user to cancel any running broken data-pipeline task (or do it automatically if we're able to get it from the index)
2. Update the hook definition with a previous version, trigger it
3. Retrigger http service deployment (we'll need a custom retriggering script, as we need to keep the dependencies on the old training tasks)

marco-c:
3 can be easily tested with rollbacking to the same version (we just need to make sure the deployment happens).

PR #649, opened 2019-06-28
Start tracking the start of the data pipeline. #648

Lothiraldan:
This way we can automate the stop of a running data pipeline during the
rollback process. The index route doesn't contains the version which would
require to update the `set_hook_version.py` script to add it.

### DIFF
--- a/infra/taskcluster-hook-pipeline-start.json
+++ b/infra/taskcluster-hook-pipeline-start.json
@@ -51,7 +51,8 @@
         "retries": 5,
         "routes": [
             "notify.email.release-mgmt-analysis@mozilla.com.on-failed",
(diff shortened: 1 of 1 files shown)


marco-c:
I wonder if this will work in all cases.

It definitely won't work when we have two pipelines running in parallel, but that's OK as we shouldn't have that in the normal case.

It also might not work when the pipeline has just started, as IIRC the tasks are indexed when they finish. Maybe we should use the Taskcluster API to index the task as soon as it starts (in spawn-pipeline.py).

Lothiraldan:
Indeed, the task is only indexed when is finished but it seems to takes only few seconds, I think we can wait a bit before checking if there is a pipeline running so we would support this case.

Issue #638, opened 2019-06-27
Try using bigrams in the NN similarity


marco-c:
In some cases, the presence of two words (or even more) words together can make a difference. E.g. in https://bugzilla.mozilla.org/show_bug.cgi?id=1320150 and https://bugzilla.mozilla.org/show_bug.cgi?id=1331130.

PR #654, opened 2019-06-28
Add bigram to vectorizer

ayush-1506:
Fixes : #638 

Also fixes the re-occurrence of bug in it's get_similar_bugs list.

### DIFF
--- a/bugbug/similarity.py
+++ b/bugbug/similarity.py
@@ -154,8 +154,8 @@ class LSISimilarity(BaseSimilarity):
 
 
 class NeighborsSimilarity(BaseSimilarity):
-    def __init__(self, k=10, vectorizer=TfidfVectorizer):
-        self.vectorizer = TfidfVectorizer()
+    def __init__(self, k=10, vectorizer=TfidfVectorizer()):
+        self.vectorizer = vectorizer
         self.similarity_calculator = NearestNeighbors(n_neighbors=k)
         text = []
         self.bug_ids = []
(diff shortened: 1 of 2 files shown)


marco-c:
> Also fixes the re-occurrence of bug in it's get_similar_bugs list.

It's preferable to do logic changes in separate commits, but it's fine this time

bugbug/similarity.py
```diff
@@ -155,7 +155,7 @@ def get_similar_bugs(self, query, k=10):
 
 class NeighborsSimilarity(BaseSimilarity):
     def __init__(self, k=10, vectorizer=TfidfVectorizer):
-        self.vectorizer = TfidfVectorizer()
+        self.vectorizer = vectorizer(ngram_range=(1, 2))
```
marco-c: I would make this an option. The caller can use whatever vectorizer they want, with a default of TfidfVectorizer.
The vectorizer can be a parameter of the __init__ function.
Either:
```
    def __init__(self, k=10, vectorizer_class=TfidfVectorizer):
        self.vectorizer = vectorizer_class()
```
or:
```
    def __init__(self, k=10, vectorizer=TfidfVectorizer()):
        self.vectorizer = vectorizer
```

bugbug/similarity.py
```diff
@@ -172,4 +172,4 @@ def get_similar_bugs(self, query):
         processed_query = self.vectorizer.transform([get_text(query)])
         _, indices = self.similarity_calculator.kneighbors(processed_query)
 
-        return [self.bug_ids[ind] for ind in indices[0]]
+        return [self.bug_ids[ind] for ind in indices[0]].remove(query["id"])
```
marco-c: Nit: this is nicer and has the same time complexity:
```
return [self.bug_ids[ind] for ind in indices[0] if self.bug_ids[ind] != query["id"]]
```

bugbug/similarity.py
ayush-1506: Isn't the first option you defined above same as what I'm doing right now?

marco-c: You are forcing the usage of ngram_range=(1, 2)

marco-c:
Can you add a comment with the results of this, compared to 1-gram?

PR #655, opened 2019-06-28
Avoid putting the version in the env. Fix #570

### LINKED ISSUE #570: Don't put VERSION in the env of the tasks
We can directly render the tag where needed with JSON-e.

Lothiraldan:
Removing the VERSION env from the docker_build task is a bit more tedious as
it requires an if condition based on a top-level variable.


marco-c:
> Removing the VERSION env from the docker_build task is a bit more tedious as it requires an if condition based on a top-level variable.

We could actually define this variable in a second let at the beginning

PR #656, opened 2019-06-28
Use the base image for training models. Fix #350

### LINKED ISSUE #350: Remove unneeded train images
They are mostly the same as the bugbug base image, plus a CMD.
We could use the bugbug base image and add the CMD in the task payload.

This will make things more flexible, as we won't need to create new repositories on docker hub for each train image. Also, caching at runtime will work better (all tasks will use the same image), and builds and pushes should get faster.

Lothiraldan:


marco-c:
@Lothiraldan we might actually still want a separate train image, so we can have a lean base image and a train image with the nn requirements (I don't know why I remembered we were already not using nn requirements in the base)

Lothiraldan:
@marco-c Do we need the nn requirements for loading models in the http service?

marco-c:
> @marco-c Do we need the nn requirements for loading models in the http service?

I imagine we do, not 100% sure though.

marco-c:
Oh, now I remember, we moved the nn requirements only, not all the training-related stuff. That's fine.

PR #658, opened 2019-06-29
Switch to alpine image

### LINKED ISSUE #292: Try switching to Python alpine for the base image
This might speed up the docker build task.

Abhi-khandelwal:
Fix #292

### DIFF
--- a/infra/dockerfile.base
+++ b/infra/dockerfile.base
@@ -1,4 +1,4 @@
-FROM python:3.7.3-slim
+FROM python:3.7.3-alpine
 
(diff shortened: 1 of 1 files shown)


Lothiraldan:
Switching to alpine based image will need to update the package manager related commands (using apk instead of apt-get).

You can try building the image locally by launching `docker-compose build` at the root directory of your bugbug clone to see if it's building.

marco-c:
Closing since the PR is not really doing anything. Feel free to reopen if you manage to actually do the switch.

PR #662, opened 2019-07-01
feature_importance_visualization for multiclass problems

chidauri:
Regarding #162

### DIFF
--- a/bugbug/model.py
+++ b/bugbug/model.py
@@ -206,7 +206,7 @@ class Model:
                 for importance, index, is_positive in important_features["average"]
             ]
 
-            important_features["classes"][f"class {class_names[num]}"] = (
+            important_features["classes"][class_names[num]] = (
                 top_item_features,
                 top_avg,
             )
@@ -214,18 +214,18 @@ class Model:
         return important_features
 
     def print_feature_importances(
-        self, important_features, feature_names, predicted_class=False
+        self, important_features, feature_names, classes=False
     ):
         # extract importance values from the top features for the predicted class
         # when classsifying
-        if predicted_class is not False:
+        if classes is not False:
             # shap_values are stored in class 0 for binary classification
-            if len(predicted_class[0]) != 2:
-                predicted_class = predicted_class.argmax(axis=-1)[0]
+            if len(classes[0]) != 2:
+                predicted_class = classes.argmax(axis=-1)[0]
             else:
                 predicted_class = 0
 
-            imp_values = important_features["classes"][f"class {predicted_class}"][0]
+            imp_values = important_features["classes"][predicted_class][0]
             shap_val = []
             top_feature_names = []
             for importance, index, is_positive in imp_values:
@@ -235,7 +235,7 @@ class Model:
                     shap_val.append("-" + str(importance))
 
                 top_feature_names.append(feature_names[int(index)])
-            shap_val = [[f"class {predicted_class}"] + shap_val]
+            shap_val = [[predicted_class] + shap_val]
 
         # extract importance values from the top features for all the classes
         # when training
@@ -444,7 +444,7 @@ class Model:
             top_indexes = [
                 int(index)
                 for importance, index, is_positive in important_features["classes"][
-                    f"class {pred_class_index}"
+                    pred_class_index
                 ][0]
             ]
 
--- a/bugbug/model.py
+++ b/bugbug/model.py
@@ -156,10 +156,12 @@ class Model:
 
         return cleaned_feature_names
 
-    def get_important_features(self, cutoff, shap_values, class_names=False):
+    def get_important_features(self, cutoff, shap_values, class_names):
 
         if not isinstance(shap_values, list):
             shap_values = [shap_values]
+
+        class_names = sorted(list(class_names))
 
         # returns top features for a shap_value matrix
         def get_top_features(cutoff, shap_values):
@@ -204,15 +206,10 @@ class Model:
                 for importance, index, is_positive in important_features["average"]
             ]
 
-            if class_names:
-                important_features["classes"][
-                    class_names[len(class_names) - num - 1]
-                ] = (top_item_features, top_avg)
-            else:
-                important_features["classes"][f"class {num}"] = (
-                    top_item_features,
-                    top_avg,
-                )
+            important_features["classes"][f"class {class_names[num]}"] = (
+                top_item_features,
+                top_avg,
+            )
 
         return important_features
 
@@ -222,6 +219,12 @@ class Model:
         # extract importance values from the top features for the predicted class
         # when classsifying
         if predicted_class is not False:
+            # shap_values are stored in class 0 for binary classification
+            if len(predicted_class[0]) != 2:
+                predicted_class = predicted_class.argmax(axis=-1)[0]
+            else:
+                predicted_class = 0
+
             imp_values = important_features["classes"][f"class {predicted_class}"][0]
             shap_val = []
             top_feature_names = []
@@ -246,7 +249,7 @@ class Model:
                 for class_name, imp_values in important_features["classes"].items()
             ]
 
-        # allow maximum of 8 columns in a row to fit the page better
+        # allow maximum of 6 columns in a row to fit the page better
         print("Top {} features:".format(len(top_feature_names)))
         for i in range(0, len(top_feature_names), 6):
             table = []
@@ -427,7 +430,7 @@ class Model:
             shap_values = explainer.shap_values(X)
 
             important_features = self.get_important_features(
-                importance_cutoff, shap_values
+                importance_cutoff, shap_values, class_names=range(len(classes[0]))
             )
 
             # Workaround: handle multi class case for force_plot to work correctly
--- a/bugbug/model.py
+++ b/bugbug/model.py
@@ -157,72 +157,72 @@ class Model:
         return cleaned_feature_names
 
     def get_important_features(self, cutoff, shap_values, class_names=False):
-        # Calculate the values that represent the fraction of the model output variability attributable
-        # to each feature across the whole dataset.
-        avg_shap_values = 0
-        if isinstance(shap_values, list):
-            avg_shap_values = np.sum(np.abs(shap_values), axis=0)
-        else:
-            avg_shap_values = shap_values
 
-        shap_sums = avg_shap_values.sum(0)
-        abs_shap_sums = np.abs(avg_shap_values).sum(0)
-        rel_shap_sums = abs_shap_sums / abs_shap_sums.sum()
+        if not isinstance(shap_values, list):
+            shap_values = [shap_values]
 
-        cut_off_value = cutoff * np.amax(rel_shap_sums)
+        # returns top features for a shap_value matrix
+        def get_top_features(cutoff, shap_values):
+            # Calculate the values that represent the fraction of the model output variability attributable
+            # to each feature across the whole dataset.
+            shap_sums = shap_values.sum(0)
+            abs_shap_sums = np.abs(shap_values).sum(0)
+            rel_shap_sums = abs_shap_sums / abs_shap_sums.sum()
 
-        # Get indices of features that pass the cut off value
-        top_feature_indices = np.where(rel_shap_sums >= cut_off_value)[0]
-        # Get the importance values of the top features from their indices
-        top_features = np.take(rel_shap_sums, top_feature_indices)
-        # Gets the sign of the importance from shap_sums as boolean
-        is_positive = (np.take(shap_sums, top_feature_indices)) >= 0
-        # Stack the importance, indices and shap_sums in a 2D array
-        top_features = np.column_stack((top_features, top_feature_indices, is_positive))
-        # Sort the array (in decreasing order of importance values)
-        top_features = top_features[top_features[:, 0].argsort()][::-1]
+            cut_off_value = cutoff * np.amax(rel_shap_sums)
 
-        if isinstance(shap_values, list):
-            important_features = {}
-            important_features["classes"] = {}
-            important_features["average"] = top_features
-            for num, item in enumerate(shap_values):
-                # top features for that class
-                top_item_features = self.get_important_features(
-                    cutoff, item, class_names
-                )
+            # Get indices of features that pass the cut off value
+            top_feature_indices = np.where(rel_shap_sums >= cut_off_value)[0]
+            # Get the importance values of the top features from their indices
+            top_features = np.take(rel_shap_sums, top_feature_indices)
+            # Gets the sign of the importance from shap_sums as boolean
+            is_positive = (np.take(shap_sums, top_feature_indices)) >= 0
+            # Stack the importance, indices and shap_sums in a 2D array
+            top_features = np.column_stack(
+                (top_features, top_feature_indices, is_positive)
+            )
+            # Sort the array (in decreasing order of importance values)
+            top_features = top_features[top_features[:, 0].argsort()][::-1]
 
-                # shap values of top average features for that class
-                abs_sums = np.abs(item).sum(0)
-                rel_sums = abs_sums / abs_sums.sum()
-                is_pos = ["+" if shap_sum >= 0 else "-" for shap_sum in item.sum(0)]
-                top_avg = [
-                    is_pos[int(index)] + str(rel_sums[int(index)])
-                    for importance, index, is_positive in important_features["average"]
-                ]
-
-                if class_names:
-                    important_features["classes"][
-                        class_names[len(class_names) - num - 1]
-                    ] = (top_item_features, top_avg)
-                else:
-                    important_features["classes"][f"class {num}"] = (
-                        top_item_features,
-                        top_avg,
-                    )
-
-            return important_features
-
-        else:
             return top_features
 
+        important_features = {}
+        important_features["classes"] = {}
+        important_features["average"] = get_top_features(
+            cutoff, np.sum(np.abs(shap_values), axis=0)
+        )
+        for num, item in enumerate(shap_values):
+            # top features for that class
+            top_item_features = get_top_features(cutoff, item)
+
+            # shap values of top average features for that class
+            abs_sums = np.abs(item).sum(0)
+            rel_sums = abs_sums / abs_sums.sum()
+            is_pos = ["+" if shap_sum >= 0 else "-" for shap_sum in item.sum(0)]
+            top_avg = [
+                is_pos[int(index)] + str(rel_sums[int(index)])
+                for importance, index, is_positive in important_features["average"]
+            ]
+
+            if class_names:
+                important_features["classes"][
+                    class_names[len(class_names) - num - 1]
+                ] = (top_item_features, top_avg)
+            else:
+                important_features["classes"][f"class {num}"] = (
+                    top_item_features,
+                    top_avg,
+                )
+
+        return important_features
+
     def print_feature_importances(
-        self, important_features, feature_names, pred_class=False
+        self, important_features, feature_names, predicted_class=False
     ):
         # extract importance values from the top features for the predicted class
         # when classsifying
-        if pred_class is not False:
-            imp_values = important_features["classes"][f"class {pred_class}"][0]
+        if predicted_class is not False:
+            imp_values = important_features["classes"][f"class {predicted_class}"][0]
             shap_val = []
             top_feature_names = []
             for importance, index, is_positive in imp_values:
@@ -232,7 +232,7 @@ class Model:
                     shap_val.append("-" + str(importance))
 
                 top_feature_names.append(feature_names[int(index)])
-            shap_val = [[f"class {pred_class}"] + shap_val]
+            shap_val = [[f"class {predicted_class}"] + shap_val]
 
         # extract importance values from the top features for all the classes
         # when training
@@ -333,9 +333,6 @@ class Model:
 
             matplotlib.pyplot.savefig("feature_importance.png", bbox_inches="tight")
 
-            if not isinstance(shap_values, list):
-                shap_values = [shap_values]
-
             important_features = self.get_important_features(
                 importance_cutoff, shap_values, self.class_names
             )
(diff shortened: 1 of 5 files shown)


marco-c:
Can you show an example output?

chidauri:
> Can you show an example output?

Selection_002

bugbug/model.py
```diff
@@ -193,18 +193,40 @@ def train(self, importance_cutoff=0.15):
             shap_values = explainer.shap_values(X_train)
 
             # TODO: Actually implement feature importance visualization for multiclass problems.
```
marco-c: You can now remove this TODO, since it is what you are doing now

bugbug/model.py
```diff
@@ -193,18 +193,40 @@ def train(self, importance_cutoff=0.15):
...
+            new_shap_values = 0
```
marco-c: I wouldn't call them `new`, `average` would be better

bugbug/model.py
```diff
@@ -193,18 +193,40 @@ def train(self, importance_cutoff=0.15):
...
+                    print(
+                        f'{i + 1}. \'{feature_names[int(index)]}\' ({"+" if (is_positive) else "-"}{importance})'
+                    )
+            else:
+                print("{:<}".format("classes"), end="   ")
+                for importance, index, is_pos in important_features:
+                    print("{:>}".format(feature_names[int(index)]), end="  ")
+                print("\n")
+                for class_num, item in enumerate(shap_values):
+                    print("class {}".format(class_num), end="   ")
```
marco-c: We should print the class name rather than the class number

chidauri: I've assumed the classes be in ascending order of their names, i don't know if it is correct :p

marco-c:
You'll need to rebase on top of master, as there have been some changes around the same lines.
Can you post another example output with the recent changes?

marco-c:
Oh, and you'll have to implement the same in the `classify` function.

chidauri:
> Oh, and you'll have to implement the same in the `classify` function.

The print functionality is in run.py, so for printing multi-class examples we'll have to return class-names as well as shap_values, so can we move the printing in model.py in the classify function itself?

marco-c:
In the classify case, we are only interested in the most important features for the predicted class (the user wants to know "why did we select this class?")

bugbug/model.py
```diff
@@ -192,19 +192,45 @@ def train(self, importance_cutoff=0.15):
...
-                importance_cutoff, shap_values
+                importance_cutoff, avg_shap_values
             )
 
-            print(f"\nTop {len(important_features)} Features:")
-            for i, [importance, index, is_positive] in enumerate(important_features):
-                print(
-                    f'{i + 1}. \'{feature_names[int(index)]}\' ({"+" if (is_positive) else "-"}{importance})'
-                )
+            if not isinstance(shap_values, list):
```
marco-c: I'd suggest refactoring this code to use tabulate, this would make it much cleaner.

chidauri: you mean, same as multi class make it  horizontal instead of vertical ?

marco-c: I meant using `tabulate`, which is a Python package to pretty-print tabular data such as this (both in the binary and multi-class cases).

chidauri:
@marco-c 
Selection_003
after using tabulate(in the default 'simple' table format)

bugbug/model.py
chidauri: @marco-c this should be ascending or descending? cause you have used descending order in summary plot

chidauri:
@marco-c  After recent changes.

Selection_005
This is how binary classification looks.
Selection_006
Multi class classification.


PS. We don't need "is_pos" parameter from get_important_features method anymore.

bugbug/model.py
```diff
@@ -203,46 +204,40 @@ def train(self, importance_cutoff=0.15):
             )
 
             matplotlib.pyplot.savefig("feature_importance.png", bbox_inches="tight")
-            
+
```
chidauri: i will fix this :p

marco-c:
The binary classification table doesn't look great, maybe there is a way for tabulate to handle better the large number of columns?

bugbug/model.py
marco-c: Not sure, we need to confirm with a real case and see if the results make sense in ascending or in descending order.

bugbug/model.py
```diff
@@ -204,20 +205,40 @@ def train(self, importance_cutoff=0.15):
...
             print(f"\nTop {len(important_features)} Features:")
-            for i, [importance, index, is_positive] in enumerate(important_features):
-                print(
-                    f'{i + 1}. \'{feature_names[int(index)]}\' ({"+" if (is_positive) else "-"}{importance})'
+            top_indexes = list(
+                int(index) for importance, index, is_pos in important_features
+            )
+
+            table = []
+            for num, item in enumerate(shap_values):
```
marco-c: Part of this should be moved to get_important_features.
get_important_features could return an object like:
```
{
    "average": most important features in average shap values,
    "classes": {
        "class1": (most important features in class1 shap values, values in class1 of the most important features in average shap values),
        "class2": (most important features in class1 shap values, values in class1 of the most important features in average shap values),
    }
}
```

This way here you can show the most important feaures in average shap values and their values for each class, and in classify you can show the most important features for the predicted class.

chidauri: so we want the list of most important features for each class, distinctly, to use in classify but we do not need that here, right?

marco-c: Exactly.
At training time, we will show the most important average features and their values for all classes.
At classify time, we will show the most important features for the predicted class.

get_important_features will return everything though, so we can easily do more complex stuff in the future if we want to.

chidauri:
> The binary classification table doesn't look great, maybe there is a way for tabulate to handle better the large number of columns?

we can use transpose(features  as rows and classes as columns) for both cases as features are always going to be greater than number of classes so we won't have to worry if features increase in multi class case too.

marco-c:
> we can use transpose(features as rows and classes as columns) for both cases as features are always going to be greater than number of classes so we won't have to worry if features increase in multi class case too.

Unfortunately in the "component" model we have a lot of classes :(

chidauri:
@marco-c
Binary classification. 
Selection_007
multi-class classify function:
Selection_008
mutli-class classification:
Selection_006

bugbug/model.py
```diff
@@ -113,11 +114,14 @@ def le(self):
     def get_feature_names(self):
         return []
 
-    def get_important_features(self, cutoff, shap_values):
+    def get_important_features(
+        self, cutoff, avg_shap_values, shap_values, class_names=False
```
marco-c: Move the avg calculation in this function, so you can only pass `shap_values` to it

bugbug/model.py
```diff
@@ -133,7 +137,40 @@ def get_important_features(self, cutoff, shap_values):
...
+                rel_sums = abs_sums / abs_sums.sum()
+                is_pos = ["+" if shap_sum >= 0 else "-" for shap_sum in item.sum(0)]
+                for index in important_features["average"]:
+                    top_avg.append(is_pos[index] + str(rel_sums[index]))
+
+                if class_names:
+                    important_features["classes"][
+                        class_names[len(class_names) - num - 1]
+                    ] = (top_item_features, top_avg)
+                else:
```
marco-c: In which case do you not have class names?

bugbug/model.py
```diff
@@ -133,7 +137,40 @@ def get_important_features(self, cutoff, shap_values):
...
+                top_avg = []
+                abs_sums = np.abs(item).sum(0)
+                rel_sums = abs_sums / abs_sums.sum()
+                is_pos = ["+" if shap_sum >= 0 else "-" for shap_sum in item.sum(0)]
+                for index in important_features["average"]:
+                    top_avg.append(is_pos[index] + str(rel_sums[index]))
+
+                if class_names:
+                    important_features["classes"][
+                        class_names[len(class_names) - num - 1]
```
marco-c: Are you sure they are in reverse order?

bugbug/model.py
```diff
@@ -133,7 +137,40 @@ def get_important_features(self, cutoff, shap_values):
...
+                    cutoff, item, item, class_names
+                )
+
+                # top features in average shap values for that class
+                top_avg = []
+                abs_sums = np.abs(item).sum(0)
+                rel_sums = abs_sums / abs_sums.sum()
+                is_pos = ["+" if shap_sum >= 0 else "-" for shap_sum in item.sum(0)]
+                for index in important_features["average"]:
+                    top_avg.append(is_pos[index] + str(rel_sums[index]))
```
marco-c: You can use a list comprehension instead of iterating and appending

bugbug/model.py
```diff
@@ -133,7 +137,40 @@ def get_important_features(self, cutoff, shap_values):
         # Sort the array (in decreasing order of importance values)
         top_features = top_features[top_features[:, 0].argsort()][::-1]
 
-        return top_features
+        if isinstance(shap_values, list):
+            important_features = {}
+            important_features["classes"] = {}
+            important_features["average"] = list(
```
marco-c: We should store `top_features` in here

bugbug/model.py
```diff
@@ -204,18 +241,34 @@ def train(self, importance_cutoff=0.15):
...
             )
 
-            print(f"\nTop {len(important_features)} Features:")
-            for i, [importance, index, is_positive] in enumerate(important_features):
+            print("Top {} features:".format(len(important_features["average"])))
+            top_feature_names = [
+                feature_names[index] for index in important_features["average"]
+            ]
+
+            # allow maximum of 8 columns in a row to fit the page better
```
marco-c: We should share as much as possible this code between model.py and run.py. You can add a new function "print_feature_importances".

bugbug/model.py
chidauri: Actually not in reverse order but currently they are already reversed in model.py so i reversed them again.

bugbug/model.py
chidauri: in classify case ??

marco-c: We should define class names as part of the models, so you can use it here

chidauri:
@marco-c  after the recent commit , printing is working fine for training , but will have to fix it for classify.
These are the results for train:
defect train:
defect_train
defect_enhancement_task train:
defectenhancetask_train

bugbug/model.py
```diff
@@ -204,19 +288,14 @@ def train(self, importance_cutoff=0.15):
 
             matplotlib.pyplot.savefig("feature_importance.png", bbox_inches="tight")
 
-            # TODO: Actually implement feature importance visualization for multiclass problems.
-            if isinstance(shap_values, list):
-                shap_values = np.sum(np.abs(shap_values), axis=0)
+            if not isinstance(shap_values, list):
```
marco-c: Move this to get_important_features

bugbug/model.py
```diff
@@ -300,16 +379,19 @@ def classify(
             explainer = shap.TreeExplainer(self.clf)
             shap_values = explainer.shap_values(X)
 
-            # TODO: Actually implement feature importance visualization for multiclass problems.
-            if isinstance(shap_values, list):
-                shap_values = np.sum(np.abs(shap_values), axis=0)
+            if not isinstance(shap_values, list):
```
marco-c: Here too

run.py
```diff
@@ -103,12 +103,14 @@ def main(args):
...
                 feature_names = model.get_feature_names()
-                for i, (importance, index, is_positive) in enumerate(
-                    importance["importances"]
-                ):
-                    print(
-                        f'{i + 1}. \'{feature_names[int(index)]}\' ({"+" if (is_positive) else "-"}{importance})'
-                    )
+
+                # shap values of top features for the predicted class
+                pred_class_index = int(np.where(probas == np.amax(probas))[1])
```
marco-c: The index should just be probas.argmax(axis=-1)[0]

bugbug/model.py
chidauri: if we move this to the function itself , we will have to write separate code for single class and multiclass, i mean how will the recursion work?

run.py
```diff
@@ -103,12 +103,14 @@ def main(args):
...
+
+                model.print_feature_importances(
+                    importance["importances"], feature_names, pred_class_index
```
chidauri: @marco-c i need help here , when i call this (even with pred_class = pred_class_index), pred_class is being treated as False(the default values) :(

chidauri: i got it.

bugbug/model.py
```diff
@@ -386,11 +386,17 @@ def classify(
                 importance_cutoff, shap_values
             )
 
-            # TODO: pred class index
+            # Workaround: handle multi class case for force_plot to work correctly
+            if not len(classes[0]) == 2:
+                pred_class_index = classes.argmax(axis=-1)[0]
+                explainer.expected_value = explainer.expected_value[0]
+            else:
+                pred_class_index = 0
```
chidauri: @marco-c for binary case , we have values stored in  a single class i.e., ["class 0"], so we have to make the index 0, and not the one with maximum probas.

bugbug/model.py
```diff
@@ -300,24 +379,33 @@ def classify(
...
+                ][0]
             ]
 
             feature_names = self.get_feature_names()
 
             with io.StringIO() as out:
                 p = shap.force_plot(
                     explainer.expected_value,
-                    shap_values[:, top_indexes],
+                    shap_values[0][:, top_indexes],
```
marco-c: Shouldn't this be shap_values[pred_class_index]?

bugbug/model.py
```diff
@@ -329,7 +417,14 @@ def classify(
 
                 html = out.getvalue()
 
-            return classes, {"importances": top_importances, "html": html}
+            return (
+                classes,
+                {
+                    "importances": important_features,
+                    "html": html,
+                    "predicted_class": pred_class_index,
```
marco-c: This doesn't seem useful, the caller should be able to get the pred_class_index from `classes`

bugbug/model.py
```diff
@@ -300,24 +379,33 @@ def classify(
...
+            if not isinstance(shap_values, list):
+                shap_values = [shap_values]
 
-            top_importances = self.get_important_features(
+            important_features = self.get_important_features(
                 importance_cutoff, shap_values
             )
 
+            # Workaround: handle multi class case for force_plot to work correctly
+            if not len(classes[0]) == 2:
```
marco-c: if len(classes[0]) != 2

bugbug/model.py
```diff
@@ -21,6 +21,7 @@
 from sklearn.feature_extraction.text import TfidfVectorizer
 from sklearn.metrics.classification import precision_recall_fscore_support
 from sklearn.model_selection import cross_validate, train_test_split
+from tabulate import tabulate
```
marco-c: You'll need to add this to the requirements.txt file (see the failing tests)

bugbug/model.py
```diff
@@ -133,7 +140,84 @@ def get_important_features(self, cutoff, shap_values):
...
+                        top_avg,
+                    )
+
+            return important_features
+
+        else:
+            return top_features
+
+    def print_feature_importances(
+        self, important_features, feature_names, pred_class=False
```
marco-c: Nit: rename `pred_class` to `predicted_class`

bugbug/model.py
marco-c: You can split the function in two, one which calculates and returns top_features for a shap_values matrix, and one which is called with a shap_values list and calls the first one.

bugbug/model.py
chidauri: Traceback (most recent call last):
  File "run.py", line 165, in <module>
    main(parse_args(sys.argv[1:]))
  File "run.py", line 102, in main
    bug, probabilities=True, importances=True
  File "/home/harshit/Documents/bugbug/bugbug/model.py", line 398, in classify
    shap_values[pred_class_index][:, top_indexes],
IndexError: too many indices for array

bugbug/model.py
chidauri: added this, but still failing tests :(

run.py
```diff
@@ -103,12 +103,19 @@ def main(args):
...
 
                 feature_names = model.get_human_readable_feature_names()
-                for i, (importance, index, is_positive) in enumerate(
-                    importance["importances"]
-                ):
-                    print(
-                        f'{i + 1}. \'{feature_names[int(index)]}\' ({"+" if (is_positive) else "-"}{importance})'
-                    )
+
+                # shap_values are stored in class 0 for binary classification
```
marco-c: I don't like this part, users of the `print_feature_importance` function should not have to know about this. We should move it to print_feature_importances (making `probas` the parameter of print_feature_importances instead of `predicted_class`)

bugbug/model.py
```diff
@@ -140,27 +141,110 @@ def get_human_readable_feature_names(self):
...
+            abs_sums = np.abs(item).sum(0)
+            rel_sums = abs_sums / abs_sums.sum()
+            is_pos = ["+" if shap_sum >= 0 else "-" for shap_sum in item.sum(0)]
+            top_avg = [
+                is_pos[int(index)] + str(rel_sums[int(index)])
+                for importance, index, is_positive in important_features["average"]
+            ]
 
-        return top_features
+            if class_names:
```
marco-c: We should always use class_names

run.py
```diff
@@ -103,12 +103,19 @@ def main(args):
...
+                    pred_class_index = probas.argmax(axis=-1)[0]
+                else:
+                    pred_class_index = 0
+
+                model.print_feature_importances(
+                    importance["importances"],
+                    feature_names,
+                    predicted_class=pred_class_index,
+                )
+
```
marco-c: Nit: don't add newline here

run.py
```diff
@@ -112,12 +112,10 @@ def main(args):
...
                 feature_names = model.get_human_readable_feature_names()
-                for i, (importance, index, is_positive) in enumerate(
-                    importance["importances"]
-                ):
-                    print(
-                        f'{i + 1}. \'{feature_names[int(index)]}\' ({"+" if (is_positive) else "-"}{importance})'
-                    )
+
+                model.print_feature_importances(
+                    importance["importances"], feature_names, predicted_class=probas
```
marco-c: The name of the parameter should change too, since it's no longer `predicted_class`

bugbug/model.py
```diff
@@ -156,27 +156,113 @@ def get_human_readable_feature_names(self):
...
+            # shap values of top average features for that class
+            abs_sums = np.abs(item).sum(0)
+            rel_sums = abs_sums / abs_sums.sum()
+            is_pos = ["+" if shap_sum >= 0 else "-" for shap_sum in item.sum(0)]
+            top_avg = [
+                is_pos[int(index)] + str(rel_sums[int(index)])
+                for importance, index, is_positive in important_features["average"]
+            ]
+
+            important_features["classes"][f"class {class_names[num]}"] = (
```
marco-c: Let's directly use the class_name as a key here, instead of "class CLASS_NAME"

bugbug/model.py
```diff
@@ -349,16 +427,23 @@ def classify(
...
             shap_values = explainer.shap_values(X)
 
-            # TODO: Actually implement feature importance visualization for multiclass problems.
-            if isinstance(shap_values, list):
-                shap_values = np.sum(np.abs(shap_values), axis=0)
-
-            top_importances = self.get_important_features(
-                importance_cutoff, shap_values
+            important_features = self.get_important_features(
+                importance_cutoff, shap_values, class_names=range(len(classes[0]))
```
marco-c: I've filed #732 to use the real class names here.

bugbug/model.py
```diff
@@ -156,31 +165,116 @@ def get_human_readable_feature_names(self):
...
-        # Get the importance values of the top features from their indices
-        top_features = np.take(rel_shap_sums, top_feature_indices)
-        # Gets the sign of the importance from shap_sums as boolean
-        is_positive = (np.take(shap_sums, top_feature_indices)) >= 0
-        # Stack the importance, indices and shap_sums in a 2D array
-        top_features = np.column_stack((top_features, top_feature_indices, is_positive))
-        # Sort the array (in decreasing order of importance values)
-        top_features = top_features[top_features[:, 0].argsort()][::-1]
+    def print_feature_importances(
+        self, important_features, feature_names, classify=False
```
marco-c: `classify` is not a great name, and it's not great that it can be either a boolean or a list, it should be consistent.
You could call it `class_probabilities` and make it None in the non-classify case and a list in the classify case.

.isort.cfg
```diff
@@ -5,3 +5,4 @@ force_grid_wrap=0
 use_parentheses=True
 line_length=88
 known_third_party = apispec,apispec_webframeworks,dateutil,flask,flask_cors,hglib,imblearn,jsone,jsonschema,libmozdata,marshmallow,matplotlib,microannotate,models,numpy,pandas,pkg_resources,pyemd,pytest,redis,requests,responses,rq,setuptools,shap,sklearn,tabulate,taskcluster,tqdm,xgboost,yaml,zstandard
+
```
marco-c: This change is unrelated to the PR, can you revert it?

bugbug/model.py
```diff
@@ -156,31 +165,116 @@ def get_human_readable_feature_names(self):
 
         return cleaned_feature_names
 
-    def get_important_features(self, cutoff, shap_values):
-        # Calculate the values that represent the fraction of the model output variability attributable
-        # to each feature across the whole dataset.
-        shap_sums = shap_values.sum(0)
-        abs_shap_sums = np.abs(shap_values).sum(0)
-        rel_shap_sums = abs_shap_sums / abs_shap_sums.sum()
+    def get_important_features(self, cutoff, shap_values, class_names):
```
marco-c: Since class_names are stored in self, class_names doesn't need to be a parameter of this function. You can just use self.class_names instead of class_names when needed

scripts/commit_classifier.py
```diff
@@ -109,9 +109,15 @@ def classify(self, diff_id):
 
         feature_names = self.model.get_human_readable_feature_names()
 
+        if len(probs[0]) > 2:
```
marco-c: We know that this is a binary case, so we don't need to check

scripts/commit_classifier.py
```diff
@@ -109,9 +109,15 @@ def classify(self, diff_id):
...
+            pred_class_index = probs.argmax(axis=-1)[0]
+        else:
+            pred_class_index = 0
+
+        pred_class = self.model.class_names[pred_class_index]
         features = []
         for i, (val, feature_index, is_positive) in enumerate(
-            importance["importances"]
+            importance["importances"]["classes"][pred_class][0]
```
marco-c: Here too, we know the class name as it's a binary classifier, I think you can just hardcode "1" instead of `pred_class`

scripts/commit_classifier.py
```diff
@@ -121,6 +127,10 @@ def classify(self, diff_id):
                 ]
             )
 
+        self.model.print_feature_importances(
```
marco-c: We don't need to print the features here

Issue #637, opened 2019-06-27
Try word2vec with WMD as a similarity algorithm


marco-c:
The position of the words can be important (e.g. https://bugzilla.mozilla.org/show_bug.cgi?id=1320150 and https://bugzilla.mozilla.org/show_bug.cgi?id=1331130 are different bugs, despite having similar words), so WMD can help improve the precision.

An implementation of word2vec+WMD (in another context) can be seen in https://github.com/marco-c/crashsimilarity/blob/master/crashsimilarity/models/word2vec.py and https://github.com/marco-c/crashsimilarity/blob/master/crashsimilarity/models/distances.py.

Issue #631, opened 2019-06-26
Investigate results of the similarity model


marco-c:
We need to check a few results where we are wrong.

Is there something in common to bugs we miss?

Is there something in common to bugs we found but shouldn't have found?

ayush-1506:
Some examples of missed bugs: (bugs that are actually duplicate but aren't classified as duplicate by the model)

1. [1320978](https://bugzilla.mozilla.org/show_bug.cgi?id=1320978) and [1320006](https://bugzilla.mozilla.org/show_bug.cgi?id=1320006) Distance calculated : 0.5698

2. [1320041](https://bugzilla.mozilla.org/show_bug.cgi?id=1320041) and [1320112](https://bugzilla.mozilla.org/show_bug.cgi?id=1320112) Distance calculated : 0.6324  (Although this pair is pretty hard to classify).

3. [1320112](https://bugzilla.mozilla.org/show_bug.cgi?id=1320112) and [1230147](https://bugzilla.mozilla.org/show_bug.cgi?id=1320147) Distance calculated : 0.36344

ayush-1506:
4. [1320332](https://bugzilla.mozilla.org/show_bug.cgi?id=1320332) and [1336375](https://bugzilla.mozilla.org/show_bug.cgi?id=1320332) Distance calculated : 0.544

ayush-1506:
Some other examples:

5.[1320500](https://bugzilla.mozilla.org/show_bug.cgi?id=1320500) and [1320485](https://bugzilla.mozilla.org/show_bug.cgi?id=1320485)

6. [1320401](https://bugzilla.mozilla.org/show_bug.cgi?id=1320401) and [1324301](https://bugzilla.mozilla.org/show_bug.cgi?id=1324301)

7. [1322357](https://bugzilla.mozilla.org/show_bug.cgi?id=1322357) and [1320048](https://bugzilla.mozilla.org/show_bug.cgi?id=1320048)

8. [1320148](https://bugzilla.mozilla.org/show_bug.cgi?id=1320148) and [1320786](https://bugzilla.mozilla.org/show_bug.cgi?id=1320786)

marco-c:
> 1320978 and 1320006 Distance calculated : 0.5698

I wouldn't have said they were duplicate :P

> 1320041 and 1320112 Distance calculated : 0.6324 (Although this pair is pretty hard to classify).

Indeed!

> 1320112 and 1230147 Distance calculated : 0.36344

Same here...

[...]

PR #666, opened 2019-07-02
Add word2vec with wmd distance similarity

ayush-1506:
Fixes : #637

To do : 

- [x] Make this faster

- [ ] get_similar_bugs returns the 5 closest bugs. Experiment with minimum distance cut off along with this.

### DIFF
--- a/bugbug/similarity.py
+++ b/bugbug/similarity.py
@@ -3,11 +3,13 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
+import bisect
+import random
 import re
 from collections import defaultdict
 
-from gensim.models import Word2Vec
-from gensim.similarities import WmdSimilarity
+import numpy as np
+from pyemd import emd
 from sklearn.feature_extraction.text import TfidfVectorizer
 from sklearn.neighbors import NearestNeighbors
 
@@ -19,7 +21,9 @@ OPT_MSG_MISSING = (
 
 try:
     import nltk
+    import gensim
     from gensim import models, similarities
+    from gensim.models import Word2Vec
     from gensim.corpora import Dictionary
     from nltk.corpus import stopwords
     from nltk.stem.porter import PorterStemmer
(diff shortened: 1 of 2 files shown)


marco-c:
Can you post some example results?

What part is slow? And how slow?

marco-c:
> Can you post some example results?

I mean the evaluation results.

ayush-1506:
Ok, so the first part is done (implemented r-wmd with cosine distance instead of wmd), So it's comparatively fast now.

marco-c:
Good! Let me know when you are done with an evaluation so we can compare the results with the others.

ayush-1506:
Looks like there's an edge case that I'm overlooking. This causes an error in a few specific bugs. I'll try to narrow it down.

ayush-1506:
@marco-c 
```
Recall: 24.818577648766325%
Precision: 3.189405949827474%
```

marco-c:
Are these recall and precision with the same number of bugs you used for the other evaluations? If not, can you run at least the nn evaluation with the same number of bugs so we can compare?

marco-c:
The next step will be #631, we need to understand why the results are like this and if/how we can improve them.

ayush-1506:
I tried nn similarity with the same number of bugs:
```
Recall: 30.327056491575817%
Precision: 3.912793299661147%
```

bugbug/similarity.py
```diff
@@ -6,6 +6,8 @@
 import re
 from collections import defaultdict
 
+import numpy as np
+from gensim.models import Word2Vec
```
marco-c: You'll need to move the import where the other gensim imports are

bugbug/similarity.py
```diff
@@ -88,6 +90,7 @@ def evaluation(self):
         hits_p = 0
 
         for bug in bugzilla.get_bugs():
+
```
marco-c: Please remove this newline as it's unrelated to your changes

bugbug/similarity.py
```diff
@@ -175,3 +178,46 @@ def get_similar_bugs(self, query):
...
+
+class Word2VecWmdSimilarity(BaseSimilarity):
+    def __init__(self):
+        self.corpus = []
+        self.bug_ids = []
+        for bug in bugzilla.get_bugs():
+            self.corpus.append(text_preprocess(get_text(bug)))
+            self.bug_ids.append(bug["id"])
+        self.w2vmodel = Word2Vec(self.corpus, size=100, min_count=1)
```
marco-c: Maybe better to use min_count=5?

marco-c: That is, the default

bugbug/similarity.py
```diff
@@ -175,3 +178,46 @@ def get_similar_bugs(self, query):
...
+
+    def get_similar_bugs(self, query):
+        self.w2vmodel.init_sims()
```
marco-c: Maybe we can call this in the constructor?

bugbug/similarity.py
marco-c: Maybe we should shuffle the corpus

bugbug/similarity.py
marco-c: And you can use replace=True to reduce the memory usage

bugbug/similarity.py
```diff
@@ -175,3 +178,46 @@ def get_similar_bugs(self, query):
...
+                self.w2vmodel.wv.vectors_norm,
+                self.w2vmodel.wv.vectors_norm[
+                    [self.w2vmodel.wv.vocab[word].index for word in words]
+                ].transpose(),
+            ),
+            dtype=np.double,
+        )
+        distances = []
+        for i in range(len(self.corpus)):
+            idx = [self.w2vmodel.wv.vocab[word].index for word in self.corpus[i]]
```
marco-c: Nit: it's plural, so maybe use `indexes`

bugbug/similarity.py
```diff
@@ -175,3 +178,46 @@ def get_similar_bugs(self, query):
...
+                word_dists = all_distances[idx]
+                rwmd = max(
+                    np.sum(np.min(word_dists, axis=0)),
+                    np.sum(np.min(word_dists, axis=1)),
+                )
+
+                distances.append((self.bug_ids[i], rwmd))
+
+        distances.sort(key=lambda v: v[1])
+
```
marco-c: You should now apply real WMD, like at https://github.com/marco-c/crashsimilarity/blob/master/crashsimilarity/models/base.py#L165.

bugbug/similarity.py
```diff
@@ -175,3 +180,111 @@ def get_similar_bugs(self, query):
...
+    def __init__(self):
+        self.corpus = []
+        self.bug_ids = []
+        for bug in bugzilla.get_bugs():
+            self.corpus.append(text_preprocess(get_text(bug)))
+            self.bug_ids.append(bug["id"])
+        self.w2vmodel = Word2Vec(self.corpus, size=100, min_count=5)
+        self.w2vmodel.init_sims(replace=True)
+
+    def wmdistance(self, document1, document2, all_distances, distance_metric="cosine"):
```
marco-c: Can you check the differences between this function and the equivalent present in gensim? I don't remember exactly why I rewrote it.

ayush-1506:
Results with the latest implementation:
```
Recall: 25.82417582417583%
Precision: 3.0921052631578947%
```

bugbug/similarity.py
ayush-1506: There is a subtle difference. Here we've computed the cosine distances before calling wmdistance and stored them in all_distances. (and then we use this inside the wmdistance method.). On the other hand the word2vec [wmdistance](https://tedboy.github.io/nlps/_modules/gensim/models/word2vec.html#Word2Vec.wmdistance) calculates them inside the method. 

The other major difference is that `word2vec.wmdistance` calculates only the euclidean distance. (There isn't an option to calculate cosine distances.), so we need to implement it separately over here.

marco-c: Can you add a comment explaining that the function partially comes from gensim and explaining the differences?

PR #680, opened 2019-07-08
Fix small bug in word2vec similarity

ayush-1506:
We also need to remove words from the corpus in the `get_similar_bugs` method which have a count < 5, and are neglected by the word2vec model.

I was playing around with `min_count = 1`, hence I never encountered this issue before.


ayush-1506:
Oh, and `get_similar_bugs(query)` also returns `query["id"]` in the result, so that should be removed.

PR #681, opened 2019-07-08
more automatic labelled examples for the stepstoreproduce model

### LINKED ISSUE #426: Add more automatic labelled examples for the stepstoreproduce model
We should add the following to the training set of stepstoreproduce:
1) Bugs that had 'stepswanted' keyword (rollbacked to the time before the keyword was removed)
2) Bugs that had 'Has STR' set to no (rollbacked to the time before 'Has STR' turned to yes)

This is not really easy to implement. The bugs that had 'Has STR=no' and now have 'Has STR=yes' are already in the training set as examples of bugs *with* steps to reproduce.
So, we'll need to consider some bugs twice, once rollbacked and once with the full history.

chidauri:
Fixes #426

### DIFF
--- a/bugbug/model.py
+++ b/bugbug/model.py
@@ -560,15 +560,6 @@ class BugModel(Model):
 
             assert len(commit_map) > 0
 
-        if "to_rollback" in classes:
-            for bug in bugzilla.get_bugs():
-                bug_id = bug["id"]
-
-                if bug_id not in classes["to_rollback"]:
-                    continue
-
-                yield bug, classes["to_rollback"][bug_id]
-
         for bug in bugzilla.get_bugs():
             bug_id = bug["id"]
             if bug_id not in classes:
@@ -581,6 +572,15 @@ class BugModel(Model):
                     bug["commits"] = []
 
             yield bug, classes[bug_id]
+
+        if "to_rollback" in classes:
+            for bug in bugzilla.get_bugs():
+                bug_id = bug["id"]
+                if bug_id not in classes["to_rollback"]:
+                    continue
+
+                rollback_bug, rollback_label = self.rollback_gen(bug, classes)
+                yield rollback_bug, rollback_label
 
 
 class CommitModel(Model):
--- a/bugbug/model.py
+++ b/bugbug/model.py
@@ -285,7 +285,7 @@ class Model:
         self.class_names = sort_class_names(self.class_names)
 
         # Get items and labels, filtering out those for which we have no labels.
-        X_iter, y_iter = split_tuple_iterator(self.items_gen(classes))
+        X_iter, y_iter = split_tuple_iterator(self.items_gen(classes, self.augment()))
 
         # Extract features from the items.
         X = self.extraction_pipeline.fit_transform([item for item in X_iter])
@@ -545,7 +545,7 @@ class BugModel(Model):
         Model.__init__(self, lemmatization)
         self.commit_data = commit_data
 
-    def items_gen(self, classes):
+    def items_gen(self, classes, bugs):
         if not self.commit_data:
             commit_map = None
         else:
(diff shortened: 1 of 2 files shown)


marco-c:
I think you might be rollbacking all bugs this way, are you sure you aren't?

chidauri:
> I think you might be rollbacking all bugs this way, are you sure you aren't?

https://github.com/mozilla/bugbug/blob/b456bd3b2d00aad453b42cc8761fc04fbf0e3866/bugbug/bug_features.py#L444-L449
This helped :)

marco-c:
I think relying on that is a little brittle:
1) there might be cases when you want to rollback the same bug multiple times in different ways;
2) you are going to rollback the bug the first time it appears, so for example for the stepswanted case you might rollback the bug with label 1 while you should have rollbacked the bug with label 0. Right now it doesn't happen as you are yielding the bugs to rollback first, but it's brittle to rely on this.

bugbug/model.py
```diff
@@ -367,6 +367,15 @@ def items_gen(self, classes):
 
             assert len(commit_map) > 0
 
+        if "to_rollback" in classes:
+            for bug in bugzilla.get_bugs():
+                bug_id = bug["id"]
+
+                if bug_id not in classes["to_rollback"]:
+                    continue
+
```
chidauri: @marco-c 
1.one solution could be to use bug_snapshot.rollback() here , and not use it in pipeline.This way it won't matter what bugs yield first,
2. We can introduce a new field to the bugs we want to rollback(cause we need something to differentiate them from the original bug) , and rollback them in bug_features.py, but we might want to remove "already_rollbacked" parameter this way.

bugbug/models/stepstoreproduce.py
```diff
@@ -92,6 +111,15 @@ def get_labels(self):
 
         return classes, [0, 1]
 
+    def rollback_gen(self, bug, classes):
+        for change in classes["to_rollback"][int(bug["id"])]:
+            self.rollback_change = change
+            rollbacked_bug = bug_snapshot.rollback(bug, self.rollback_when)
+            if change["field_name"] == "cf_has_str" and change["removed"] == "no":
+                return rollbacked_bug, 0
```
chidauri: @marco-c now we can rollback the same bug multiple times , there's this one thing to fix though , we need to find a way to "yield" instead of "return".

chidauri:
**Before changes**

> 73 bugs have no steps to reproduce
2642 bugs have steps to reproduce
X: (2715, 68672), y: (2715,)
Cross Validation scores:
Accuracy: f0.6930046232725715 (+/- 0.08506114027939689)
Precision: f0.9889862871434468 (+/- 0.011123332687863755)
Recall: f0.6927315347191507 (+/- 0.09585654167878219)
X_train: (128, 68672), y_train: (128,)
X_test: (272, 68672), y_test: (272,)

**After changes**

> 73 bugs have no steps to reproduce
2642 bugs have steps to reproduce
X: (2730, 68672), y: (2730,)
Cross Validation scores:
Accuracy: f0.7285596012782939 (+/- 0.08060902831036675)
Precision: f0.9897260455187465 (+/- 0.008173384646726092)
Recall: f0.7275707767383682 (+/- 0.08338297307686995)
X_train: (150, 68672), y_train: (150,)
X_test: (273, 68672), y_test: (273,)

bugbug/models/stepstoreproduce.py
chidauri: Also , i can move rollback_gen function to model.py if that's better.

marco-c (changes requested):
I've been thinking about this, and I feel it would be nicer to separate the "data augmentation" and the "get labels" steps.

You could implement this in this way:
1) Each model defines a `augment` function, which performs data augmentation:
```Python
def augment(self, bugs):
    for bug in bugs:
        # Do what it wants with the bug, yielding all generated bugs (from a given bug, it could generate 0, 1 or even more than 1 bug).
        yield generated_bugs
```

2) The base Model class calls `augment` *before* `get_labels`, so the input to get_labels will be the normal bugs + the augmented bugs.

PR #689, opened 2019-07-12
replace deprecated download_bugs_between with get_ids_between

cklyyung:
Looks like `bugzilla.download_bugs_between` was removed in commit 735fccc, and replaced with `bugzilla.get_ids_between`, but it was missed in this file.

I ran it and was able to generate the csv successfully (Got about ~1156 bugs).


kanurag94:
@marco-c  Defect/enhancement/task Output: 
image

Issue #523, opened 2019-05-30
Use some categorical features too for duplicate detection


marco-c:
Right now we are only using text, but some categorical features might be interesting too. For example, if two bugs have the same affected OS, the same product, and so on.

marco-c:
Some possible features:
- [ ] "same_product", which is True when the product of the two bugs is the same;
- [ ] "same_component", which is True when both the product and the component of the two bugs are the same;
- [ ] "same_platform";
- [ ] "same_os";
- [ ] "same_version";
- [ ] "same_target";
- [ ] "same_first_affected_version";
- [ ] "shared_keywords" list of keywords that are in both bugs;
- [ ] "shared_whiteboard" list of whiteboard strings that are in both bugs;
- [ ] something about the crash signature?
- ...

marco-c:
We can call this fixed, we can open new issues if we think about additional features to consider.

PR #692, opened 2019-07-12
Add support for feature extraction across pairs of bugs

ayush-1506:
First part of #523.

### DIFF
--- a/bugbug/bug_features.py
+++ b/bugbug/bug_features.py
@@ -21,28 +21,36 @@ def field(bug, field):
     return None
 
 
-class has_str(object):
+class single_bug_feature(object):
+    pass
+
+
+class couple_bug_feature(object):
+    pass
+
+
+class has_str(single_bug_feature):
     name = "Has STR"
 
     def __call__(self, bug, **kwargs):
         return field(bug, "cf_has_str")
 
 
-class has_regression_range(object):
+class has_regression_range(single_bug_feature):
     name = "Has Regression Range"
 
     def __call__(self, bug, **kwargs):
         return field(bug, "cf_has_regression_range")
 
 
-class has_crash_signature(object):
+class has_crash_signature(single_bug_feature):
     name = "Crash signature present"
 
     def __call__(self, bug, **kwargs):
         return "cf_crash_signature" in bug and bug["cf_crash_signature"] != ""
 
 
-class keywords(object):
+class keywords(single_bug_feature):
     def __init__(self, to_ignore=set()):
         self.to_ignore = to_ignore
 
@@ -62,19 +70,19 @@ class keywords(object):
         return keywords + subkeywords
 
 
-class severity(object):
+class severity(single_bug_feature):
     def __call__(self, bug, **kwargs):
         return field(bug, "severity")
 
 
-class number_of_bug_dependencies(object):
+class number_of_bug_dependencies(single_bug_feature):
     name = "# of bug dependencies"
 
     def __call__(self, bug, **kwargs):
         return len(bug["depends_on"])
 
 
-class is_coverity_issue(object):
+class is_coverity_issue(single_bug_feature):
     name = "Is Coverity issue"
 
     def __call__(self, bug, **kwargs):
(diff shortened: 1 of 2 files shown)


bugbug/bug_features.py
```diff
@@ -509,28 +542,38 @@ def apply_transform(bug):
...
+                }
 
         for bug in bugs:
             if isinstance(bug, dict):
                 results.append(apply_transform(bug))
             elif isinstance(bug, tuple):
                 result1 = apply_transform(bug[0])
                 result2 = apply_transform(bug[1])
+
+                res = apply_transform(bug, tuples=True)
```
marco-c: Nit: `result`

bugbug/bug_features.py
```diff
@@ -548,6 +591,7 @@ def apply_transform(bug):
                             "first_comment2": result2["first_comment"],
                             "comments1": result1["comments"],
                             "comments2": result2["comments"],
+                            "bug_tuples_features": res["data"],
```
marco-c: Nit: `couple_data` instead of `bug_tuples_features`, and move it after `data2`

bugbug/bug_features.py
```diff
@@ -475,21 +488,41 @@ def transform(self, bugs):
...
 
             for feature_extractor in self.feature_extractors:
-                res = feature_extractor(
-                    bug,
-                    reporter_experience=reporter_experience_map[bug["creator"]],
-                    author_ids=author_ids,
-                )
+                res = None
+                if isinstance(feature_extractor, single_bug_feature) and not tuples:
+
```
marco-c: Nit: no newline after `if`

bugbug/bug_features.py
```diff
@@ -475,21 +488,41 @@ def transform(self, bugs):
...
+                    res = feature_extractor(
+                        bug,
+                        reporter_experience=reporter_experience_map[bug["creator"]],
+                        author_ids=author_ids,
+                    )
+
+                elif isinstance(feature_extractor, tuple_bug_feature) and tuples:
+                    res = feature_extractor(bug)
```
marco-c: Remove `and tuples` and do `assert tuples, "Couple features can only be used for couples"` in the branch

bugbug/bug_features.py
```diff
@@ -21,28 +21,36 @@ def field(bug, field):
     return None
 
 
-class has_str(object):
+class single_bug_feature(object):
+    pass
+
+
+class tuple_bug_feature(object):
```
marco-c: Nit: replace `tuple` with `couple` everywhere, since tuple can have many elements in general

bugbug/bug_features.py
marco-c: This way we ensure single bug models are not trying to use features for couples.

marco-c: Actually we can't do this here, as in the couples case we run both kinds of feature extractors.

bugbug/models/duplicate.py
```diff
@@ -48,7 +48,12 @@ def __init__(self, training_size=14000, lemmatization=False):
 
         self.extraction_pipeline = Pipeline(
             [
-                ("bug_extractor", bug_features.BugExtractor([], cleanup_functions)),
+                (
+                    "bug_extractor",
+                    bug_features.BugExtractor(
+                        [bug_features.is_same_product()], cleanup_functions
```
marco-c: Nit: define the extractors earlier, before cleanup_functions (to have the same style as the other models).

bugbug/bug_features.py
```diff
@@ -476,20 +489,41 @@ def transform(self, bugs):
         already_rollbacked = set()
 
         def apply_transform(bug):
-            bug_id = bug["id"]
 
-            if self.rollback and bug_id not in already_rollbacked:
-                bug = bug_snapshot.rollback(bug, self.rollback_when)
-                already_rollbacked.add(bug_id)
+            tuples = isinstance(bug, tuple)
```
marco-c: Nit: `is_couple` instead of `tuples`, it's clearer

PR #695, opened 2019-07-13
consider significant "files" for features

### LINKED ISSUE #685: Add 'files' to regressor features
I've removed it in 66c496bb25b69ca4b42333ee365755f49faf62fd so that I could enable feature importance for the model.
In order to add it back, we need to implement something similar to TfIdfVectorizer's min_df parameter and only consider files that are most often modified.

chidauri:
Fixes #685

### DIFF
--- a/bugbug/commit_features.py
+++ b/bugbug/commit_features.py
@@ -3,8 +3,12 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
+from collections import defaultdict
+
 import pandas as pd
 from sklearn.base import BaseEstimator, TransformerMixin
+
+from bugbug import repository
 
 EXPERIENCE_TIMESPAN = 90
 EXPERIENCE_TIMESPAN_TEXT = f"{EXPERIENCE_TIMESPAN}_days"
@@ -178,8 +182,21 @@ class directory_touched_prev(object):
 
 
 class files(object):
+    def __init__(self, min_freq=0.00003):
+        self.min_freq = min_freq
+        self.count = defaultdict(int)
+
+        for commit in repository.get_commits():
+            for file in commit["files"]:
+                self.count[file] += 1
+        self.total_files = sum(self.count.values())
+
     def __call__(self, commit, **kwargs):
-        return commit["files"]
+        return [
+            file
+            for file in commit["files"]
+            if (self.count[file] / self.total_files) > self.min_freq
+        ]
 
 
 class file_touched_prev(object):
--- a/bugbug/commit_features.py
+++ b/bugbug/commit_features.py
@@ -180,24 +180,21 @@ class directory_touched_prev(object):
 
 
 class files(object):
-    def __init__(self, min_freq=0.00003, commits=[]):
-        self.commits = commits
+    def fit(self, commits, min_freq=0.00003):
         self.min_freq = min_freq
         self.count = defaultdict(int)
 
-        for commit in self.commits:
+        for commit in commits:
             for file in commit["files"]:
                 self.count[file] += 1
         self.total_files = sum(self.count.values())
 
     def __call__(self, commit, **kwargs):
-        if self.commits:
-            return [
-                file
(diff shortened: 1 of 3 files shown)


chidauri:
> Before changes

36746 commits caused regressions
79403 commits did not cause regressions
X: (116149, 10859), y: (116149,)
Cross Validation scores:
Accuracy: f0.7654554703869444 (+/- 0.022148284613653764)
Precision: f0.6089912519123126 (+/- 0.03095320271302983)
Recall: f0.7436657681940699 (+/- 0.030561806540287586)
X_train: (7420, 10859), y_train: (7420,)
X_test: (1162, 10859), y_test: (1162,)

> After changes

36746 commits caused regressions
79403 commits did not cause regressions
X: (116149, 13935), y: (116149,)
Cross Validation scores:
Accuracy: f0.7657998526512577 (+/- 0.022987720883341913)
Precision: f0.6095466486891623 (+/- 0.03219886009248239)
Recall: f0.7436657681940702 (+/- 0.03046656851955754)
X_train: (7420, 13935), y_train: (7420,)
X_test: (1162, 13935), y_test: (1162,)

**why is there such a insignificant difference upon adding like 3000 files, do they have such a low weight?**

bugbug/commit_features.py
```diff
@@ -178,8 +182,21 @@ def __call__(self, commit, **kwargs):
 
 
 class files(object):
+    def __init__(self, min_freq=0.00003):
+        self.min_freq = min_freq
+        self.count = defaultdict(int)
+
+        for commit in repository.get_commits():
```
marco-c: We should only use commits from the training set to calculate the counts (otherwise we leak information from outside the training set into the model).

marco-c:
> why is there such a insignificant difference upon adding like 3000 files, do they have such a low weight?

Yeah I guess so. I thought they might be more important, but apparently they aren't.

bugbug/models/regressor.py
```diff
@@ -110,6 +110,15 @@ def get_labels(self):
             )
         )
 
+        self.feature_extractors += [
```
marco-c: The reponsibility of the `get_labels` function is to only get the labels, we shouldn't overload it to do unrelated things.

What we want to add here is a kind of `fit`, so we should probably do it in `CommitExtractor::fit` (that is, what you are currently doing in files::__init__ you can do in files::fit).

bugbug/commit_features.py
```diff
@@ -178,8 +180,21 @@ def __call__(self, commit, **kwargs):
 
 
 class files(object):
+    def fit(self, commits, min_freq=0.00003):
+        self.min_freq = min_freq
```
marco-c: Since this is an option, you can set it in __init__ (so each model can decide what it wants).

bugbug/commit_features.py
```diff
@@ -178,8 +180,21 @@ def __call__(self, commit, **kwargs):
...
+        self.count = defaultdict(int)
+
+        for commit in commits:
+            for file in commit["files"]:
```
marco-c: Nit: Use `f` instead of `file` as `file` is a keyword

bugbug/commit_features.py
```diff
@@ -202,6 +217,10 @@ def __init__(self, feature_extractors, cleanup_functions):
         self.cleanup_functions = cleanup_functions
 
     def fit(self, x, y=None):
+        for item in self.feature_extractors:
+            if item.__class__ == files:
```
marco-c: Instead of making it specific for `files`, check if the item has a `fit` function

bugbug/model.py
```diff
@@ -170,7 +170,7 @@ def train(self, importance_cutoff=0.15):
         X_iter, y_iter = split_tuple_iterator(self.items_gen(classes))
 
         # Extract features from the items.
-        X = self.extraction_pipeline.fit_transform(X_iter)
+        X = self.extraction_pipeline.fit_transform([item for item in X_iter])
```
marco-c: We can probably simplify a few things if we do this, but let's do it later as a follow-up.

PR #700, opened 2019-07-15
Retrigger microannotate hook if the generation process is not fully done 

### LINKED ISSUE #652: Automatically trigger a new microannotate task if the previous one didn't finish
We are transforming 10000 commits at a time. The first time we run it, we'd have to run the task almost 50 times to get all mozilla-central commits. It'd be better to automate this.

marco-c:
Fixes #652.

### DIFF
--- a/infra/trigger_again.py
+++ b/infra/trigger_again.py
@@ -1,0 +1,91 @@
+#!/bin/env python
+# -*- coding: utf-8 -*-
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import argparse
+import logging
+import os
+import sys
+from urllib.request import urlretrieve
+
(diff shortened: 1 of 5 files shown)


infra/trigger_again.py
```diff
@@ -0,0 +1,91 @@
...
+
+    task = queue.task(os.environ["TASK_ID"])
+    assert len(task["dependencies"]) > 0, "No task dependencies"
+
+    download_artifacts(queue, task["dependencies"])
+
+    should_trigger = False
+    for task_id in task["dependencies"]:
+        with open(task_id, "r") as f:
+            done = True if int(f.read()) == 1 else False
```
La0: Easier to read as `done = int(f.read()) == 1`

infra/trigger_again.py
```diff
@@ -0,0 +1,91 @@
...
+
+        if done:
+            should_trigger = True
+            break
+
+    if should_trigger:
+        hooks = taskcluster.Hooks(options)
```
La0: I would create the service & ping it at the start of task

PR #703, opened 2019-07-15
is_same_component for bug couples

ayush-1506:
Part of #523.

### DIFF
--- a/bugbug/bug_features.py
+++ b/bugbug/bug_features.py
@@ -453,6 +453,14 @@ class is_same_product(couple_bug_feature):
         return bugs[0]["product"] == bugs[1]["product"]
 
 
+class is_same_component(couple_bug_feature):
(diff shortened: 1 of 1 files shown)


bugbug/bug_features.py
```diff
@@ -453,6 +453,11 @@ def __call__(self, bugs, **kwargs):
         return bugs[0]["product"] == bugs[1]["product"]
 
 
+class is_same_component(couple_bug_feature):
+    def __call__(self, bugs, **kwargs):
+        return bugs[0]["component"] == bugs[1]["component"]
```
marco-c: I know it's counter-intuitive, but in this case we should check both product and component (a component is just a sub-group of product, so if you want to check that two components are the same you need to check the parent group, that is 'product', first).

PR #707, opened 2019-07-16
Added labels to confusion matrix

### LINKED ISSUE #704: Show labelled confusion matrix
Currently the output is something like:
```
[[ 924   36]
 [  54 1117]]
```
Instead, it would be better to use tabulate to show something like the table on this page: https://en.wikipedia.org/wiki/Confusion_matrix#Example.

kanurag94:
Fixes #704


kanurag94:
This is implemented using tabulate.
For labelling entire columns as Predicted and rows as True, we can add simple print statement over and left of the matrix.
Although, It looks better this way

marco-c:
Can you show an example output for the defect model and the defect/enhancement/task model?

If it takes too long or runs out of memory, you can restrict the size of the dataset to 5000 bugs by replacing https://github.com/mozilla/bugbug/blob/master/bugbug/models/defect.py#L248 with:
```Python
        return {bug_id: label for bug_id, label in itertools.islice(classes.items(), 5000) if bug_id in bug_ids}
```

bugbug/model.py
```diff
@@ -254,8 +255,18 @@ def train(self, importance_cutoff=0.15):
 
         print(f"No confidence threshold - {len(y_test)} classified")
         confusion_matrix = metrics.confusion_matrix(y_test, y_pred, labels=class_names)
-        print(confusion_matrix)
         tracking_metrics["confusion_matrix"] = confusion_matrix.tolist()
+        for i in range(len(tracking_metrics["confusion_matrix"])):
+            tracking_metrics["confusion_matrix"][i].insert(0, class_names[i])
```
marco-c: We shouldn't modify the `tracking_metrics` object

bugbug/model.py
```diff
@@ -254,8 +255,18 @@ def train(self, importance_cutoff=0.15):
...
+        headers_for_visualisation = class_names.copy()
+        headers_for_visualisation.insert(0, "True/False")
+        print(
+            tabulate(
+                tracking_metrics["confusion_matrix"],
+                headers=headers_for_visualisation,
```
marco-c: You don't need to create a new variable for this, you can simply do:
```Python
headers=class_names + ["True/False"]
```

marco-c:
There is also another place printing a confusion matric a few lines after, you should convert that one too.

bugbug/model.py
kanurag94: This won't append in beginning but in the end. If that's okay ?

bugbug/model.py
kanurag94: Got it

kanurag94:
image
Here is a preview. If okay, I'll do changes as per your review

bugbug/model.py
marco-c: You can do ["True/False"] + class_names to append in the beginning

marco-c:
>  Here is a preview. If okay, I'll do changes as per your review 

I'd like to have the "Predicted class" / "Actual class" like in the Wikipedia table at https://en.wikipedia.org/wiki/Confusion_matrix#Example.

Another thing, "True/False" might work for binary models, but it doesn't for multiclass models. I'd just leave that empty.

kanurag94:
> > Here is a preview. If okay, I'll do changes as per your review
> 
> I'd like to have the "Predicted class" / "Actual class" like in the Wikipedia table at https://en.wikipedia.org/wiki/Confusion_matrix#Example.
> 
> Another thing, `"True/False"` might work for binary models, but it doesn't for multiclass models. I'd just leave that empty.

It should have been `Actual/Pred` in place of `True/False` depicting the column wise classes are real and row wise predicted.
Tabulate doesn't have a way to have different number of sections in different rows. Although this can be done 
image

marco-c:
> Tabulate doesn't have a way to have different number of sections in different rows. Although this can be done

OK! That looks good enough to me! Just double check that the format of the table is correct with what scikit-learn outputs.

kanurag94:
Output:
image

marco-c:
> Output:

Nice! Could you generate it for the defect/enhancement/task model too?

kanurag94:
@marco-c  Output for Defect/enhancement/task :

image

bugbug/model.py
```diff
@@ -254,8 +255,21 @@ def train(self, importance_cutoff=0.15):
 
         print(f"No confidence threshold - {len(y_test)} classified")
         confusion_matrix = metrics.confusion_matrix(y_test, y_pred, labels=class_names)
-        print(confusion_matrix)
         tracking_metrics["confusion_matrix"] = confusion_matrix.tolist()
+        confusion_matrix_reformatted = tracking_metrics["confusion_matrix"].copy()
```
marco-c: Call the variable `confusion_matrix_table`, and replace `tracking_metrics["confusion_matrix"].copy()` with `confusion_matrix.tolist()`

bugbug/model.py
```diff
@@ -254,8 +255,21 @@ def train(self, importance_cutoff=0.15):
...
+        class_names_string = [str(item) for item in class_names]
```
marco-c: You don't really need this, you can use f-strings later to handle the conversion

marco-c: E.g.:
```Python
f"{class_name} (Predicted)"
```

bugbug/model.py
```diff
@@ -254,8 +255,21 @@ def train(self, importance_cutoff=0.15):
...
+        for i in range(len(confusion_matrix_reformatted)):
+            confusion_matrix_reformatted[i].insert(
+                0, str(class_names_string[i]) + " (Actual)"
+            )
+        headers_for_cm = [s + " (Predicted)" for s in class_names_string]
```
marco-c: Rename this `confusion_matrix_header`.

bugbug/model.py
```diff
@@ -284,9 +298,19 @@ def train(self, importance_cutoff=0.15):
                 f"\nConfidence threshold > {confidence_threshold} - {len(y_test_filter)} classified"
             )
             if len(y_test_filter) != 0:
+                confusion_matrix = metrics.confusion_matrix(
```
marco-c: Since you are doing the same thing twice, it might be better to define a helper function which takes in input y_test, y_pred and class_names and returns the output to print.

marco-c: E.g. `def get_labeled_confusion_matrix(y_test, y_pred, class_names):`

bugbug/model.py
kanurag94: f-strings produce results like:
`f"{class_names}"` outputs `'[0, 1]'` instead of `['0', '1']` .

This is how I can still make it work
`        for i in range(len(confusion_matrix_table)):
            confusion_matrix_header.append(f"{class_names[i]} (Predicted)")`

Let me know if that's okay

marco-c: Of course, that was just an example. f-string is just a way to format a variable, if you pass the list you will format the list, if you pass an element of the list you will format the element of the list.

Issue #717, opened 2019-07-16
Calculate other metrics for the similarity models


marco-c:
In the evaluation phase, we should compute:
- recall rate @ 1
- recall rate @ 5
- recall rate @ 10
- precision @ 1
- precision @ 5
- precision @ 10
- MAP

PR #718, opened 2019-07-16
Use nltk's tokenizer instead of split()

ayush-1506:
Fixes : #713

### DIFF
--- a/bugbug/similarity.py
+++ b/bugbug/similarity.py
@@ -29,6 +29,7 @@ try:
     from gensim.corpora import Dictionary
     from nltk.corpus import stopwords
     from nltk.stem.porter import PorterStemmer
+    from nltk.tokenize import word_tokenize
 except ImportError:
     raise ImportError(OPT_MSG_MISSING)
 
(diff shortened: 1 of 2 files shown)


marco-c:
Is there any difference in the evaluation?

ayush-1506:
@marco-c Results:
Without these changes:
```
Recall: 37.17948717948718%
Precision: 4.754098360655737%
```

After these changes:
```
Recall: 35.8974358974359%
Precision: 4.57516339869281%
```

marco-c:
Since it appears to be worse, let's make it optional. Then we'll reevaluate once we have more metrics (#717). Or, let's keep this open, fix #717 and then re-evaluate it right away.

ayush-1506:
Results before changes:
```
Recall @ 1: 26.21951219512195%
Recall @ 5: 30.48780487804878%
Recall @ 10: 30.48780487804878%
Precision @ 1: 19.28571428571429%
Precision @ 5: 6.142857142857143%
Precision @ 10: 3.57142857142857%
Recall: 30.48780487804878%
Precision: 5.213764337851929%
MAP@k : 22.955215419501137%
```

Results after changes:
```
Recall @ 1: 25.0%
Recall @ 5: 28.65853658536585%
Recall @ 10: 28.65853658536585%
Precision @ 1: 17.142857142857142%
Precision @ 5: 6.000000000000002%
Precision @ 10: 3.3571428571428563%
Recall: 28.65853658536585%
Precision: 4.7283702213279675%
MAP@k : 21.076643990929707%
```

marco-c:
So it is indeed worse! Let's make it an option then (False by default).

PR #719, opened 2019-07-16
Add more evaluation metrics

ayush-1506:
Recall at {1, 5, 10}
Precision at {1, 5, 10}

### DIFF
--- a/bugbug/similarity.py
+++ b/bugbug/similarity.py
@@ -93,24 +93,62 @@ class BaseSimilarity:
         total_p = 0
         hits_p = 0
 
+        recall_rate_1 = 0
+        recall_rate_5 = 0
+        recall_rate_10 = 0
+        precision_rate_1 = 0
+        precision_rate_5 = 0
+        precision_rate_10 = 0
+
+        queries = 0
+        apk = []
         for bug in bugzilla.get_bugs():
             if duplicates[bug["id"]]:
-                similar_bugs = self.get_similar_bugs(bug)
+                score = 0
+                num_hits = 0
+                queries += 1
+                similar_bugs = self.get_similar_bugs(bug)[:10]
 
                 # Recall
-                for item in duplicates[bug["id"]]:
+                for idx, item in enumerate(duplicates[bug["id"]]):
                     total_r += 1
                     if item in similar_bugs:
                         hits_r += 1
+                        if idx == 0:
+                            recall_rate_1 += 1
+                        if idx < 5:
+                            recall_rate_5 += 1
+                        if idx < 10:
+                            recall_rate_10 += 1
 
                 # Precision
-                for element in similar_bugs:
+                for idx, element in enumerate(similar_bugs):
                     total_p += 1
                     if element in duplicates[bug["id"]]:
                         hits_p += 1
+                        if idx == 0:
+                            precision_rate_1 += 1
 
+                        if idx < 5:
+                            precision_rate_5 += 1 / 5
+
+                        if idx < 10:
+                            precision_rate_10 += 1 / 10
+
+                        num_hits += 1
+                        score += num_hits / (idx + 1)
+
+                apk.append(score / min(len(duplicates[bug["id"]]), 10))
+
+        print(f"Recall @ 1: {recall_rate_1/total_r * 100}%")
+        print(f"Recall @ 5: {recall_rate_5/total_r * 100}%")
+        print(f"Recall @ 10: {recall_rate_10/total_r * 100}%")
+        print(f"Precision @ 1: {precision_rate_1/queries * 100}%")
+        print(f"Precision @ 5: {precision_rate_5/queries * 100}%")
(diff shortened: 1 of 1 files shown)


marco-c:
I think you are calculating the recall correctly, as you need to divide the actual duplicates found in the first K by all the duplicates.
For precision instead, you need to divide the actual duplicates found by all the bugs found. So e.g. if K=1, you need to divide "# of actual duplicates found" / 1, if K=5, you need to divide "# of actual duplicates found" / 5, instead you are dividing by all bugs found.

marco-c:
https://ils.unc.edu/courses/2013_spring/inls509_001/lectures/10-EvaluationMetrics.pdf can be a good reference.

bugbug/similarity.py
```diff
@@ -93,24 +93,58 @@ def evaluation(self):
...
+
+                        if idx in list(range(5)):
+                            precision_rate_5 += 1 / 5
+
+                        if idx in list(range(10)):
+                            precision_rate_10 += 1 / 10
+
+                        preckq += 1 / ((idx + 1) * len(similar_bugs))
 
+        print(f"Recall @ 1: {recall_rate_1/total_r * 100}%")
```
marco-c: I think the code would be more easy to understand if we treated recall_rate like precision_rate (so calculating a recall for each query and then dividing by the number of queries)

bugbug/similarity.py
```diff
@@ -93,24 +93,58 @@ def evaluation(self):
...
+                    if item in similar_bugs[:10]:
+                        recall_rate_10 += 1
 
                 # Precision
-                for element in similar_bugs:
+                for idx, element in enumerate(similar_bugs):
                     total_p += 1
                     if element in duplicates[bug["id"]]:
                         hits_p += 1
+                        if idx == 1:
```
marco-c: The first index is actually 0

bugbug/similarity.py
```diff
@@ -93,24 +93,58 @@ def evaluation(self):
...
+                            precision_rate_1 += 1
+
+                        if idx in list(range(5)):
```
marco-c: This is a bit strange, `if idx < 5` is better

bugbug/similarity.py
```diff
@@ -93,24 +93,58 @@ def evaluation(self):
...
+                            precision_rate_5 += 1 / 5
+
+                        if idx in list(range(10)):
```
marco-c: Same here

bugbug/similarity.py
```diff
@@ -93,24 +93,58 @@ def evaluation(self):
...
+                            precision_rate_10 += 1 / 10
+
+                        preckq += 1 / ((idx + 1) * len(similar_bugs))
```
marco-c: I'm not sure this works, maybe you can take inspiration from https://github.com/benhamner/Metrics/blob/master/Python/ml_metrics/average_precision.py#L3 (check if your code works with a few examples from https://github.com/benhamner/Metrics/blob/master/Python/ml_metrics/test/test_average_precision.py#L10.

bugbug/similarity.py
```diff
@@ -93,24 +93,63 @@ def evaluation(self):
...
+                        if idx == 0:
+                            precision_rate_1 += 1
 
+                        if idx < 5:
+                            precision_rate_5 += 1 / 5
+
+                        if idx < 10:
+                            precision_rate_10 += 1 / 10
+
+                        if element not in similar_bugs[:idx]:
```
marco-c: This check should be unneeded, as we never suggest twice the same bug

ayush-1506:
```Recall @ 1: 25.609756097560975%
Recall @ 5: 29.878048780487802%
Recall @ 10: 29.878048780487802%
Precision @ 1: 17.857142857142858%
Precision @ 5: 6.000000000000002%
Precision @ 10: 3.499999999999999%
Recall: 29.878048780487802%
Precision: 5.020491803278689%
MAP : 21.282596371882086%```

(Testing on a subset of the bugs)

File: bugbug/similarity.py [hunk 7]
```diff
@@ -93,24 +93,62 @@ def evaluation(self):
...
+
+        print(f"Recall @ 1: {recall_rate_1/total_r * 100}%")
+        print(f"Recall @ 5: {recall_rate_5/total_r * 100}%")
+        print(f"Recall @ 10: {recall_rate_10/total_r * 100}%")
+        print(f"Precision @ 1: {precision_rate_1/queries * 100}%")
+        print(f"Precision @ 5: {precision_rate_5/queries * 100}%")
+        print(f"Precision @ 10: {precision_rate_10/queries * 100}%")
         print(f"Recall: {hits_r/total_r * 100}%")
         print(f"Precision: {hits_p/total_p * 100}%")
+        print(f"MAP : {np.mean(apk) * 100}%")
```
marco-c: Since it's not exactly MAP (as we restrict the number of results), write `MAP@k:` here.

PR #722, opened 2019-07-17
Add option to not cleanup urls when training duplicate model

### LINKED ISSUE #709: Add an option not to cleanup URLs for the Duplicate model
In the duplicate case, having two equal URL can be a strong indication that two bugs are the same.

cklyyung:
Closes #709.

### DIFF
--- a/run.py
+++ b/run.py
@@ -30,7 +30,7 @@ def parse_args(args):
         help="The size of the training set for the duplicate model",
     )
     parser.add_argument(
-        "--dont-cleanup-urls",
+        "--disable-url-cleanup",
         help="Don't cleanup urls when training the duplicate model",
         dest="cleanup_urls",
         default=True,
--- a/run.py
+++ b/run.py
@@ -29,6 +29,13 @@ def parse_args(args):
         type=int,
         help="The size of the training set for the duplicate model",
     )
+    parser.add_argument(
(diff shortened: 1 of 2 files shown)


cklyyung:
If I'm reading this correctly, slightly more duplicates are found when the urls are not cleaned up.

With url cleanup
cleanup_urls

Without url cleanup
dont_cleanup_urls

This is for a training set size of 14000.

marco-c:
@cklyyung the option should actually be added to the similarity script :)
We can add it to the Duplicate model too in the future, but we need to fix #724 first for that.

marco-c:
Well, let's do it for the Duplicate model too anyway, and we'll block adding new additional options on fixing #724.

run.py
```diff
@@ -29,6 +29,13 @@ def parse_args(args):
         type=int,
         help="The size of the training set for the duplicate model",
     )
+    parser.add_argument(
+        "--dont-cleanup-urls",
```
marco-c: Nit: `--disable-url-cleanup`

cklyyung:
> @cklyyung the option should actually be added to the similarity script :)
> We can add it to the Duplicate model too in the future, but we need to fix #724 first for that.

Ah shoot, sorry my bad.

PR #728, opened 2019-07-18
Add option to disable url cleanup for similarity

### LINKED ISSUE #725: Add an option not to cleanup URLs for similarity
In the duplicate case, having two equal URL can be a strong indication that two bugs are the same.

See also #709.

cklyyung:
Fixes #725.

I decided to make `cleanup_urls` a variable in the `BaseSimilarity` class, because I thought that passing an extra variable to all of `__init__`, `text_preprocess`, and `get_ similar_bugs` look a bit messy. Thoughts?

I ran the `lsi` algorithm on 5000 bugs.
With cleanup
```
Recall: 41.156462585034014%
Precision: 4.481481481481482%
```
Without cleanup
```
Recall: 41.83673469387755%
Precision: 4.555555555555555%
```

### DIFF
--- a/bugbug/similarity.py
+++ b/bugbug/similarity.py
@@ -115,7 +115,7 @@ class BaseSimilarity:
 
 class LSISimilarity(BaseSimilarity):
     def __init__(self, cleanup_urls=True):
-        super().__init__(cleanup_urls)
+        super().__init__(cleanup_urls=cleanup_urls)
         self.corpus = []
 
         for bug in bugzilla.get_bugs():
@@ -162,7 +162,7 @@ class LSISimilarity(BaseSimilarity):
 
 class NeighborsSimilarity(BaseSimilarity):
     def __init__(self, k=10, vectorizer=TfidfVectorizer(), cleanup_urls=True):
-        super().__init__(cleanup_urls)
+        super().__init__(cleanup_urls=cleanup_urls)
         self.vectorizer = vectorizer
         self.similarity_calculator = NearestNeighbors(n_neighbors=k)
         text = []
@@ -187,7 +187,7 @@ class NeighborsSimilarity(BaseSimilarity):
 
 class Word2VecWmdSimilarity(BaseSimilarity):
     def __init__(self, cut_off=0.2, cleanup_urls=True):
-        super().__init__(cleanup_urls)
+        super().__init__(cleanup_urls=cleanup_urls)
         self.corpus = []
         self.bug_ids = []
         self.cut_off = cut_off
--- a/bugbug/similarity.py
+++ b/bugbug/similarity.py
@@ -34,16 +34,6 @@ nltk.download("stopwords")
 
 REPORTERS_TO_IGNORE = {"intermittent-bug-filer@mozilla.bugs", "wptsync@mozilla.bugs"}
 
-cleanup_functions = [
-    feature_cleanup.responses(),
-    feature_cleanup.hex(),
-    feature_cleanup.dll(),
-    feature_cleanup.fileref(),
-    feature_cleanup.url(),
-    feature_cleanup.synonyms(),
-    feature_cleanup.crash(),
-]
-
 # A map from bug ID to its duplicate IDs
 duplicates = defaultdict(set)
 all_ids = set(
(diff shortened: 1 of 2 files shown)


marco-c:
@ayush-1506 could you do a first review here?

cklyyung:
> The current changes look good, though I'm wondering if it'd be easier to just add a new argument to `text_preprocess` (boolean argument that indicates including `feature_cleanup.url()` or not), which is equal to the value of the `cleanup_urls` attribute.
> 
> However, If the current implementation is the better way, the PR has my approval. 👍

Hey @ayush-1506, thanks for the review! I agree that just adding the new argument would be easier, and I did start with that. However, I thought that it looked a bit messy since I was passing it to `__init__`, `text_preprocess`, and `get_similar_bugs`. I decided that it would be cleaner if I only passed `cleanup_urls` during the initialization. I could make another branch with the other method to compare if you want?

bugbug/similarity.py
```diff
@@ -62,30 +52,40 @@
...
-        text = func(text)
+class BaseSimilarity:
+    def __init__(self, cleanup_urls=True):
+        self.cleanup_urls = cleanup_urls
+
+    def get_text(self, bug):
+        return "{} {}".format(bug["summary"], bug["comments"][0]["text"])
+
+    def text_preprocess(self, text, join=False):
+        cleanup_functions = [
```
marco-c: Better to define these once in __init__ and reuse them here, otherwise you'll rerun the constructors of the cleanup functions for each text

scripts/evaluate_similarity.py
```diff
@@ -18,18 +18,28 @@ def parse_args(args):
...
+        help="Don't cleanup urls when training the similarity model",
+        dest="cleanup_urls",
+        default=True,
+        action="store_false",
+    )
     return parser.parse_args(args)
 
 
 def main(args):
     if args.algorithm == "lsi":
```
marco-c: It's not great that we have to pass the same argument to the different constructors.
I'd do something like:
```
if args.algorithm == "lsi":
    model_creator = LSISimilarity
elif args.algorithm == "neighbors_tfidf":
   model_creator = NeighborsSimilarity
...

model = model_creator(cleanup_urls=args.cleanup_urls)
```

bugbug/similarity.py
```diff
@@ -62,30 +52,39 @@
...
-def text_preprocess(text, join=False):
-    for func in cleanup_functions:
-        text = func(text)
+    def get_text(self, bug):
+        return "{} {}".format(bug["summary"], bug["comments"][0]["text"])
 
-    text = re.sub("[^a-zA-Z0-9]", " ", text)
+    def text_preprocess(self, text, join=False):
+        if self.cleanup_urls:
+            self.cleanup_functions.append(feature_cleanup.url())
```
marco-c: This should be done in __init__ too

scripts/evaluate_similarity.py
```diff
@@ -18,19 +18,33 @@ def parse_args(args):
...
     elif args.algorithm == "neighbors_tfidf_bigrams":
-        model = NeighborsSimilarity(vectorizer=TfidfVectorizer(ngram_range=(1, 2)))
+        model_creator = NeighborsSimilarity
     elif args.algorithm == "word2vec_wmd":
-        model = Word2VecWmdSimilarity()
-
+        model_creator = Word2VecWmdSimilarity
+
+    if args.algorithm == "neighbors_tfidf_bigrams":
+        model = model_creator(
```
marco-c: This kind of defies the purpose of having a common model_creator.
In the `neighbors_tfidf_bigrams` above, you can define model_creator as a function that calls NeighborsSimilarity with the vectorizer argument + other arguments passed to the function.

scripts/evaluate_similarity.py
```diff
@@ -18,19 +18,33 @@ def parse_args(args):
...
     if args.algorithm == "lsi":
-        model = LSISimilarity()
+        model_creator = LSISimilarity
     elif args.algorithm == "neighbors_tfidf":
-        model = NeighborsSimilarity()
+        model_creator = NeighborsSimilarity
     elif args.algorithm == "neighbors_tfidf_bigrams":
-        model = NeighborsSimilarity(vectorizer=TfidfVectorizer(ngram_range=(1, 2)))
+
+        def model_creator(cleanup_urls):
```
marco-c: Again, it shouldn't be specific to cleanup_urls, otherwise whenever we add a new option we need to update both this function and line 47.

marco-c: You can use kwargs

scripts/evaluate_similarity.py
```diff
@@ -18,19 +18,33 @@ def parse_args(args):
...
     if args.algorithm == "lsi":
-        model = LSISimilarity()
+        model_creator = LSISimilarity
     elif args.algorithm == "neighbors_tfidf":
-        model = NeighborsSimilarity()
+        model_creator = NeighborsSimilarity
     elif args.algorithm == "neighbors_tfidf_bigrams":
-        model = NeighborsSimilarity(vectorizer=TfidfVectorizer(ngram_range=(1, 2)))
+
+        def model_creator(**kwargs):
```
marco-c: Your latest change doesn't really make a difference, you still have a model_creator function where you are hardcoding the cleanup_urls parameter, so whenever we need to add a new parameter we need to update this function (just like before!).
You should instead add vectorizer to kwargs, and then call NeighborsSimilarity with kwargs.

cklyyung: Ack sorry, I'm still a bit confused. If I understand correctly; I should create `kwargs` with `cleanup_urls`, then if the algorithm is `neighbors_tfidf_bigrams`, add the vectorizer argument. Finally, create the model by passing in `kwargs` to `model_creator`; this method would mean I don't have the custom `model_creator` function defined in the `neighbors_tfidf_bigrams` option. Or were you suggesting a different method?

marco-c: @cklyyung the call at line 47 will stay what it is (so it will pass kwargs with `cleanup_urls` to the `model_creator` function at line 38). Then in the `model_creator` function you will update kwargs to add the `vectorizer` parameter and call NeighborsSimilarity with the kwargs.

PR #731, opened 2019-07-18
Add couple target feature

ayush-1506:
Addresses issue : #523


bugbug/bug_features.py
```diff
@@ -476,6 +476,11 @@ def __call__(self, bugs, **kwargs):
         return bugs[0]["op_sys"] == bugs[1]["op_sys"]
 
 
+class is_same_target(couple_bug_feature):
```
marco-c: `is_same_target_milestone`

Issue #360, opened 2019-05-08
Check training metrics are not getting too bad


marco-c:
Depends on #342.

marco-c:
For this we need:
1) Plot of the training metrics;
2) Alert when some metric is much worse compared to before.

We can do both in a script that we run in a task at the end of the pipeline.

marco-c:
It should be both at per-version and per-date (especially per-date).

PR #734, opened 2019-07-19
Retrieve metric date and later. Fix #614

### LINKED ISSUE #614: Retrieve all metrics from a date and later
See https://github.com/mozilla/bugbug/pull/604#discussion_r296311845

Lothiraldan:
I've updated the `scripts/retrieve_training_metrics.py` file to include a `date and later` download mode.

You can test it like that:

```
python scripts/retrieve_training_metrics.py component v0.0.69 2018
python scripts/retrieve_training_metrics.py component v0.0.69 2019.07
python scripts/retrieve_training_metrics.py component v0.0.69 2019.07.17
python scripts/retrieve_training_metrics.py component v0.0.69 2019.07.16.00.01.03
```

There are several questions around this feature:
- Should we keep the download a single metrics file given an exact date? If we pass a date in the past, the `date and later` mode will download more than 1 file.
- The `date and later` mode doesn't support the print to stdout option either which might be useful for scripts.
- Do we have a specific file name pattern that could helps for https://github.com/mozilla/bugbug/issues/360? Currently we save it like that:

```
INFO:__main__:Retrieving metrics from https://index.taskcluster.net/v1/task/project.relman.bugbug.train_component.v0.0.69.2019.07.16.00.01.03/artifacts/public/metrics.json
INFO:__main__:Metrics saved to 'metric_project.relman.bugbug.train_component.v0.0.69.2019.07.16.00.01.03.json'
INFO:__main__:Retrieving metrics from https://index.taskcluster.net/v1/task/project.relman.bugbug.train_component.v0.0.69.2019.07.18.11.54.00/artifacts/public/metrics.json
INFO:__main__:Metrics saved to 'metric_project.relman.bugbug.train_component.v0.0.69.2019.07.18.11.54.00.json'
```


marco-c:
> Should we keep the download a single metrics file given an exact date? If we pass a date in the past, the date and later mode will download more than 1 file.

Probably unnecessary? We could add it later if needed.

> The date and later mode doesn't support the print to stdout option either which might be useful for scripts.

We can add this later if needed too.

> Do we have a specific file name pattern that could helps for #360? Currently we save it like that:

Maybe dividing the different parts with "_" character (which can't be in the version nor in the date, so it should be safe to split the name by "_").

scripts/retrieve_training_metrics.py
```diff
@@ -3,20 +3,108 @@
...
 import requests
+import taskcluster
+
+from bugbug.utils import get_taskcluster_options
 
 LATEST_URI = "train_{}.latest"
 VERSIONED_URI = "train_{}.{}"
 DATED_VERSIONED_URI = "train_{}.{}.{}"
-BASE_URL = "https://index.taskcluster.net/v1/task/project.relman.bugbug.{}/artifacts/public/metrics.json"
+PROJECT_PREFIX = "project.relman.bugbug.{}"
```
marco-c: You're not using this anywhere

scripts/retrieve_training_metrics.py
```diff
@@ -26,47 +114,16 @@ def main():
         "version",
         nargs="?",
         help="Which bugbug version should we retrieve training metrics from.",
-        default=None,
     )
     parser.add_argument(
         "date",
         nargs="?",
         help="Which date should we retrieve training metrics from. Default to latest",
```
marco-c: There is no default anymore, right?

scripts/retrieve_training_metrics.py
```diff
@@ -26,47 +114,16 @@ def main():
         "version",
         nargs="?",
         help="Which bugbug version should we retrieve training metrics from.",
```
marco-c: I think for the version we should default to "latest". The main use case is getting how metrics evolve over time, even while the versions of bugbug change

scripts/retrieve_training_metrics.py
```diff
@@ -3,20 +3,108 @@
...
+    options = get_taskcluster_options()
+
+    index = taskcluster.Index(options)
+
+    index.ping()
+
+    # Split the date
+    from_date = date.split(".")
+
+    uris = []
```
marco-c: I wonder if we can't find nicer names for the variables, I've had to read the code a few times to understand it

scripts/retrieve_training_metrics.py
```diff
@@ -3,20 +3,108 @@
...
+
+        for namespace in get_namespaces(index, index_uri):
+            new_uri = uri.copy()
+            new_uri.append(namespace["name"])
+
+            if not is_later_or_equal(new_uri, from_date):
+                LOGGER.debug("NEW URI %s is before %s", new_uri, from_date)
+                continue
+
+            # Temp
```
marco-c: ?

scripts/retrieve_training_metrics.py
Lothiraldan: I will try to come with better variable names

scripts/retrieve_training_metrics.py
Lothiraldan: Yeah there is no default anymore

scripts/retrieve_training_metrics.py
```diff
@@ -3,70 +3,120 @@
...
 import requests
+import taskcluster
+
+from bugbug.utils import get_taskcluster_options
 
 LATEST_URI = "train_{}.latest"
-VERSIONED_URI = "train_{}.{}"
-DATED_VERSIONED_URI = "train_{}.{}.{}"
-BASE_URL = "https://index.taskcluster.net/v1/task/project.relman.bugbug.{}/artifacts/public/metrics.json"
+VERSIONED_URI = "train_{}.per_date"
```
marco-c: Nit: we shouldn't call it `VERSIONED_URI`, since the version is no longer there

scripts/retrieve_training_metrics.py
```diff
@@ -3,70 +3,120 @@
...
+DATED_VERSIONED_URI = "train_{}.per_date.{}"
```
marco-c: Same here

Issue #735, opened 2019-07-19
Test switching from ProcessPoolExecutor to ThreadPoolExecutor in the repository mining script


marco-c:
The work done in the workers is actually I/O bound (communication with the HG command server), so using threads might be much better than using processes.

marco-c:
Partially done in https://github.com/mozilla/bugbug/pull/737, we need to check if it makes sense for the other usage too.

marco-c:
The other usage is faster with multiple processes, probably because we're parsing the patches.

PR #737, opened 2019-07-19
Use threads instead of processes for hg_log

marco-c:
Part of #735

Tested it on 50k commits, the thread version took ~14 minutes, the processes version took ~18 minutes.


marco-c:
Merging as I need this for another script, please do a post-review ;)

PR #739, opened 2019-07-19
Issue 733 - Use db.download in trainer.py

### LINKED ISSUE #733: In the trainer script, use db.download instead of implementing download manually
script/trainer.py is currently implementing DB download manually, it should instead use db.download (similarly to scripts/commit_classifier.py).

kanurag94:
Fixes #733

### DIFF
--- a/scripts/trainer.py
+++ b/scripts/trainer.py
@@ -4,11 +4,10 @@ import argparse
 import json
 import os
 from logging import INFO, basicConfig, getLogger
-from urllib.request import urlretrieve
 
 import zstandard
 
-from bugbug import get_bugbug_version, model
+from bugbug import db, get_bugbug_version, model
 from bugbug.models import get_model_class
 from bugbug.utils import CustomJsonEncoder
 
@@ -19,13 +18,6 @@ BASE_URL = "https://index.taskcluster.net/v1/task/project.relman.bugbug.data_{}.
 
 
 class Trainer(object):
-    def decompress_file(self, path):
-        dctx = zstandard.ZstdDecompressor()
-        with open(f"{path}.zst", "rb") as input_f:
-            with open(path, "wb") as output_f:
-                dctx.copy_stream(input_f, output_f)
-        assert os.path.exists(path), "Decompressed file exists"
-
     def compress_file(self, path):
         cctx = zstandard.ZstdCompressor()
         with open(path, "rb") as input_f:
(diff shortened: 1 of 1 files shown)


marco-c:
@kanurag94 did you test the script?

kanurag94:
@marco-c I tested `bugbug-train defect` and `backout` with success

scripts/trainer.py
```diff
@@ -36,11 +28,8 @@ def download_db(self, db_type):
         path = f"data/{db_type}.json"
         formatted_base_url = BASE_URL.format(db_type, f"v{get_bugbug_version()}")
         url = f"{formatted_base_url}/{db_type}.json.zst"
-        logger.info(f"Downloading {db_type} database from {url} to {path}.zst")
-        urlretrieve(url, f"{path}.zst")
-        assert os.path.exists(f"{path}.zst"), "Downloaded file exists"
-        logger.info(f"Decompressing {db_type} database")
-        self.decompress_file(path)
+        db.register(path, url, get_bugbug_version())
```
marco-c: We shouldn't register the db here, it should be registered by the bugzilla.py or repository.py script.

scripts/trainer.py
```diff
@@ -36,11 +28,8 @@ def download_db(self, db_type):
...
+        db.download(path, force=True, support_files_too=True)
```
marco-c: I think we don't need support_files here.

scripts/trainer.py
kanurag94: Got it

scripts/trainer.py
kanurag94: Agreed

scripts/trainer.py
```diff
@@ -34,13 +26,7 @@ def compress_file(self, path):
...
     def download_db(self, db_type):
         path = f"data/{db_type}.json"
-        formatted_base_url = BASE_URL.format(db_type, f"v{get_bugbug_version()}")
-        url = f"{formatted_base_url}/{db_type}.json.zst"
-        logger.info(f"Downloading {db_type} database from {url} to {path}.zst")
-        urlretrieve(url, f"{path}.zst")
-        assert os.path.exists(f"{path}.zst"), "Downloaded file exists"
-        logger.info(f"Decompressing {db_type} database")
-        self.decompress_file(path)
+        db.download(path, force=True, support_files_too=False)
```
marco-c: Nit: support_files_too is False by default, you can just avoid passing it.

scripts/trainer.py
```diff
@@ -34,13 +26,7 @@ def compress_file(self, path):
 
     def download_db(self, db_type):
         path = f"data/{db_type}.json"
```
marco-c: Instead of passing the path manually, we should use bugzilla.BUGS_DB instead of "bugs" and repository.COMMITS_DB instead of "commits".

kanurag94: I was trying to avoid unnecessary imports. But yes, it gives more structure to the code. I'll do it.

marco-c: They are not unnecessary, as we want to use them to reduce code duplication

PR #766, opened 2019-07-22
Add same first affected version feature.

ayush-1506:
Addresses issue : #523

### DIFF
--- a/bugbug/bug_features.py
+++ b/bugbug/bug_features.py
@@ -258,35 +258,40 @@ class ever_affected(single_bug_feature):
         return False
 
 
+def get_versions_statuses(bug):
+    unaffected = []
+    affected = []
+    for key, value in bug.items():
+        version = None
+        if key.startswith("cf_status_firefox_esr"):
+            version = key[len("cf_status_firefox_esr") :]
+        elif key.startswith("cf_status_firefox"):
+            version = key[len("cf_status_firefox") :]
+
+        if version is None:
+            continue
+
+        if value == "unaffected":
+            unaffected.append(version)
+        elif value in [
+            "affected",
+            "fixed",
+            "wontfix",
+            "fix-optional",
+            "verified",
+            "disabled",
+            "verified disabled",
+        ]:
+            affected.append(version)
+
+    return unaffected, affected
+
+
 class affected_then_unaffected(single_bug_feature):
     name = "status has ever been set to 'affected' and 'unaffected'"
 
     def __call__(self, bug, **kwargs):
-        unaffected = []
-        affected = []
-        for key, value in bug.items():
-            version = None
-            if key.startswith("cf_status_firefox_esr"):
-                version = key[len("cf_status_firefox_esr") :]
-            elif key.startswith("cf_status_firefox"):
-                version = key[len("cf_status_firefox") :]
-
-            if version is None:
-                continue
-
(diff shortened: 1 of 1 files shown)


bugbug/bug_features.py
```diff
@@ -481,6 +481,22 @@ def __call__(self, bugs, **kwargs):
         return bugs[0]["target_milestone"] == bugs[1]["target_milestone"]
 
 
+class is_first_affected_same(couple_bug_feature):
+    def __call__(self, bugs, **kwargs):
+        version_1 = [
+            re.sub("cf_status_firefox_esr|cf_status_firefox", "", change)[:2]
+            for change in bugs[0].keys()
```
marco-c: Why the name `change`? The variable here is the name of the field.

bugbug/bug_features.py
```diff
@@ -481,6 +481,22 @@ def __call__(self, bugs, **kwargs):
         return bugs[0]["target_milestone"] == bugs[1]["target_milestone"]
 
 
+class is_first_affected_same(couple_bug_feature):
+    def __call__(self, bugs, **kwargs):
+        version_1 = [
+            re.sub("cf_status_firefox_esr|cf_status_firefox", "", change)[:2]
```
marco-c: The field can have different values, `affected` is just one of them. This will not work in many cases, e.g. if bug 1 has cf_status_firefox51=unaffected and cf_status_firefox52=affected and bug 2 has cf_status_firefox52=affected, this will return false, but instead it should be true (the first affected version is 52).

bugbug/bug_features.py
```diff
@@ -481,6 +481,22 @@ def __call__(self, bugs, **kwargs):
         return bugs[0]["target_milestone"] == bugs[1]["target_milestone"]
 
 
+class is_first_affected_same(couple_bug_feature):
+    def __call__(self, bugs, **kwargs):
```
marco-c: I suggest moving the code at https://github.com/mozilla/bugbug/blob/7f5f43d4c63337e2a475998bd1d8003dca752b8c/bugbug/bug_features.py#L265-L288 in a helper function and reusing it here.

bugbug/bug_features.py
```diff
@@ -481,6 +484,13 @@ def __call__(self, bugs, **kwargs):
...
+        return min(
+            map(int, affected_then_unaffected().get_lists(bugs[0], **kwargs)[1])
```
marco-c: You are creating a new object of class affected_then_unaffected simply to call `get_lists`. It'd be better instead to define get_lists (with a better name, as `get_lists` is *really* generic :P) as a helper function outside of any class, given that it is useful for many classes and it's not directly linked to `affected_then_unaffected`.

marco-c: Not sure you need to convert to int, but if you do do it directly in the helper function when you get the list of affected versions.

ayush-1506: Oh, I just realized comparison operations work between strings, so I guess there's no need for the `map(int ..)`

bugbug/bug_features.py
```diff
@@ -258,35 +258,40 @@ def __call__(self, bug, **kwargs):
         return False
 
 
+def get_versions_statuses(bug, **kwargs):
```
marco-c: You are not using kwargs

PR #778, opened 2019-07-24
Add couple delta creation time

ayush-1506:
Addresses issue: #523

### DIFF
--- a/bugbug/bug_features.py
+++ b/bugbug/bug_features.py
@@ -5,9 +5,10 @@
 
 import re
 from collections import defaultdict
-from datetime import datetime, timezone
+from datetime import datetime, timedelta, timezone
 
 import pandas as pd
+from dateutil import parser
 from libmozdata import versions
 from sklearn.base import BaseEstimator, TransformerMixin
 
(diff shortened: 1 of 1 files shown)


bugbug/bug_features.py
```diff
@@ -505,6 +505,14 @@ def __call__(self, bugs, **kwargs):
         )
 
 
+class couple_delta_creation_date(couple_bug_feature):
+    def __call__(self, bugs, **kwargs):
+        delta = datetime.strptime(bugs[0], "%Y-%m-%dT%H:%M:%SZ") - datetime.strptime(
```
marco-c: It'd be better to use `dateutil.parser`, so we'd be more robust to possible changes in the format.

bugbug/bug_features.py
```diff
@@ -505,6 +505,14 @@ def __call__(self, bugs, **kwargs):
...
+            bugs[1], "%Y-%m-%dT%H:%M:%SZ"
+        )
+        return delta.days + delta.seconds / (24 * 60 * 60)
```
marco-c: `delta / timedelta(days=1)` is nicer

PR #786, opened 2019-07-25
Clean cached data when the bug gets updated. Fix #765

### LINKED ISSUE #765: Invalidate cache when a bug changes
Otherwise we risk making an obsolete prediction for a bug.

Lothiraldan:


bugbug/bugzilla.py
```diff
@@ -104,7 +104,9 @@ def bughandler(bug):
     return all_ids
 
 
-def get(ids_or_query):
+def get(
+    ids_or_query, download_comment=True, download_attachment=True, download_history=True
```
marco-c: Nit: rename the parameters `get_comment` or `comment` instead of `download_comment` (same with attachment and history)

http_service/app.py
```diff
@@ -132,6 +130,39 @@ def is_running(model_name, bug_id):
     return False
 
 
+def get_bug_ids_last_change_time(bug_ids):
+    query = {"id": bug_ids, "include_fields": ["last_change_time", "id"]}
+    bug = bugzilla.get(
```
marco-c: Nit: bugs

marco-c: I wonder if it wouldn't even be faster to do the query directly with requests instead of libmozdata. libmozdata is chunking by 100 bugs at a time simply because it's getting a lot of data for each bug, but if we just get id and last_change_time we might be able to get all of the bugs in the same query.

marco-c: I wonder if it wouldn't even be faster to perform the query directly with requests instead of with libmozdata. libmozdata is first performing a query to get the number of results, then chunking in batches of 100 bugs, simply because often people retrieve a lot of data from bugs. In this case our query only requests the id and the last_change_time, so we might as well do one query to bugzilla instead of chunking.

http_service/app.py
```diff
@@ -132,6 +130,39 @@ def is_running(model_name, bug_id):
...
+        query, download_comment=False, download_attachment=False, download_history=False
+    )
+
+    return bug
+
+
+def check_bug_updated_since_last_cached(model_name, bug_id, change_time):
```
marco-c: We could call the function something like `is_bug_cache_invalidated`

marco-c: `is_prediction_invalidated`

marco-c: Or something else a bit shorter than `check_bug_updated_since_last_cached`

http_service/app.py
```diff
@@ -132,6 +130,39 @@ def is_running(model_name, bug_id):
     return False
 
 
+def get_bug_ids_last_change_time(bug_ids):
```
marco-c: `get_bugs_last_change_time`

http_service/app.py
```diff
@@ -132,6 +130,39 @@ def is_running(model_name, bug_id):
...
+        # We can have a result without a cache time
+        if redis_conn.get(result_key(model_name, bug_id)):
+            return True
+
+        return False
+
+    return saved_change_time.decode("utf-8") != change_time
+
+
+def check_bug_updated_and_clean(model_name, bug_id, change_time):
```
marco-c: Not sure if it's better to have a function called `clean_prediction_cache`, and then in each call site do:
```
if is_prediction_invalidated(model_name, bug_id, change_time):
    clean_prediction_cache(model_name, bug_id)
```

Or maybe what you are doing now, but changing the function name to `clean_prediction_cache` (even if it's not always doing it).

I just feel the current name is too complex.

http_service/app.py
Lothiraldan: That might indeed be faster, will Bugzilla accept returning 1000 result at the same time? Do you have some pointers where we are querying bugzilla directly with requests?

marco-c: > That might indeed be faster, will Bugzilla accept returning 1000 result at the same time?

Yes, we're doing that in relman-auto-nag (with the chunk size parameter of get_bugs).

> Do you have some pointers where we are querying bugzilla directly with requests?

You can just do something like `requests.get('https://bugzilla.mozilla.org/rest/bug', params=query)`

Lothiraldan: I went the requests way, thank you for the idea

http_service/app.py
```diff
@@ -51,6 +52,9 @@
 
 BUGZILLA_TOKEN = os.environ.get("BUGBUG_BUGZILLA_TOKEN")
 
+# Keep an HTTP client around for persistent connections
+HTTP_CLIENT, BUGZILLA_API_URL = get_bugzilla_http_client()
```
marco-c: Nit: BUGZILLA_HTTP_CLIENT

http_service/app.py
```diff
@@ -126,12 +128,54 @@ def is_running(model_name, bug_id):
...
+def is_prediction_invalidated(model_name, bug_id, change_time):
+    # First get the saved change time
+    change_key = change_time_key(model_name, bug_id)
+
+    saved_change_time = redis_conn.get(change_key)
+
+    # If we have no last changed time, the bug was not classified yet or the bug was classified by an old worker
+    if not saved_change_time:
+        # We can have a result without a cache time
+        if redis_conn.get(result_key(model_name, bug_id)):
```
marco-c: I wonder if we couldn't avoid getting the result twice from redis (once here and once in get_bug_classification).

Lothiraldan: We would get it twice only if the cache key is not in Redis, which could happens in three cases:
- The bug was not classified yet
- The bug was classified but expired
- The bug was classified with an old version of the code and it won't happen again.

I will see if I can check if I can check only if the redis key exists.

Issue #298, opened 2019-04-19
Run the linting task only on modified files


Lothiraldan:
We could reduce the duration of linting tasks by running pre-commit only on modified / new files. But we need to be careful about edge-cases like people pushing several commits at the same time.

marco-c:
This is a good-first-bug for somebody who's really familiar with git.

thedeveloperr:
Hey I would like to work on it. I am new to bugbug, would like to take a try. Can you please assign it to me ? Also if possible some guidance or tips for an absolute new contributor to bugbug like me. 😅

marco-c:
Hello @thedeveloperr, you can work on any issue that you see open, but issues are not assigned until there is a PR open to fix them (see CONTRIBUTING.md).

To fix this, https://github.com/mozilla/bugbug/blob/331aa50f1fa5e142bd0870ede25721c8bda6a54a/.taskcluster.yml#L46 needs to be done.

dishvyas:
So i need to find a way to run pre-commit on modified files only which will require me to write a function to get the modified files only and then run the command. Am i understanding it wrong or missing something here??

marco-c:
@dishvyas there is already a PR open to fix this, #791.

PR #791, opened 2019-07-26
Modify linting task to only run on modified files

cklyyung:
Addresses issue #298 .
I'm comparing against the master branch of the original repo; `upstream/master...` will compare the original repo to what's currently checked out, from their common ancestor. The `diff-filter` is an exclusive filter, that filters out files that have been deleted.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -21,6 +21,14 @@ tasks:
         then: ${event.after}
         else: ${event.release.tag_name}
 
+    base_rev:
+      $if: 'tasks_for == "github-pull-request"'
+      then: ${event.pull_request.base.sha}
+      else:
+        $if: 'tasks_for == "github-push"'
+        then: ${event.before}
+        else: ${event.release.tag_name}
+
     repository:
       $if: 'tasks_for == "github-pull-request"'
       then: ${event.pull_request.head.repo.html_url}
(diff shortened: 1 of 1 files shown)


marco-c:
I think to be more precise, you could probably use the target branch information from the github event.

cklyyung:
I decided to use `head_rev` instead of `head_branch`, since I would have to add an additional command to fetch the branch from origin.

marco-c:
> I think to be more precise, you could probably use the target branch information from the github event.

This comment still applies.
You are getting all files modified compared to master, but PRs can be against other branches, master is not the only possible target branch.

cklyyung:
For a `github-push` event, I'm grabbing `event.before`. According to the [docs](https://developer.github.com/v3/activity/events/types/#pushevent), this is the most recent commit on the branch before the push. 
I'm not sure if this will be sufficient; consider a case where commit `A` is pushed with a linting error on file `Z`. The check will fail. Then commit `B` is pushed, which only modifies file `Y`, and not file `Z`. The check will succeed, since file `Z` won't be caught in the diff between commits `A` and `B`.
Also, in the original issue, this was suggested as a way to reduce the runtime of the linting task. Looking at the logs, this still runs at around the same time as checking all files.

marco-c:
> I'm not sure if this will be sufficient; consider a case where commit A is pushed with a linting error on file Z. The check will fail. Then commit B is pushed, which only modifies file Y, and not file Z. The check will succeed, since file Z won't be caught in the diff between commits A and B.

This case should never happen, since we never push commits or merge PRs with linting errors.

>  Also, in the original issue, this was suggested as a way to reduce the runtime of the linting task. Looking at the logs, this still runs at around the same time as checking all files.

Yeah, maybe it's not worth it then. There are also other complexities that might cause this to fail, for example, what if you have a pull request which wants to merge branchA into master, where they share some parts of the history? `git diff --name-only --diff-filter=d ${base_rev}...${head_rev}` would probably be wrong.

So, I'd say let's not do this, and let's just remove the TODO.

PR #794, opened 2019-07-26
Validate batch endpoint inputs

Lothiraldan:


setup.py
```diff
@@ -29,7 +29,6 @@ def read_requirements(file_):
 for extra in extras:
     extras_require[extra] = read_requirements("extra-%s-requirements.txt" % extra)
 
-
```
marco-c: Nit: unnecessary change

http_service/app.py
```diff
@@ -379,6 +381,19 @@ def batch_prediction(model_name):
     # TODO Check is JSON is valid and validate against a request schema
     batch_body = json.loads(request.data)
 
+    # Validate
+    schema = {
+        "bugs": {
+            "type": "list",
+            "minlength": 1,
+            "maxlength": 1000,
```
marco-c: Do we really need a limit now?

Lothiraldan: I really think so, if we get a too big batch, it will slow down the web worker, we might generate a query string too big for Bugzilla and the resulting background job will takes more time to run, slowing down the rest.

We need to find the right balance between processing each bug individually and processing them as enormous batches.

marco-c: Is there really a difference? In the end we'd still be queueing all bugs for classification, even if we do it in batches, no?

Lothiraldan: I would prefer to limit the batch size at the bugbug level, even if we batch the query to bugzilla or fallbacking to limzodata, consumers might still be able to do a request with an enormous amount of bug ids. Worst case scenario someone to a request asking to classify all bug ids in bugzilla. This will make make the service hit a timeout or overloading Bugzilla itself.

marco-c: > I would prefer to limit the batch size at the bugbug level, even if we batch the query to bugzilla or fallbacking to limzodata, consumers might still be able to do a request with an enormous amount of bug ids. Worst case scenario someone to a request asking to classify all bug ids in bugzilla.

The thing is that they can still do that, by performing 1000 requests each with a batch size of 1000. If we want to avoid it, we need to implement some rate limiting.

> This will make make the service hit a timeout or overloading Bugzilla itself.

Where will the timeout happen?

PR #798, opened 2019-07-26
Add method to Word2VecWmdSimilarity model to get distance between two bugs

ayush-1506:


bugbug/similarity.py
```diff
@@ -349,3 +349,30 @@ def get_similar_bugs(self, query):
             for similar in sorted(similarities, key=lambda v: v[1])[:10]
             if similar[0] != query["id"] and similar[1] < self.cut_off
         ]
+
+    def get_distance(self, query1, query2):
```
marco-c: It might be worth adding this function for all Similarity models, but for now you could just define an abstract function for in the BaseSimilarity class and then we'll implement it for the other models later.

marco-c: You are duplicating quite a bit of code between this function and `get_similar_bugs`, I suggest writing an helper function which you can call from both.

bugbug/similarity.py
```diff
@@ -148,6 +148,9 @@ def evaluation(self):
         print(f"Precision: {hits_p/total_p * 100}%")
         print(f"MAP@k : {np.mean(apk) * 100}%")
 
+    def get_distance(self, query1, query2):
+        pass
```
marco-c: Define this as an abstract method (@abstractmethod)

bugbug/similarity.py
```diff
@@ -349,3 +356,22 @@ def get_similar_bugs(self, query):
...
+    def get_distance(self, query1, query2):
+
+        words1 = self.text_preprocess(self.get_text(query1))
+        words1 = [word for word in words1 if word in self.w2vmodel.wv.vocab]
+        words2 = self.text_preprocess(self.get_text(query2))
+        words2 = [word for word in words2 if word in self.w2vmodel.wv.vocab]
+
+        all_distances = self.all_distances(words1)
+
+        doc_words_clean = [
```
marco-c: This is the same as words2, I think you don't need it.

bugbug/similarity.py
```diff
@@ -288,12 +291,8 @@ def nbow(document):
 
         return emd(d1, d2, distance_matrix)
 
-    def get_similar_bugs(self, query):
-
-        words = self.text_preprocess(self.get_text(query))
-        words = [word for word in words if word in self.w2vmodel.wv.vocab]
-
-        all_distances = np.array(
+    def all_distances(self, words):
```
marco-c: Nit: `calculate_all_distances`

PR #802, opened 2019-07-28
Switching to alpine as base image

kanurag94:
Addresses #292


infra/dockerfile.base
```diff
@@ -1,13 +1,13 @@
...
+FROM python:3.7.3-alpine
 
 # Setup dependencies in a cacheable step
 COPY requirements.txt /requirements.txt
-RUN apt-get update && \
-		apt-get install -y --no-install-recommends gcc g++ libgomp1 && \
+
+RUN apk update && apk upgrade && \
+		echo "manylinux1_compatible = True" > /usr/local/lib/python3.7/site-packages/_manylinux.py && \
+		apk add --no-cache gcc g++ libgomp lapack-dev git libexecinfo-dev make libc6-compat && \
```
marco-c: You are not removing these packages anymore. This is going to increase the size of the resulting image.

kanurag94: Sorry, dropped out of my mind. Updated PR

infra/dockerfile.base
```diff
@@ -1,13 +1,14 @@
...
 		pip install --disable-pip-version-check --no-cache-dir -r /requirements.txt && \
-		apt-get remove -y gcc g++ && \
-		apt-get autoremove -y && \
-		rm -rf /var/lib/apt/lists/*
+		apk del gcc g++ libgomp lapack-dev git libexecinfo-dev make libc6-compat && \
+		rm -rf /var/cache/apk/*
 
 COPY . /tmp/bugbug
-RUN pip install --disable-pip-version-check --no-cache-dir /tmp/bugbug
```
marco-c: Why did you remove these options?

infra/dockerfile.commit_retrieval
```diff
@@ -4,18 +4,18 @@ COPY infra/git_hyper_blame_patch /root/git_hyper_blame_patch
 
 # Mercurial need Python2 :(
 # git and git-hyper-blame (from depot_tools) are required by the annotate pipeline.
-RUN apt-get update && \
-    apt-get install -y python python-pip git curl && \
+RUN apk update && apk upgrade && \
+    apk add --no-cache gcc g++ libgomp lapack-dev git libexecinfo-dev make libc6-compat && \
```
marco-c: Are you sure you need gcc here? It wasn't needed before

infra/dockerfile.commit_retrieval
```diff
@@ -4,18 +4,18 @@ COPY infra/git_hyper_blame_patch /root/git_hyper_blame_patch
...
+    apk add python-dev py-pip curl && \
```
marco-c: Why a new line to add these? You can add them together with the other packages above

marco-c: Or at least you have to use ` --no-cache` here too

infra/dockerfile.commit_retrieval
```diff
@@ -4,18 +4,18 @@ COPY infra/git_hyper_blame_patch /root/git_hyper_blame_patch
...
     python2 -m pip install --disable-pip-version-check --no-cache-dir mercurial==4.8 && \
     hg clone -r 0d68b0deeca2 https://hg.mozilla.org/hgcustom/version-control-tools /version-control-tools/ && \
-    rm -r /version-control-tools/.hg /version-control-tools/ansible /version-control-tools/docs /version-control-tools/testing && \
+    rm -rf /version-control-tools/.hg /version-control-tools/ansible /version-control-tools/docs /version-control-tools/testing && \
```
marco-c: Why this change? We want to know if one of the directories doesn't exist, so we can stop removing it

infra/dockerfile.base
kanurag94: I mistakenly took an old branch for reference. I'll update the patch

infra/dockerfile.commit_retrieval
kanurag94: Yes. Mercurial setup requires gcc to be installed.
I'll still look if I can reduce some dependency

infra/dockerfile.commit_retrieval
kanurag94: Will do. Thanks

infra/dockerfile.commit_retrieval
marco-c: Mmmh, that's strange, how is it working without your changes?

kanurag94: Not sure, but python package earlier was may be sufficient to compile headers required for mercurial install.

infra/dockerfile.commit_retrieval
kanurag94: This is a mistake. Rectified in new PR. Thanks for pointing out.

infra/dockerfile.commit_retrieval
kanurag94: Oh, So earlier slim image used to install gcc as a part of "apt-get update" command

`The following additional packages will be installed:
  binutils build-essential bzip2 cpp cpp-6 dbus dirmngr dpkg-dev fakeroot file
  g++ g++-6 gcc gcc-6`

While this isn't the case with alpine image

marco-c: Can you check if you can run the `commit_retriever` script in a Docker container with this image?

kanurag94: I got some errors installing numpy in bug-http_service. 
I'll update you as soon as I resolve them

marco-c:
Closing as the PR is stale. Feel free to reopen if you still intend to work on this.

PR #803, opened 2019-07-29
Use revision title when the commit message is missing from the diff commit

marco-c:


scripts/commit_classifier.py
```diff
@@ -69,18 +69,30 @@ def apply_phab(self, hg, diff_id):
...
+            diff_data = diffs_data.get(diff_phid)
+
+            commits = diff_data["commits"]
+            revision = diff_data["revision"]
 
-            message = ""
-            if commit:
-                message += "{}\n".format(commit[0]["message"])
+            if commits and commits[0]["message"]:
+                message = commits[0]["message"]
```
La0: You lost the `\n` here. Is that wanted ?

marco-c: Yeah, I thought it's actually unnecessary

Issue #804, opened 2019-07-29
Error running the LSI/Neighbors similarity model


ayush-1506:
Looks like adding an abstractmethod to BaseSimilarity is causing other derived classes (that don't overwrite `get_distance`) to give an error :

```
python scripts/evaluate_similarity.py --algorithm=lsi
INFO:gensim.summarization.textcleaner:'pattern' package not found; tag filters are not available for English
[nltk_data] Downloading package stopwords to
[nltk_data]     /home/ayush99/nltk_data...
[nltk_data]   Package stopwords is already up-to-date!
Traceback (most recent call last):
  File "scripts/evaluate_similarity.py", line 58, in <module>
    main(parse_args(sys.argv[1:]))
  File "scripts/evaluate_similarity.py", line 52, in main
    cleanup_urls=args.cleanup_urls, nltk_tokenizer=args.nltk_tokenizer
TypeError: Can't instantiate abstract class LSISimilarity with abstract methods get_distance
```

PR #805, opened 2019-07-29
Override abstract method

ayush-1506:
Fixes : #804 

Overriding the method for now. We need to implement these methods later, I'll open an issue to address this.

### DIFF
--- a/bugbug/similarity.py
+++ b/bugbug/similarity.py
@@ -207,6 +207,9 @@ class LSISimilarity(BaseSimilarity):
         # Get IDs of the k most similar bugs
         return [self.corpus[j[0]][0] for j in sims[:k]]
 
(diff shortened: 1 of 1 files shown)


bugbug/similarity.py
```diff
@@ -207,6 +207,9 @@ def get_similar_bugs(self, query, k=10):
         # Get IDs of the k most similar bugs
         return [self.corpus[j[0]][0] for j in sims[:k]]
 
+    def get_distance(self, query1, query2):
+        pass
```
marco-c: Raise a NotImplemented exception instead of "pass", to specify that the function is not implemented

marco-c: https://docs.python.org/3.7/library/exceptions.html#NotImplementedError

Issue #799, opened 2019-07-26
Try using https://github.com/src-d/wmd-relax package to calculate WMD


marco-c:
It's probably faster than our custom implementation, and we can remove a lot of code.

PR #813, opened 2019-07-31
Use wmd-relax instead of pyemd

ayush-1506:
Fixes: #799


marco-c:
It should also be possible to use wmd-relax without using spacy, e.g. https://github.com/RaRe-Technologies/gensim/blob/develop/docs/notebooks/soft_cosine_tutorial.ipynb.

ayush-1506:
I tried that earlier but got a few errors.
```
    distances = wmd.WMD(embeddings, nbow, vocabulary_min=1).nearest_neighbors("query")
  File "/home/ayush99/GitHub/bbug/lib/python3.7/site-packages/wmd/__init__.py", line 523, in nearest_neighbors
    d = self._estimate_WMD_centroid_batch(avg, i2)
  File "/home/ayush99/GitHub/bbug/lib/python3.7/site-packages/wmd/__init__.py", line 406, in _estimate_WMD_centroid_batch
    avg2 = self._get_centroid_by_index(i2)
  File "/home/ayush99/GitHub/bbug/lib/python3.7/site-packages/wmd/__init__.py", line 402, in _get_centroid_by_index
    words, weights = self._get_vocabulary(index)
  File "/home/ayush99/GitHub/bbug/lib/python3.7/site-packages/wmd/__init__.py", line 370, in _get_vocabulary
    _, words, weights = self.nbow[index]
ValueError: not enough values to unpack (expected 3, got 1)
```

ayush-1506:
Although I'm not sure why this is happening.
`self.nbow[index]` returns a tuple of length 3, so this shouldn't be happening.

ayush-1506:
Oh, turns out it was because of an edge case I'd been overlooking.

ayush-1506:
Results before:
```
Recall @ 1: 15.853658536585366%
Recall @ 5: 16.585365853658537%
Recall @ 10: 16.585365853658537%
Precision @ 1: 13.20754716981132%
Precision @ 5: 3.5579514824797807%
Precision @ 10: 1.8328840970350382%
Recall: 16.585365853658537%
Precision: 6.519654841802493%
MAP@k : 14.80340563898515%
```
After:
```
Recall @ 1: 43.604651162790695%
Recall @ 5: 48.25581395348838%
Recall @ 10: 49.41860465116279%
Precision @ 1: 0.6622516556291391%
Precision @ 5: 9.205298013245011%
Precision @ 10: 5.629139072847673%
Recall: 49.41860465116279%
Precision: 5.629139072847682%
MAP@k : 19.848707032481865%
```

bugbug/similarity.py
```diff
@@ -268,134 +267,95 @@ def __init__(self, cut_off=0.2, cleanup_urls=True, nltk_tokenizer=False):
...
+        query = dict(self.tfidf[self.dictionary.doc2bow(query)])
+        query = [
+            (new_index, query[dict_index])
+            for new_index, dict_index in enumerate(indices)
+            if dict_index in query
+        ]
+        documents = [
+            dict(self.tfidf[self.dictionary.doc2bow(document)])
+            for document in self.corpus
+        ]
```
marco-c: I wonder if this couldn't be done just once in the constructor and cached, it doesn't seem to use any data from the query itself.
How long does `get_similar_bugs` take if you move this out? Is it a measurable improvement?

bugbug/similarity.py
```diff
@@ -268,134 +267,95 @@ def __init__(self, cut_off=0.2, cleanup_urls=True, nltk_tokenizer=False):
...
-                self.w2vmodel.wv.vectors_norm[
-                    [self.w2vmodel.wv.vocab[word].index for word in words]
-                ].transpose(),
-            ),
-            dtype=np.double,
-        )
+        self.dictionary = Dictionary(self.corpus)
+        self.tfidf = TfidfModel(dictionary=self.dictionary)
 
     def get_similar_bugs(self, query):
```
marco-c: It's a pity that the code became so complex :(
I was hoping by switching to r-wmd library we could simplify it, but it seems like we ended up making it even more complex than before :D
Anyway, there is an improvement in the metrics, so I guess it's a good improvement nevertheless.

ayush-1506:
Some more results:
After changes, no cut off:
```
Recall @ 1: 31.463414634146343%
Recall @ 5: 34.146341463414636%
Recall @ 10: 34.390243902439025%
Precision @ 1: 0.5390835579514826%
Precision @ 5: 6.1455525606468875%
Precision @ 10: 3.8005390835579425%
Recall: 34.390243902439025%
Precision: 3.8005390835579513%
MAP@k : 13.606779189663287%
```

ayush-1506:
Before changes:

No cut off :
```
Recall @ 1: 15.853658536585366%
Recall @ 5: 16.82926829268293%
Recall @ 10: 16.82926829268293%
Precision @ 1: 14.016172506738545%
Precision @ 5: 3.6118598382749285%
Precision @ 10: 1.8598382749326121%
Recall: 16.82926829268293%
Precision: 6.641000962463909%
MAP@k : 15.191674153938303%
```

Default cut off :
```
Recall @ 1: 15.609756097560975%
Recall @ 5: 16.585365853658537%
Recall @ 10: 16.585365853658537%
Precision @ 1: 13.477088948787062%
Precision @ 5: 3.5579514824797807%
Precision @ 10: 1.8328840970350382%
Recall: 16.585365853658537%
Precision: 6.507177033492822%
MAP@k : 14.877208745133272%
```

marco-c:
It's pretty strange that the cut-off decreases both recall and precision, only the recall should decrease but the precision should increase (or remain the same). Can you investigate this?

Also, can you compare with the Spacy implementation (the one you had at the beginning). If it is better, we might use it (but using Spacy all the way, not a mix of gensim and Spacy; for the evaluation you can use the mix).

ayush-1506:
@marco Results of spacy + gensim implementation (the one I had earlier) with default cutoff:
```
Recall @ 1: 5.609756097560976%
Recall @ 5: 5.853658536585367%
Recall @ 10: 5.853658536585367%
Precision @ 1: 0.5390835579514826%
Precision @ 5: 1.2938005390835583%
Precision @ 10: 0.6469002695417791%
Recall: 5.853658536585367%
Precision: 5.286343612334802%
MAP@k : 3.2794249775381847%
```

ayush-1506:
Also spacy + gensim is too slow. Took me 3 hours to evaluate results of just 6K bugs.

bugbug/similarity.py
ayush-1506: With the current implementation, it takes 5.4607 seconds to test on the first 10 bugs.
Taking it ti the constructor (making this an attribute) actually increases the time to 5.7524 seconds.

Issue #339, opened 2019-05-03
Get rid of run.py and move its functionality to scripts/


marco-c:
The train functionality is already partly implemented in script/trainer.py.

ayush-1506:
@marco-c How is the script/trainer.py supposed to run? (I see that we can't run it from the command line directly.)

marco-c:
You can run it like this: `python3 -c 'from scripts import trainer; trainer.main()'`.

This issue is a bit problematic, as we need to make a few decisions before moving stuff, so let's not do it yet.

marco-c:
We can now do this, so I've added back the 'good-first-bug' label.

Lothiraldan:
You can also launch it with `bugbug-train`

asmaknikar:
can I claim this issue?

asmaknikar:
can I claim this issue?

marco-c:
@asmaknikar you can work on any issue which is unassigned and has no open PR linked to it (read CONTRIBUTING.md).

Abhi-khandelwal:
Hi @marco-c , So I was kind of interesting to take this issue, Actually I have one doubt, that do we have to run this functionality by some Bash script or by direct script/trainer.py ?

marco-c:
We need to split the run.py functionality into multiple scripts under script/.
(Part of) The train functionality is already in trainer.py.

cklyyung:
Seems like args.classify and args.generate_sheet could be split into scripts called bug_classifier.py and generate_sheet.py (or some other names along those lines). However, both of these require a trained model. Do you want these scripts to call the trainer.py script to train the models if the file can't be loaded? Or should the user be informed that the model should be trained beforehand by running trainer.py?

marco-c:
> Or should the user be informed that the model should be trained beforehand by running trainer.py?

This one!

cklyyung:
I think this will be easier to review if it's broken into multiple PRs. I plan to divide this into:

- [ ] generate_sheet.py

- [ ] bug_classifier.py

- [ ] add extra functionality to trainer.py

- [ ] remove run.py and update readme

I can do it in one PR if you'd prefer though.

marco-c:
Yeah, breaking into multiple PRs is the best option!

PR #816, opened 2019-08-01
Separate generate_sheet from run.py

cklyyung:
Addresses #339
I moved the code to load the model and everything under `if args.generate_sheet:` from run.py into this script.
I was able to run the command with 
`python3 -c 'from scripts import generate_sheet; generate_sheet.main()' model token`.

Here's an example of the csv that was generated.
image


marco-c:
Can you remove the original generate_sheet code from run.py?

Issue #792, opened 2019-07-26
Restrict the training set of the StepsToReproduce model only to defects


marco-c:
Given that STRs don't apply to enhancement or task.

sosa-e:
Hello! I am new to open source and am currently working on this bug after following the steps outlined in #1092 . I read through #817 and have some questions. May I post questions in this issue?

suhaibmujahid:
> Hello! I am new to open source and am currently working on this bug after following the steps outlined in #1092 . I read through #817 and have some questions. May I post questions in this issue?

Welcome, @sosa-e! You can post questions related to the issues here. Also, you can chat with us in the [bugbug](https://chat.mozilla.org/#/room/#bugbug:mozilla.org) Matrix room.

naoya2000:
Turns out that there is an issue; I tried to use the model to classify a certain bug but it didn't work. However, it seems like that there are some steps that may help in the [Discussion](https://github.com/mozilla/bugbug/discussions/3338) so I will try those out first. Thanks!
`
(base) naoyaokamoto@Naoyas-MacBook-Air-393 bugbug % python3 -m scripts.bug_classifier stepstoreproduce --bug-id 1635837                                        
2023-12-03 18:38:17,677:INFO:numexpr.utils:NumExpr defaulting to 8 threads.
2023-12-03 18:38:19,688:INFO:__main__:stepstoreproducemodel does not exist. Downloading the model....
2023-12-03 18:38:19,703:INFO:bugbug.utils:Downloading https://community-tc.services.mozilla.com/api/index/v1/task/project.bugbug.train_stepstoreproduce.latest/artifacts/public/stepstoreproducemodel.tar.zst...
tar: Error inclusion pattern: Failed to open 'zstdmt'
2023-12-03 18:38:21,680:WARNING:bugbug.utils:Command '['tar', '-I', 'zstdmt', '-xf', 'stepstoreproducemodel.tar.zst']' returned non-zero exit status 1.. Falling back to zstandard API, which could be slower.
Traceback (most recent call last):
  File "/Users/naoyaokamoto/mambaforge/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/Users/naoyaokamoto/mambaforge/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
[...]

PR #817, opened 2019-08-01
Restrict training set of stepstoreproduce model only to defects

chidauri:
Regarding #792

### DIFF
--- a/bugbug/models/stepstoreproduce.py
+++ b/bugbug/models/stepstoreproduce.py
@@ -67,7 +67,7 @@ class StepsToReproduceModel(BugModel):
         classes = {}
 
         for bug_data in bugzilla.get_bugs():
-            if (bug_data["id"], 1) in defects.items():
+            if bug_data["id"] in defects:
                 if "cf_has_str" in bug_data:
                     if bug_data["cf_has_str"] == "no":
                         classes[int(bug_data["id"])] = 0
--- a/bugbug/models/stepstoreproduce.py
+++ b/bugbug/models/stepstoreproduce.py
@@ -65,7 +65,7 @@ class StepsToReproduceModel(BugModel):
         classes = {}
 
         for bug_data in bugzilla.get_bugs():
-            if "type" in bug_data and bug_data["type"] != "defect":
+            if bug_data["type"] != "defect":
                 continue
             if "cf_has_str" in bug_data:
                 if bug_data["cf_has_str"] == "no":
--- a/bugbug/models/stepstoreproduce.py
+++ b/bugbug/models/stepstoreproduce.py
@@ -11,7 +11,6 @@ from sklearn.pipeline import Pipeline
 
 from bugbug import bug_features, bugzilla, feature_cleanup
 from bugbug.model import BugModel
-from bugbug.models.defect import DefectModel
 
 
 class StepsToReproduceModel(BugModel):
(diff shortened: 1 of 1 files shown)


chidauri:
**After this change**

> 13 bugs have no steps to reproduce
1438 bugs have steps to reproduce
X: (1451, 50719), y: (1451,)
Cross Validation scores:
Accuracy: f0.5418039385083343 (+/- 0.07154716421783254)
Precision: f0.9926759918094075 (+/- 0.010011230145725573)
Recall: f0.5413815809164646 (+/- 0.06940490718276383)
X_train: (24, 50719), y_train: (24,)
X_test: (146, 50719), y_test: (146,)

**Before this change**

> 73 bugs have no steps to reproduce
2642 bugs have steps to reproduce
X: (2715, 68671), y: (2715,)
Cross Validation scores:
Accuracy: f0.6966856048676635 (+/- 0.09750364886404748)
Precision: f0.9890742501710503 (+/- 0.01088263716211331)
Recall: f0.6965130473241927 (+/- 0.10856018472515121)
X_train: (128, 68671), y_train: (128,)
X_test: (272, 68671), y_test: (272,)

marco-c:
Could you show here the confusion matrices too?

bugbug/models/stepstoreproduce.py
```diff
@@ -62,21 +63,23 @@ def __init__(self, lemmatization=False):
         self.clf.set_params(predictor="cpu_predictor")
 
     def get_labels(self):
+        defects = DefectModel.get_bugbug_labels(self, "bug")
         classes = {}
 
         for bug_data in bugzilla.get_bugs():
-            if "cf_has_str" in bug_data:
-                if bug_data["cf_has_str"] == "no":
+            if bug_data["id"] in defects:
```
marco-c: Rewrite this as:
```Python
if bug_data["id"] not in defects:
    continue
```

So you don't need to increase the indentation for the rest of the code.

bugbug/models/stepstoreproduce.py
```diff
@@ -62,21 +63,23 @@ def __init__(self, lemmatization=False):
         self.clf.set_params(predictor="cpu_predictor")
 
     def get_labels(self):
+        defects = DefectModel.get_bugbug_labels(self, "bug")
```
marco-c: I think you might be able to just use the `bug_type` field, checking `bug["bug_type"] == "defect"` instead of `bug["id"] in defects`.
Can you check how many bugs you'd have in the training set in the first vs the second case?

chidauri: Do i need to train defect model first? because there wasn't a single bug with 'bug_type' field when i ran this.

marco-c: Sorry, the field name is actually "type", not "bug_type".

marco-c: (And make sure you have the latest bugs DB)

chidauri: It worked! The dataset actually increased a bit, the comment above contains the current result.

marco-c:
> Could you show here the confusion matrices too?

The `before` ones too.

marco-c:
BTW, there isn't much change, but this could make a difference when we have PULearning, so I'll keep it open and we'll re-evaluate after that.

chidauri:
> > Could you show here the confusion matrices too?
> 
> The `before` ones too.

Done

PR #819, opened 2019-08-01
Also check for job timeout when checking if a job is running

Lothiraldan:
It seems some RQ job can be stuck and blocks classification of single bug ids.
If a job timeout is crossed and the job is not running, clean it so it can be
requeued. Also bump the job timeout to 30 minutes.

### DIFF
--- a/http_service/app.py
+++ b/http_service/app.py
@@ -7,6 +7,7 @@ import json
 import logging
 import os
 import uuid
+from datetime import datetime, timedelta
 
 from apispec import APISpec
 from apispec.ext.marshmallow import MarshmallowPlugin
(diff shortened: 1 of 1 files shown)


Lothiraldan:
> Should we have a shorter timeout?

We can but we need to make sure the jobs duration is lesser than the timeout. I am not sure if RQ is actually enforcing the job timeout by itself or not. This shouldn't be a problem right now as we check the job timeout only for enqueued but not started jobs.

That makes me think that if a job starts but doesn't finished or isn't cleaned properly, we might hit a deadlock once again. But if it's the case, we should probably reevaluate RQ reliability.

PR #820, opened 2019-08-01
Swap xp & transform to be able to move file_stuff in _transform

### LINKED ISSUE #422: Don't get file_copies from hg log
It's pretty slow.
We need it for experience calculation, so we could move this step after the parse patch step.

calixteman:
Fixes #422.

### DIFF
--- a/bugbug/repository.py
+++ b/bugbug/repository.py
@@ -55,8 +55,6 @@ class Commit:
         backedoutby,
         author_email,
         reviewers,
-        files=[],
-        file_copies={},
         ignored=False,
     ):
         self.node = node
@@ -68,9 +66,6 @@ class Commit:
         self.backedoutby = backedoutby
         self.ever_backedout = backedoutby != ""
         self.author_email = author_email
-        if files:
-            self.set_files(files)
-        self.file_copies = file_copies
         self.reviewers = reviewers
         self.ignored = ignored
         self.added = 0
@@ -92,12 +87,14 @@ class Commit:
     def __hash__(self):
         return hash(self.node)
 
-    def set_files(self, files):
+    def set_files(self, files, file_copies):
         self.files = files
+        self.file_copies = file_copies
         self.components = list(
             set(path_to_component[path] for path in files if path in path_to_component)
         )
         self.directories = get_directories(files)
+        return self
 
     def set_experience(
         self, exp_type, commit_type, timespan, exp_sum, exp_max, exp_min
@@ -223,7 +220,7 @@ def hg_modified_files(hg, commit):
     x = hg.rawcommand(args)
     files_str, file_copies_str = x.split(b"\x00")[:-1]
 
-    commit.file_copies = f_copies = {}
+    file_copies = {}
     for file_copy in file_copies_str.decode("utf-8").split("|"):
         if not file_copy:
             continue
@@ -231,9 +228,11 @@ def hg_modified_files(hg, commit):
         parts = file_copy.split(" (")
         copied = parts[0]
         orig = parts[1][:-1]
-        f_copies[sys.intern(orig)] = sys.intern(copied)
+        file_copies[sys.intern(orig)] = sys.intern(copied)
 
-    commit.set_files([sys.intern(f) for f in files_str.decode("utf-8").split("|")])
+    commit.set_files(
+        [sys.intern(f) for f in files_str.decode("utf-8").split("|")], file_copies
+    )
 
 
 def _transform(commit):
@@ -331,7 +330,7 @@ def hg_log(hg, revs):
         else:
             pushdate = datetime.utcnow()
 
-        bug = rev[4]
+        bug_id = int(rev[4].decode("ascii")) if rev[4] else None
 
         revs.append(
             Commit(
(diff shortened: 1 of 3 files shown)


bugbug/repository.py
```diff
@@ -51,24 +51,39 @@ def __init__(
...
     ):
         self.node = node
         self.author = author
+        self.bug_id = bug_id
         self.desc = desc
         self.date = date
         self.pushdate = pushdate
-        self.bug = bug
         self.backedoutby = backedoutby
+        self.ever_backedout = backedoutby != ""
```
marco-c: Nit: you could make this a property function

bugbug/repository.py
```diff
@@ -51,24 +51,39 @@ def __init__(
...
         self.author_email = author_email
-        self.files = files
+        if files:
```
marco-c: Do we have this parameter only for tests? Could you remove it and call `set_files` in tests directly?

bugbug/repository.py
```diff
@@ -324,11 +340,9 @@ def hg_log(hg, revs):
                 desc=rev[2].decode("utf-8"),
                 date=date,
                 pushdate=pushdate,
-                bug=rev[4],
+                bug_id=int(bug.decode("ascii")) if bug else None,
```
marco-c: Nit: move this directly when you define the `bug` variable

bugbug/repository.py
```diff
@@ -653,10 +659,16 @@ def hg_log_multi(repo_dir, revs):
     os.chdir(repo_dir)
 
     CHUNK_SIZE = 256
-    revs_groups = [revs[i : (i + CHUNK_SIZE)] for i in range(0, len(revs), CHUNK_SIZE)]
+    processes = os.cpu_count() + 1
```
marco-c: Nit: threads_num (since it's threads and not processes)

bugbug/repository.py
```diff
@@ -653,10 +659,16 @@ def hg_log_multi(repo_dir, revs):
...
+    CHUNK_SIZE = int(math.ceil(len(revs) / processes))
+    REVS_COUNT = len(revs)
+    revs_groups = [
+        (revs[i], revs[min(i + CHUNK_SIZE, REVS_COUNT) - 1])
+        for i in range(0, REVS_COUNT, CHUNK_SIZE)
+    ]
```
marco-c: Is this change in particular improving performance?

calixteman: Yeah, it isn't crazy but it has an impact

bugbug/repository.py
```diff
@@ -709,14 +719,20 @@ def download_commits(repo_dir, rev_start=0, ret=False, save=True):
         commits = executor.map(_transform, commits, chunksize=64)
         commits = tqdm(commits, total=commits_num)
 
-        if ret:
-            commits = list(commits)
+    commits = list(commits)
```
marco-c: Shouldn't this be under the `with`?

bugbug/repository.py
```diff
@@ -709,14 +719,20 @@ def download_commits(repo_dir, rev_start=0, ret=False, save=True):
...
 
-        if save:
-            db.append(COMMITS_DB, commits)
+    calculate_experiences(commits, first_pushdate, save)
+
+    if ret:
```
marco-c: Since you're collecting commits to `list` anyway, I'd remove `ret` completely and consider it as always True

bugbug/repository.py
```diff
@@ -691,10 +703,8 @@ def download_commits(repo_dir, rev_start=0, ret=False, save=True):
     commits_to_ignore = get_commits_to_ignore(repo_dir, commits)
     print(f"{len(commits_to_ignore)} commits to ignore")
 
-    calculate_experiences(commits, commits_to_ignore, first_pushdate, save)
-
-    # Exclude commits to ignore.
-    commits = [commit for commit in commits if commit not in commits_to_ignore]
+    for commit in commits:
+        commit.ignored = commit in commits_to_ignore
```
marco-c: If you're not using commits_to_ignore anymore, maybe you can set this directly in get_commits_to_ignore

tests/test_repository.py
```diff
@@ -119,6 +119,55 @@ def test_get_revs(fake_hg_repo):
...
+    hg.push(dest=bytes(remote, "ascii"))
+    revs = repository.get_revs(hg, revision1)
+    commits = repository.hg_log(hg, revs)
+
+    for c in commits:
+        repository.hg_modified_files(hg, c)
+
+    print((commits[2].node, revision1))
+    print(commits[2].files)
+    print(commits[2].file_copies)
```
marco-c: Nit: remove the pring statements

bugbug/repository.py
calixteman: I thought about that but since we use this value and need it at the end I thought it was "better" to have it instead of computing it again and again.
But it's as you wish.

bugbug/repository.py
marco-c: I'm asking because we have the same code in `microannotate` :)

marco-c:
@calixteman I forgot scripts/regressor_finder.py needs to be updated too (just the `get_commits_to_ignore` function).

Issue #821, opened 2019-08-01
Add configuration parameters to generate_sheet script


marco-c:
We should make a couple of things configurable:
- the time span (currently it's one week);
- the confidence threshold (currently it's 0.7).

WaxXwing:
Hey @marco-c!
This is my first time in open source and I would really like to contribute to this. 
I guess passing the time span and threshold as arguments will do?

marco-c:
Yes!

WaxXwing:
So what next?

WaxXwing:
Hey @marco-c ,
I wanted to know what might be the range values for the time span?
Will it always be in days?

marco-c:
Yeah, I think we can make it an int which specifies the number of days.

WaxXwing:
Sorry if this is silly, but I can't seem to push my changes to the remote repo

marco-c:
@WaxXwing you'll need to push your changes to your fork of the repo. I suggest reading a guide on how to use git.

PR #827, opened 2019-08-01
added pramaters for time span and confidence threshold

WaxXwing:
For #821

### DIFF
--- a/scripts/generate_sheet.py
+++ b/scripts/generate_sheet.py
@@ -9,7 +9,7 @@ from bugbug import bugzilla
 from bugbug.models import get_model_class
 
 
-def generate_sheet(model_name, token):
+def generate_sheet(model_name, token, span, threshold):
     model_file_name = f"{model_name}model"
 
     assert os.path.exists(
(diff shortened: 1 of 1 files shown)


scripts/generate_sheet.py
```diff
@@ -53,12 +53,18 @@ def generate_sheet(model_name, token):
...
 
 def main():
-    description = "Perform evaluation on bugs from last week on the specified model and generate a csv file "
+    description = "Perform evaluation on bugs from specified days back on the specified model and generate a csv file "
     parser = argparse.ArgumentParser(description=description)
 
     parser.add_argument("model", help="Which model to generate a csv for.")
     parser.add_argument("token", help="Bugzilla token")
+    parser.add_argument(
+        "span", type=int, help="No. of days back from which bugs will be evaluated"
```
marco-c: Let's use `days` as a name instead of `span`

scripts/generate_sheet.py
```diff
@@ -9,7 +9,7 @@
 from bugbug.models import get_model_class
 
 
-def generate_sheet(model_name, token):
+def generate_sheet(model_name, token, span, threshold):
```
marco-c: And let's call the variable days instead of span.

WaxXwing: ok, making all the `span` s into `days`

scripts/generate_sheet.py
```diff
@@ -53,12 +53,18 @@ def generate_sheet(model_name, token):
...
     parser = argparse.ArgumentParser(description=description)
 
     parser.add_argument("model", help="Which model to generate a csv for.")
     parser.add_argument("token", help="Bugzilla token")
+    parser.add_argument(
+        "days", type=int, help="No. of days back from which bugs will be evaluated"
+    )
+    parser.add_argument(
+        "threshold", type=float, help="Confidence threshold for the model"
+    )
```
marco-c: Forgot to tell you before: let's use 7 as default for the days, and 0.7 as default for the threshold (to keep the same behavior as before if the arguments are not passed).

PR #829, opened 2019-08-01
Seperate bug_classifier.py from run.py

cklyyung:
Addresses #339 
Moved code from under `args.classify` in run.py into this script. Also, I removed the model_loading code since it was no longer being used in run.py. 
I was able to run this script with the command;
`python3 -c 'from scripts import bug_classifier; bug_classifier.main()' model`

Example output:
```https://bugzilla.mozilla.org/show_bug.cgi?id=1322441 - The Internet Download Manager add-on can seriously slow down Firefox with e10s 
Top 11 features:
+-----------+----------------------------+---------------------------------+--------------------------+-----------------------------------------------------------------+-------------------------+
|   classes |   Comments contain 'fixed' |   Comments contain 'regression' |   Comments contain 'str' |   status has ever been set to 'affected' and 'unaffected'=False |   Comments contain '52' |
+===========+============================+=================================+==========================+=================================================================+=========================+
|         1 |                   0.139018 |                       -0.104707 |                0.0750531 |                                                      -0.0537768 |                0.052669 |
+-----------+----------------------------+---------------------------------+--------------------------+-----------------------------------------------------------------+-------------------------+

+------------------------------+---------------------------+-------------------------------+--------------------------+--------------------------------+---------------------------+
|   Comments contain 'support' |   Comments contain 'with' |   First comment contains 'we' |   Comments contain 'per' |   First comment contains 'not' |   Comments contain 'some' |
+==============================+===========================+===============================+==========================+================================+===========================+
|                    0.0516413 |                 0.0378344 |                     0.0293846 |               -0.0273952 |                      0.0267647 |                -0.0223058 |
+------------------------------+---------------------------+-------------------------------+--------------------------+--------------------------------+---------------------------+

Positive! [[0.19689035 0.80310965]]
```

### DIFF
--- a/scripts/bug_classifier.py
+++ b/scripts/bug_classifier.py
@@ -8,11 +8,15 @@ import numpy as np
 from bugbug import bugzilla
 from bugbug.models import get_model_class
 
+MODELS_WITH_TYPE = ("component",)
+
 
 def classify_bugs(model_name, classifier):
-    classifier_models = ["component"]
+    if classifier != "default":
+        assert (
+            model_name in MODELS_WITH_TYPE
+        ), f"{classifier} is not a valid classifier type for {model_name}"
 
-    if classifier != "default" and model_name in classifier_models:
         model_file_name = f"{model_name}{classifier}model"
         model_name = f"{model_name}_{classifier}"
     else:
--- a/scripts/bug_classifier.py
+++ b/scripts/bug_classifier.py
@@ -1,0 +1,55 @@
+# -*- coding: utf-8 -*-
+
+import argparse
+import os
+
+import numpy as np
+
+from bugbug import bugzilla
+from bugbug.models import get_model_class
+
+
+def classify_bugs(model_name):
+    model_file_name = f"{model_name}model"
+
+    assert os.path.exists(
+        model_file_name
+    ), f"{model_file_name} does not exist. Train the model with trainer.py first."
+
+    model_class = get_model_class(model_name)
+    model = model_class.load(model_file_name)
+
+    for bug in bugzilla.get_bugs():
+        print(
+            f'https://bugzilla.mozilla.org/show_bug.cgi?id={bug["id"]} - {bug["summary"]} '
+        )
+
+        if model.calculate_importance:
+            probas, importance = model.classify(
+                bug, probabilities=True, importances=True
+            )
+
+            feature_names = model.get_human_readable_feature_names()
+
+            model.print_feature_importances(
+                importance["importances"], feature_names, class_probabilities=probas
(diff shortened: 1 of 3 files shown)


scripts/bug_classifier.py
```diff
@@ -0,0 +1,55 @@
...
+import os
+
+import numpy as np
+
+from bugbug import bugzilla
+from bugbug.models import get_model_class
+
+
+def classify_bugs(model_name):
+    model_file_name = f"{model_name}model"
```
marco-c: In the previous code, we were sometimes adding the classifier to the name, we should do the same here.

cklyyung: The classifier type only applies to the component_nn right? Can't the user just pass in component_nn as the model, instead of component --classifier nn?

marco-c: > The classifier type only applies to the component_nn right?

For now, yes.

> Can't the user just pass in component_nn as the model, instead of component --classifier nn?

They could, but it's less clear for them and there's no advantage for us in doing that.

scripts/bug_classifier.py
```diff
@@ -0,0 +1,67 @@
...
+import os
+
+import numpy as np
+
+from bugbug import bugzilla
+from bugbug.models import get_model_class
+
+
+def classify_bugs(model_name, classifier):
+    classifier_models = ["component"]
```
cklyyung: My idea here is that this will make it easier to add more models that have a classifier argument. I can change it to what was originally in run.py if you prefer.

setup.py
```diff
@@ -49,6 +49,7 @@ def read_requirements(file_):
             "bugbug-check = scripts.check:main",
             "bugbug-microannotate-generate = scripts.microannotate_generator:main",
             "bugbug-classify-commit = scripts.commit_classifier:main",
+            "bugbug-classify-bug = scripts.bug_classifier:main",
```
cklyyung: I saw that there's another PR to add the scripts to setup.py; I added it here, but I can remove it if this is the wrong PR to do this.

scripts/bug_classifier.py
marco-c: No, that's a good idea. My only suggestion:
1) If classifier != "default", assert model_name is in this list, to make sure the parameters passed by the user are valid
2) Rename the variable, as `classifier_models` is too generic (maybe `MODELS_WITH_TYPE`, and make it a constant (UPPERCASE and at the top of the file after the imports) tuple.

setup.py
marco-c: No, this is the right PR ;)

PR #836, opened 2019-08-02
Check metrics evolution

Lothiraldan:


.taskcluster.yml
```diff
@@ -136,6 +136,36 @@ tasks:
           owner: mcastelluccio@mozilla.com
           source: ${repository}/raw/${head_rev}/.taskcluster.yml
 
+      - taskId: {$eval: as_slugid("check_metrics")}
```
marco-c: I'd put this task at the end of the data pipeline, so we generate new graphs whenever we train the models.

scripts/check_all_metrics.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
marco-c: We could move the script into scripts/analyze_training_metrics.py

scripts/check_all_metrics.sh
```diff
@@ -0,0 +1,16 @@
...
+set -euox pipefail
+
+all_metrics="backout component defectenhancementtask duplicate regression regressionrange regressor stepstoreproduce tracking"
```
marco-c: We should try to avoid hardcoding the models, maybe we could set the training tasks as dependencies of the "check_metrics" task, and the "check_metrics" task can get the metrics from its dependencies.

.taskcluster.yml
Lothiraldan: Got it. I put it temporarily in this file so I can iron out the taskcluster running with artifacts declaration. I will put it in the data pipeline once it's working properly.

scripts/check_all_metrics.sh
Lothiraldan: In order to do that, it would be better if we moved the content of the `retrieve_training_metrics` in bugbug itself (something like `bugbug/metrics.py`) so we won't have to mangle with the Python path, would you agree?

scripts/check_all_metrics.sh
Lothiraldan: As long as the naming scheme for the training tasks doesn't change, it should holds. I will work on that

scripts/check_all_metrics.sh
marco-c: Yeah, but I'd like to keep the Taskcluster dependency out of bugbug core code.
Maybe we can have a TaskIndex base class, a TaskclusterTaskIndex extending the base class, and allow users to use other services if they want (implementing another extension of TaskIndex).
Anyway, we can keep the script for now and file an issue about this.

infra/data-pipeline.yml
```diff
@@ -444,6 +444,46 @@ tasks:
...
+        maxRunTime: 3600
+        image: python
+        command:
+          - "/bin/bash"
+          - "-lcx"
+          - "git clone --quiet https://github.com/mozilla/bugbug /code &&
+             cd /code &&
+             git checkout ${version} &&
+             pip install --quiet . &&
+             pip install --quiet -r test-requirements.txt &&
```
marco-c: Can we use the bugbug base image instead of cloning bugbug and installing requirements?

scripts/analyze_training_metrics.py
```diff
@@ -0,0 +1,168 @@
...
+) -> bool:
+    sorted_metrics = sorted(values_dict.items())
+    x, y = zip(*sorted_metrics)
+
+    # Compute the threshold
+    if len(y) >= 2:
+        before_last_value = y[-2]
+    else:
+        before_last_value = y[-1]
+    ninety_five_percent = before_last_value * 0.95
```
marco-c: Nit: define the 0.95 in a constant at the top, and rename `ninety_five_percent` to use a generic name

scripts/analyze_training_metrics.py
```diff
@@ -0,0 +1,168 @@
...
+    for single_x, single_y in zip(x, y):
+        label = "{:.4f}".format(single_y)
+
+        plt.annotate(
+            label,  # this is the text
+            (single_x, single_y),  # this is the point to label
+            textcoords="offset points",  # how to position the text
+            xytext=(0, 10),  # distance from text to points (x,y)
+            ha="center",
+        )  # horizontal alignment can be left, right or center
```
marco-c: Nit: I'd remove this comment

scripts/analyze_training_metrics.py
```diff
@@ -0,0 +1,168 @@
...
+
+    axes.plot_date(x, y, marker=".", fmt="-")
+    # plt.show(block=True)
```
marco-c: Nit: remove commented out code

scripts/analyze_training_metrics.py
```diff
@@ -0,0 +1,168 @@
...
+    # Check if the threshold has been crossed
+    return y[-1] < ninety_five_percent
+
+
+def analyze_metrics(metrics_directory: str, output_directory: str):
+    root = Path(metrics_directory)
+
+    metrics = defaultdict(
+        lambda: defaultdict(dict)
+    )  # type: Dict[str, Dict[str, Dict[datetime, float]]]
```
marco-c: Can't the type annotation be actually used here? Or is the defaultdict preventing that?

scripts/analyze_training_metrics.py
```diff
@@ -0,0 +1,168 @@
...
+        date = datetime(
+            date_parts[0],
+            date_parts[1],
+            date_parts[2],
+            date_parts[3],
+            date_parts[4],
+            date_parts[5],
+            tzinfo=timezone.utc,
+        )
+        version = file_path_parts[14:]  # noqa TODO: Use version
```
marco-c: Why the noqa?

scripts/analyze_training_metrics.py
```diff
@@ -0,0 +1,168 @@
...
+
+        # Then process the report
```
marco-c: Let's do the plotting and the checking only for accuracy, precision and recall for now (list them in a constant at the top).

scripts/analyze_training_metrics.py
```diff
@@ -0,0 +1,168 @@
...
+                )
+
+                threshold_ever_crossed = threshold_ever_crossed or threshold_crossed
+
+    if threshold_ever_crossed:
+        sys.exit(1)
+
+
+def main():
+    description = "Retrieve a model training metrics"
```
marco-c: Update desc

scripts/check_all_metrics.py
```diff
@@ -0,0 +1,122 @@
...
+    download_script_path = CURRENT_DIR / "retrieve_training_metrics.py"
+
+    cli_args = [
+        sys.executable,
+        str(download_script_path),
+        model_name,
+        "2019",
+        "-d",
+        metric_directory,
+    ]  # type: List[str]
```
marco-c: Can we actually use the type annotation?

scripts/check_all_metrics.py
```diff
@@ -0,0 +1,122 @@
...
+
+    LOGGER.info("Download metrics for %r", model_name)
+
+    subprocess.check_output(cli_args)
```
marco-c: Nit: subprocess.run(cli_args, check=True)

scripts/check_all_metrics.py
```diff
@@ -0,0 +1,122 @@
...
+            "Loading task dependencies {}/{} {}".format(
+                i + 1, len(task["dependencies"]), task_id
+            )
+        )
+
+        dependency_task = queue.task(task_id)
+
+        # Check dependency route to see if we should download it
+        for route in dependency_task["routes"]:
+            if fnmatch(route, QUEUE_ROUTE_PATTERN):
```
marco-c: Let's assert here, so we notice if somebody adds a bad dependency

scripts/analyze_training_metrics.py
Lothiraldan: Because it is not used yet and flake8 is emitting a warning

scripts/analyze_training_metrics.py
Lothiraldan: Including or not the `test_*` metrics?

scripts/check_all_metrics.py
Lothiraldan: So you mean raising an exception if a dependency has no matching route? That would mean that we cannot add a non training task dependency to the checking job.

scripts/analyze_training_metrics.py
marco-c: Ah, then I'd comment the whole line

marco-c: So we can avoid the noqa

scripts/analyze_training_metrics.py
marco-c: Including

scripts/check_all_metrics.py
marco-c: Yeah, I guess that should be OK, no?

Lothiraldan: We just need to remember to remove the exception raising if ever we add a non-training dependency, I will add a comment to remember.

scripts/check_all_metrics.py
Lothiraldan: Do you prefer not to see the process output?

scripts/analyze_training_metrics.py
Lothiraldan: Done, but the report doesn't seems to contains the accuracy right now, need to check why

infra/data-pipeline.yml
Lothiraldan: It should works but we may not ship the scripts directory in a very friendly manner. We could also add the three scripts as console scripts but we already have 7 console scripts this might be a bit too much.

marco-c: Yeah, we have many... But I guess it's fine, I wouldn't worry too much about that

scripts/analyze_training_metrics.py
marco-c: Yeah the report doesn't (from the report we just need `precision` and `recall`), the accuracy is just for `test_*`.

scripts/analyze_training_metrics.py
```diff
@@ -0,0 +1,177 @@
...
+from typing import Dict
+
+import matplotlib.dates as mdates
+import matplotlib.pyplot as plt
+
+LOGGER = logging.getLogger(__name__)
+
+logging.basicConfig(level=logging.INFO)
+
+# If the latest metric point is 5% lower then the one before, show a warning and exists with 1.
```
marco-c:
```suggestion
# If the latest metric point is 5% lower than the previous one, show a warning and exit with 1.
```

scripts/check_all_metrics.py
marco-c: `subprocess.run(cli_args, check=True)` shows the output, `subprocess.check_output(cli_args)` doesn't.

Lothiraldan: Got it, it is catching stdout but not stderr and most output how the other scripts are on stderr, I will make the change.

PR #838, opened 2019-08-02
Differentiate file types and sizes according to whether they are source code or tests

marco-c:


bugbug/repository.py
```diff
@@ -241,24 +241,47 @@ def _transform(commit):
...
     commit.average_file_size = (
         commit.total_file_size / len(sizes) if len(sizes) > 0 else 0
     )
     commit.maximum_file_size = max(sizes, default=0)
     commit.minimum_file_size = min(sizes, default=0)
 
-    commit.files_modified_num = len(patch_data)
+    commit.total_test_file_size = sum(test_sizes)
+    commit.average_test_file_size = (
+        commit.total_test_file_size / len(test_sizes) if len(test_sizes) > 0 else 0
```
calixteman: nit: we should have a function for that stuff since we've this pattern several times

bugbug/repository.py
```diff
@@ -241,24 +241,47 @@ def _transform(commit):
...
+                test_sizes.append(size)
+        else:
+            commit.files_modified_num += 1
+
+            commit.added += stats["added_lines"]
+            commit.deleted += stats["deleted_lines"]
+
+            if size is not None:
+                sizes.append(size)
+            commit.types.add(type_)
```
calixteman: Is it on purpose to not add type for a test ?

marco-c: Yeah, I initially added it, but then I removed it as almost all tests are JS, so I think there isn't much info in this.

calixteman: ok, just add a comment to not forget why we don't add them

marco-c: Done.

PR #845, opened 2019-08-03
Remove 'feature_names' parameter from model.print_feature_importances

### LINKED ISSUE #834: Remove `feature_names` parameter from model.print_feature_importances
The `print_feature_importances` function can simply access the feature names of the model directly (self.get_human_readable_feature_names()), so the `feature_names` argument is not needed.

zephyrmoth:
Fixes #834

### DIFF
--- a/bugbug/model.py
+++ b/bugbug/model.py
@@ -229,7 +229,7 @@ class Model:
         return important_features
 
     def print_feature_importances(
-        self, important_features, feature_names, class_probabilities=None
+        self, important_features, class_probabilities=None
     ):
         # extract importance values from the top features for the predicted class
         # when classifying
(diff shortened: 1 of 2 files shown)


chidauri:
@zephyrmoth can you post the results before/after? It'll be useful.

scripts/bug_classifier.py
```diff
@@ -42,7 +42,7 @@ def classify_bugs(model_name, classifier):
             feature_names = model.get_human_readable_feature_names()
```
marco-c: You're not using this anymore, so you should remove it

bugbug/model.py
```diff
@@ -239,7 +239,7 @@ def print_feature_importances(
                 predicted_class_index = class_probabilities.argmax(axis=-1)[0]
             else:
                 predicted_class_index = 0
-
+            feature_names = self.get_human_readable_feature_names()
```
marco-c: You are defining this in the `if` branch, which means it won't be available in the `else` where it is used...

marco-c:
Closing PR as it's stale.

Issue #769, opened 2019-07-22
Close connections to the hg server once the process pool is terminated


marco-c:
Similar to ec6ffeb38f6778f95f5a76ebc2ba331d27cf5cac, but for the process pool.

harsh020:
Hey, as there are no PRs open for this issue can I take it? If yes, I need a bit help as this will be my first open source contribution. (I'm reading the contribution docs). Just a question, the clear function (@ line 757), tends to open the hg server, we need to close it there?
The clone function (@ line 776), is it building the repo if not already existing, so we also need to close the hg server there also?

Thanks.

harsh020:
@marco-c Hey, as there is no PR for this issue I would like to take it.
As this is my very first open source contribution I would be very grateful if you helped me out. I went through the file repository.py but still am not clear where should I close the hg server, is it in the clear function at line 757?

marco-c:
Yes, it's 2.
In 1 we are already closing the hg server, as we are using a context manager https://github.com/mozilla/bugbug/blob/17ff6ed31569fa893e13b8a1140bd1f458765a70/bugbug/repository.py#L758.

@harsh020 note that this is a good-first-issue as it is self-contained, but it is not a good-first-issue if you don't have experience with Python and knowledge of how multiprocessing works. Feel free to propose a PR if you manage to fix it, or to switch to some other good-first-issue if this one is too hard.

> I think it would be the second one, if so, should I use the global variable and append it every time , like in the link you provided and close all the servers before returning the commits?

If you use a global variable, each process will have its own copy, so this won't work.

harsh020:
> In 1 we are already closing the hg server, as we are using a context manager
@marco-c Sorry, I did not notice that.

> if you don't have experience with Python and knowledge of how multiprocessing works. 
Well, I've been working with python for about two years (yup, not too much experience) and I'll read more about multiprocessing, not a problem.

> If you use a global variable, each process will have its own copy, so this won't work.
I'm not thinking to declare a global variable, but the `_init` function itself has a `global HG`, should I simply append them in a list and close all of the servers before returning the commits?

marco-c:
> yup, not much experience, but I wanna learn, and I'll read more about multiprocessing, not a problem.

OK! Note: for `multiprocessing` I'm not just talking about the `multiprocessing` module, but concepts about working with multiple processes in general, if you don't know them you'll need to read about them too.

[...]

PR #846, opened 2019-08-03
Close mercurial servers in process pool

dmitrypolo:
This addresses #769

### DIFF
--- a/bugbug/repository.py
+++ b/bugbug/repository.py
@@ -10,6 +10,7 @@ import itertools
 import json
 import logging
 import math
+import multiprocessing
 import os
 import pickle
 import sys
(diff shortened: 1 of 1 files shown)


dmitrypolo:
@marco-c just to clarify, since I have never used the `initializer` arg of the `ProcessPoolExecutor`,  are all the processes currently using the same instance of `hglib`? If that is the case, would the last `Future` object to touch that instance be the one responsible for closing it? Unfortunately can't take the same approach as the thread pool since the instance of `hglib` is a non-pickleable object.

marco-c:
Each process is using a different instance.

@dmitrypolo unfortunately this is not a trivial issue to solve (see the discussion in the issue itself). It is a good-first-issue as it's totally self-contained, but only if you have experience with multi processing concepts and with the multiprocessing Python library.
So, I suggest picking up another issue to solve. Or feel free to come back to this once you know how to do it.

PR #847, opened 2019-08-03
show the values of features too when showing importance values

### LINKED ISSUE #835: When calculating and showing feature importance, also show the values of the features
@chidauri you might be interested in this.

chidauri:
Fixes #835

### DIFF
--- a/bugbug/model.py
+++ b/bugbug/model.py
@@ -229,7 +229,7 @@ class Model:
         return important_features
 
     def print_feature_importances(
-        self, important_features, feature_names, class_probabilities=None, X=None
+        self, important_features, feature_names, class_probabilities=None
     ):
         # extract importance values from the top features for the predicted class
         # when classifying
@@ -250,8 +250,11 @@ class Model:
                 else:
                     shap_val.append("-" + str(importance))
 
+                feature_value = np.squeeze(
+                    important_features["values"].toarray()[:, int(index)]
+                )
                 top_feature_names.append(
-                    f"{feature_names[int(index)]} \nvalue: {X.toarray()[:,int(index)]}"
+                    f"{feature_names[int(index)]} = {feature_value.round(decimals=5)}"
                 )
             shap_val = [[predicted_class] + shap_val]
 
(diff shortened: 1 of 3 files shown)


chidauri:
Selection_006
@marco-c

bugbug/model.py
```diff
@@ -250,7 +250,9 @@ def print_feature_importances(
                 else:
                     shap_val.append("-" + str(importance))
 
-                top_feature_names.append(feature_names[int(index)])
+                top_feature_names.append(
+                    f"{feature_names[int(index)]} \nvalue: {X.toarray()[:,int(index)]}"
```
marco-c: Instead of:
```
FEATURE_NAME 
value: [VALUE]
```
Let's do:
```
FEATURE_NAME = VALUE
```

The other characters (newline, `value` and the parenthesis) are not really adding info but just making the output more crowded.

bugbug/model.py
```diff
@@ -518,6 +520,7 @@ def classify(
             return (
                 classes,
                 {
+                    "bug": X,
```
marco-c: We don't need to return this here, as the caller definitely has `X` when it calls `classify` (and `print_feature_importances` is called after `classify`).

bugbug/model.py
chidauri: I tried this but then the length of feature names become too long and it doesn't align good then

marco-c: Can you show me? The other option is FEATURE_NAME\nVALUE

bugbug/model.py
chidauri: But we don't have the transformed bug in the caller, which is required.

bugbug/model.py
chidauri: Selection_007

marco-c: Couldn't we just reduce the 6 at https://github.com/mozilla/bugbug/pull/847/files#diff-5602bf3d29e24bb4a2603298fae3ebbdR272?

(Remove the "[" and "]" too)

bugbug/model.py
marco-c: Ah I see, then you could store the values of the features directly as part of important_features

Issue #826, opened 2019-08-01
Clone repositories in parallel


marco-c:
In scripts/regressor_finder.py and scripts/microannotate_generator.py we could clone the various repositories in parallel using a ThreadPoolExecutor.

PR #849, opened 2019-08-04
Clone repositories in parallel

kanurag94:
Addresses #826


scripts/microannotate_generator.py
```diff
@@ -25,7 +26,8 @@ def __init__(self, cache_root, repo_url, tokenize, remove_comments):
         self.repo_dir = os.path.join(cache_root, "mozilla-central")
 
     def generate(self):
-        repository.clone(self.repo_dir)
+        executor = concurrent.futures.ThreadPoolExecutor(max_workers=os.cpu_count() + 1)
```
marco-c: We're only executing two tasks, what is the point of having so many workers? :P

marco-c: You shoud use the executor as a context manager to make sure the threads are cleaned up at the end.

scripts/microannotate_generator.py
```diff
@@ -37,9 +39,11 @@ def generate(self):
         )
         git_repo_path = os.path.basename(self.repo_url)
 
-        retry(
-            lambda: subprocess.run(
-                ["git", "clone", self.repo_url, git_repo_path], check=True
+        executor.submit(
```
marco-c: You are passing to `executor.submit` the result of executing the clone, this is not going to work.

marco-c:
Re-request review when you're done with the changes.

Note: the README update should be done in a separate PR.

kanurag94:
@marco-c This is working locally. Can you please re review the PR again. I will remove the readme from this one

scripts/microannotate_generator.py
```diff
@@ -29,37 +30,20 @@ def __init__(self, cache_root, repo_url, tokenize, remove_comments):
         self.repo_dir = os.path.join(cache_root, "mozilla-central")
 
     def generate(self):
-        repository.clone(self.repo_dir)
 
-        logger.info("mozilla-central cloned")
+        with concurrent.futures.ThreadPoolExecutor(max_workers=2) as executor:
+            executor.submit(repository.clone, self.repo_dir)
+            logger.info("Cloning mozilla-central")
```
marco-c: You can keep the "mozilla-central cloned" log at the end of cloning, by using https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Future.add_done_callback on the future returned by executor.submit

kanurag94: Thanks, I have updated PR.
Can you please re review ?

PR #854, opened 2019-08-05
Add option to bug_classifier script to download model when it doesn't exist

### LINKED ISSUE #832: Add an option to the bug_classifier script to download the model when it doesn't exist

chidauri:
Fixes #832

### DIFF
--- a/scripts/bug_classifier.py
+++ b/scripts/bug_classifier.py
@@ -4,11 +4,10 @@ import argparse
 import os
 
 import numpy as np
-import requests
-import zstandard
 
 from bugbug import bugzilla
 from bugbug.models import get_model_class
+from bugbug.utils import download_check_etag, zstd_decompress
 
 MODELS_WITH_TYPE = ("component",)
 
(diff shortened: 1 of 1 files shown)


scripts/bug_classifier.py
```diff
@@ -22,9 +24,23 @@ def classify_bugs(model_name, classifier):
     else:
         model_file_name = f"{model_name}model"
 
-    assert os.path.exists(
-        model_file_name
-    ), f"{model_file_name} does not exist. Train the model with trainer.py first."
+    if not os.path.exists(model_file_name):
```
marco-c: You should use the functions in `utils` for downloading and for extracting.

scripts/bug_classifier.py
```diff
@@ -22,9 +23,12 @@ def classify_bugs(model_name, classifier):
...
+        print(f"{model_file_name} does not exist. Downloading the model....")
```
marco-c: Nit: use logger.info (you'll also need to setup the logger like other scripts do).

scripts/bug_classifier.py
```diff
@@ -22,9 +23,12 @@ def classify_bugs(model_name, classifier):
...
+        download_url = f"https://index.taskcluster.net/v1/task/project.relman.bugbug.train_{model_name}.latest/artifacts/public/{model_file_name}.zst"
```
marco-c: Nit: I'd remove this variable and replace its usage with its value
`download_check_etag(f"https://index.tas...", f"{model_file...")`

marco-c:
What happens after your changes if the model is not available for download?

chidauri:
> What happens after your changes if the model is not available for download?

requests.exceptions.HTTPError: 404 Client Error: Not Found for url: https://index.taskcluster.net/v1/task/project.relman.bugbug.train_uplift.latest/artifacts/public/upliftmodel.zst

marco-c:
> requests.exceptions.HTTPError: 404 Client Error

We should catch the exception and log the error: "A pre-trained model is not available, you will need to train it yourself using the trainer script"

scripts/bug_classifier.py
```diff
@@ -22,9 +28,21 @@ def classify_bugs(model_name, classifier):
...
+        try:
+            download_check_etag(
+                f"https://index.taskcluster.net/v1/task/project.relman.bugbug.train_{model_name}.latest/artifacts/public/{model_file_name}.zst",
+                f"{model_file_name}.zst",
+            )
+        except requests.HTTPError:
+            logger.error(
+                f"A pre-trained model is not available, you will need to train it yourself using the trainer script"
+            )
+            raise SystemExit(0)
```
marco-c: Since it's an error, the error code should not be 0 (0 means no error).

chidauri: ohhh :P

PR #856, opened 2019-08-05
Add arguments to trainer script

cklyyung:
Addresses #339 
In the test, the duplicate model took ~5 min to train while the defect model took ~13, so I used the one that took a shorter time.


marco-c:
> In the test, the duplicate model took ~5 min to train while the defect model took ~13, so I used the one that took a shorter time.

The test is using a mock DB, so it takes just a few seconds. It's better to use the `defect` model in it, as it's more stable.

marco-c:
The trainer.py script is always downloading the bugs and commits DB. We should instead re-download them only if necessary (that is, if they don't exist) to match the old behavior of run.py.
Remove the `force=True` parameter from the db.download calls to achieve that.

PR #857, opened 2019-08-05
[BUG] Close mercurial servers in process pool

dmitrypolo:
This addresses #769 . This keeps the same logic intact that is currently present but adds a `finalizer` to each `Process` using [Finalize](https://github.com/python/cpython/blob/master/Lib/multiprocessing/util.py#L159) which will shutdown the respective `hglib` client.


marco-c:
Closing as the PR is stale. Feel free to reopen if you still intend to work on this.

PR #863, opened 2019-08-06
Add a script to get similar bugs

ayush-1506:
Fixes : #860 

Usage : `python scripts/similar_bugs.py --algorithm=word2vec_wmdrelax --bug_id=1322441`.

### DIFF
--- a/bugbug/similarity.py
+++ b/bugbug/similarity.py
@@ -12,6 +12,7 @@ from itertools import chain
 
 import numpy as np
 from pyemd import emd
+from sklearn.externals import joblib
 from sklearn.feature_extraction.text import TfidfVectorizer
 from sklearn.neighbors import NearestNeighbors
 from tqdm import tqdm
@@ -162,6 +163,13 @@ class BaseSimilarity(abc.ABC):
     @abc.abstractmethod
     def get_distance(self, query1, query2):
         return
+
+    def save(self):
+        joblib.dump(self, f"{self.__class__.__name__.lower()}.similaritymodel")
+
+    @staticmethod
+    def load(model_file_name):
+        return joblib.load(model_file_name)
 
 
 class LSISimilarity(BaseSimilarity):
@@ -516,3 +524,13 @@ class Word2VecSoftCosSimilarity(Word2VecSimilarityBase):
 
     def get_distance(self, query1, query2):
         raise NotImplementedError
+
+
+model_name_to_class = {
+    "lsi": LSISimilarity,
+    "neighbors_tfidf": NeighborsSimilarity,
+    "neighbors_tfidf_bigrams": NeighborsSimilarity,
+    "word2vec_wmdrelax": Word2VecWmdRelaxSimilarity,
+    "word2vec_wmd": Word2VecWmdSimilarity,
+    "word2vec_softcos": Word2VecSoftCosSimilarity,
+}
--- a/scripts/similarity_query.py
+++ b/scripts/similarity_query.py
@@ -1,0 +1,32 @@
+# -*- coding: utf-8 -*-
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import argparse
+import sys
+
+from bugbug import bugzilla, similarity
+
+
+def parse_args(args):
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--algorithm",
+        help="Similarity algorithm to use",
+        choices=similarity.model_name_to_class.keys(),
+    )
+    parser.add_argument("--bug_id", help="Bug to test")
+    return parser.parse_args(args)
+
+
+def main(args):
+    model = similarity.model_name_to_class[args.algorithm].load(
+        f"{similarity.model_name_to_class[args.algorithm].__name__.lower()}.similaritymodel"
+    )
+
+    print(model.get_similar_bugs(bugzilla.get(int(args.bug_id))[int(args.bug_id)]))
+
+
+if __name__ == "__main__":
+    main(parse_args(sys.argv[1:]))
--- a/scripts/similarity_trainer.py
+++ b/scripts/similarity_trainer.py
@@ -35,6 +35,7 @@ def parse_args(args):
 
 
 def main(args):
+
     if args.algorithm == "neighbors_tfidf_bigrams":
         model = similarity.model_name_to_class[args.algorithm](
             vectorizer=TfidfVectorizer(ngram_range=(1, 2)),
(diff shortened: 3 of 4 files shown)


marco-c:
We'd actually need two parts (so maybe two scripts), one to train the model, one to get most similar bugs to a given bug.
In production, we will only train once (every week or every two weeks), but we will do many queries on the same trained model.

ayush-1506:
@marco-c So the first script will train and save the model to disk, and the second will read this model and do the prediction?

marco-c:
Yep!

ayush-1506:
So the updated usage is:
`python scripts/train_similar_model.py --algorithm=word2vec_wmdrelax` to train and save the model to disk, and `python scripts/get_similar_bugs.py --algorithm=word2vec_wmdrelax --bug_id=1322441` to get the similar bugs for the corresponding bug id using the specified algorithm.

scripts/train_similar_model.py
```diff
@@ -0,0 +1,77 @@
...
+    elif args.algorithm == "word2vec_wmdrelax":
+        model_creator = Word2VecWmdRelaxSimilarity
+    elif args.algorithm == "word2vec_softcos":
+        model_creator = Word2VecSoftCosSimilarity
+    model = model_creator(
+        cleanup_urls=args.cleanup_urls, nltk_tokenizer=args.nltk_tokenizer
+    )
+
+    with open(f"{args.algorithm}.model", "wb") as f:
+        pickle.dump(model, f)
```
marco-c: You can use joblib (like we are doing for the classification models.
I'd move it to the similarity module directly (like we're doing for the classification models).

marco-c: I'd move it to the similarity module directly (like we're doing for the classification models).

scripts/train_similar_model.py
```diff
@@ -0,0 +1,77 @@
+# -*- coding: utf-8 -*-
```
marco-c: Nit: call the file `similarity_trainer`

scripts/get_similar_bugs.py
```diff
@@ -0,0 +1,44 @@
+# -*- coding: utf-8 -*-
```
marco-c: Nit: Call the file `similarity_query`

bugbug/similarity.py
```diff
@@ -163,6 +164,9 @@ def evaluation(self):
     def get_distance(self, query1, query2):
         return
 
+    def load(self):
+        return joblib.load(f"{self.__class__.__name__.lower()}.similarity")
```
marco-c: Let's end it with `similaritymodel` instead

bugbug/similarity.py
```diff
@@ -163,6 +164,9 @@ def evaluation(self):
     def get_distance(self, query1, query2):
         return
 
+    def load(self):
```
marco-c: Make this a static method, otherwise users will have to instantiate an object first just to load another object.

bugbug/similarity.py
```diff
@@ -195,6 +199,8 @@ def __init__(self, cleanup_urls=True, nltk_tokenizer=False):
             output_prefix="simdata.shdat", corpus=corpus_lsi, num_features=300
         )
 
+        joblib.dump(self, f"{self.__class__.__name__.lower()}.similarity")
```
marco-c: We should have this in common between all classes (so, the base class should do this).
You can either:
1) Add a `save` method to the base class which does this, and make the trainer script call the `save` function
2) Move the training code into a `train_model` function for each class, the base class will have a `train` function and a `train_model` abstract function. The `train` function will call `train_model` and then use joblib.dump. The trainer script will just call `train`.

scripts/similarity_query.py
```diff
@@ -0,0 +1,58 @@
...
+from bugbug import bugzilla
+from bugbug.similarity import (
+    LSISimilarity,
+    NeighborsSimilarity,
+    Word2VecSoftCosSimilarity,
+    Word2VecWmdRelaxSimilarity,
+    Word2VecWmdSimilarity,
+)
+
+algo_to_model = {
```
marco-c: Define this in `similarity`, so you can re-use it in both scripts.

marco-c: And in the evaluation script

scripts/similarity_query.py
```diff
@@ -0,0 +1,58 @@
...
+    "word2vec_softcos": Word2VecSoftCosSimilarity,
+}
+
+
+def parse_args(args):
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--algorithm",
+        help="Similarity algorithm to use",
+        choices=[
```
marco-c: `choices=algo_to_model.keys()`, so you don't have to repeat the same names.

scripts/similarity_query.py
marco-c: Nit: call it `model_name_to_class`

scripts/similarity_query.py
```diff
@@ -0,0 +1,37 @@
...
+        "--algorithm",
+        help="Similarity algorithm to use",
+        choices=similarity.model_name_to_class.keys(),
+    )
+    parser.add_argument("--bug_id", help="Bug to test")
+    return parser.parse_args(args)
+
+
+def main(args):
+    model = similarity.model_name_to_class[args.algorithm]().load(
```
marco-c: You made `load` a static method, so you don't need to instantiate an object in order to call it

scripts/similarity_query.py
```diff
@@ -0,0 +1,37 @@
...
+        f"{similarity.model_name_to_class[args.algorithm].__name__.lower()}.similaritymodel"
+    )
+
+    for bug in bugzilla.get_bugs():
```
marco-c: I'd download the bug instead of trying to find it in the DB.

scripts/similarity_trainer.py
```diff
@@ -0,0 +1,69 @@
...
+        help="Use nltk's tokenizer for text preprocessing",
+        dest="nltk_tokenizer",
+        default=False,
+    )
+    return parser.parse_args(args)
+
+
+def main(args):
+    if args.algorithm == "lsi":
+        model_creator = LSISimilarity
```
marco-c: You should be able to get model_creator like you're doing in the other script, using the model_name_to_class dict.

marco-c: (With just an if to check if neighbors_tfidf_bigrams)

scripts/similarity_query.py
marco-c: Don't forget to also use it in the evaluation script (scripts/evaluate_similarity.py).

scripts/similarity_query.py
ayush-1506: Ahh, sorry, missed this.

PR #864, opened 2019-08-08
Option to classify a specific bug 

### LINKED ISSUE #833: Add an option to the bug_classifier script to classify a specific bug

cklyyung:
Fixes #833 
I was able to successfully fetch one bug : 
`python3 -c 'from scripts import bug_classifier; bug_classifier.main()' defect --bug_id 1322944`

If the `--bug_id` argument isn't specified, `bug_id=None` is passed to the `classify_bugs` function and all bugs are fetched.

### DIFF
--- a/scripts/bug_classifier.py
+++ b/scripts/bug_classifier.py
@@ -17,7 +17,7 @@ basicConfig(level=INFO)
 logger = getLogger(__name__)
 
 
-def classify_bugs(model_name, classifier):
+def classify_bugs(model_name, classifier, bug_id):
     if classifier != "default":
         assert (
             model_name in MODELS_WITH_TYPE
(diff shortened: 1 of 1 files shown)


scripts/bug_classifier.py
```diff
@@ -83,7 +89,7 @@ def main():
         choices=["default", "nn"],
         default="default",
     )
+    parser.add_argument("--bug_id", help="Classify the given bug id")
```
marco-c: Nit: I prefer `--bug-id`

scripts/bug_classifier.py
```diff
@@ -83,7 +89,7 @@ def main():
...
 
     args = parser.parse_args()
-
```
marco-c: Nit: don't remove newline

PR #867, opened 2019-08-08
Switch to pickle DBs

### LINKED ISSUE #749: Switch to pickle DBs
They should be faster to read and write (we need to confirm).

chidauri:
Closes #749


scripts/bug_retriever.py
```diff
@@ -122,7 +122,7 @@ def retrieve_bugs(self):
             bugzilla.delete_bugs(lambda bug: bug["id"] in inconsistent_bug_ids)
             bugzilla.download_bugs(inconsistent_bug_ids)
 
-        zstd_compress("data/bugs.json")
+        zstd_compress("data/bugs.pickle")
```
marco-c: I think you can replace "data/bugs.pickle" with bugzilla.BUGS_DB

scripts/commit_retriever.py
```diff
@@ -35,7 +35,7 @@ def retrieve_commits(self):
 
         logger.info("commit data extracted from repository")
 
-        zstd_compress("data/commits.json")
+        zstd_compress("data/commits.pickle")
```
marco-c: Same here with repository.COMMITS_DB

marco-c:
We need to think on how to manage the transition between the old DB and the new DB.

marco-c:
Can you check:
1) What is the size of the bugs.json.zst, bugs.pickle.zst, bugs.json, bugs.zst files (same with commits.*);
2) How long it takes to iterate through all bugs with JSON vs pickle?

marco-c:
I'd handle the transition this way (so that it is fully generic).
1) Change the `register` function to define the name of the DB without the format + the preferred format separately.
2) Change the DB code to try downloading the DB in the preferred format first, and fallback to the other supported formats in case of 404.

chidauri:
@marco-c If we change the DB code to download the json format if pickle format not found, we need to convert the json to pickle, right?

marco-c:
Correct!

chidauri:
@marco-c  Now the only test failing are test_db.py, those are pretty complex :P, should i fix them too?

chidauri:
@marco-c now i've hardcoded .json as fallback path, another way is we pass this parameter fallback_path from the caller(example bug_retriever.py) , i think this is better cause hardcoding is brittle, wdyt?

marco-c:
> @marco-c now i've hardcoded .json as fallback path, another way is we pass this parameter fallback_path from the caller(example bug_retriever.py) , i think this is better cause hardcoding is brittle, wdyt?

In the db module, we have a list of all formats we support (SERIALIZATION_FORMATS), we can just use that.

chidauri:
@marco-c i've made some changes to make it more generic and it is working as expected(it first downloads `.json `and then converts it to `.pickle`), but the tests are failing because the path sent is not a string in tests, can we not change them to string format?

marco-c:
> the tests are failing because the path sent is not a string in tests, can we not change them to string format?

Yes, feel free to change them to be strings!

chidauri:
@marco-c The `test_download_missing` is failing with the following error:
testfail
I tried using pickle url but no luck.

marco-c:
@chidauri let me know if you need any help or clarification ;)

chinu0710:
Use pickles db instead of json

PR #873, opened 2019-08-09
Calculate Recall better for Confidence Threshold cases

### LINKED ISSUE #509: Calculate recall better when a confidence threshold is set
We are basically overestimating the recall for the confidence threshold cases: https://github.com/mozilla/bugbug/blob/f0547925391cf082ee31f2031f2d7a35c715c884/bugbug/model.py#L139.

We should take into account that some elements are not classified when we set a threshold.

chidauri:
Fixes #509


chidauri:
Selection_008
Looks like this

bugbug/model.py
```diff
@@ -381,6 +381,7 @@ def train(self, importance_cutoff=0.15):
             confusion_matrix = metrics.confusion_matrix(
                 y_test, y_pred, labels=self.class_names
             )
+            tn, fp, fn, tp = metrics.confusion_matrix(y_test, y_pred).ravel()
```
marco-c: This will only work for the binary case.

marco-c:
If you want to implement it only for the binary case, it's fine, but instead of calculating an adjusted recall, I'd change y_pred_filter to have 0s where the confidence threshold is not hit. This way the correct recall is shown directly in the normal report.

chidauri:
@marco-c what should we do for multi-class models(this method won't work there)?

marco-c:
Maybe in the multi-class case you can try using a string instead of 0 (e.g. "___NOT_CLASSIFIED___").

marco-c:
Can you check if that works?

marco-c:
Even better, you could convert class names to strings when you calculate confusion matrix and classification_report_imbalanced, and always use "___NOT_CLASSIFIED___" when the confidence is lower than the threshold.

chidauri:
@marco-c  is this okay?
```
1695 defects
60 enhancements
221 tasks
X: (1976, 29003), y: (1976,)
Cross Validation scores:
Accuracy: f0.9077607216331698 (+/- 0.025236667804356565)
X_train: (4575, 29003), y_train: (4575,)
X_test: (198, 29003), y_test: (198,)
Test Set scores:
No confidence threshold - 198 classified

     defect       0.96      0.96      0.75      0.96      0.85      0.74       170
enhancement       1.00      0.14      1.00      0.25      0.38      0.13         7
       task       0.65      0.81      0.95      0.72      0.88      0.76        21

avg / total       0.93      0.92      0.78      0.91      0.84      0.72       198

╒══════════════════════╤══════════════════════╤═══════════════════════════╤════════════════════╕
│                      │   defect (Predicted) │   enhancement (Predicted) │   task (Predicted) │
╞══════════════════════╪══════════════════════╪═══════════════════════════╪════════════════════╡
│ defect (Actual)      │                  164 │                         0 │                  6 │
├──────────────────────┼──────────────────────┼───────────────────────────┼────────────────────┤
│ enhancement (Actual) │                    3 │                         1 │                  3 │
├──────────────────────┼──────────────────────┼───────────────────────────┼────────────────────┤
│ task (Actual)        │                    4 │                         0 │                 17 │
╘══════════════════════╧══════════════════════╧═══════════════════════════╧════════════════════╛


Confidence threshold > 0.6 - 198 classified
/home/harshit/.local/lib/python3.7/site-packages/sklearn/metrics/classification.py:1439: UndefinedMetricWarning: Recall and F-score are ill-defined and being set to 0.0 in labels with no true samples.
  'recall', 'true', average, warn_for)
/home/harshit/.local/lib/python3.7/site-packages/imblearn/metrics/_classification.py:241: UndefinedMetricWarning: Sensitivity is ill-defined and being set to 0.0 in labels with no true samples.
  average, warn_for)

        defect       0.97      0.94      0.82      0.96      0.88      0.78       170
   enhancement       1.00      0.14      1.00      0.25      0.38      0.13         7
          task       0.75      0.71      0.97      0.73      0.83      0.68        21
NOT_CLASSIFIED       0.00      0.00      0.94      0.00      0.00      0.00         0

   avg / total       0.95      0.89      0.84      0.91      0.86      0.75       198

╒══════════════════════╤══════════════════════╤═══════════════════════════╤════════════════════╤══════════════════════════════╕
│                      │   defect (Predicted) │   enhancement (Predicted) │   task (Predicted) │   NOT_CLASSIFIED (Predicted) │
╞══════════════════════╪══════════════════════╪═══════════════════════════╪════════════════════╪══════════════════════════════╡
│ defect (Actual)      │                  160 │                         0 │                  4 │                            6 │
├──────────────────────┼──────────────────────┼───────────────────────────┼────────────────────┼──────────────────────────────┤
│ enhancement (Actual) │                    3 │                         1 │                  1 │                            2 │
├──────────────────────┼──────────────────────┼───────────────────────────┼────────────────────┼──────────────────────────────┤
│ task (Actual)        │                    2 │                         0 │                 15 │                            4 │
╘══════════════════════╧══════════════════════╧═══════════════════════════╧════════════════════╧══════════════════════════════╛


Confidence threshold > 0.7 - 198 classified

        defect       0.98      0.92      0.89      0.95      0.91      0.83       170
   enhancement       1.00      0.14      1.00      0.25      0.38      0.13         7
          task       0.88      0.67      0.99      0.76      0.81      0.64        21
NOT_CLASSIFIED       0.00      0.00      0.89      0.00      0.00      0.00         0

   avg / total       0.97      0.87      0.91      0.91      0.88      0.78       198

╒══════════════════════╤══════════════════════╤═══════════════════════════╤════════════════════╤══════════════════════════════╕
│                      │   defect (Predicted) │   enhancement (Predicted) │   task (Predicted) │   NOT_CLASSIFIED (Predicted) │
╞══════════════════════╪══════════════════════╪═══════════════════════════╪════════════════════╪══════════════════════════════╡
│ defect (Actual)      │                  157 │                         0 │                  1 │                           12 │
├──────────────────────┼──────────────────────┼───────────────────────────┼────────────────────┼──────────────────────────────┤
│ enhancement (Actual) │                    2 │                         1 │                  1 │                            3 │
├──────────────────────┼──────────────────────┼───────────────────────────┼────────────────────┼──────────────────────────────┤
│ task (Actual)        │                    1 │                         0 │                 14 │                            6 │
╘══════════════════════╧══════════════════════╧═══════════════════════════╧════════════════════╧══════════════════════════════╛


Confidence threshold > 0.8 - 198 classified

        defect       0.98      0.86      0.89      0.92      0.88      0.76       170
   enhancement       1.00      0.14      1.00      0.25      0.38      0.13         7
          task       0.91      0.48      0.99      0.62      0.69      0.45        21
NOT_CLASSIFIED       0.00      0.00      0.81      0.00      0.00      0.00         0

   avg / total       0.97      0.79      0.91      0.86      0.84      0.71       198

╒══════════════════════╤══════════════════════╤═══════════════════════════╤════════════════════╤══════════════════════════════╕
│                      │   defect (Predicted) │   enhancement (Predicted) │   task (Predicted) │   NOT_CLASSIFIED (Predicted) │
╞══════════════════════╪══════════════════════╪═══════════════════════════╪════════════════════╪══════════════════════════════╡
│ defect (Actual)      │                  146 │                         0 │                  0 │                           24 │
├──────────────────────┼──────────────────────┼───────────────────────────┼────────────────────┼──────────────────────────────┤
│ enhancement (Actual) │                    2 │                         1 │                  1 │                            3 │
├──────────────────────┼──────────────────────┼───────────────────────────┼────────────────────┼──────────────────────────────┤
│ task (Actual)        │                    1 │                         0 │                 10 │                           10 │
╘══════════════════════╧══════════════════════╧═══════════════════════════╧════════════════════╧══════════════════════════════╛


Confidence threshold > 0.9 - 198 classified

        defect       0.98      0.73      0.93      0.84      0.82      0.66       170
   enhancement       1.00      0.14      1.00      0.25      0.38      0.13         7
          task       0.83      0.24      0.99      0.37      0.49      0.22        21
NOT_CLASSIFIED       0.00      0.00      0.67      0.00      0.00      0.00         0

   avg / total       0.97      0.66      0.94      0.77      0.77      0.60       198

╒══════════════════════╤══════════════════════╤═══════════════════════════╤════════════════════╤══════════════════════════════╕
│                      │   defect (Predicted) │   enhancement (Predicted) │   task (Predicted) │   NOT_CLASSIFIED (Predicted) │
╞══════════════════════╪══════════════════════╪═══════════════════════════╪════════════════════╪══════════════════════════════╡
│ defect (Actual)      │                  124 │                         0 │                  0 │                           46 │
├──────────────────────┼──────────────────────┼───────────────────────────┼────────────────────┼──────────────────────────────┤
│ enhancement (Actual) │                    2 │                         1 │                  1 │                            3 │
├──────────────────────┼──────────────────────┼───────────────────────────┼────────────────────┼──────────────────────────────┤
│ task (Actual)        │                    0 │                         0 │                  5 │                           16 │
╘══════════════════════╧══════════════════════╧═══════════════════════════╧════════════════════╧══════════════════════════════╛

I0903 15:21:24.471544 140559380772672 trainer.py:67] Training done
I0903 15:21:24.603002 140559380772672 trainer.py:73] Model compressed

```

bugbug/model.py
```diff
@@ -402,12 +407,15 @@ def train(self, importance_cutoff=0.15):
         # Evaluate results on the test set for some confidence thresholds.
         for confidence_threshold in [0.6, 0.7, 0.8, 0.9]:
             y_pred_probas = self.clf.predict_proba(X_test)
+            confidence_class_names = self.class_names + ["NOT_CLASSIFIED"]
 
             y_test_filter = []
             y_pred_filter = []
+            not_classified_indices = []
```
marco-c: This implementation seems more complex than necessary, couldn't you simply `y_pred_filter.append("___NOT_CLASSIFIED___")` when `y_pred_probas[i][argmax] < confidence_threshold`?

bugbug/model.py
```diff
@@ -93,6 +93,11 @@ def classification_report_imbalanced_values(
 
 def print_labeled_confusion_matrix(confusion_matrix, labels, is_multilabel=False):
     confusion_matrix_table = confusion_matrix.tolist()
+    try:
+        confusion_matrix_table.pop(labels.index("NOT_CLASSIFIED"))
+    except ValueError:
+        pass
```
marco-c: When is the exception thrown?

bugbug/model.py
```diff
@@ -104,7 +109,7 @@ def print_labeled_confusion_matrix(confusion_matrix, labels, is_multilabel=False
             table_labels = labels
 
         confusion_matrix_header = []
-        for i in range(len(table)):
+        for i in range(len(table[0])):
```
marco-c: Why this change?

marco-c:
The output looks good, but I'd replace `NOT_CLASSIFIED (Predicted)` with `Not classified`.

bugbug/model.py
marco-c: Is it for when there is no confidence threshold?

marco-c: If so, it'd be better to do:
```
if "___NOT_CLASSIFIED___" in labels:
    confusion_matrix_table.pop(labels.index("___NOT_CLASSIFIED___"))
```

And a comment before the if to explain that we don't want to show the "Not classified" row in the table output.

bugbug/model.py
chidauri: This is the problem https://github.com/mozilla/bugbug/blob/6426361dbff7e06a4f677af4b631687005e1fc9f/bugbug/model.py#L420

bugbug/model.py
chidauri: This is because the table has different sizes of rows and columns,earlier they're same.

chidauri:
Done.

bugbug/model.py
```diff
@@ -93,6 +93,9 @@ def classification_report_imbalanced_values(
 
 def print_labeled_confusion_matrix(confusion_matrix, labels, is_multilabel=False):
     confusion_matrix_table = confusion_matrix.tolist()
+    if "Not classified" in labels:
```
marco-c: Don't forget the comment here.

bugbug/model.py
```diff
@@ -402,12 +409,15 @@ def train(self, importance_cutoff=0.15):
         # Evaluate results on the test set for some confidence thresholds.
         for confidence_threshold in [0.6, 0.7, 0.8, 0.9]:
             y_pred_probas = self.clf.predict_proba(X_test)
+            confidence_class_names = self.class_names + ["Not classified"]
```
marco-c: Use "___NOT_CLASSIFIED___" (so it's clear that it is just an internal name). Then when you print it you use "Not classified"

bugbug/model.py
marco-c: OK, makes sense.

bugbug/model.py
```diff
@@ -402,12 +409,15 @@ def train(self, importance_cutoff=0.15):
...
             y_test_filter = []
             y_pred_filter = []
+            not_classified_indices = []
             for i in range(0, len(y_test)):
                 argmax = np.argmax(y_pred_probas[i])
                 if y_pred_probas[i][argmax] < confidence_threshold:
+                    not_classified_indices.append(i)
                     continue
 
                 y_test_filter.append(y_test[i])
```
marco-c: We are no longer using a subset of y_test, so you can just remove y_test_filter altogether and use y_test (converting it to a string array when needed with `y_test.astype(str)`).

marco-c:
Could you double check this works in all cases? Binary, multiclass, and multilabel.

bugbug/model.py
chidauri: There are two problems 
1. The subset of y_test is still being used for multi-label model.
2. I've appended the not-classified y_test labels to the last of y_test_filter because we're concatening in case of y_pred_filter too. To do what you're proposing, we'll have to find a way to insert "NOT_CLASSIFIED" labels at the corresponding positions and not concatenate them at the last.

chidauri:
It is working fine in all cases

chidauri:
Works for all types of models now

bugbug/model.py
```diff
@@ -418,6 +430,17 @@ def train(self, importance_cutoff=0.15):
 
             if not is_multilabel:
                 y_pred_filter = self.le.inverse_transform(y_pred_filter)
+                y_pred_filter = np.asarray(
+                    np.concatenate(
+                        (
+                            y_pred_filter,
+                            np.array(["NOT_CLASSIFIED"] * len(not_classified_indices)),
```
marco-c: Use "___NOT_CLASSIFIED___" so it's clearer it's an internal name (no class is ever going to be called like this).

marco-c: Note: with \_\_\_ before and after NOT_CLASSIFIED

bugbug/model.py
marco-c: For 1, can't we support the multi-label case too?

For 2, it's pretty easy, in the loop instead of `not_classified_indices.append(i)` you append \_\_\_NOT_CLASSIFIED\_\_\_

marco-c:
>  Btw why is docker build failing?

It was just a temporary failure

chidauri:
@marco-c Multi-label models can't be treated the same way, it gives an error (mix of multi-label and multi-class multi-output not supported), so i've included `classified_indices ` parameter which we can get rid of, otherwise.

bugbug/model.py
```diff
@@ -402,44 +411,54 @@ def train(self, importance_cutoff=0.15):
...
+                y_pred_filter[classified_indices] = self.le.inverse_transform(
+                    np.array(y_pred_filter[classified_indices], dtype=int)
+                )
 
             print(
-                f"\nConfidence threshold > {confidence_threshold} - {len(y_test_filter)} classified"
+                f"\nConfidence threshold > {confidence_threshold} - {len(y_test)} classified"
             )
-            if len(y_test_filter) != 0:
+            if len(y_test) != 0:
```
marco-c: I think we can remove this check now, as len(y_test) will always be > 0

marco-c:
(Just cleanup that check and then I'll merge)

PR #882, opened 2019-08-13
Update model.py

Sejal-G:
Removed feature_names parameter from print_feature_importances function in models.py file
In correspondence to #834 issue in mozilla/bugbug.


marco-c:
Did you close the PR on purpose?

Issue #870, opened 2019-08-08
Make the `similarity_query` script download a pre-trained model


marco-c:
Depends on #859.

Similarly to the bug_classifier script, the similarity_query should download a pre-trained model when available (or assert that it should be trained first).

PR #874, opened 2019-08-09
BM25 similarity task

ayush-1506:
Fixes: #859

PR #885, opened 2019-08-15
Download similarity model if available

ayush-1506:
Fixes: #870.

Need to figure out the url of the models in case they aren't available locally.

### DIFF
--- a/scripts/similarity_query.py
+++ b/scripts/similarity_query.py
@@ -4,9 +4,19 @@
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import argparse
+import os
 import sys
+from logging import INFO, basicConfig, getLogger
+
+import requests
 
 from bugbug import bugzilla, similarity
+from bugbug.utils import download_check_etag, zstd_decompress
+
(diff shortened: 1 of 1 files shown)


marco-c:
This looks good so far, but needs to be tested after we merge and run #874.

scripts/similarity_query.py
```diff
@@ -21,6 +29,25 @@ def parse_args(args):
...
 
 def main(args):
+
+    model_file_name = f"{similarity.model_name_to_class[args.algorithm].__name__.lower()}.similaritymodel"
+
+    if not os.path.exists(model_file_name):
+        logger.info(f"{model_file_name} does not exist. Downloading the model....")
+        try:
+            download_check_etag(
+                f"https://index.taskcluster.net/v1/task/project.relman.bugbug.train_similarity.latest/artifacts/public/{model_file_name}.zst",
```
marco-c: Nit: put `https://index.taskcluster.net/v1/task/project.relman.bugbug.train_similarity.latest/artifacts/public/{}.zst` in a URL variable at the top, and then here do URL.format(model_file_name)

PR #889, opened 2019-08-17
Add couple features to duplicate model

ayush-1506:
Part of #523 

The results before and after are exactly the same. I'm looking into this.

### DIFF
--- a/bugbug/bug_features.py
+++ b/bugbug/bug_features.py
@@ -463,11 +463,13 @@ class had_severity_enhancement(single_bug_feature):
 
 class couple_common_whiteboard_keywords(couple_bug_feature):
     def __call__(self, bugs, **kwargs):
-        return [
-            keyword
-            for keyword in whiteboard_keywords(bugs[0])
-            if keyword in whiteboard_keywords(bugs[1])
-        ]
+        return len(
+            [
+                keyword
+                for keyword in whiteboard_keywords(bugs[0])
+                if keyword in whiteboard_keywords(bugs[1])
+            ]
+        )
 
 
 class is_same_product(couple_bug_feature):
@@ -522,16 +524,44 @@ class couple_delta_creation_date(couple_bug_feature):
         return delta / timedelta(days=1)
 
 
+class couple_common_words_summary(couple_bug_feature):
+    def __init__(self, to_ignore=set()):
+        self.to_ignore = to_ignore
+
+    def __call__(self, bugs):
+        return len(
+            set(bugs[0]["summary"].split()).intersection(
+                set(bugs[1]["summary"].split())
+            )
+        )
+
+
+class couple_common_words_comments(couple_bug_feature):
+    def __init__(self, to_ignore=set()):
+        self.to_ignore = to_ignore
+
+    def __call__(self, bugs):
+        text1 = " ".join(
+            bugs[0]["comments"][idx]["text"] for idx in range(len(bugs[0]["comments"]))
+        )
+        text2 = " ".join(
+            bugs[1]["comments"][idx]["text"] for idx in range(len(bugs[1]["comments"]))
+        )
+        return len(set(text1.split()).intersection(set(text2.split())))
+
+
 class couple_common_keywords(couple_bug_feature):
     def __init__(self, to_ignore=set()):
         self.to_ignore = to_ignore
 
     def __call__(self, bugs, **kwargs):
-        return [
-            keyword
-            for keyword in bugs[0]["keywords"]
-            if keyword in bugs[1]["keywords"] and keyword not in self.to_ignore
-        ]
+        return len(
+            [
+                keyword
+                for keyword in bugs[0]["keywords"]
+                if keyword in bugs[1]["keywords"] and keyword not in self.to_ignore
+            ]
+        )
 
 
 def get_author_ids():
--- a/bugbug/bug_features.py
+++ b/bugbug/bug_features.py
@@ -472,12 +472,12 @@ class couple_common_whiteboard_keywords(couple_bug_feature):
 
 class is_same_product(couple_bug_feature):
     def __call__(self, bugs, **kwargs):
-        return bugs[0]["product"] == bugs[1]["product"]
+        return int(bugs[0]["product"] == bugs[1]["product"])
 
 
 class is_same_component(couple_bug_feature):
     def __call__(self, bugs, **kwargs):
-        return (
+        return int(
             bugs[0]["product"] == bugs[1]["product"]
             and bugs[0]["component"] == bugs[1]["component"]
         )
@@ -485,22 +485,22 @@ class is_same_component(couple_bug_feature):
 
 class is_same_platform(couple_bug_feature):
     def __call__(self, bugs, **kwargs):
-        return bugs[0]["platform"] == bugs[1]["platform"]
+        return int(bugs[0]["platform"] == bugs[1]["platform"])
 
 
 class is_same_version(couple_bug_feature):
     def __call__(self, bugs, **kwargs):
-        return bugs[0]["version"] == bugs[1]["version"]
+        return int(bugs[0]["version"] == bugs[1]["version"])
 
 
 class is_same_os(couple_bug_feature):
     def __call__(self, bugs, **kwargs):
-        return bugs[0]["op_sys"] == bugs[1]["op_sys"]
+        return int(bugs[0]["op_sys"] == bugs[1]["op_sys"])
 
 
 class is_same_target_milestone(couple_bug_feature):
     def __call__(self, bugs, **kwargs):
-        return bugs[0]["target_milestone"] == bugs[1]["target_milestone"]
+        return int(bugs[0]["target_milestone"] == bugs[1]["target_milestone"])
 
 
 class is_first_affected_same(couple_bug_feature):
--- a/bugbug/bug_features.py
+++ b/bugbug/bug_features.py
@@ -463,23 +463,21 @@ class had_severity_enhancement(single_bug_feature):
 
 class couple_common_whiteboard_keywords(couple_bug_feature):
     def __call__(self, bugs, **kwargs):
-        return len(
-            [
-                keyword
-                for keyword in whiteboard_keywords(bugs[0])
-                if keyword in whiteboard_keywords(bugs[1])
-            ]
-        )
+        return [
+            keyword
+            for keyword in whiteboard_keywords(bugs[0])
+            if keyword in whiteboard_keywords(bugs[1])
+        ]
 
 
 class is_same_product(couple_bug_feature):
     def __call__(self, bugs, **kwargs):
-        return int(bugs[0]["product"] == bugs[1]["product"])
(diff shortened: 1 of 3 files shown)


ayush-1506:
Update : I'm trying to figure out a way to add the couple_data as a feature from inside the extraction pipeline.

ayush-1506:
Ok, so that works now. However, there are a few other problems:

1. Values such as "True", "False" need to be converted to integers so that we get the resultant sparse matrix. For now, I'm making the feature extractors return `int`, but I'm not sure if this is the best way.

2. How should I handle feature extractors that don't returns "true"/"false", but a list of keywords (such as `couple_common_keywords`)? We need a way to convert these keywords to numbers. (Is using a vectorizer the best option in this case?)

@marco-c

ayush-1506:
Maybe number of common keywords make more sense than list of the common keywords?

ayush-1506:
Results:

Adding common keywords:
```
{'test_accuracy': {'mean': 0.9663222337896421, 'std': 0.00617975339630379}, 'test_precision': {'mean': 
0.8653523448151335, 'std': 0.06047868681910123}, 'test_recall': {'mean': 0.7488813898262717, 'std': 
0.08162211314077335}, 'report': {'targets': {'1': {'precision': 0.8493150684931506, 'recall': 0.7045454545454546, 
'specificity': 0.9839181286549707, 'f1': 0.7701863354037267, 'geo_mean': 0.8325953670233395, 'iba': 
0.6738485110979969, 'support': 88}, '0': {'precision': 0.9628040057224606, 'recall': 0.9839181286549707, 'specificity': 
0.7045454545454546, 'f1': 0.9732465654374548, 'geo_mean': 0.8325953670233395, 'iba': 0.7125815792794619, 
'support': 684}}, 'average': {'precision': 0.9498674429294823, 'recall': 0.9520725388601037, 'specificity': 
0.7363910443403217, 'f1': 0.9500998034646982, 'geo_mean': 0.8325953670233395, 'iba': 0.7081664108857199, 
'support': 772}}, 'confusion_matrix': [[62, 26], [11, 673]]}
```

common keywords + common whiteboard keyword:

```
{'test_accuracy': {'mean': 0.9661783488975558, 'std': 0.006543452994697574}, 'test_precision': {'mean': 
0.8634883097274141, 'std': 0.05722827724494728}, 'test_recall': {'mean': 0.7488813898262717, 'std': 
0.08162211314077335}, 'report': {'targets': {'1': {'precision': 0.8611111111111112, 'recall': 0.7045454545454546, 'specificity': 
0.9853801169590644, 'f1': 0.775, 'geo_mean': 0.8332137075222522, 'iba': 0.6747482740680824, 'support': 88}, '0': 
{'precision': 0.9628571428571429, 'recall': 0.9853801169590644, 'specificity': 0.7045454545454546, 'f1': 
0.9739884393063584, 'geo_mean': 0.8332137075222522, 'iba': 0.7137418907378721, 'support': 684}}, 'average': 
{'precision': 0.9512591496011185, 'recall': 0.9533678756476683, 'specificity': 0.7365576958568505, 'f1': 
0.9513058192817995, 'geo_mean': 0.8332137075222521, 'iba': 0.7092970225164452, 'support': 772}}, 'confusion_matrix': 
[[62, 26], [10, 674]]}
```

Will keep adding more results as they come.

ayush-1506:
@marco-c I'm using the following couple features right now:

> 
> common keywords + common whiteboard keyword + common words in summary + common words in comments:
> 

Let me know more features to add here.

bugbug/bug_features.py
```diff
@@ -463,44 +463,46 @@ def __call__(self, bug, **kwargs):
 
 class couple_common_whiteboard_keywords(couple_bug_feature):
```
marco-c, line 464: I'd define a new feature (`couple_number_of_common_whiteboard_keywords`) instead of changing this one

bugbug/bug_features.py
```diff
@@ -463,44 +463,46 @@ def __call__(self, bug, **kwargs):
...
+                for keyword in whiteboard_keywords(bugs[0])
+                if keyword in whiteboard_keywords(bugs[1])
+            ]
+        )
 
 
 class is_same_product(couple_bug_feature):
     def __call__(self, bugs, **kwargs):
-        return bugs[0]["product"] == bugs[1]["product"]
+        return int(bugs[0]["product"] == bugs[1]["product"])
```
marco-c, line 477: Why an int here and in the other features?

bugbug/bug_features.py
marco-c: Or maybe not, if we are not using the old in any other place

bugbug/model.py
```diff
@@ -248,10 +248,7 @@ def get_top_features(cutoff, shap_values):
 
             class_name = self.le.inverse_transform([num])[0]
 
-            important_features["classes"][class_name] = (
-                top_item_features,
-                top_avg,
-            )
+            important_features["classes"][class_name] = (top_item_features, top_avg)
```
marco-c, line 251: This change seems unrelated, could you revert it?

bugbug/models/duplicate.py
```diff
@@ -28,6 +30,28 @@ def fit(self, X, y):
         self._le.fit(y)
 
 
+class BypassFeatures(TransformerMixin, BaseEstimator):
```
marco-c, line 33: Why was this needed?

marco-c:
> Values such as "True", "False" need to be converted to integers so that we get the resultant sparse matrix. For now, I'm making the feature extractors return int, but I'm not sure if this is the best way.

> How should I handle feature extractors that don't return "true"/"false", but a list of keywords (such as couple_common_keywords)? We need a way to convert these keywords to numbers. (Is using a vectorizer the best option in this case?)

You can solve these problems like this: https://github.com/mozilla/bugbug/blob/07590f145d5707c0efd212a72635d1e619bfe98d/bugbug/commit_features.py#L666-L672.

bugbug/bug_features.py
marco-c: OK, read your earlier comment and I replied in https://github.com/mozilla/bugbug/pull/889#issuecomment-670501994 :)

marco-c:
> Adding common keywords:

What are the results before adding them?

Could you try all combinations of these new features with/without xgboost?

(Could you report just accuracy, precision and recall so they're easier to read?)

bugbug/models/duplicate.py
ayush-1506: Right now our pipeline looks like:
```
self.extraction_pipeline = Pipeline(
            [
                (
                    "bug_extractor",
                    bug_features.BugExtractor(feature_extractors, cleanup_functions),
                ),
                (
                    "union",
                    ColumnTransformer(
                        [
                            ("text", self.text_vectorizer(), "text"),
                        ]
                    ),
                ),
            ]
        )
```

BugFeatures returns a dictionary {"text":... , "couple_data"..}, but the `ColumnTransformer` in the next step just takes the "text" data from the dict and ignores everything else. Hence, we end up loosing the "couple_data" data, which has all the bug couple features. I confirmed this by completely removing all bug couple data, but I still get same dimensional features as obtained by using them (implying that they weren't being used at all).

One way to circumvent this is by introducing the `BypassFeatures` transformer that appends the couple data features to the current tf-idf text vectors, thus expanding our features. Is there a better way to do this? @marco-c

bugbug/model.py
ayush-1506: pre-commit seems to be making this change, not sure why it is detecting this

marco-c: You should be able to revert it now, it was probably a different version of black

ayush-1506:
xgboost + all feature extractors:
```
Accuracy: f0.9310097201147786 (+/- 0.036759555908687874)
Precision: f0.8938016215308388 (+/- 0.07994774867892024)
Recall: f0.8761904761904763 (+/- 0.08575999315359989)
```

ayush-1506:
@marco-c Updated results above.

ayush-1506:
@marco-c Not sure what's up with black: it seems to be changing a lot of files.

marco-c:
> @marco-c Not sure what's up with black: it seems to be changing a lot of files.

You probably need to update black, you can try to:
```
pre-commit uninstall
pre-commit clean
pre-commit gc
pre-commit install
pre-commit run --all-files
```

marco-c:
> @marco-c Updated results above.

They are looking pretty good, I hope we are not making mistakes in the evaluation code :P

marco-c:
The patch looks good, I'll approve it once you manage to get rid of the black crazyness :P

ayush-1506:
@marco-c r? (Had to manually revert changes for each file, not sure what's wrong)

bugbug/bug_features.py
```diff
@@ -522,6 +522,30 @@ def __call__(self, bugs, **kwargs):
...
+        )
+
+
+class couple_common_words_comments(couple_bug_feature):
+    def __init__(self, to_ignore=set()):
+        self.to_ignore = to_ignore
+
+    def __call__(self, bugs):
+        text1 = " ".join(
+            bugs[0]["comments"][idx]["text"] for idx in range(len(bugs[0]["comments"]))
```
marco-c, line 541: You could rewrite as:
```
comment["text"] for comment in bug[0]["comments"]

File: bugbug/models/duplicate.py [hunk 6]
```diff
@@ -8,9 +8,10 @@
 
 from sklearn.calibration import CalibratedClassifierCV
 from sklearn.compose import ColumnTransformer
+from sklearn.feature_extraction import DictVectorizer
 from sklearn.pipeline import Pipeline
 from sklearn.preprocessing import LabelEncoder
-from sklearn.svm import LinearSVC
+from xgboost import XGBClassifier
```
marco-c, line 14: Could you split the classifier change in a separate PR?

bugbug/models/duplicate.py
```diff
@@ -55,16 +69,23 @@ def __init__(self, training_size=14000, lemmatization=False, cleanup_urls=True):
...
+                            ("text", self.text_vectorizer(), "text"),
+                            ("couple_data", DictVectorizer(), "couple_data"),
+                        ]
+                    ),
                 ),
             ]
         )
 
-        self.clf = LinearSVCWithLabelEncoding(LinearSVC())
+        self.clf = XGBClassifier(n_jobs=4)
```
marco-c, line 88:
```suggestion
        self.clf = XGBClassifier(n_jobs=utils.get_physical_cpu_count())
```

bugbug/models/duplicate.py
```diff
@@ -29,15 +30,28 @@ def fit(self, X, y):
 
 
 class DuplicateModel(BugCoupleModel):
-    def __init__(self, training_size=14000, lemmatization=False, cleanup_urls=True):
+    def __init__(self, training_size=2000, lemmatization=False, cleanup_urls=True):
```
marco-c, line 33: Let's revert this back to 14000

ayush-1506:
Made the changes. Will open a new PR adding XGBoost.

marco-c (approved):
Thanks!

PR #890, opened 2019-08-18
Fix is_first_affected_same when get_versions_statuses returns empty list

ayush-1506:
I'd been overlooking the case when get_version_statuses returns an empty list, and hence calling `min` on this gives an error.

### DIFF
--- a/bugbug/bug_features.py
+++ b/bugbug/bug_features.py
@@ -501,9 +501,13 @@ class is_same_target_milestone(couple_bug_feature):
 
 class is_first_affected_same(couple_bug_feature):
     def __call__(self, bugs, **kwargs):
-        return min(get_versions_statuses(bugs[0])[1]) == min(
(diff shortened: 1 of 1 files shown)


bugbug/bug_features.py
```diff
@@ -501,9 +501,13 @@ def __call__(self, bugs, **kwargs):
 
 class is_first_affected_same(couple_bug_feature):
     def __call__(self, bugs, **kwargs):
-        return min(get_versions_statuses(bugs[0])[1]) == min(
-            get_versions_statuses(bugs[1])[1]
-        )
+        version_status1 = get_versions_statuses(bugs[0])[1]
+        version_status2 = get_versions_statuses(bugs[1])[1]
+
+        if version_status1 != [] and version_status2 != []:
```
marco-c: `if len(version_status1) > 0 and len(version_status2) > 0` is nicer

Issue #612, opened 2019-06-21
Try BM25 as a similarity algorithm


marco-c:
We can take inspiration from https://github.com/marco-c/uplift-analysis/blob/master/bm25.py.

Depends on #611.

ayush-1506:
@marco-c The link doesn't seem to work.

marco-c:
@ayush-1506 it should be available now.

ayush-1506:
I'll try to see how this compares to other algorithms using https://pypi.org/project/rank-bm25/.

PR #891, opened 2019-08-18
Add BM25 similarity algorithm

ayush-1506:
Fixes : #612

This model seems to outperform all other models in terms of both speed and accuracy. The evaluation results are better than any other model, and evaluation on 6k bugs takes about 40 seconds (as compared to 55 minutes for word2vecwmdrelax model).


ayush-1506:
Results:
```
Recall @ 1: 44.44444444444444%
Recall @ 5: 50.0%
Recall @ 10: 50.641025641025635%
Precision @ 1: 28.676470588235293%
Precision @ 5: 9.50980392156863%
Precision @ 10: 5.8088235294117805%
Recall: 50.641025641025635%
Precision: 6.454248366013073%
MAP@k : 34.1643694514843%
```

Issue #522, opened 2019-05-30
Consider more duplicate bugs by generating all combinations of bugs in a duplicate group


marco-c:
See https://github.com/mozilla/bugbug/pull/484#discussion_r288724543.

PR #894, opened 2019-08-19
Use more combinations to get duplicate pairs

ayush-1506:
Fixes: #522

### DIFF
--- a/bugbug/models/duplicate.py
+++ b/bugbug/models/duplicate.py
@@ -4,6 +4,7 @@
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import random
+from itertools import combinations
 
 from sklearn.calibration import CalibratedClassifierCV
 from sklearn.compose import ColumnTransformer
@@ -84,6 +85,8 @@ class DuplicateModel(BugCoupleModel):
         duplicates_num = 0
         for bug_data in bugzilla.get_bugs():
             bug_id = bug_data["id"]
+            local_duplicates = []
+
             if bug_id not in all_ids:
                 continue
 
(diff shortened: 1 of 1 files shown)


bugbug/models/duplicate.py
```diff
@@ -94,10 +97,17 @@ def get_labels(self):
                 if duplicate_bug_id not in all_ids:
                     continue
 
+                local_duplicates.append(duplicate_bug_id)
                 duplicate_ids.append(duplicate_bug_id)
 
                 if duplicates_num < self.num_duplicates:
                     classes[(bug_id, duplicate_bug_id)] = 1
```
marco-c: Do we still need this now that we have the for loop?

marco-c: Or is this particular tuple already set to 1 in the for loop?

marco-c: I guess it isn't right now, but you could do that simply by initializing `local_duplicates` this way:
```
local_duplicates = [bug_id]
```

bugbug/models/duplicate.py
```diff
@@ -84,6 +85,8 @@ def get_labels(self):
         duplicates_num = 0
         for bug_data in bugzilla.get_bugs():
             bug_id = bug_data["id"]
+            local_duplicates = []
```
marco-c:
```suggestion
            current_duplicates = []
```

bugbug/models/duplicate.py
ayush-1506: We already have `(bug_id, duplicate_bug_id)` = 1. We need to add all pairs of duplicate_bug_ids to this set. Hence, I'm filling local_duplicates with only the duplicate_ids and using itertools.combinations to generate more labels. Is this not right?

marco-c: Yeah, I was just suggesting to have:
```
local_duplicates = [bug_id]

...

if duplicates_num < self.num_duplicates:
    for duplicate_pair in combinations(local_duplicates, 2):
        classes[tuple(duplicate_pair)] = 1
        duplicates_num += 1
```
as an alternative implementation, so that you can remove the `classes[(bug_id, duplicate_bug_id)] = 1` line.

ayush-1506: Ah, I see. Will do 👍

Issue #822, opened 2019-08-01
Make generate_sheet support multiclass models


marco-c:
At the moment, generate_sheet is assuming a binary model.

PR #896, opened 2019-08-20
Support for Multiclass models in generate_sheet.py

SanthoshBala18:
I have added support for multiclass models in generate_sheet.py. Please do review and let me know if any changes are to be made. This commit resolves issue #822

### DIFF
--- a/scripts/generate_sheet.py
+++ b/scripts/generate_sheet.py
@@ -20,7 +20,6 @@ def generate_sheet(model_name, token, days, threshold):
 
     model_class = get_model_class(model_name)
     model = model_class.load(model_file_name)
-    _, all_labels = model.get_labels()
 
     today = datetime.utcnow()
     start_date = today - timedelta(days)
@@ -37,7 +36,7 @@ def generate_sheet(model_name, token, days, threshold):
         probability = p[0]
         if len(probability) > 2:
             index = np.argmax(probability)
-            prediction = all_labels[index]
+            prediction = model.class_names[index]
         else:
             prediction = "y" if probability[1] >= threshold else "n"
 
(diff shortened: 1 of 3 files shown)


chidauri:
@SanthoshBala18 can you post the results after this change please?

SanthoshBala18:
Is there any pre-trained models available to use.? because it is getting really difficult to train the model with less memory.

chidauri:
> Is there any pre-trained models available to use.? because it is getting really difficult to train the model with less memory.

You can run "bug_classifier" script, it will download a pre-trained model if available . Else you can use itertools to reduce the training set, like for defectenhancement task model you can modify the following line:
 https://github.com/mozilla/bugbug/blob/e2810c24023be4d530f114118c597a93d0fafe47/bugbug/models/defect.py#L248
`itertools.islice(classes.items(), 20000)` and then train defectenhancementtask model

SanthoshBala18:
[sheets.zip](https://github.com/mozilla/bugbug/files/3536026/sheets.zip)
@chidauri Hi.. i have attached the sheets generated using generate_sheet.py. 2 models(stepstoreproduce and tracking) are Binary. Other 2 models(defecttaskenhancement and component) are Multi class models. Please do review and let me know if any changes are to be made.

scripts/generate_sheet.py
```diff
@@ -18,6 +20,7 @@ def generate_sheet(model_name, token, days, threshold):
 
     model_class = get_model_class(model_name)
     model = model_class.load(model_file_name)
+    _, all_labels = model.get_labels()
```
chidauri: I think we can use `model.class_names` here as class_names are already sorted, this way we don't need to sort them in the models themselves.

Issue #40, opened 2018-12-16
Try using multitask classification


marco-c:
See also #8 and #39.

Some of our classifiers are operating on the same data, it might make sense to use multitask classification.

PR #898, opened 2019-08-21
[WIP] Multi Task Learning

chidauri:
#40 
**Currently implemented for bugtype model only**

### DIFF
--- a/bugbug/models/__init__.py
+++ b/bugbug/models/__init__.py
@@ -11,6 +11,7 @@ MODELS = {
     "backout": "bugbug.models.backout.BackoutModel",
     "bug": "bugbug.model.BugModel",
     "bugtype": "bugbug.models.bugtype.BugTypeModel",
+    "bugtype_nn": "bugbug.models.bugtype_nn.BugTypeNNModel",
     "component": "bugbug.models.component.ComponentModel",
     "component_nn": "bugbug.models.component_nn.ComponentNNModel",
     "defect": "bugbug.models.defect.DefectModel",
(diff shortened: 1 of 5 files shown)


ayush-1506:
@chidauri Why are you unable to run this on your system? I'd love to see how this compares to our current bugtype model. You could try running on colab, or reducing the size of the dataset you're training on.

chidauri:
@ayush-1506 @marco-c I am only able to run it for 7000 bugs , if i go more, memory overflow occurs :(
**For 7000 bugs:**
```
6300/6300 [==============================] - 21s 3ms/step - loss: 0.1978 - acc: 0.9311
Epoch 2/2
6300/6300 [==============================] - 12s 2ms/step - loss: 0.1066 - acc: 0.9769
```

ayush-1506:
I'm no expert, but an accuracy of 0.97 sounds too good to be true.

chidauri:
> I'm no expert, but an accuracy of 0.97 sounds too good to be true.

yeah it is probably overfitting(probably due to such a small training set ), so first of all we need to focus on reducing memory requirements so that we can train on larger dataset , i tried component_nn.py , that model suffers from the same issue

ayush-1506:
I was able to run models on the entire DB on colab (though they weren't deep learning models). I guess you could wait for Marco to return so he could test it on the 64 gb RAM cluster he has.

marco-c (changes requested):
Will need to be rebased but I will keep this open in case we want to pick it up again.

PR #903, opened 2019-08-21
Update model.py

Sejal-G:
Updating model.py by removing feature_names parameter from print_feature_importances and the caller function.


ayush-1506:
You can make changes in the same PR, no need to open multiple PR's. Also, looks like lint tests are also failing. Setup pre-commit to ensure that the code is properly linted before committing.

chidauri:
Duplicate PR.

PR #904, opened 2019-08-22
Fix metrics check

Lothiraldan:


Lothiraldan:
Calling `python scripts/check_all_metrics.py --task-id KZrD5K4WTeOII6AZC2U1JQ metrics metric-outputs` now works as expected

scripts/check_all_metrics.py
```diff
@@ -59,10 +59,10 @@ def get_model_name(queue, task_id: str):
             model_name = route.split(".")[4]  # model_name = "train_component"
             return model_name[6:]
 
-    # Raise an exception if no matching route was found, this can happens when
-    # the current task has a dependency to a non-training task or if the route
+    # Show a warning if no matching route was found, this can happens when the
```
marco-c:
```suggestion
    # Show a warning if no matching route was found, this can happen when the
```

requirements.txt
```diff
@@ -10,7 +10,7 @@ python-hglib==2.6.1
 shap==0.29.3
 pandas==0.25.0
 rs_parsepatch==0.2.2
-tqdm==4.33.0
+tqdm~=4.32.2
```
marco-c: Is this change necessary?

marco-c:
```suggestion
tqdm==4.33.0
```

marco-c: I guess not, so I've removed it. Feel free to submit another PR if that's needed for some reason.

Issue #715, opened 2019-07-16
Try using Spacy lemmatization for similarity


marco-c:

ayush-1506:
Results:

Before changes:
```
Recall @ 1: 35.43689320388349%
Recall @ 5: 38.83495145631068%
Recall @ 10: 39.07766990291262%
Precision @ 1: 0.546448087431694%
Precision @ 5: 7.431693989071022%
Precision @ 10: 4.398907103825126%
Recall: 39.07766990291262%
Precision: 4.398907103825136%
MAP@k : 15.2855841790268%
```

After changes:
```
Recall @ 1: 31.55339805825243%
Recall @ 5: 34.95145631067961%
Recall @ 10: 35.19417475728155%
Precision @ 1: 0.546448087431694%
Precision @ 5: 6.612021857923482%
Precision @ 10: 3.961748633879772%
Recall: 35.19417475728155%
Precision: 3.9617486338797816%
MAP@k : 13.97627721398213%
```

PR #908, opened 2019-08-23
Use default spacy lemmatizer

ayush-1506:
Fixes: #715

### DIFF
--- a/bugbug/similarity.py
+++ b/bugbug/similarity.py
@@ -36,11 +36,13 @@ try:
     from nltk.corpus import stopwords
     from nltk.stem.porter import PorterStemmer
     from nltk.tokenize import word_tokenize
+    import spacy
     from wmd import WMD
 except ImportError:
     raise ImportError(OPT_MSG_MISSING)
 
 nltk.download("stopwords")
+nlp = spacy.load("en_core_web_sm")
 
 REPORTERS_TO_IGNORE = {"intermittent-bug-filer@mozilla.bugs", "wptsync@mozilla.bugs"}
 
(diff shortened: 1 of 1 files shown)


ayush-1506:
Results:

Before changes:
```
Recall @ 1: 35.43689320388349%
Recall @ 5: 38.83495145631068%
Recall @ 10: 39.07766990291262%
Precision @ 1: 0.546448087431694%
Precision @ 5: 7.431693989071022%
Precision @ 10: 4.398907103825126%
Recall: 39.07766990291262%
Precision: 4.398907103825136%
MAP@k : 15.2855841790268%
```

After changes:
```
Recall @ 1: 31.55339805825243%
Recall @ 5: 34.95145631067961%
Recall @ 10: 35.19417475728155%
Precision @ 1: 0.546448087431694%
Precision @ 5: 6.612021857923482%
Precision @ 10: 3.961748633879772%
Recall: 35.19417475728155%
Precision: 3.9617486338797816%
MAP@k : 13.97627721398213%
```

marco-c:
Since it appears to be worse, let's make it an option.

bugbug/similarity.py
```diff
@@ -78,21 +78,22 @@ def get_text(self, bug):
...
 
         text = re.sub("[^a-zA-Z0-9]", " ", text)
 
-        ps = PorterStemmer()
+        text = [word.lemma_ for word in nlp(text)]
 
-        tokenized_text = (
-            word_tokenize(text.lower()) if self.nltk_tokenizer else text.lower().split()
-        )
+        ps = PorterStemmer()
```
marco-c: When we lemmatize, we should not stem.

bugbug/similarity.py
```diff
@@ -78,22 +79,29 @@ def __init__(self, cleanup_urls=True, nltk_tokenizer=False):
     def get_text(self, bug):
         return "{} {}".format(bug["summary"], bug["comments"][0]["text"])
 
-    def text_preprocess(self, text, join=False):
+    def text_preprocess(self, text, spacy_lemmatization=False, join=False):
+        nlp = spacy.load("en_core_web_sm")
```
marco-c: It'd be better to do the loading only once (at the top of the file in a global variable so you can use it everywhere), and not for every call to text_preprocess

bugbug/similarity.py
```diff
@@ -78,22 +79,29 @@ def __init__(self, cleanup_urls=True, nltk_tokenizer=False):
     def get_text(self, bug):
         return "{} {}".format(bug["summary"], bug["comments"][0]["text"])
 
-    def text_preprocess(self, text, join=False):
+    def text_preprocess(self, text, spacy_lemmatization=False, join=False):
```
marco-c:
```suggestion
    def text_preprocess(self, text, lemmatization=False, join=False):
```

marco-c: The fact that it is Spacy is just an implementation detail

Issue #712, opened 2019-07-16
Try LDA as a similarity algorithm


marco-c:
It might work better in our case as the first comments are often pretty long, maybe WMD doesn't work so well in those cases.

PR #911, opened 2019-08-26
LDA with hellinger distance similarity

ayush-1506:
Fixes: #712

### DIFF
--- a/bugbug/similarity.py
+++ b/bugbug/similarity.py
@@ -28,6 +28,8 @@ try:
     import gensim
     from gensim import models, similarities
     from gensim.models import Word2Vec, WordEmbeddingSimilarityIndex, TfidfModel
+    from gensim.models.ldamodel import LdaModel
+    from gensim.matutils import sparse2full
     from gensim.similarities import SoftCosineSimilarity, SparseTermSimilarityMatrix
     from gensim.corpora import Dictionary
     from nltk.corpus import stopwords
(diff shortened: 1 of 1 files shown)


ayush-1506:
Update: Changed similarity metric to Hellinger distance (it's supposed to work better in case of probability distributions.)

ayush-1506:
Results:
```
Recall @ 1: 15.384615384615385%
Recall @ 5: 17.52136752136752%
Recall @ 10: 17.735042735042736%
Precision @ 1: 0.7352941176470588%
Precision @ 5: 2.941176470588233%
Precision @ 10: 2.034313725490193%
Recall: 17.735042735042736%
Precision: 2.034313725490196%
MAP@k : 6.410626750700279%
```

bugbug/similarity.py
```diff
@@ -526,6 +528,51 @@ def get_distance(self, query1, query2):
...
+
+        distances.sort(key=lambda v: v[1])
+
+        return [distance[0] for distance in distances[:10]]
+
+    def get_distance(self, query1, query2):
+        raise NotImplementedError
+
+
 model_name_to_class = {
```
marco-c: You'll need to add it to `model_name_to_class`, right? (Same for BM25, but you'll need to do that in another PR).

ayush-1506: Yes, I'll add them in a new PR once this is merged.

Issue #824, opened 2019-08-01
Track feature importance along with training metrics


marco-c:
If a feature suddenly becomes important when it was not important before, or when a feature which was important suddenly disappers, we have some important change in the dataset that we need to manually double check.

PR #915, opened 2019-08-29
Track feature importances in report. #824

Lothiraldan:


bugbug/model.py
```diff
@@ -285,6 +285,24 @@ def print_feature_importances(
...
+
+        feature_report = {}
+
+        for i, feature_name in enumerate(top_feature_names):
+            for class_name, imp_values in important_features["classes"].items():
+                feature_report.setdefault(feature_name, {})[class_name] = float(
+                    imp_values[1][i]
+                )
+
+        return feature_report
```
marco-c: We could save (and, in the next PR, check) the "average" features too in the report.

bugbug/model.py
```diff
@@ -285,6 +285,24 @@ def print_feature_importances(
...
+        top_feature_names = [
+            feature_names[int(index)]
+            for importance, index, is_pos in important_features["average"]
+        ]
+
+        feature_report = {}
+
+        for i, feature_name in enumerate(top_feature_names):
+            for class_name, imp_values in important_features["classes"].items():
+                feature_report.setdefault(feature_name, {})[class_name] = float(
```
marco-c: For the check that we will implement in the future, it might be better to have the dict be:
class_name -> feature_name -> value
rather than
feature_name -> class_name -> value

Since what we want to check is whether a feature which was not important before suddenly became important, or whether a feature which was important suddenly became not important.

marco-c: Your call here, I don't have a preference ;)

PR #928, opened 2019-09-04
Update dockerfile.commit_retrieval to install Mercurial without pip (#922)

### LINKED ISSUE #922: Install Mercurial without pip
Like we are doing in https://github.com/mozilla/code-coverage/blob/master/bot/Dockerfile, we should try changing infra/dockerfile.commit_retrieval to install Mercurial without pip.

dotslash21:
Updated infra/dockerfile.commit_retrieval to install Mercurial from the official homepage (https://www.mercurial-scm.org/) instead of using pip.

Fixes #922.


marco-c:
The docker build is still failing. You can test it locally (please do before you update the PR again).

infra/dockerfile.commit_retrieval
```diff
@@ -2,18 +2,22 @@ FROM mozilla/bugbug-base:latest
...
 
-# Mercurial need Python2 :(
+# Setup mercurial from its own website, without install pip2 which has a lot of dependencies
+ARG MERCURIAL_VERSION="5.1"
+
 # git and git-hyper-blame (from depot_tools) are required by the annotate pipeline.
 RUN apt-get update && \
-    apt-get install -y python python-pip git curl && \
-    python2 -m pip install --disable-pip-version-check --no-cache-dir mercurial==5.1 && \
+    apt-get install --no-install-recommends -y python python-dev python-docutils build-essential git curl && \
```
marco-c: We can use python2-minimal instead of python to avoid installing a lot of dependencies

marco-c: build-essential has many dependencies, can't you install just gcc instead like https://github.com/mozilla/code-coverage/blob/master/bot/ci/bootstrap.sh does?

infra/dockerfile.commit_retrieval
```diff
@@ -2,18 +2,22 @@ FROM mozilla/bugbug-base:latest
...
+    make -C /opt/mercurial-$MERCURIAL_VERSION/ install PYTHON=/usr/bin/python2 && \
+    hg debuginstall && \
     hg clone -r 709c897f2444 https://hg.mozilla.org/hgcustom/version-control-tools /version-control-tools/ && \
     rm -r /version-control-tools/.hg /version-control-tools/ansible /version-control-tools/docs /version-control-tools/testing && \
     git clone https://chromium.googlesource.com/chromium/tools/depot_tools.git /depot_tools && \
     cd /depot_tools && git -c advice.detachedHead=false checkout 73065b2067bc1516728a1e6251df751db98f0fba && git apply /root/git_hyper_blame_patch && cd .. && \
     rm -r /depot_tools/.git /depot_tools/recipes /depot_tools/tests /depot_tools/man /depot_tools/testing_support /depot_tools/win_toolchain /depot_tools/bootstrap /depot_tools/fetch_configs && \
     curl -L https://github.com/mozilla/rust-code-analysis/releases/download/v0.0.2/rust-code-analysis-linux-x86_64.tar.bz2 | tar -C /usr/local/bin -xjv && \
-    apt-get purge -y python-pip curl && \
+    apt-get purge -y build-essential curl && \
```
marco-c: You should remove python-dev as it will needlessly occupy space otherwise.

infra/dockerfile.commit_retrieval
marco-c: Why is pyhton-docutils needed? We are not installing it in https://github.com/mozilla/code-coverage/blob/master/bot/ci/bootstrap.sh.

infra/dockerfile.commit_retrieval
```diff
@@ -2,18 +2,22 @@ FROM mozilla/bugbug-base:latest
...
+ARG MERCURIAL_VERSION="5.1"
+
 # git and git-hyper-blame (from depot_tools) are required by the annotate pipeline.
 RUN apt-get update && \
-    apt-get install -y python python-pip git curl && \
-    python2 -m pip install --disable-pip-version-check --no-cache-dir mercurial==5.1 && \
+    apt-get install --no-install-recommends -y python python-dev python-docutils build-essential git curl && \
+    curl -L https://www.mercurial-scm.org/release/mercurial-$MERCURIAL_VERSION.tar.gz | tar -C /opt -xvz && \
+    make -C /opt/mercurial-$MERCURIAL_VERSION/ install PYTHON=/usr/bin/python2 && \
+    hg debuginstall && \
```
marco-c: Why are you installing in this way instead of just copying from https://github.com/mozilla/code-coverage/blob/master/bot/ci/bootstrap.sh? :P

infra/dockerfile.commit_retrieval
dotslash21: Ok will do that in the next edit.

dotslash21: Ok, will do that in the next edit.

infra/dockerfile.commit_retrieval
dotslash21: Ok, will do that in the next edit.

infra/dockerfile.commit_retrieval
dotslash21: Using the make method required python-docutils, otherwise, the compilation fails partway and the  docker build fails.

dotslash21: I've figured out how to solve the error that runnning "python /opt/mercurial-$MERCURIAL_VERSION/setup.py install" gives. So I'll be reverting back to the old method of installation and won't be needing python-docutils in the future.

infra/dockerfile.commit_retrieval
dotslash21: I was using the python install method earlier but It was giving a mysterious error which I finally managed to solve. So I'll be changing this in the next edit.

dotslash21: I haven't come across the method of installing mercurial 5.1 using python other than here (https://www.mercurial-scm.org/wiki/UnixInstall).
The official documentation only had the instructions to install it via make, https://www.mercurial-scm.org/wiki/UnixInstall

infra/dockerfile.commit_retrieval
```diff
@@ -2,18 +2,21 @@ FROM mozilla/bugbug-base:latest
 
 COPY infra/git_hyper_blame_patch /root/git_hyper_blame_patch
 
-# Mercurial need Python2 :(
```
marco-c: This comment still applies, can you bring it back? (and move it to before `# git and git-hyper-blame ...`)

infra/dockerfile.commit_retrieval
```diff
@@ -2,18 +2,21 @@ FROM mozilla/bugbug-base:latest
...
+# Setup mercurial from its own website, without install pip2 which has a lot of dependencies
+ARG MERCURIAL_VERSION="5.1"
+
 # git and git-hyper-blame (from depot_tools) are required by the annotate pipeline.
 RUN apt-get update && \
-    apt-get install -y python python-pip git curl && \
-    python2 -m pip install --disable-pip-version-check --no-cache-dir mercurial==5.1 && \
+    apt-get install --no-install-recommends -y python-minimal python-dev gcc bzip2 git curl && \
```
marco-c: Specify the 2 after python, so that if the default python becomes 3, we will still be installing 2.

infra/dockerfile.commit_retrieval
dotslash21: Ok, Will do that in the next edit

infra/dockerfile.commit_retrieval
dotslash21: Ok, Will do that in the next edit.

marco-c:
Unfortunately it looks like the improvement in build time is negligible, and the size of the resulting image got bigger :(
We have two options:
1) Figure out if there is some package left out that is increasing the size of the image and remove it too;
2) Don't do this at all.

dotslash21:
I need to know what packages are absolutely required for the docker image, for example, I can see that Python 2, Git and Mercurial are kept installed in the dockerfile in question at the master branch. I couldn't find any relevant documentation about it. Meanwhile, I'll try to find potential optimizations.

dotslash21:
@marco-c the latest changes have made the build time much faster. I've timed it on my laptop and found the original dockerfile build time to be 14 minutes approximately and the new one 6 minutes approx. But the size increased from 1.04GB to 1.07GB. Can you please review it and confirm the same?

marco-c:
@dotslash21 the build time will be shorter, but given that we have caching it won't make much of a difference when you rebuild unless you change something in one of the layers. We change the layers pretty rarely though, so we don't care much about the build time.

dotslash21:
I guess using pip is more space-efficient method. But If I had any way of installing Mercurial 5.1 binaries rather than building from source then I could make the docker image size smaller. I can make a .deb package from source, but it'll require hosting of that package to access it during the build. So I guess this idea would be out of scope, what are your thoughts @marco-c?

marco-c:
> I guess using pip is more space-efficient method. But If I had any way of installing Mercurial 5.1 binaries rather than building from source then I could make the docker image size smaller. I can make a .deb package from source, but it'll require hosting of that package to access it during the build. So I guess this idea would be out of scope, what are your thoughts @marco-c?

Yeah this seems like too much work for a small improvement :)
We should rather try to find out why using pip is more space-efficient than installing directly.

marco-c:
Feel free to reopen or comment if you find out why using pip is more space-efficient than installing directly.

PR #930, opened 2019-09-05
Check metric long time

Lothiraldan:
Internet suggested that using a linear regression on a timeseries data is a
first good step to detect the trend on the data. It seems to works fine but I
wasn't unable to distinguish between the metric getting significantly worse or
just some noise.

Because the range of value is limited, the slope is also very small.

I updated the checking script to emit a warming message instead of failing so
we can tune it if needing before activating it.


Lothiraldan:
I have sent the code implementing the smoothing over 1 month of data and checking if the local trend is up or down. Checking the local trend seems to work well to avoid any past outlier to indefinitely impact the check.

Here are all the case where the previous threshold check was triggered (extracted from [output.txt](https://github.com/mozilla/bugbug/files/3579690/output.txt)):
```
WARNING:__main__:Metric 'test_accuracy_std' for model backout seems to be decreasing
WARNING:__main__:Last metric 'test_accuracy_std' for model backout is at least 5.000000% worse than the previous one
--
WARNING:__main__:Metric 'test_precision_std' for model backout seems to be decreasing
WARNING:__main__:Last metric 'test_precision_std' for model backout is at least 5.000000% worse than the previous one
--
WARNING:__main__:Metric 'test_accuracy_std' for model regression seems to be decreasing
WARNING:__main__:Last metric 'test_accuracy_std' for model regression is at least 5.000000% worse than the previous one
--
WARNING:__main__:Metric 'test_precision_std' for model regressionrange seems to be decreasing
WARNING:__main__:Last metric 'test_precision_std' for model regressionrange is at least 5.000000% worse than the previous one
--
WARNING:__main__:Metric 'test_accuracy_std' for model regressor seems to be decreasing
WARNING:__main__:Last metric 'test_accuracy_std' for model regressor is at least 5.000000% worse than the previous one
--
INFO:__main__:Metric 'test_precision_std' for model regressor seems to be increasing
WARNING:__main__:Last metric 'test_precision_std' for model regressor is at least 5.000000% worse than the previous one
--
WARNING:__main__:Metric 'recall' for model stepstoreproduce seems to be decreasing
WARNING:__main__:Last metric 'recall' for model stepstoreproduce is at least 5.000000% worse than the previous one
--
INFO:__main__:Metric 'test_accuracy_std' for model stepstoreproduce seems to be increasing
WARNING:__main__:Last metric 'test_accuracy_std' for model stepstoreproduce is at least 5.000000% worse than the previous one
--
INFO:__main__:Metric 'test_precision_std' for model stepstoreproduce seems to be increasing
WARNING:__main__:Last metric 'test_precision_std' for model stepstoreproduce is at least 5.000000% worse than the previous one
--
INFO:__main__:Metric 'test_recall_std' for model stepstoreproduce seems to be increasing
WARNING:__main__:Last metric 'test_recall_std' for model stepstoreproduce is at least 5.000000% worse than the previous one
```

Over 10 cases, the smoothed analysis detected an increasing local trend. In all cases it's concerning the std metric:

```
INFO:__main__:Metric 'test_precision_std' for model regressor seems to be increasing
WARNING:__main__:Last metric 'test_precision_std' for model regressor is at least 5.000000% worse than the previous one
--
INFO:__main__:Metric 'test_accuracy_std' for model stepstoreproduce seems to be increasing
WARNING:__main__:Last metric 'test_accuracy_std' for model stepstoreproduce is at least 5.000000% worse than the previous one
--
INFO:__main__:Metric 'test_precision_std' for model stepstoreproduce seems to be increasing
WARNING:__main__:Last metric 'test_precision_std' for model stepstoreproduce is at least 5.000000% worse than the previous one
--
INFO:__main__:Metric 'test_recall_std' for model stepstoreproduce seems to be increasing
WARNING:__main__:Last metric 'test_recall_std' for model stepstoreproduce is at least 5.000000% worse than the previous one
```

Here are some of the generated graph (also available in [metrics_out.zip](https://github.com/mozilla/bugbug/files/3579692/metrics_out.zip)).

Before smoothing: 
stepstoreproduce_test_recall_std_before_smoothing

After smoothing:
stepstoreproduce_test_recall_std_after_smoothing


It feels off to do this kind of analysis on the standard deviation of a metric, that might explains why the std is varying so much and why it doesn't work well.

@marco-c Does the output and generated graph are helpful to know when you need to dig and check on the model?

marco-c:
> It feels off to do this kind of analysis on the standard deviation of a metric, that might explains why the std is varying so much and why it doesn't work well.

What if we not do it for the std but just for mean? That ought to be at least a bit better.

calixteman:
@marco, hard to say.
AFAIK, the starting and ending points are always problematic since they've only short context (starting point has no past and ending point has no future).
You may have a spike for the ending point but the future will tell "ok it was just an artefact because the trend is good finally".
Maybe we should only analyze past and just say "Oh it was bad 2 weeks ago and the trend after is bad too" then we can check what happened 2 weeks ago because it introduced a regressing trend.
Anyway I'm not sure that's possible to have a kinda "real-time" feedback.
A way to "detect" spikes is to remove the trend (to have something almost flat), compute the mean and stddev and when a "value - guessed value (according to the trend) - mean" is X times greater than stddev then you've a spike.
But here we've only few values so...

marco-c:
We don't need to have a real-time feedback actually, so I'm thinking we could just leave the first part and the last part out (still considering them to smooth, but not consider them for "spike" detection). Would that work?

Lothiraldan:
I will try to remove the first and last point from the local min/max detection and see how it behave

Lothiraldan:
Here are the latest alerts:

```
WARNING:__main__:Metric 'precision' for model backout seems to be decreasing
WARNING:__main__:Metric 'precision' for model component seems to be decreasing
WARNING:__main__:Metric 'test_accuracy_mean' for model regressionrange seems to be decreasing
WARNING:__main__:Last metric 'test_accuracy_mean' for model regressionrange is at least 1.000000% worse than the previous one
WARNING:__main__:Metric 'test_recall_mean' for model regressionrange seems to be decreasing
WARNING:__main__:Metric 'precision' for model regressor seems to be decreasing
WARNING:__main__:Metric 'recall' for model regressor seems to be decreasing
WARNING:__main__:Metric 'test_precision_mean' for model regressor seems to be decreasing
WARNING:__main__:Metric 'test_recall_mean' for model regressor seems to be decreasing
WARNING:__main__:Metric 'recall' for model stepstoreproduce seems to be decreasing
WARNING:__main__:Last metric 'recall' for model stepstoreproduce is at least 1.000000% worse than the previous one
WARNING:__main__:Metric 'test_precision_mean' for model stepstoreproduce seems to be decreasing
```

Trying to exclude the first and last points for the analysis now raises those warnings:

```
WARNING:__main__:Metric 'test_recall_mean' for model backout seems to be decreasing
WARNING:__main__:Metric 'precision' for model defectenhancementtask seems to be decreasing
WARNING:__main__:Metric 'precision' for model regressionrange seems to be decreasing
WARNING:__main__:Metric 'recall' for model regressionrange seems to be decreasing
WARNING:__main__:Last metric 'test_accuracy_mean' for model regressionrange is at least 1.000000% worse than the previous one
WARNING:__main__:Metric 'test_accuracy_mean' for model regressor seems to be decreasing
WARNING:__main__:Metric 'test_precision_mean' for model regressor seems to be decreasing
WARNING:__main__:Metric 'precision' for model stepstoreproduce seems to be decreasing
WARNING:__main__:Metric 'recall' for model stepstoreproduce seems to be decreasing
WARNING:__main__:Last metric 'recall' for model stepstoreproduce is at least 1.000000% worse than the previous one
WARNING:__main__:Metric 'test_accuracy_mean' for model stepstoreproduce seems to be decreasing
WARNING:__main__:Metric 'test_recall_mean' for model stepstoreproduce seems to be decreasing
WARNING:__main__:Metric 'test_recall_mean' for model tracking seems to be decreasing
```

test_recall_mean for the backout model seems to follow a spike cycle:

backout_test_recall_mean_before_smoothing

backout_test_recall_mean_after_smoothing

For other models, excluding the last point seems to trigger false alarms:

regressionrange_precision_before_smoothing
regressionrange_precision_after_smoothing

regressionrange_recall_before_smoothing
regressionrange_recall_after_smoothing

For the same model, it seems to help ignore a potential bad last result:

regressionrange_test_accuracy_mean_before_smoothing

regressionrange_test_accuracy_mean_after_smoothing

@marco-c @calixteman I'm wondering if we would have more data, the analysis would be easier?

PR #931, opened 2019-09-05
Convert similarity_trainer to console scipt

ayush-1506:
And giving it an alias `bugbug-train-similarity`.

### DIFF
--- a/scripts/similarity_trainer.py
+++ b/scripts/similarity_trainer.py
@@ -35,8 +35,8 @@ def parse_args(args):
     return parser.parse_args(args)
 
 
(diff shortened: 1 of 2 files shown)


scripts/similarity_trainer.py
```diff
@@ -49,7 +49,3 @@ def main(args):
         )
 
     model.save()
-
-
-if __name__ == "__main__":
```
marco-c: You can keep this and make it call `main()`, like https://github.com/mozilla/bugbug/blob/master/scripts/trainer.py#L122.

Issue #935, opened 2019-09-05
Similarity trainer script should download the bugs DB


marco-c:
Similarly to https://github.com/mozilla/bugbug/blob/6181dd11af35c406da160f7c0cfad3bac0d12f51/scripts/regressor_finder.py#L204-L206.

PR #937, opened 2019-09-06
Download bugs DB in similarity_trainer

ayush-1506:
Fixes: #935

### DIFF
--- a/scripts/similarity_trainer.py
+++ b/scripts/similarity_trainer.py
@@ -6,10 +6,14 @@
 
 import argparse
 import sys
+from logging import INFO, basicConfig, getLogger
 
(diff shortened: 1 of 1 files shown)


scripts/similarity_trainer.py
```diff
@@ -6,10 +6,13 @@
...
 import argparse
 import sys
+from logging import getLogger
 
 from sklearn.feature_extraction.text import TfidfVectorizer
 
-from bugbug import similarity
+from bugbug import bugzilla, db, similarity
+
+logger = getLogger(__name__)
```
marco-c: Also add https://github.com/mozilla/bugbug/blob/6181dd11af35c406da160f7c0cfad3bac0d12f51/scripts/regressor_finder.py#L28

PR #952, opened 2019-09-13
Fix integration tests assertion. Fix 948

Lothiraldan:


marco-c:
I'm thinking that these kinds of assertions could be problematic. If we train only on 100 bugs, we might be training a bad model and predict wrong.

Issue #939, opened 2019-09-07
similarity task run, but didn't do anything


marco-c:
https://tools.taskcluster.net/groups/EtQ7qTyZTEOtwTxBsnTA0A/tasks/SLjLm837QmeXAEQaW45-wA/runs/0/logs/public%2Flogs%2Flive.log

The artifact is not available. I think the training didn't happen at all.

@ayush-1506 you can debug this locally, just build and run the docker image.
```
docker build -t mozilla/bugbug-base-nlp -f infra/dockerfile.base-nlp .
docker run -it --rm mozilla/bugbug-base-nlp bash
```
Then in the Docker container, you can run the command from the task (https://github.com/mozilla/bugbug/blob/029d93ada4221642f2bc65b1ce546ec1f0d40fec/infra/data-pipeline.yml#L481-L483):
```
bugbug-train-similarity --algorithm bm25
```

Another issue that we definitely need to solve is that the similarity module is building the `duplicates` map as soon as it is imported, and when it is imported the bugs database is not available yet (since we import at https://github.com/mozilla/bugbug/blob/029d93ada4221642f2bc65b1ce546ec1f0d40fec/scripts/similarity_trainer.py#L13, and download the DB later at https://github.com/mozilla/bugbug/blob/029d93ada4221642f2bc65b1ce546ec1f0d40fec/scripts/similarity_trainer.py#L47-L48).
We can simply move the `duplicates` map building to the `evaluation` function, since that's the only user of the map.

marco-c:
Unfortunately it still happened :(

https://tools.taskcluster.net/groups/FECJFy8CTIy-ZCrAt8jgIA/tasks/fBO2GNGbRC2sSwAVcKkuOQ/runs/0/logs/public%2Flogs%2Flive.log

marco-c:
Fixed by 16e36cb54b45058352dd202daf116feb0f0d63db.

PR #966, opened 2019-09-19
Move duplicates map inside evaluation method

ayush-1506:
Addresses : #939


marco-c:
@ayush-1506 this is solving only the second part of that issue, right?

ayush-1506:
@marco-c Yes, the part which consists of running the docker image is taking too much time. Is there a way to run the model in the docker environment but on a reduced set of bugs? (Just to test if the model is being trained)

marco-c:
@ayush-1506 you can simply test by modifying https://github.com/mozilla/bugbug/blob/91db314240bd53f093218c47d5274e54d75f456d/bugbug/similarity.py#L559 to use just a subset of bugs.

ayush-1506:
Oh, I thought I might have to make changes in the dockerfile. Thanks!

PR #974, opened 2019-09-24
Move exp_queue from repository.py to utils.py 

### LINKED ISSUE #971: Move exp_queue from repository.py to utils.py
Since it is now being used in other places too and it is not directly related to repository.py, we should move exp_queue to utils.py.

horacix:
Fixes #971


tests/test_repository.py
```diff
@@ -13,7 +13,7 @@
 import responses
 from dateutil.relativedelta import relativedelta
 
-from bugbug import repository
+from bugbug import repository, utils
```
marco-c: Need to move the ExpQueue tests to test_utils.py, as test_repository.py is only meant to test repository.py functions.

marco-c:
Looks good, just one minor thing to fix.

tests/test_repository.py
horacix: Of course. I'll push an update tomorrow.

horacix: done

PR #976, opened 2019-09-25
Removed /tmp/bugbug

### LINKED ISSUE #970: Don't store /tmp/bugbug in Docker image
We are copying bugbug (https://github.com/mozilla/bugbug/blob/f749fc8b95ff86238beaed502d305bb950fddaff/infra/dockerfile.base#L12) and then installing it with pip (https://github.com/mozilla/bugbug/blob/f749fc8b95ff86238beaed502d305bb950fddaff/infra/dockerfile.base#L13). In the end we end up with two copies of the bugbug source code, one in /tmp/bugbug and one in /usr/lib.

imskr:
Removed the command to not create copy of it. Fixes #970


imskr:
@marco-c I will fix it & update & also I will test before updating PR.

imskr:
@marco-c I'm can't run `docker build -t mozilla/bugbug-base -f infra/dockerfile.base` 

Error:
"docker build" requires exactly 1 argument

marco-c:
@imskr unfortunately fixing this issue does require a bit of Docker knowledge. If you don't even know how to run it, it'll be really hard to fix this. I'd suggest coming back to the issue after a bit of reading and experimentation with Docker.

infra/dockerfile.base
```diff
@@ -11,3 +11,4 @@ RUN apt-get update && \
 
 COPY . /tmp/bugbug
 RUN pip install --disable-pip-version-check --no-cache-dir /tmp/bugbug
+RUN rm -rf /tmp/bugbug
```
marco-c: If you rm the directory after another Docker layer, the size of the image won't be affected, so this is not useful.

imskr: If I use `RUN pip install --disable-pip-version-check --no-cache-dir /tmp/bugbug && rm -rf /tmp/bugbug`

This will not store it. Am I right?

marco-c: No, it will still be stored, as you are storing it in the previous layer.

PR #981, opened 2019-09-26
Ensure that there is no stale data when running the integration test

Lothiraldan:


marco-c:
It seems like it's working! Could you revert 44394fe so I can merge this?

PR #986, opened 2019-09-27
Compress and save similarity models too

ayush-1506:
Fixes: #939


scripts/similarity_trainer.py
```diff
@@ -59,6 +61,8 @@ def main():
         )
 
     model.save()
```
marco-c: Make `save` return the path to the model, so you don't have to re-create it

Issue #438, opened 2019-05-15
Add basic tests for defect_enhancement_task model


ayush-1506:

hmrishavbandy:
Can you please assign this issue to me? Ma a noobie and would like some help alongside my contribution

marco-c:
@hmrishavbandy issues are assigned only when there is a PR open to fix them. You can feel free to work on any issue which you see unassigned and for which there is no PR linked, there is no need to be assigned first.

marco-c:
In order to fix this issue, you'll need to do something really similar to https://github.com/mozilla/bugbug/blob/master/tests/test_tracking.py, but for the defect/enhancement/task model.

hmrishavbandy:
@marco-c thanks a lot.. I'll be working on this now

hmrishavbandy:
Can anyone please elaborate on what is meant by "basic" tests in here?I am getting the concept of tests with respect to the model specified but  what tests are classfied as basic tests? Thanks for your help!

marco-c:
@hmrishavbandy see my previous comment, it contains an example of a basic test.

hmrishavbandy:
@marco-c I am probably slow, but tests as in using assert? If that is so , then in https://github.com/mozilla/bugbug/blob/master/tests/test_tracking.py , I didn't get the place where you use the numbers in the square brackets. Thanks for your patience !

marco-c:
@hmrishavbandy you are going to have to investigate a bit yourself, that's the nice part :P

hmrishavbandy:
oh! Okay... I'll surely give my best :smile:

PR #1019, opened 2019-10-13
Add basic tests for defect_enhancement_task model

con-cat:
Addresses #438


tests/test_defect_enhancement_task.py
```diff
@@ -0,0 +1,14 @@
...
+
+from bugbug.models.defect_enhancement_task import DefectEnhancementTaskModel
+
+
+def test_get_defect_enhancement_task_labels():
+    model = DefectEnhancementTaskModel()
+    classes, _ = model.get_labels()
+    assert classes[1042414] == "defect"
+    assert classes[1531080] == "task"
+    assert classes[1042734] != "enhancement"
```
marco-c: Try to find a bug which is an enhancement instead

con-cat:
I've added a bug that's an enhancement to bugs.json in the fixtures and updated the test to include that.

marco-c:
> I've added a bug that's an enhancement to bugs.json in the fixtures and updated the test to include that.

There should already by some enhancement bug, can you remove your addition and just use one that is already there?

con-cat:
It doesn't seem to be finding any enhancement bugs otherwise - here's the stdout call from the test with the existing data and a print statement added for classes:
```
---------------------------- Captured stdout call ----------------------------
14 defects
0 enhancements
2 tasks
{1042414: 'defect', 1042734: 'defect', 1046231: 'defect', 1047180: 'defect', 1053944: 'defect', 1087488: 'defect', 518272: 'defect', 528988: 'defect', 1037762: 'defect', 1042096: 'defect', 1488307: 'task', 1488310: 'defect', 1531080: 'task', 1388990: 'defect', 1325215: 'defect', 1364870: 'defect'}
```

marco-c:
Oh, OK, that's unexpected :/
So can you add just the one you need, instead of two?

con-cat:
I've now got just one new bug in the json file.

PR #1029, opened 2019-10-19
Switch dockerfile.base to alpine

dotslash21:
**Fixes Issue:** #292 
**Old Image Size:** 907MB
**New Image Size:** 780MB (127MB Decrement!)


infra/dockerfile.base
```diff
@@ -1,13 +1,12 @@
-FROM python:3.7.4-slim
+FROM alpine:latest
```
marco-c: There is a python alpine image which has Python already installed, I think it's preferable to use that one.

dotslash21:
@marco-c, Yeah It seems that I need to change the `dockerfile.base-nlp` and `dockerfile.commit_retrieval` to work with the new base image.

infra/dockerfile.base
dotslash21: Changed in the latest commit.

dotslash21:
It seems the dockerfiles under http_service also needs to be checked for any bugs due to the switch :(

marco-c:
As discussed in #292, this is not really worth it given the small win in size compared to the Debian-based images.

PR #1032, opened 2019-10-21
 While using pip in .taskcluster.yml and in docker files, added options --disable-pip-version-check, --no-cache-dir and --quiet

### LINKED ISSUE #1028: Always use --no-cache-dir option for pip
Whenever we are using pip in .taskcluster.yml and in docker files, we should use the options `--disable-pip-version-check`, `--no-cache-dir` and `--quiet`.

snowmanstark:
Closes #1028

### DIFF
--- a/http_service/Dockerfile
+++ b/http_service/Dockerfile
@@ -4,7 +4,7 @@ FROM mozilla/bugbug-base:$BUGBUG_VERSION
 
 COPY requirements.txt /code/http_service/
 
-RUN pip install -r /code/http_service/requirements.txt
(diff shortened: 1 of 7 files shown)


http_service/Dockerfile
```diff
@@ -4,7 +4,7 @@ FROM mozilla/bugbug-base:$BUGBUG_VERSION
 
 COPY requirements.txt /code/http_service/
 
-RUN pip install -r /code/http_service/requirements.txt
+RUN pip install --disable-pip-version-check --quiet --no-cache-dir-r /code/http_service/requirements.txt
```
marco-c: There's an error here, `--no-cache-dir-r` should be `--no-cache-dir -r`

Issue #1013, opened 2019-10-09
Identify spam/fake bugs


emceeaich:
We get too many spam bugs filed by students at schools with instructors telling them to learn how to use Bugzilla by filing junk bugs here. 

Most of the junk bugs have been moved to the Invalid Bugs::General component. 

Can we add a model we can run against incoming bugs that can mark as "potential spam" bug and divert to a moderator for review?

marco-c:
Yes, I think it should be feasible! We have plenty of "good" bugs (we can just use past RESOLVED FIXED bugs as examples of "good" bugs), and we can get "bad" bugs by using "Invalid Bugs::General" (hoping there are enough to train the model).

ayush-1506:
This shouldn't be very tough to implement. @marco-c Please let me know if I should go ahead with this. (I noticed the good-first-bug label, so maybe I should let someone else work on this?)

m-agboola:
@marco-c I'm interested in working on this if no one is assigned to it

marco-c:
@ayush-1506 you can work on it, don't worry (I mark them as good-first-issues because they can be worked on in isolation, but it doesn't mean only newcomers can work on them).

@m-agboola you can work on it too, any issue is unassigned until there is a PR open to fix it. It is not the simplest issue though, you might want to start with another one.

m-agboola:
Thanks @marco-c I understand and would give it a try!

ayush-1506:
@marco-c Correct me if I'm wrong:
Good bugs are the ones with `STATUS` as `FIXED` or `RESOLVED`.
Bad bugs are the ones where `Invalid Bugs` was added at any time in the history, right?

If so, there aren't a lot of bad bugs. Maybe we should consider other indicative fields too?

marco-c:
I think the problem is simply that we are not downloading the bad bugs in our DB, as "Invalid Bugs" is not in the list of considered components at https://github.com/mozilla/bugbug/blob/a36edbbe14bd1590d42ac5b171a17c894c066301/bugbug/bugzilla.py#L25.

ayush-1506:
@marco-c So we'd need to add it to the products, regenerate the bugs db and upload it to tc, right? Should this be a two step PR? (First one adding the label to the products and the second one implementing the model)

marco-c:
It can be a two step PR or a single PR with two separate commits.
[...]

PR #1033, opened 2019-10-22
[WIP] Spam bugs classifier

ayush-1506:
Fixes: #1013


bugbug/bugzilla.py
```diff
@@ -39,6 +39,7 @@
     "Firefox Health Report",
     # 'Focus',
     # 'Hello (Loop)',
+    "Invalid Bugs::General",
```
marco-c: Are you sure this is working?
This is a list of products, the product should be just "Invalid Bugs" ("General" is a component).

bugbug/models/spambug.py
```diff
@@ -0,0 +1,91 @@
...
+from sklearn.feature_extraction import DictVectorizer
+from sklearn.pipeline import Pipeline
+
+from bugbug import bug_features, bugzilla, feature_cleanup
+from bugbug.model import BugModel
+
+
+class SpamBugModel(BugModel):
+    def __init__(self, lemmatization=False):
+        BugModel.__init__(self, lemmatization, commit_data=True)
```
marco-c: We shouldn't need the `commit_data` for this model.

bugbug/models/spambug.py
```diff
@@ -0,0 +1,91 @@
...
+
+            # Spam bugs
+            elif (
+                bug_data["product"] == "Invalid Bugs"
+                and bug_data["component"] == "General"
+            ):
+                classes[bug_id] = 1
+
+        return classes, [0, 1]
+
```
marco-c: We have to rollback to the bug opening time, as we want to detect spam bugs as soon as they are filed

bugbug/bugzilla.py
ayush-1506: Ah, sorry forgot to update this.

marco-c:
We should also make sure we don't use these invalid bugs in other models. Perhaps add an argument `include_invalid` (false by default) to the get_bugs function.

bugbug/bugzilla.py
```diff
@@ -75,8 +76,16 @@ def get_bug_fields():
     return r.json()["fields"]
 
 
-def get_bugs():
-    return db.read(BUGS_DB)
+def get_bugs(include_invalid=False):
+    bugs = db.read(BUGS_DB)
+    for bug in bugs:
```
marco-c: You should be able to implement it in a more concise way, something like:
```Python
yield from (bug for bug in bugs if include_invalid or bug["product"] != "Invalid Bugs"
```
Note: we have to exclude all bugs from "Invalid Bugs", not only "Invalid Bugs::General".

bugbug/bugzilla.py
```diff
@@ -75,8 +76,16 @@ def get_bug_fields():
     return r.json()["fields"]
 
 
-def get_bugs():
-    return db.read(BUGS_DB)
+def get_bugs(include_invalid=False):
```
marco-c: Please add a test in tests/test_bugzilla.py to make sure this is OK.

bugbug/models/spambug.py
marco-c: Don't forget about this.

marco-c: Ah you did it, OK!

bugbug/models/spambug.py
```diff
@@ -0,0 +1,93 @@
...
+from bugbug import bug_features, bugzilla, feature_cleanup
+from bugbug.model import BugModel
+
+
+class SpamBugModel(BugModel):
+    def __init__(self, lemmatization=False):
+        BugModel.__init__(self, lemmatization)
+
+        self.cross_validation_enabled = False
```
marco-c: Why did you disable cross validation? Do we have too many bugs?

bugbug/models/spambug.py
```diff
@@ -0,0 +1,93 @@
...
+from imblearn.under_sampling import RandomUnderSampler
+from sklearn.compose import ColumnTransformer
+from sklearn.feature_extraction import DictVectorizer
+from sklearn.pipeline import Pipeline
+
+from bugbug import bug_features, bugzilla, feature_cleanup
+from bugbug.model import BugModel
+
+
+class SpamBugModel(BugModel):
```
marco-c: Can you paste here the current results?

bugbug/models/spambug.py
ayush-1506: I figured it might be faster to test the model that way.

ayush-1506:
@marco-c I unable to run the bug_retriever.py script since I always end up with `requests.exceptions.SSLError`, irrespective of the network. Not sure why this is happening though.

bugbug/models/spambug.py
```diff
@@ -0,0 +1,93 @@
...
+    def __init__(self, lemmatization=False):
+        BugModel.__init__(self, lemmatization)
+
+        self.cross_validation_enabled = True
```
marco-c: No need to enable it, it's True by default.

bugbug/bugzilla.py
```diff
@@ -75,8 +76,11 @@ def get_bug_fields():
...
+    bugs = db.read(BUGS_DB)
+    yield from (
+        bug for bug in bugs if include_invalid or bug["product"] != "Invalid Bugs"
```
marco-c: Nit: bug for bug in db.read(BUGS_DB), so you don't need the `bugs` variable

bugbug/models/spambug.py
```diff
@@ -0,0 +1,93 @@
...
+        )
+
+        self.clf = xgboost.XGBClassifier(n_jobs=16)
+        self.clf.set_params(predictor="cpu_predictor")
+
+    def get_labels(self):
+        classes = {}
+
+        for bug_data in bugzilla.get_bugs():
+            bug_id = int(bug_data["id"])
```
marco-c: Nit: this should already be an int

tests/test_bugzilla.py
```diff
@@ -0,0 +1,20 @@
...
+    legitimate_bugs = {int(bug["id"]) for bug in bugzilla.get_bugs()}
+
+    assert 1541482 in all_bugs
+    assert 1541482 not in legitimate_bugs
+
+    assert 1559674 in all_bugs
+    assert 1559674 not in legitimate_bugs
+
+    assert 1549207 in all_bugs
+    assert 1549207 not in legitimate_bugs
```
marco-c: Also assert one which is in both

PR #1040, opened 2019-10-24
Add arch as test_scheduling feature

BBloggsbott:
Fix for Issue#1018


bugbug/test_scheduling_features.py
```diff
@@ -108,3 +108,18 @@ def __call__(self, test_job, **kwargs):
                 "failures_past_56_pushes_in_components"
             ],
         }
+
+class arch(object):
+    def __call__(self, test_job, **kwargs):
+        archs = []
+        architectures = ("64", "32", "x86", "i386", "arm", "aarch64")
```
marco-c: 32, x86 and i386 should all be considered 32 bit

bugbug/test_scheduling_features.py
```diff
@@ -108,3 +108,18 @@ def __call__(self, test_job, **kwargs):
...
+        for ps in (("linux",), ("windows", "win"), ("android",), ("macosx",)):
+            for p in ps:
+                os_info = test_job["name"][: test_job["name"].index("/")].split("-")[1]
+                if p in os_info:
+                    archs.append(os_info.replace(p, ""))
+                    break
+        assert sum([a in architectures for a in archs]) == 1, "Wrong architectures ({}) in {}".format(
+            archs, test_job["name"]
+        )
+        return archs[0]
```
marco-c: You should assert that there is only one element in `archs`.

marco-c: As each job can only be for a single architecture

bugbug/test_scheduling_features.py
```diff
@@ -108,3 +108,18 @@ def __call__(self, test_job, **kwargs):
                 "failures_past_56_pushes_in_components"
             ],
         }
+
+class arch(object):
```
marco-c: The implementation seems a bit too complex, e.g. we should be able to get the archs without using `linux`, `win`, `windows`, etc.

BBloggsbott:
Okay. I'll make the necessary changes. Thanks for the inputs

bugbug/test_scheduling_features.py
BBloggsbott: `aarch64` is a 64 bit implementation of ARM. Do you want me to classify this as ARM or 64 bit?

BBloggsbott:
@marco-c The changes have been made

bugbug/test_scheduling_features.py
```diff
@@ -108,3 +108,17 @@ def __call__(self, test_job, **kwargs):
...
+    def __call__(self, test_job, **kwargs):
+        archs = []
+        architectures = (("arm",), ("64", "aarch64"), ("32", "x86", "i386"))
+        for arcs in architectures:
+            for a in arcs:
+                if a in test_job["name"][: test_job["name"].index("/")]:
+                    archs.append(arcs[0])
+                    break
```
marco-c: You are kind of "cheating" here, this way the assertion I asked you to add https://github.com/mozilla/bugbug/pull/1040/files#r339354740 will clearly be met, but simply because you are not trying other matches.

bugbug/test_scheduling_features.py
marco-c: It should be a separate one (Aarch64 or ARM64).

bugbug/test_scheduling_features.py
BBloggsbott: I used the implementation of [`platform`](https://github.com/mozilla/bugbug/blob/2740992b9159747c6168dee56116789bb134a151/bugbug/test_scheduling_features.py#L12) as a reference. 

So, if a `test_job`has multiple architectures in its name, you want this function to throw an assertion error?

BBloggsbott:
@marco-c Made the change to check match with every architecture.

BBloggsbott:
@marco-c The comments in all the reviews have been fixed.

marco-c:
There is some linting failure to fix (see the Taskcluster failing check). You can use pre-commit for that.

marco-c:
Did you try to apply this to check if the assertion is always holding?

BBloggsbott:
@marco-c Fixed the failing pylint

marco-c:
> @marco-c Fixed the failing pylint

Thanks! What about:
> Did you try to apply this to check if the assertion is always holding?

BBloggsbott:
@marco-c Yes. I tested it with a small list of dictionaries:
```python
test_jobs = [{"name":"test-linux64-ccov/xpcshell"}, {"name":"test-linux32-ccov/xpcshell"}, {"name":"test-android64-ccov/xpcshell"}, {"name":"test-linuxaarch64-ccov/xpcshell"}, {"name":"test-linuxarmx86-ccov/xpcshell"}]
```
The assertion failed only for the last one.

BBloggsbott:
Those were just some fake data. I'll test them on the whole dataset and share the results.

BBloggsbott:
@marco-c What do I do with jobs that have names without architectures? Like `build-android-api-16/pgo`

marco-c:
> @marco-c What do I do with jobs that have names without architectures? Like build-android-api-16/pgo and build-linux-rusttests/opt? Shall I just return None?

> I identified a few other architectures that I had to add. Apart from those architectures, these jobs are the only ones that raise an AssertionError

Are they the only two jobs? If so, we could hardcode their architecture.

BBloggsbott:
There are a lot of jobs like this. From a brief look, the majority of them have the following terms in them:
* `linux-rusttests`
* `linux-shippable`
* `linux` (In this case, the job name is mostly `linux/debug`)
* `android-api-16`

Also a new problem. Some jobs names don't have `/` in them which will cause a problem during slicing. I think this will also cause an issue with the `platform` feature as even that slices based on the index of `/`

marco-c:
> There are a lot of jobs like this. From a brief look, the majority of them have the following terms in them:

>     linux-rusttests
>     linux-shippable
>     linux (In this case, the job name is mostly linux/debug)
>     android-api-16

> These are a few of the job names that raised an assertion error.

It looks like all of them are "build-"? If so, we can just return no architecture when it's a build job.

marco-c:
> Also a new problem. Some jobs names don't have / in them which will cause a problem during slicing. I think this will also cause an issue with the platform feature as even that slices based on the index of /. I can fix this by using the entire name of the job if / is not present.

For example which job?

BBloggsbott:
These were a few job names that did not have a `/` in them.
```
build-docker-image-debian7-base
build-docker-image-lint
build-docker-image-debian9-packages
build-docker-image-pipfile-updates
build-docker-image-image_builder
build-docker-image-debian7-amd64-build-base
build-docker-image-android-build
build-docker-image-webrender
build-docker-image-custom-v8
build-docker-image-mozapkpublisher
build-docker-image-debian9-base
build-docker-image-valgrind-build
build-docker-image-debian7-i386-raw
build-docker-image-fetch
build-docker-image-firefox-snap
build-docker-image-mingw32-build
build-docker-image-debian7-raw
build-docker-image-debian9-raw
build-docker-image-diffoscope
build-docker-image-partner-repack
build-docker-image-debian9-amd64-build
build-docker-image-debian9-arm64-build
build-docker-image-periodic-updates
build-docker-image-debian7-amd64-build
build-docker-image-index-task
build-docker-image-debian7-i386-build
build-docker-image-toolchain-build
build-docker-image-funsize-update-generator
build-docker-image-update-verify
build-docker-image-desktop1604-test
build-docker-image-debian7-mozjs-rust-build
build-docker-image-debian7-packages
build-docker-image-debian7-i386-packages
build-docker-image-toolchain-arm64-build
build-docker-image-debian9-base
```

marco-c:
OK, you can ignore those, they'll be removed soon.

BBloggsbott:
In that case only the jobs without the architecture in the names are the problem. How should I handle those?

marco-c:
> In that case only the jobs without the architecture in the names are the problem. How should I handle those?

See my comment above, https://github.com/mozilla/bugbug/pull/1040#issuecomment-549058813:
> It looks like all of them are "build-"? If so, we can just return no architecture when it's a build job.

BBloggsbott:
@marco-c Fixed it

BBloggsbott:
Also tested it on the dataset. Faced no problems.

bugbug/test_scheduling_features.py
```diff
@@ -108,3 +108,29 @@ def __call__(self, test_job, **kwargs):
...
+class arch(object):
+    def __call__(self, test_job, **kwargs):
+        if "build-" in test_job["name"]:
+            return []
+        archs = set()  # Used set to eliminate duplicates like in case of aarch64
+        for arcs in (
+            ("arm",),
+            ("aarch64", "arm64"),
+            ("64", "x86_64"),
+            ("32", "x86", "i386", "arm7"),
```
marco-c: `arm7` should be `arm`

scripts/regressor_finder.py
```diff
@@ -120,9 +120,10 @@ def init_mapping(self):
         logger.info("Downloading Mercurial <-> git mapping file...")
```
marco-c: Please rebase on top of master, as the changes in this and the following file should not be here

BBloggsbott:
@marco-c Changed the `arm7` to the `arm` group and updated the branch from master.

PR #1052, opened 2019-10-29
Overwrite items_gen for spambug model

ayush-1506:
Overwriting the items_gen method to include spam bugs for the `SpamBug` model.

### DIFF
--- a/bugbug/models/spambug.py
+++ b/bugbug/models/spambug.py
@@ -3,15 +3,13 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
-from collections import defaultdict
-
 import xgboost
 from imblearn.under_sampling import RandomUnderSampler
 from sklearn.compose import ColumnTransformer
 from sklearn.feature_extraction import DictVectorizer
 from sklearn.pipeline import Pipeline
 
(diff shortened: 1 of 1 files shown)


bugbug/models/spambug.py
```diff
@@ -87,5 +89,33 @@ def get_labels(self):
 
         return classes, [0, 1]
 
+    def items_gen(self, classes):
```
marco-c: It is overly complex for this case, as we already know commit_data is false here, so we can just get rid of all that code

bugbug/models/spambug.py
```diff
@@ -87,5 +87,14 @@ def get_labels(self):
...
```
marco-c: Please add a comment to say that we overwrite the method so that we can include invalid bugs.

bugbug/models/spambug.py
```diff
@@ -87,5 +87,14 @@ def get_labels(self):
...
+
+        for bug in bugzilla.get_bugs(include_invalid=True):
+            bug_id = bug["id"]
+            if bug_id not in classes:
+                continue
+
+            yield bug, classes[bug_id]
```
marco-c: Nit: You can rewrite this as:
`return ((bug, classes[bug["id"]]) for bug in bugzilla.get_bugs(include_invalid=True) if bug["id"] in classes)`

PR #1053, opened 2019-10-29
Remove unwanted feature extractors for the spambug model

ayush-1506:
Removed the commit features that aren't of any use for the spambug model. Also added the include_invalid = True option to the get_labels method.


marco-c:
Thanks!

PR #1076, opened 2019-11-05
Update .taskcluster.yml for community cluster

imbstack:
(context: [bug 1578409](https://bugzilla.mozilla.org/show_bug.cgi?id=1578409))

We'll still need to make the secrets in the new deployment in order for this to work but that is blocked on [bug 1593896](https://bugzilla.mozilla.org/show_bug.cgi?id=1593896) at the moment. I wanted to get this up for review to see if it made sense to you all. @La0 and @marco-c in particular.

*Note: this will fail until the github app that provides tc is switched which we would do right before we land this.*


marco-c:
@imbstack will we need to update Taskcluster URLs all over the place?

marco-c:
There are task definitions under "infra/" too.

.taskcluster.yml
```diff
@@ -276,147 +276,14 @@ tasks:
             owner: mcastelluccio@mozilla.com
             source: ${repository}/raw/${head_rev}/.taskcluster.yml
 
-      - $if: 'tasks_for == "github-push" && head_branch[:10] == "refs/tags/"'
```
marco-c: Did you remove all these tasks on purpose?

imbstack:
Ok, as discussed in irc today, bugbug will now run entirely in the community cluster. This PR now adds back the tasks that were removed before and sets `TASKCLUSTER_ROOT_URL` for task-boot tasks so they deploy in the correct deployment. I will be adding the roles for this in a separate PR to the tc config repo

marco-c:
@imbstack note that this project is using Taskcluster in many places, not only for CI.
E.g. infra/data-pipeline.yml, infra/check-pipeline.yml, infra/taskcluster-hook-classify-patch.json, etc.
Also, there are artifacts URLs such as https://github.com/mozilla/bugbug/blob/8b4cfd2dc4320a1f2265c5cf1d808746a913a00b/scripts/retrieve_training_metrics.py#L19 that I guess should be updated?

imbstack:
Are all of the `infra/` tasks run as "production" rather than ci? How do they get scheduled? If they are there, I'll need to change their workerType from what I have set here. Probably need to figure this out with releng tomorrow. I'll also update the taskcluster.net urls and the artifact urls

marco-c:
> Are all of the infra/ tasks run as "production" rather than ci? How do they get scheduled? If they are there, I'll need to change their workerType from what I have set here. Probably need to figure this out with releng tomorrow. I'll also update the taskcluster.net urls and the artifact urls

Yep, they are scheduled via hooks (the hooks themselves are created and updated by CI, from the definitions in infra/).

.taskcluster.yml
```diff
@@ -171,15 +171,16 @@ tasks:
...
+        provisionerId: proj-relman
+        workerType: ci
         payload:
           capabilities:
             privileged: true
           maxRunTime: 10800
           image: "${taskboot_image}"
           env:
             REGISTRY: registry.hub.docker.com
+            TASKCLUSTER_ROOT_URL: '${taskcluster_root_url}'
```
marco-c: As for task-boot, this can be removed, right?

La0: Yes, it should be set automatically by the worker

imbstack:
I've opted for now to not do the falling-back-to-legacy-cluster stuff for index calls here just due to running out of time on my part. If we really want that maybe we can make it happen during your day tomorrow?

Other than that I'm just going to be making some admin changes on our side to allow this to create hooks and not have them removed and I _think_ we're set with this.

tests/fixtures/bug_features/comment_length.json
```diff
@@ -1,2 +1,2 @@
...
...18&tree=all", "id": 13765378, "creation_time": "2018-11-19T17:21:04Z", "author": "orangefactor@bots.tld"}, {"text": "1 failures in 1862 pushes (0.001 failures/push) were associated with this bug in the last 7 days.\n\nRepository breakdown:\n* mozilla-inbound: 1\n\nPlatform breakdown:\n* windows7-32: 1\n\nFor more details, see:\nhttps://treeherder.mozilla.org/intermittent-failures.html#/bugdetails?bug=1383552&startday=2018-12-03&endday=2018-12-09&tree=all", "id": 13802912, "creation_time": "2018-12-10T06:19:30Z", "author": "orangefactor@bots.tld"}, {"text": "2 failures in 2648 pushes (0.001 failures/push) were associated with this bug in the last 7 days.\n\nRepository breakdown:\n* mozilla-central: 1\n* autoland: 1\n\nPlatform breakdown:\n* windows7-32: 2\n\nFor more details, see:\nhttps://treeherder.mozilla.org/intermittent-failures.html#/bugdetails?bug=1383552&startday=2018-12-17&endday=2018-12-23&tree=all", "id": 13829666, "creation_time": "2018-12-24T06:15:21Z", "author": "orangefactor@bots.tld"}, {"text": "https://wiki.mozilla.org/Bug_Triage#Intermittent_Test_Failure_Cleanup", "id": 13871675, "creation_time": "2019-01-18T22:24:54Z", "author": "bug-husbandry-bot@mozilla.bugs"}]}
```
marco-c: As this is a test fixture, you can keep the old link here and avoid changing this file and tests/test_bug_features.py

marco-c:
> I've opted for now to not do the falling-back-to-legacy-cluster stuff for index calls here just due to running out of time on my part. If we really want that maybe we can make it happen during your day tomorrow?

I've thought about this, and it should be pretty simple. All the code which is downloading some artifact is going through https://github.com/mozilla/bugbug/blob/master/bugbug/utils.py#L134-L153, so we can simply change it to check with the community-tc URL first and fallback on the old URL if 404.

marco-c:
> Other than that I'm just going to be making some admin changes on our side to allow this to create hooks and not have them removed and I think we're set with this.

Let me know when that is ready.

imbstack:
@marco-c ok, I can try to make the fix for fallback stuff this morning. The changes to allow hooks creation are still in progress but actually this can make hooks right now, it just risks having them deleted by our admin tools if people aren't careful

imbstack:
@marco-c I've added a very hacky way to fall back to the old deployment in the place you specified. Is that roughly what you were thinking?

bugbug/utils.py
```diff
@@ -143,6 +143,13 @@ def download_check_etag(url, path):
 
     if old_etag != new_etag:
         r = requests.get(url, stream=True)
+
+        # TODO Remove this hack after a few runs on the new deployment
```
marco-c: Nit: looks good, only small improvement: move this at the head request

imbstack:
No problem! Glad to try to make it a bit less of a pain for you all.

marco-c:
Oops, it looks like there is a test failure and a linting failure, but I'll take care of it.

imbstack:
Thanks, sounds great. btw the hooks creation stuff is done now, so I think we're all set!

PR #1094, opened 2019-11-09
output prediction class name in bug_classifier script

### LINKED ISSUE #1070: Support multiclass classifiers in bug_classifier script
The bug_classifier script at scripts/bug_classifier.py currently assumes the model is a binary model. We need to make it work for multiclass models too (e.g. defectenhancementtask).

In particular, https://github.com/mozilla/bugbug/blob/65bf1b4604ca55a67490d27adc99c6441bad38c8/scripts/bug_classifier.py#L75-L78 needs to be changed.

To test your changes, simply run `python3 -m scripts.bug_classifier defectenhancementtask`

arjun-krishna:
Fix #1070 

Would it be better add an argument to the script to print probability optionally?

### DIFF
--- a/scripts/bug_classifier.py
+++ b/scripts/bug_classifier.py
@@ -53,8 +53,6 @@ def classify_bugs(model_name, classifier, bug_id):
     else:
         bugs = bugzilla.get_bugs()
 
-    class_names = model.class_names
-
     for bug in bugs:
         print(
             f'https://bugzilla.mozilla.org/show_bug.cgi?id={bug["id"]} - {bug["summary"]} '
@@ -71,8 +69,8 @@ def classify_bugs(model_name, classifier, bug_id):
         else:
             probas = model.classify(bug, probabilities=True, importances=False)
 
-        predicted_class = class_names[np.argmax(probas)]
-        print(f"{predicted_class} {probas}")
+        pred_class = model.le.inverse_transform([np.argmax(probas)])[0]
+        print(f"{pred_class} {probas}")
         input()
 
 
--- a/scripts/bug_classifier.py
+++ b/scripts/bug_classifier.py
@@ -70,11 +70,12 @@ def classify_bugs(model_name, classifier, bug_id):
             probas = model.classify(bug, probabilities=True, importances=False)
 
         probability = probas[0]
+        pred_index = np.argmax(probability)
         if len(probability) > 2:
-            pred_class = model.le.inverse_transform([np.argmax(probability)])[0]
+            pred_class = model.le.inverse_transform([pred_index])[0]
         else:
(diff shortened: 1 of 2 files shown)


arjun-krishna:
> We can't use class_names like this, as we can't be sure that the index in the prediction list matches the index in the class_names list.

Hmm... I saw a similar pattern used in [scripts/generate_sheet.py](https://github.com/mozilla/bugbug/blob/master/scripts/generate_sheet.py#L39), I'll go through the train function in the Model class to check this out again.

marco-c:
> Hmm... I saw a similar pattern used in scripts/generate_sheet.py, I'll go through the train function in the Model class to check this out again.

Yeah, that's an error. It works most of the time, but we can't really rely on it. You can do something like https://github.com/mozilla/bugbug/blob/2124f60a9a40931ed0729d8ced2ed75b4ccf2d7f/scripts/commit_classifier.py#L377 instead.

arjun-krishna:
Ah, that makes sense. Will fix it. Thanks for the quick response!

scripts/bug_classifier.py
```diff
@@ -69,10 +69,8 @@ def classify_bugs(model_name, classifier, bug_id):
         else:
             probas = model.classify(bug, probabilities=True, importances=False)
 
-        if np.argmax(probas) == 1:
-            print(f"Positive! {probas}")
-        else:
-            print(f"Negative! {probas}")
+        pred_class = model.le.inverse_transform([np.argmax(probas)])[0]
+        print(f"{pred_class} {probas}")
```
marco-c: For 1 and 0, we can still print Positive and Negative (or True and False), it's a bit more user-friendly.

scripts/generate_sheet.py
```diff
@@ -35,8 +35,7 @@ def generate_sheet(model_name, token, days, threshold):
         p = model.classify(bug, probabilities=True)
         probability = p[0]
         if len(probability) > 2:
-            index = np.argmax(probability)
-            prediction = model.class_names[index]
+            prediction = model.le.inverse_transform([np.argmax(probability)])[0]
```
marco-c: Thanks for this fix, could you move it to a separate PR though, as it's a separate fix?

scripts/bug_classifier.py
```diff
@@ -69,10 +69,12 @@ def classify_bugs(model_name, classifier, bug_id):
         else:
             probas = model.classify(bug, probabilities=True, importances=False)
 
-        if np.argmax(probas) == 1:
-            print(f"Positive! {probas}")
+        probability = probas[0]
```
marco-c: You can calculate np.argmax(probability) here, so you don't have to repeat it twice in the if-else

marco-c:
Thanks!

Issue #1109, opened 2019-11-19
Make the bug_classifier script automatically download the required DB


marco-c:
The bug_classifier script is using the bugs DB (https://github.com/mozilla/bugbug/blob/d1016a351603d948230188fa3f370a057a1a8127/scripts/bug_classifier.py#L54), but the bugs DB might not be available yet at this point.
So, we should make the bug_classifier script download the bugs DB before trying to use it.

PR #1111, opened 2019-11-20
fix-1109 : Make the bug_classifier script automatically download the required DB

jeffin143:
Handling : https://github.com/mozilla/bugbug/issues/1109

### DIFF
--- a/scripts/bug_classifier.py
+++ b/scripts/bug_classifier.py
@@ -7,7 +7,7 @@ from logging import INFO, basicConfig, getLogger
 import numpy as np
 import requests
 
-from bugbug import bugzilla
+from bugbug import bugzilla, db
 from bugbug.models import get_model_class
 from bugbug.utils import download_check_etag, zstd_decompress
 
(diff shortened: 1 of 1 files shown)


scripts/bug_classifier.py
```diff
@@ -51,6 +51,7 @@ def classify_bugs(model_name, classifier, bug_id):
         bugs = bugzilla.get(bug_id).values()
         assert bugs, f"A bug with a bug id of {bug_id} was not found"
     else:
+        bugzilla.download_bugs_db()
```
marco-c: You can just call `db.download(bugzilla.BUGS_DB)` like the other scripts, also make sure you assert on its result

marco-c:
Make sure you fix the linting issues too, you can use pre-commit for that.

scripts/bug_classifier.py
jeffin143: Resolved

jeffin143:
> Make sure you fix the linting issues too, you can use pre-commit for that (see https://github.com/mozilla/bugbug#auto-formatting).

Thanks, I went through it

PR #1159, opened 2019-12-09
WIP: add code metrics

calixteman:


bugbug/utils.py
```diff
@@ -309,7 +309,7 @@ def __setitem__(self, day, value):
 
 class LMDBDict:
     def __init__(self, path):
-        self.db = lmdb.open(path, map_size=68719476736, metasync=False, sync=False)
+        self.db = lmdb.open(path, map_size=68_719_476_736, metasync=False, sync=False)
```
marco-c: Don't change this unrelated file.

bugbug/models/regressor.py
```diff
@@ -49,6 +49,7 @@ def __init__(self, lemmatization=False, interpretable=False):
 
         feature_extractors = [
             commit_features.source_code_file_size(),
+            commit_features.source_code_file_metrics(),
```
marco-c: Don't add this yet to the two models, I want to add it in a separate PR so that we can measure the improvement in accuracy.

bugbug/repository.py
```diff
@@ -77,9 +77,7 @@
     "Build System File": [".build", ".mk", ".in"],
 }
 
-HARDCODED_TYPES = {
-    ".eslintrc.js": ".eslintrc.js",
-}
+HARDCODED_TYPES = {".eslintrc.js": ".eslintrc.js"}
```
marco-c: Nit: don't change this, let black handle the formatting how it wants

bugbug/repository.py
```diff
@@ -372,6 +378,23 @@ def _transform(commit):
 
             if size is not None:
                 source_code_sizes.append(size)
+                metrics = code_analysis_server.metrics(path, after)
+                if "error" in metrics:
```
marco-c: Why this if?

bugbug/repository.py
```diff
@@ -372,6 +378,23 @@ def _transform(commit):
...
+                    ]
+                    commit.average_halstead_operands += metrics["halstead"]["operands"]
+                    commit.average_halstead_unique_operators += metrics["halstead"][
+                        "unique_operators"
+                    ]
+                    commit.average_halstead_operators += metrics["halstead"][
+                        "operators"
+                    ]
+                    commit.average_source_loc += metrics["loc"]["sloc"]
+                    commit.average_logical_loc += metrics["loc"]["lloc"]
```
marco-c: Are these at the file level or the function level?

bugbug/repository.py
```diff
@@ -372,6 +378,23 @@ def _transform(commit):
...
+                        "unique_operands"
+                    ]
+                    commit.average_halstead_operands += metrics["halstead"]["operands"]
+                    commit.average_halstead_unique_operators += metrics["halstead"][
+                        "unique_operators"
+                    ]
+                    commit.average_halstead_operators += metrics["halstead"][
+                        "operators"
+                    ]
+                    commit.average_source_loc += metrics["loc"]["sloc"]
```
marco-c: Is this the number of lines of code, excluding comments?

bugbug/repository.py
marco-c: They are at file level, as discussed on IRC, let's get mean/max/min/avg complexities both at function level and at file level

bugbug/repository.py
calixteman: good catch: `not in` is better

bugbug/commit_features.py
```diff
@@ -128,6 +128,47 @@ def __call__(self, commit, **kwargs):
         }
 
 
+class source_code_file_metrics(object):
+    name = "metrics on source code file"
+
+    def __call__(self, commit, **kwargs):
+        return {
```
marco-c: We could add the total (sum) too, like for other metrics

bugbug/repository.py
```diff
@@ -378,6 +465,9 @@ def _transform(commit):
 
             if size is not None:
                 source_code_sizes.append(size)
+                metrics = code_analysis_server.metrics(path, after, unit=False)
```
marco-c: I forgot what the unit parameter is, could you document it in code_analysis_server?

bugbug/repository.py
```diff
@@ -325,6 +411,7 @@ def _transform(commit):
     source_code_sizes = []
     other_sizes = []
     test_sizes = []
+    metrics_file_count = 0
```
marco-c: You are never increasing this :P

bugbug/repository.py
```diff
@@ -415,6 +505,23 @@ def _transform(commit):
...
+    if metrics_file_count:
+        commit.average_cyclomatic /= metrics_file_count
+        commit.average_halstead_unique_operands /= metrics_file_count
+        commit.average_halstead_operands /= metrics_file_count
+        commit.average_halstead_unique_operators /= metrics_file_count
+        commit.average_halstead_operators /= metrics_file_count
+        commit.average_source_loc /= metrics_file_count
+        commit.average_logical_loc /= metrics_file_count
+    else:
+        commit.minimum_cyclomatic = 0
```
marco-c: Add a comment explaining that we need to do this when we didn't get any metrics and so the minimum would be sys.maxsize

bugbug/rust_code_analysis_server.py
```diff
@@ -45,7 +45,7 @@ def start_process(self):
 
         try:
             self.proc = subprocess.Popen(
-                ["rust-code-analysis", "--serve", "--port", str(self.port),]
+                ["rust-code-analysis", "--serve", "--port", str(self.port)]
```
marco-c: WTH, why is black giving different results for you and me? :P

bugbug/commit_features.py
```diff
@@ -128,6 +128,54 @@ def __call__(self, commit, **kwargs):
...
+                "minimum_halstead_unique_operands"
+            ],
+            "Minimum number of operands": commit["minimum_halstead_operands"],
+            "Minimum number of unique operators": commit[
+                "minimum_halstead_unique_operators"
+            ],
+            "Minimum number of operators": commit["minimum_halstead_operators"],
+            "Minimum number of source loc": commit["minimum_source_loc"],
+            "Minimum number of logical loc": commit["minimum_logical_loc"],
+            "Sum of number of operands": commit["sum_halstead_operands"],
```
marco-c: Use `Total` instead of `Sum of`, as there is some code which simplifies the names and relies on `Maximum`, `Minimum`, `Total` and `Average` being there

bugbug/rust_code_analysis_server.py
calixteman: I don't know... my editor was open before black ran and the comma is there... so it removed it

bugbug/repository.py
calixteman: sloc is just function span length and lloc is the number of line with something of interest.
For example, for c++:
https://github.com/mozilla/rust-code-analysis/blob/master/src/sloc.rs#L164-L178

PR #1160, opened 2019-12-09
Add rust-code-analysis server

calixteman:


bugbug/rust_code_analysis_server.py
```diff
@@ -0,0 +1,82 @@
...
+                r = requests.get(url)
+                r.raise_for_status()
+                print("Rust code analysis server is ready to accept queries")
+                self.ok = True
+                return
+            except Exception:
+                time.sleep(1)
+
+    @property
+    def available(self):
```
marco-c: We could merge ping and available in the constructor itself, so we don't allow to construct the object at all if the server is not available.

bugbug/rust_code_analysis_server.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def __str__(self):
+        return f"Server running at {self.base_url}"
+
+    def metrics(self, filename, code, unit=True):
+        unit = 1 if unit else 0
+        url = f"{self.base_url}/metrics?file_name={filename}&unit={unit}"
+        r = requests.post(url, data=code, headers=self.headers)
+
+        if r.status_code != 404:
```
marco-c: `if r.ok` so we ignore all other errors.

bugbug/rust_code_analysis_server.py
```diff
@@ -0,0 +1,82 @@
...
+            return r.json()
+        return {}
+
+    def function(self, filename, code):
+        url = f"{self.base_url}/function?file_name={filename}"
+        r = requests.post(url, data=code, headers=self.headers)
+
+        if r.status_code != 404:
```
marco-c: Same here, `if r.ok`

PR #1161, opened 2019-12-09
Collect information about functions while analyzing commits

marco-c:


bugbug/repository.py
```diff
@@ -299,8 +301,34 @@ def _transform(commit):
         elif type_ in SOURCE_CODE_TYPES_TO_EXT:
             commit.source_code_files_modified_num += 1
 
-            commit.source_code_added += stats["added_lines"]
-            commit.source_code_deleted += stats["deleted_lines"]
+            if after is not None:
```
calixteman: nit: put this stuff in a function and add a test to check the job is done correctly

marco-c: I've split it in a function, and filed a follow-up to introduce a test: #1163.

Issue #1184, opened 2019-12-18
Stop using sklearn.metrics.classification module


marco-c:
`The sklearn.metrics.classification module is  deprecated in version 0.22 and will be removed in version 0.24. The corresponding classes / functions should instead be imported from sklearn.metrics. Anything that cannot be imported from sklearn.metrics is now part of the private API.`

PR #1185, opened 2019-12-18
Stop using sklearn.metrics.classification as it is deprecated

abhaykatheria:
sklearn.metrics.classification is deprecated #1184


abhaykatheria:
i only found one occurence of sklearn.metrics.classification and the function that was used is moved to sklearn.metrics (precision_recall_fscore_support)

Issue #1082, opened 2019-11-06
Add a few more features to the spambug model


marco-c:
There are a few features that might make sense, we should try to add them one by one and see if they improve results:
- [ ] keywords;
- [ ] number_of_bug_dependencies;
- [ ] blocked_bugs_number;
- [ ] product;
- [ ] component;
- [x] is_mozillian;
- ...

marco-c:
> is_mozillian

Actually for this one it might be better to just overwrite the decision (#1084).

marco-c:
@ayush-1506 do you want to look into this too?

ayush-1506:
Ok!

ayush-1506:
Sure!

ayush-1506:
Results:

Before any additions:
```
Accuracy: f0.995783670252694 (+/- 0.0007449414872651147)
Precision: f0.9833024639219158 (+/- 0.004560331424210395)
Recall: f0.9944979418565302 (+/- 0.0017856201890962399)
```

After adding `keywords()`:
```

marco-c:
Thanks for the investigation, so it looks like the increase is minimal for all. I guess the metrics are already too high :)

I'd add keywords, product and component anyway. They would probably be useful if we had more examples of spam bugs.

PR #1202, opened 2019-12-24
Add a few more features to spambug model

ayush-1506:
Fixes: #1082


bugbug/models/spambug.py
```diff
@@ -35,6 +35,9 @@ def __init__(self, lemmatization=False):
             bug_features.component(),
             bug_features.num_words_title(),
             bug_features.num_words_comments(),
+            bug_features.keywords(),
+            bug_features.product(),
+            bug_features.component(),
```
marco-c: Oops, product and component were already there :)

ayush-1506: Ah, missed that. Will add only keywords then.

Issue #1108, opened 2019-11-19
Remove DBs when the schema of the downloaded DB is older than the current schema


marco-c:
In the download method, https://github.com/mozilla/bugbug/blob/d1016a351603d948230188fa3f370a057a1a8127/bugbug/db.py#L89, we should handle the case where the DB is already downloaded but is using an older schema version than the current one.

marco-c:
It's possible that this is already kind of supported, as we check the ETag and the ETag changes when the schema version changes. We need to add a test to confirm.

Nathan-Wisner:
For this, would we want to update and download a newer version or just exit the function with false like 
``` if is_old_schema(path):
     return False
```

marco-c:
We should try to download a new version.

mohanish2504:
Hi,
I want to contribute and I am a beginner, I wanted to know is the issue resolved? I am stuck here 

'if is_old_schema(path): return False'

what to do if it is an old schema?

mohanish2504:
Hi,
I want to contribute and I am a beginner, I wanted to know is the issue resolved? I am stuck here

`if is_old_schema(path): return False`

what to do if it is an old schema?

Luis-gd:
Can i take this work?

marco-c:
> I want to contribute and I am a beginner, I wanted to know is the issue resolved? I am stuck here
> if is_old_schema(path): return False

No, the issue is still here. If you look at the code, you'll see that `is_old_schema` is not checking the schema version of the downloaded DB, but the schema version of the remote DB.

> what to do if it is an old schema?

Please read all the comments, it is explained above.

marco-c:
> Can i take this work?

You can work on any issue which is open and has no PR linked to it.
Did you read https://github.com/mozilla/bugbug/issues/1092 or CONTRIBUTING.md?

avijit1258:
Hello @marco-c ,

I am thinking to implement a new function download_check_version() similar to download_check_etag() in the bugbug/utils.py which will be called after downloading the current DB. Then return false if that is the case. Am I going to the right direction?

Avijit

marco-c:
@avijit1258 the goal is to remove a DB when its schema version is old compared to the current schema. I'm not sure how what you proposed is going to do this, but I'm not sure exactly what you mean. Proposing a PR is the best way to show what you mean ;)

avijit1258:
[...]

PR #1206, opened 2019-12-29
added check for downloaded DB version with current version

avijit1258:
@marco-c I have added the code here. Can you kindly check if I am getting things write?

### DIFF
--- a/bugbug/db.py
+++ b/bugbug/db.py
@@ -82,12 +82,14 @@ def download_support_file(path, file_name):
         logger.info(
             f"{file_name} is not yet available to download for {path}", exc_info=True
         )
-        return False
+
+    return False
 
 
 # Download and extract databases.
 def download(path, support_files_too=False):
     # If a DB with the current schema is not available yet, we can't download.
+    print("Path: ", path)
     if is_old_schema(path):
         return False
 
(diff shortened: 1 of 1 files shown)


abhaykatheria:
```
            with open(f"{os.path.basename(path)}.version", "r") as f:
                current_version = f.read()
```
suppose path was "C:/new1/new2/prova.json"
the current schema existed there and not at os.path.basename(path) ///which is just prova.json

there are cases where files are downloaded in current directory where your code have not given error at some places paths have been specified 
 
and when you extract the downloaded file the earlier prova.json is overwritten by new prova.json
so you are comparing to equal things instead.

p.s.: I am just a beginner so pardon me if i got something wrong and do correct me.
Happy To Help

avijit1258:
@abhaykatheria Thank you for your response. I think you are right. I got that wrong. Can you help me to identify how db.py differentiates the current DB file and the downloaded one? Or I have to create a temp path, download there instead and then replace the DB if the condition applies?

Thanks I highly appreciate your time.
Avijit

abhaykatheria:
Creating a temp directory is the way to go I guess..

marco-c:
See https://github.com/mozilla/bugbug/issues/1108#issuecomment-570610247, this should already be covered by the etag check, so we just need to write a test.

Issue #1195, opened 2019-12-20
Add the SpamBug model to the HTTP service


marco-c:
We need the HTTP service to support the SpamBug model too (it should just be a matter of adding the trainining task as a dependency of the http service task, and add the spambug model to the http service models in https://github.com/mozilla/bugbug/blob/75939c52a010666a776a9d1c4f41e01397c9918e/http_service/models.py#L23).

This is the last step required before being able to put the model in production in https://github.com/mozilla/relman-auto-nag. Well, the actual real last step is going to be in https://github.com/mozilla/relman-auto-nag itself :)

ayush-1506:
@marco-c Where is the http-service task defined? I can't find it in infra/data-pipeline.yml .

marco-c:
@ayush-1506 it's this task: https://github.com/mozilla/bugbug/blob/1d4f0dada360641fb9406dbcfd43199a35068be1/infra/data-pipeline.yml#L983.

ayush-1506:
@marco-c Could you let me know the final steps to implement in relman-auto-nag?

marco-c:
@ayush-1506 first of all, we need to fix https://github.com/mozilla/bugbug/issues/1200, otherwise the model fails during training.

Then, in relman-auto-nag, we need something like https://github.com/mozilla/relman-auto-nag/blob/master/auto_nag/scripts/stepstoreproduce.py and https://github.com/mozilla/relman-auto-nag/blob/57fae814cb1d5c56cbb4e808784b3a21031aff28/auto_nag/scripts/configs/tools.json#L580-L603.
Assuming you will name the script "spam.py", you can run it this way: `python -m auto_nag.scripts.spam --dryrun`. You might need some minimal configuration for it to run, let me know (here or on IRC) if you run into issues and I'll help you.

PR #1213, opened 2020-01-03
Add spambug to http service

ayush-1506:
Fixes: #1195


ayush-1506:
@marco-c Why are the tests failing? I've merely added another dependency to http-service task.

marco-c:
I think it was just an intermittent issue with one of the dependencies. They are passing now.

Issue #1211, opened 2020-01-02
Make sure features are not used twice


marco-c:
This is a minor enhancement, the BugExtractor and CommitExtractor classes should make sure there are no duplicate feature_extractors (and cleanup_functions).

PR #1215, opened 2020-01-04
#1211 made sure duplicate features are not used

abhaykatheria:
I made sure even if there are duplicate feature extractors only unique ones are used.

### DIFF
--- a/bugbug/bug_features.py
+++ b/bugbug/bug_features.py
@@ -12,7 +12,7 @@ from dateutil import parser
 from libmozdata import versions
 from sklearn.base import BaseEstimator, TransformerMixin
 
-from bugbug import bug_snapshot, repository,feature_cleanup
+from bugbug import bug_snapshot, repository, feature_cleanup
 
 
 def field(bug, field):
@@ -539,62 +539,63 @@ class couple_common_keywords(couple_bug_feature):
             if keyword in bugs[1]["keywords"] and keyword not in self.to_ignore
         ]
 
+
 FEATURE_CLASSES = [
-            had_severity_enhancement,
-            is_reporter_a_developer,
-            op_sys,
-            platform,
-            components_touched_num,
-            components_touched,
-            commit_no_of_backouts,
-            commit_reviewer_experience_90_days,
-            commit_reviewer_experience,
-            commit_author_experience_90_days,
-            commit_author_experience,
-            commit_files_modified_num,
-            commit_types,
-            commit_deleted,
-            commit_added,
-            has_image_attachment,
-            has_image_attachment_at_bug_creation,
-            num_words_comments,
-            num_words_title,
-            affected_then_unaffected,
-            ever_affected,
-            reporter_experience,
-            comment_length,
-            comment_count,
-            has_cve_in_alias,
-            priority,
-            blocked_bugs_number,
-            delta_request_merge,
-            bug_reporter,
-            is_mozillian,
-            component,
-            product,
-            landings,
-            patches,
-            whiteboard,
-            has_github_url,
-            has_w3c_url,
-            has_url,
-            is_coverity_issue,
-            number_of_bug_dependencies,
-            severity,
-            keywords,
-            has_crash_signature,
-            has_regression_range,
-            has_str,
-            couple_common_whiteboard_keywords,
-            is_same_product,
-            is_same_component,
-            is_same_platform,
-            is_same_version,
-            is_same_os,
-            is_same_target_milestone,
-            is_first_affected_same,
-            couple_delta_creation_date,
-            couple_common_keywords,
+    had_severity_enhancement,
+    is_reporter_a_developer,
+    op_sys,
+    platform,
+    components_touched_num,
+    components_touched,
+    commit_no_of_backouts,
+    commit_reviewer_experience_90_days,
+    commit_reviewer_experience,
+    commit_author_experience_90_days,
+    commit_author_experience,
+    commit_files_modified_num,
+    commit_types,
+    commit_deleted,
+    commit_added,
+    has_image_attachment,
+    has_image_attachment_at_bug_creation,
+    num_words_comments,
+    num_words_title,
+    affected_then_unaffected,
+    ever_affected,
+    reporter_experience,
+    comment_length,
+    comment_count,
+    has_cve_in_alias,
+    priority,
+    blocked_bugs_number,
+    delta_request_merge,
+    bug_reporter,
+    is_mozillian,
+    component,
+    product,
+    landings,
+    patches,
+    whiteboard,
+    has_github_url,
+    has_w3c_url,
+    has_url,
+    is_coverity_issue,
+    number_of_bug_dependencies,
+    severity,
+    keywords,
+    has_crash_signature,
+    has_regression_range,
+    has_str,
+    couple_common_whiteboard_keywords,
+    is_same_product,
+    is_same_component,
+    is_same_platform,
+    is_same_version,
+    is_same_os,
+    is_same_target_milestone,
+    is_first_affected_same,
+    couple_delta_creation_date,
+    couple_common_keywords,
 ]
 
 CLEANUP_FUNCTIONS = [
@@ -604,7 +605,7 @@ CLEANUP_FUNCTIONS = [
     feature_cleanup.hex,
     feature_cleanup.dll,
     feature_cleanup.synonyms,
-    feature_cleanup.crash
+    feature_cleanup.crash,
 ]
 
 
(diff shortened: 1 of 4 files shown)


marco-c:
> I made sure even if there are duplicate feature extractors only unique ones are used.

It'd be better to have an assertion that fails when there are duplicate features, instead of silently ignoring the duplication.

abhaykatheria:
I made a list of all the feature classes that are being used.

`        self.feature_extractors = []
        used_features=[]
        for feature_extractor in feature_extractors:
            for feature in FEATURE_CLASSES:
                if isinstance(feature_extractor,feature):
                    if feature not in used_features:
                        used_features.append(feature)
                        self.feature_extractors.append(feature_extractor)
                    else :
                        assert False,("Duplicate extractors used")`

abhaykatheria:
`        self.feature_extractors = []
        used_features=[]
        for feature_extractor in feature_extractors:
            for x in FEATURE_CLASSES:
                if isinstance(feature_extractor,x):
                    if x not in used_features:
                        used_features.append(x)
                        self.feature_extractors.append(feature_extractor)
                    else :
                        assert False,("Duplicate extractors used")`

abhaykatheria:
-i am getting zero_divison error

bugbug/bug_features.py
```diff
@@ -557,8 +557,24 @@ def __init__(
...
-        self.feature_extractors = feature_extractors
-        self.cleanup_functions = cleanup_functions
+        self.feature_extractors = []
+        used_features = []
+        for feature_extractor in feature_extractors:
+            if type(feature_extractor) not in used_features:
+                used_features.append(type(feature_extractor))
+                self.feature_extractors.append(feature_extractor)
+            else:
+                assert False, "Duplicate features are used"
```
marco-c: You should be able to greatly simplify the implementation by using a set:
```Python
assert len(set(type(fe) for fe in feature_extractors)) == len(feature_extractors), "Duplicate feature extractors"
self.feature_extractors = feature_extractors
```

marco-c:
Please also add tests to make sure the assertion is hit when it should be hit.

tests/test_bug_features.py
```diff
@@ -163,6 +165,19 @@ def test_comment_length(read):
...
 
 
+def test_bug_extractor():
+    feature_extractors = [blocked_bugs_number(), bug_reporter(), bug_reporter()]
+    cleanup_functions = [
+        fileref(),
+        url(),
+        url(),
+    ]
+    try:
```
marco-c: Your code is not making sure that the assertion is thrown. You'd have to do something like:
```
try:
    BugExtractor(feature_extractors, cleanup_functions)
    assert False, "An assertion should have failed"
except AssertionError:
    pass
```

Even better, use the pytest utility to do this, like in:
https://github.com/mozilla/bugbug/blob/abf47d3b1c69e782450a57918d5950b2bec22cb4/tests/test_db.py#L82-L83

tests/test_bug_features.py
```diff
@@ -163,6 +165,19 @@ def test_comment_length(read):
     read("comment_length.json", comment_length, [566, 5291])
 
 
+def test_bug_extractor():
```
marco-c: Nit: add your tests at the end of the file

tests/test_bug_features.py
```diff
@@ -163,6 +165,19 @@ def test_comment_length(read):
...
+    cleanup_functions = [
+        fileref(),
+        url(),
+        url(),
+    ]
+    try:
+        BugExtractor(feature_extractors, cleanup_functions)
+    except AssertionError:
+        print("exception handled")
+
```
marco-c: There should be three tests:
1) Two feature extractors and two cleanup functions, without duplicates;
2) Two duplicate feature extractors and two non-duplicate cleanup functions;
3) Two non-duplicate feature extractors and two duplicate cleanup functions.

tests/test_bug_features.py
marco-c: A similar test should be written for CommitExtractor

abhaykatheria:
i ran the flake 8 testing there are no changes to commit further but still this fails

abhaykatheria:
Sorry for the mess actually i wrote this in bit hurry and thanks for your patience.

abhaykatheria:
image

tests/test_bug_features.py
```diff
@@ -266,3 +268,16 @@ def test_is_same_target_milestone(test_data, expected):
 @pytest.mark.parametrize("test_data, expected", FIRST_AFFECTED_PARAMS)
 def test_is_first_affected_same(test_data, expected):
     assert is_first_affected_same()(test_data) == expected
+
+
+BUG_EXTRACTOR_PARAMS = [
+    ([has_str, has_url], [fileref, url]),
```
marco-c: This case should not fail the assertion.

tests/test_commit_features.py
```diff
@@ -0,0 +1,24 @@
...
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import pytest
+
+from bugbug.bug_features import has_str, has_url
+from bugbug.commit_features import CommitExtractor
+from bugbug.feature_cleanup import fileref, url
+
+COMMIT_EXTRACTOR_PARAMS = [
+    ([has_str, has_url], [fileref, url]),
```
marco-c: You should not test bug_features functions in the commit_features test

abhaykatheria:
image
features i used are different but it still raises the assertion

marco-c:
Add some logging to figure out why:
```
print(set(type(fe) for fe in feature_extractors))
```

abhaykatheria:
its strange the features were read as class not objects as soon as i added the parenthesis the error went away but i did the same in test _bug_features and it ran flawlessly so i added the parenthesis everywhere to make it uniform.:)

tests/test_bug_features.py
```diff
@@ -266,3 +268,16 @@ def test_is_same_target_milestone(test_data, expected):
...
+
+BUG_EXTRACTOR_PARAMS = [
+    ([has_str(), has_str()], [fileref(), url()]),
+    ([has_str(), has_url()], [fileref(), fileref()]),
+]
+
+
+@pytest.mark.parametrize("feature_extractors,cleanup_functions", BUG_EXTRACTOR_PARAMS)
+def test_BugExtractor(feature_extractors, cleanup_functions):
+    BugExtractor([has_str(), has_url()], [fileref(), url()])
```
marco-c: You are running this twice because of the parametrization. Just remove the parametrization and run the checks directly in the test

tests/test_commit_features.py
```diff
@@ -0,0 +1,23 @@
...
+    ([author_experience(), author_experience()], [fileref(), url()]),
+    ([reviewers_num(), author_experience()], [fileref(), fileref()]),
+]
+
+
+@pytest.mark.parametrize(
+    "feature_extractors,cleanup_functions", COMMIT_EXTRACTOR_PARAMS
+)
+def test_CommitExtractor(feature_extractors, cleanup_functions):
+    CommitExtractor([reviewers_num(), author_experience()], [fileref(), url()])
```
marco-c: Same here

marco-c:
> its strange the features were read as class not objects as soon as i added the parenthesis the error went away but i did the same in test _bug_features and it ran flawlessly so i added the parenthesis everywhere to make it uniform.:)

It is because they are classes, so you need to "call" them to create objects. I hadn't noticed you weren't in the test.

tests/test_bug_features.py
```diff
@@ -266,3 +268,10 @@ def test_is_same_target_milestone(test_data, expected):
 @pytest.mark.parametrize("test_data, expected", FIRST_AFFECTED_PARAMS)
 def test_is_first_affected_same(test_data, expected):
     assert is_first_affected_same()(test_data) == expected
+
+
+def test_BugExtractor(feature_extractors, cleanup_functions):
+    BugExtractor([has_str(), has_url()], [fileref(), url()])
+    with pytest.raises(AssertionError):
+        BugExtractor([has_str(), has_str()], [fileref(), url()])
+        BugExtractor([has_str(), has_url()], [fileref(), fileref()])
```
marco-c: This way you are only testing the first call, as it raises the assertion and exits.

You should instead do:
```
with pytest.raises(AssertionError):
    BugExtractor([has_str(), has_str()], [fileref(), url()])


with pytest.raises(AssertionError):
    BugExtractor(...)
```

tests/test_commit_features.py
```diff
@@ -0,0 +1,16 @@
...
+
+from bugbug.commit_features import CommitExtractor, author_experience, reviewers_num
+from bugbug.feature_cleanup import fileref, url
+
+
+def test_CommitExtractor(feature_extractors, cleanup_functions):
+    CommitExtractor([reviewers_num(), author_experience()], [fileref(), url()])
+    with pytest.raises(AssertionError):
+        CommitExtractor([reviewers_num(), author_experience()], [fileref(), fileref()])
+        CommitExtractor([author_experience(), author_experience()], [fileref(), url()])
```
marco-c: Same here

bugbug/commit_features.py
```diff
@@ -495,7 +495,17 @@ def merge_commits(commits):
 
 class CommitExtractor(BaseEstimator, TransformerMixin):
     def __init__(self, feature_extractors, cleanup_functions):
+        print(feature_extractors)
+        print(type(feature_extractors[0]))
+        print(set(type(fe) for fe in feature_extractors))
```
marco-c: Don't forget to remove the logging statements

PR #1243, opened 2020-01-13
Make bugbug classify patch and test select hook support different Phabricator deployments

marco-c:


infra/taskcluster-hook-test-select.json
```diff
@@ -74,6 +76,10 @@
     "triggerSchema": {
         "additionalProperties": false,
         "properties": {
+            "PHABRICATOR_DEPLOYMENT": {
+                "type": "string",
+                "enum": ["prod", "dev"]
```
La0: Why not stick to production/testing and use `APP_CHANNEL` on the code-review side ?

scripts/commit_classifier.py
```diff
@@ -234,9 +234,14 @@ def has_revision(revision):
             except hglib.error.CommandError:
                 return False
 
-        phabricator_api = PhabricatorAPI(
-            api_key=get_secret("PHABRICATOR_TOKEN"), url=get_secret("PHABRICATOR_URL")
-        )
+        if phabricator_deployment == "prod":
```
La0: Nit: Use some constant

scripts/commit_classifier.py
```diff
@@ -708,6 +713,13 @@ def main():
     parser.add_argument("model", help="Which model to use for evaluation")
     parser.add_argument("cache_root", help="Cache for repository clones.")
     parser.add_argument("diff_id", help="diff ID to analyze.", type=int)
+    # TODO: The "" choice can be removed when all users have been updated to pass a correct phabricator_deployment.
+    parser.add_argument(
+        "phabricator_deployment",
+        help="Which Phabricator deployment to hit.",
+        type=str,
+        choices=["", "prod", "dev"],
```
La0: Using the empty string will default to `dev`, maybe better to set a `default="dev"`

marco-c: It will default to "prod" because of https://github.com/mozilla/bugbug/pull/1243/files#diff-9bdbe400fc491c12c4cfbf9444a10336R733

infra/taskcluster-hook-test-select.json
marco-c: Having a testing channel that does exactly the same things as production would be quite expensive, so I only have a production channel for bugbug

marco-c: Ah you meant passing the APP_CHANNEL from code-review. That's a problem because of https://github.com/mozilla/code-review/pull/433#discussion_r365708075

scripts/commit_classifier.py
marco-c: Done

PR #1245, opened 2020-01-14
Upload artifacts as we go to avoid losing work

marco-c:
I need to test this.


marco-c:
Unfortunately this won't work, as we can not overwrite artifacts.

PR #1248, opened 2020-01-14
Use tenacity to refactor retry function

dalyl-zero:


dalyl-zero:
Thanks for the feedback. I don't, however, understand why you have not merged [this](https://github.com/mozilla/bugbug/commit/9e8d27ec1fd07d0f2bd02613abc376f7c1986109) pull request. My initial assumption was that the code was a bit redundant, so I decided to refactor it by decorating the _retry_for_ function with _tenacity.retry_ which indeed does everything. I will now adjust as expected.

marco-c:
> I don't, however, understand why you have not merged this pull request

I *have* merged it (in fact, you are pointing to a commit on master, not to a pull request :D), but it was reverted because it broke things, as mentioned in the issue.

dalyl-zero:
> > I don't, however, understand why you have not merged this pull request
> 
> I _have_ merged it (in fact, you are pointing to a commit on master, not to a pull request :D), but it was reverted because it broke things, as mentioned in the issue.

Right ><

May I get some insights on understanding what's broken, please? Because I think my final solution might look identical to his.

marco-c:
tenacity was used wrongly, `tenacity.retry` is a decorator, it can't be used to call a function (see tenacity's documentation).

marco-c:
I'm closing as it's been a while since you updated the PR or replied, but feel free to reopen if you still want to work on this.

PR #1251, opened 2020-01-20
Set label as 0 in the QANeeded model when one of the qa flags are removed

### LINKED ISSUE #1249: Set label as 0 in the QANeeded model when one of the qa flags are removed

abhaykatheria:
Fixes #1249.

### DIFF
--- a/bugbug/models/qaneeded.py
+++ b/bugbug/models/qaneeded.py
@@ -92,16 +92,16 @@ class QANeededModel(BugModel):
                 for entry in bug_data["history"]:
                     for change in entry["changes"]:
                         if any(
-                            change["added"].startswith(label)
-                            for label in ["qawanted", "qe-verify", "qaurgent"]
-                        ):
-                            classes[bug_id] = 1
-
-                        if any(
                             change["removed"].startswith(label)
                             for label in ["qawanted", "qe-verify", "qaurgent"]
                         ):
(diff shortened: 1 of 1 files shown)


bugbug/models/qaneeded.py
```diff
@@ -96,6 +96,16 @@ def get_labels(self):
                             for label in ["qawanted", "qe-verify", "qaurgent"]
                         ):
                             classes[bug_id] = 1
+
+            if not found_qa:
```
marco-c: No need to check the condition again and loop through the changes again, just add the inner if (the one about the change) in the earlier loop

bugbug/models/qaneeded.py
```diff
@@ -96,6 +96,13 @@ def get_labels(self):
                             for label in ["qawanted", "qe-verify", "qaurgent"]
                         ):
                             classes[bug_id] = 1
+
+                        if any(
+                            change["removed"].startswith(label)
+                            for label in ["qawanted", "qe-verify", "qaurgent"]
+                        ):
+                            classes[bug_id] = 0
```
marco-c: Move it *before* the other if which sets the class to 1.
This way for example if qawanted is removed but qe-verify added, we will still have the class be 1.

abhaykatheria: oh okay ...

PR #1252, opened 2020-01-21
set label as zero if qa flag is removed

avirlrma:
solves #1249


avirlrma:
@marco-c please review.

marco-c:
Sorry but there is already a PR to fix this, #1251.

PR #1269, opened 2020-02-05
Duplicate detection using Elastic search

chidauri:
Results:

**all_comments = False**
```
145038it [04:57, 486.92it/s]
Recall @ 1: 22.561892352410545%
Recall @ 5: 25.999799538939563%
Recall @ 10: 26.33056028866393%
Precision @ 1: 15.653798256537982%
Precision @ 5: 5.679950186799941%
Precision @ 10: 3.2851805728521253%
Recall: 26.440813871905384%
Precision: 3.695738972674227%
MAP@k : 18.787663264247172%
```

**all_comments = True**
```
145038it [11:39, 207.39it/s]
Recall @ 1: 42.623032975844446%
Recall @ 5: 47.82499749423674%
Recall @ 10: 48.11566603187331%
Precision @ 1: 27.40348692403487%
Precision @ 5: 10.388542963886938%
Precision @ 10: 5.993773349938669%
Recall: 48.24095419464769%
Precision: 6.672905618522755%
MAP@k : 35.02103374333683%
```

### DIFF
--- a/bugbug/similarity.py
+++ b/bugbug/similarity.py
@@ -33,6 +33,8 @@ try:
     from gensim.similarities import SoftCosineSimilarity, SparseTermSimilarityMatrix
     from gensim.summarization.bm25 import BM25
     from gensim.corpora import Dictionary
+    from elasticsearch.helpers import bulk
+    from elasticsearch import Elasticsearch
     from nltk.corpus import stopwords
     from nltk.stem.porter import PorterStemmer
     from nltk.tokenize import word_tokenize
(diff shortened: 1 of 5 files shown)


bugbug/similarity.py
```diff
@@ -626,6 +628,65 @@ def get_distance(self, query1, query2):
...
+                "bug_id": bug["id"],
+                "description": self.text_preprocess(self.get_text(bug)),
+            }
+
+    def index(self):
+        self.elastic_search.indices.delete(index="bugbug", ignore=[400, 404])
+        bulk(self.elastic_search, self.make_documents())
+
+    def get_similar_bugs(self, query):
+        find_similar = self.text_preprocess(self.get_text(query))
```
marco-c: Have you tried considering summary and comments separately?

bugbug/similarity.py
```diff
@@ -626,6 +628,65 @@ def get_distance(self, query1, query2):
...
+        assert (
+            self.elastic_search.ping()
+        ), "Check if Elastic Search Server is running by visiting http://localhost:9200"
+
+    def get_text(self, bug):
+        return "{} {}".format(
+            bug["summary"], " ".join([comment["text"] for comment in bug["comments"]])
+        )
+
+    def text_preprocess(self, text):
```
marco-c: Just use the `text_preprocess` from the base class (if you want to avoid stemming, make it optional like lemmatization).

bugbug/similarity.py
```diff
@@ -626,6 +628,65 @@ def get_distance(self, query1, query2):
...
 
+class ElasticSearchSimilarity(BaseSimilarity):
+    def __init__(self, cleanup_urls=True, nltk_tokenizer=False):
+        super().__init__(cleanup_urls=cleanup_urls, nltk_tokenizer=nltk_tokenizer)
+        self.elastic_search = Elasticsearch()
+        assert (
+            self.elastic_search.ping()
+        ), "Check if Elastic Search Server is running by visiting http://localhost:9200"
+
+    def get_text(self, bug):
```
marco-c: Same here, add the option in the base `get_text` to either get only the first comment or all comments.

bugbug/similarity.py
chidauri: nope, i will give it a try today

chidauri: The results are almost same.

marco-c:
Could you write here the results you're seeing with Elasticsearch (both with all_comments=True and with all_comments=False)? This way we can track future improvements.

chidauri:
> Could you write here the results you're seeing with Elasticsearch (both with all_comments=True and with all_comments=False)? This way we can track future improvements.

Done.

marco-c:
Thanks! I'll merge this as a first step, then we need to think about how to deploy this.

marco-c:
And, how to further improve the results.

Issue #1234, opened 2020-01-07
Use tenacity (or similar) instead of re-implementing a retry function


marco-c:
We are currently implementing https://github.com/mozilla/bugbug/blob/c84c38f964aef514171adac10d54f4a33f3477e9/bugbug/utils.py#L202-L211 and using it in a few places.
We should instead use a library (e.g. tenacity) and remove this custom retry function.

Pranay144:
```
def clone_git_repo(self, repo_url, repo_dir, rev="master", max_attempt=5):
        if not os.path.exists(repo_dir):
            tenacity.retry(
              lambda: subprocess.run(
                ["git","clone","--quiet",repo_url,repo_dir],
                check =True,
                stop = tenacity.stop_after_attempt(max_attempt)
              )
            )
```
@marco-c This is example code from 'commit_classifier.py line 191' with 'retry' function removed and implemented 'tenacity'. 
I want to ask if this code is right? so i can replace retry function from other places.

marco-c:
@Pranay144 it's easier for me to review your code if you simply submit a PR. Don't worry if it's not perfect right from the start, you can make updates to the PR!
From a quick glance, it seems more or less correct (but we don't want to add the max_attempt parameter to clone_git_repo, and "stop" must be an argument of tenacity.retry not subprocess.run).

amit6663:
can i work on this issue? @marco-c

marco-c:
Anyone can work on any issue, read the guidance for first contributors, unless there is already a PR open to fix an issue.
@Pranay144 hasn't opened a PR yet, but it's almost as if he did, so he would have priority.

marco-c:
Reopening as the PR is being reverted (as it breaks things).

NitishNaineni:
@marco-c  i would like to work on this issue. Seems this issue was already fixed but is being reverted, what seems to be the issue here?

marco-c:
@NitishNaineni usage of tenacity was not implemented correctly.

aditya-hari:
@marco-c Would how it is implemented in mozilla/code-convergence#366 work?

marco-c:
Yes

PR #1272, opened 2020-02-07
migrating the retry functions to tenacity

NitishNaineni:
for issue #1234


marco-c:
Looks reasonable, but you should also remove the `retry` function from bugbug/utils.py.

PR #1280, opened 2020-02-11
Install git from Debian testing

La0:


infra/testing.pinning
```diff
@@ -0,0 +1,7 @@
+Package: *
+Pin: release a=buster
+Pin-Priority: 700
+
+Package: *
```
marco-c: What if we pinned just `git` to testing?

marco-c: Could we get rid of Pin-Priority if we do that?

La0: I don't think so, if you don't set priority, packages may end up being installed from testing instead of buster (that's what happenned to me initially as i used `stable` instead of `buster`, and wrecking the setup of mercurial).

+ it's only 3Kb more in your image :smirk:

marco-c: If we do ` Package: git`, don't we make it sure that only git can be installed from testing?

marco-c:
This was needed for the latest version of pydriller (https://github.com/mozilla/bugbug/pull/1271), in particular for https://github.com/ishepard/pydriller/pull/74.

Issue #1279, opened 2020-02-11
Use `rca` from the whiteboard as a bug type


marco-c:
Example in https://bugzilla.mozilla.org/show_bug.cgi?id=1556911.

PR #1283, opened 2020-02-13
Add rca bugtype

ayush-1506:
Fixes: #1279 

At the moment, I'm checking for the occurrence of `[rca` in the whiteboard of the bug, and setting the corresponding index=1 if this exists.

Also, none of the bugs in the bugs db I have at present have this in the whiteboard, so it won't affect the results.


bugbug/models/bugtype.py
```diff
@@ -27,6 +27,7 @@
     "crash": "crash",
     "crashreportid": "crash",
     "perf": "performance",
+    "rca": "rca",
```
marco-c: "rca" is not a type, it's just a way to specify what the type is (e.g. `[rca - Unhandled exceptions]`).
You can find a few example bugs with this query: https://bugzilla.mozilla.org/buglist.cgi?status_whiteboard_type=allwordssubstr&status_whiteboard=%5Brca&query_format=advanced&resolution=FIXED.

marco-c:
> Also, none of the bugs in the bugs db I have at present have this in the whiteboard, so it won't affect the results.

You can download a few bugs locally yourself to test the PR (using bugzilla.download_bugs and the bug IDs from the query above).

bugbug/models/bugtype.py
ayush-1506: Ok, I see. If I have a bug with whiteboard `[rca - XYZ]`, how should I infer which category XYZ comes in (from the list of bug types we have at the moment).
Is there a list somewhere?

marco-c: For now, consider XYZ to be the category. Then we can see which categories we have and check if we can merge any of them with the existing ones.

ayush-1506:
@marco-c I'm iterating through all whiteboards to collect such cases (and add them to keywords_dict) at the moment. Do you think this is a good approach? If so, I'll move to testing this.

marco-c:
> @marco-c I'm iterating through all whiteboards to collect such cases (and add them to keywords_dict) at the moment. Do you think this is a good approach? If so, I'll move to testing this.

Sorry for the delay!
I think you should first check what are all the possible cases with a manual script, then hardcode them in the code (and add an assertion to make sure there are no new cases other than the hardcoded ones).

ayush-1506:
@marco-c Thanks for getting back. I found some time and was able to make a few changes. These include:

1. Hardcoding all the rca types I found. At the moment `rca - XYZ` is treated as a bug of type `XYZ`.
2. Apparently rca whiteboards is of two types: starting with `rca - ` and `rca:`. Added the second type (I was overlooking this in my initial commit)
3. I found a few cases of the kind `rca - coding error - logical error`. (Multiple rca types within a single whiteboard entry.) Added support for these types.
4. In case we come across an rca entry that isn't present in out current list of rca_types, it prints a message. This should help us deal with addition of newer whiteboard entries.

ayush-1506:
I think at the moment rca isn't used precisely, as there are many inconsistencies in the way rca is assigned. For example, I get `rca - coding error`, `rca: coding error`, `rca - coding`, all of which are the same but are used differently.

Also, rather than having two different rca's (`[rca - X][rca - Y]`), I found cases where it is used as `[rca - X - Y]`, which isn't quite accurate.
This is forcing me to handle various if/else cases while getting the rca.

marco-c:
> I found a few cases of the kind rca - coding error - logical error. (Multiple rca types within a single whiteboard entry.) Added support for these types.

In this case, `logical error` should be considered as a subcategory of `coding error`. We can add an option to the model to choose between considering only the top category or both categories.

marco-c:
> In case we come across an rca entry that isn't present in out current list of rca_types, it prints a message. This should help us deal with addition of newer whiteboard entries.

Sounds good, you can make it an assertion instead so we don't risk not noticing the message.

bugbug/models/bugtype.py
```diff
@@ -28,15 +30,52 @@
...
+    "rca - syntax error",
+    "rca - performance error",
+    "rca - testing error",
+    "rca - design error",
+    "rca - external software affecting firefox",
+    "rca - semantic error",
+    "rca - product decision",
+]
+
+KEYWORD_LIST = sorted(set(KEYWORD_DICT.values()).union(set(RCA_TYPES)))
```
marco-c: We can't reuse KEYWORD_LIST, as the rca types are not keywords

bugbug/models/bugtype.py
```diff
@@ -28,15 +30,52 @@
...
 
 
 class BugTypeModel(BugModel):
-    def __init__(self, lemmatization=False, historical=False):
+    def __init__(
+        self, lemmatization=False, historical=False, consider_all_rca_categories=True
```
marco-c:
```suggestion
        self, lemmatization=False, historical=False, rca_subcategories_enabled=False
```

bugbug/models/bugtype.py
```diff
@@ -28,15 +30,52 @@
...
+    ):
         BugModel.__init__(self, lemmatization)
 
         self.calculate_importance = False
 
+        # should be consider only the main category or all sub categories
```
marco-c:
```suggestion
        # should we consider only the main category or all sub categories
```

bugbug/models/bugtype.py
```diff
@@ -28,15 +30,52 @@
...
+    "rca - non-essential requirement",
+    "rca - coding error",
+    "rca - syntax error",
+    "rca - performance error",
+    "rca - testing error",
+    "rca - design error",
+    "rca - external software affecting firefox",
+    "rca - semantic error",
+    "rca - product decision",
+]
```
marco-c: Since there is no match between the bug type categories and the RCA categories, maybe better to define a separate model

marco-c: There is a match with "performance", but not more than that

bugbug/models/bugtype.py
```diff
@@ -102,6 +141,32 @@ def get_labels(self):
             for keyword in bug_data["keywords"]:
                 if keyword in KEYWORD_DICT:
                     target[KEYWORD_LIST.index(KEYWORD_DICT[keyword])] = 1
+            whiteboard_data = bug_data["whiteboard"]
+            for whiteboard in whiteboard_data.split("["):
+                if whiteboard.startswith("rca - "):
+                    # Cases of type rca - X - Y
+                    rca_whiteboards = (
+                        re.sub("]", "", whiteboard[6:]).lower().split(" - ")
```
marco-c: whiteboard[6:] is a bit brittle, maybe better to split by "-" so we also support cases like `rca-`

bugbug/models/bugtype.py
```diff
@@ -102,6 +141,32 @@ def get_labels(self):
             for keyword in bug_data["keywords"]:
                 if keyword in KEYWORD_DICT:
                     target[KEYWORD_LIST.index(KEYWORD_DICT[keyword])] = 1
+            whiteboard_data = bug_data["whiteboard"]
```
marco-c: We should lowercase the string, so we also support cases like `RCA - `

bugbug/models/bugtype.py
ayush-1506: I couldn't find any cases with `rca-`, all of them had `rca - `.

ayush-1506: On second thoughts, I think it is possible someone adds `rca-` , since there is already a lot of irregularity.

ayush-1506:
Ok, so I realized that splitting at `-` isn't a good idea, since we can have cases like `rca - x-y` ("rca - non-essential requirement"), so splitting at "-" won't be helpful here. I've added a case for handling both `rca-` and `rca - `.

ayush-1506:
@marco-c I've made the required changes (as discussed on matrix). One thing I could add to make the solution better (and reduce code duplication) is to change all rca types in RCA_TYPES from `rca - X` to `X` (get rid of the `rca-`), this should make things a bit easier and avoid calling the same function multiple times. Let me know if you think that's a good idea.

ayush-1506:
Changes made, ready for review. I'm running the model at the moment, and it is taking considerable time (maybe because this is a OneVsRest classifier). I'll update here when I have the results.

ayush-1506:
I'm getting an error on running the model:
```
  File "<stdin>", line 1, in <module>
  File "/Users/ayush99/github/bugbug/bugbug/model.py", line 504, in train
    y_test[classified_indices], np.asarray(y_pred_filter)
  File "/Users/ayush99/github/bbug/lib/python3.7/site-packages/sklearn/metrics/classification.py", line 396, in multilabel_confusion_matrix
    y_type, y_true, y_pred = _check_targets(y_true, y_pred)
  File "/Users/ayush99/github/bbug/lib/python3.7/site-packages/sklearn/metrics/classification.py", line 81, in _check_targets
    "and {1} targets".format(type_true, type_pred))
ValueError: Classification metrics can't handle a mix of multilabel-indicator and binary targets```

**Comment by ayush-1506** [comment]
Going deeper, it look to be a general issue with printing confidence score for multilabel prediction. Running the bugtype model (with cross validation disabled) leads to the same error.

**Comment by emceeaich** [comment]
FYI, so we can coordinate on values, this is the list of RCAs I have from tmaity. 

https://bugzilla.mozilla.org/show_bug.cgi?id=1613784

**Comment by ayush-1506** [comment]
Oh, nice! However I'm unsure if I should add this to the rca model right now, or wait until we have a few examples of these. @marco-c ?

**Comment by marco-c** [comment]
> FYI, so we can coordinate on values, this is the list of RCAs I have from tmaity.

Thanks, that helps!

> Oh, nice! However I'm unsure if I should add this to the rca model right now, or wait until we have a few examples of these. @marco-c ?

Let's add them, adding an assertion which fails if we have an unexpected RCA.

File: bugbug/models/rcatype.py [hunk 8]
```diff
@@ -0,0 +1,156 @@
...
+from bugbug import bug_features, bugzilla, feature_cleanup
+from bugbug.model import BugModel
+
+# For the moment, rca - XYZ is treated of bugtype XYZ,
+# so we don't need to store it in a dictionary.
+RCA_TYPES = [
+    "externalapifailure",
+    "logicalerror",
+    "codingerror",
+    "unhandledexceptions",
```
marco-c: We should handle subcategories more clearly to avoid mistakes, e.g. this is a sub-category of Coding Error.

I think we should have two lists, one for categories and one for subcategories, and use the first when subcategories are disabled and the second when subcategories are enabled.

marco-c:
@ayush-1506 could you check what the results are if you temporarily disable printing the confusion matrix?
Anyway, we should defintely fix #1373!

ayush-1506:
@marco-c Just confirming: If I have enabled subcategories, I'll have to replace every category in the main categories list with the corresponding sub categories, is that what you implied?

marco-c:
> @marco-c Just confirming: If I have enabled subcategories, I'll have to replace every category in the main categories list with the corresponding sub categories, is that what you implied?

I meant we should have two lists, one with just the categories and one with both category and subcategory, e.g.:
```
RCA_TYPES = [
    'category1',
    'category2',
]

RCA_SUBCATEGORY_TYPES = [
    'category1 - subcategory1',
    'category1 - subcategory2',
    'category1 - subcategory3',
    'category2 - subcategory1',
]
```

Then, if subcategories are disabled you use RCA_TYPES, if subcategories are enabled you use RCA_SUBCATEGORY_TYPES.

ayush-1506:
@marco-c I'm not adding assert statements for now since (based on the conversation on matrix) the list isn't quite complete at the moment. We can keep adding the rca types as the list is updated. Let me know if that works.

marco-c:
> @marco-c I'm not adding assert statements for now since (based on the conversation on matrix) the list isn't quite complete at the moment. We can keep adding the rca types as the list is updated. Let me know if that works.

Let's add a warning then instead of an assertion with logger.warning

bugbug/models/rcatype.py
```diff
@@ -0,0 +1,167 @@
...
+        )
+
+        self.RCA_LIST = sorted(set(self.RCA_TYPES))
+
+        feature_extractors = [
+            bug_features.has_str(),
+            bug_features.severity(),
+            # Ignore keywords that would make the ML completely skewed
+            # (we are going to use them as 100% rules in the evaluation phase).
+            bug_features.keywords(self.RCA_LIST),
```
marco-c: The RCA values are not keywords, they should instead be ignored from the whiteboard (if you want we can do it in a follow-up, but for now we'd need to comment out the whiteboard feature)

bugbug/models/rcatype.py
```diff
@@ -0,0 +1,167 @@
...
+            classes[int(bug_data["id"])] = target
+
+        return classes, self.RCA_LIST
+
+    def get_feature_names(self):
+        return self.extraction_pipeline.named_steps["union"].get_feature_names()
+
+    def overwrite_classes(self, bugs, classes, probabilities):
+        for i, bug in enumerate(bugs):
+            for keyword in bug["keywords"]:
```
marco-c: Similarly here, the RCA values are in the whiteboard, not keywords

bugbug/models/rcatype.py
```diff
@@ -0,0 +1,167 @@
...
+    "codingerror-syntaxerror",
+    "codingerror-logicalerror",
+    "codingerror-semanticerror",
+    "codingerror-runtimeerror",
+    "codingerror-unhandledexceptions",
+    "codingerror-internalapiissue",
+    "codingerror-networkissue",
+    "codingerror-compatibilityissue",
+    "codingerror-other",
+]
```
marco-c: In RCA_SUBCATEGORIES you don't have things like `cornercase` or `undocumentedchange`, you should still handle them when subcategories are enabled.
When subcategories are enabled, we want to be able to differentiate between all possible subcategories. If a category doesn't have any subcategory, it should still be taken into consideration.

bugbug/models/rcatype.py
```diff
@@ -0,0 +1,167 @@
...
+                        target[self.RCA_LIST.index(rca_whiteboard.split("-", 1)[1])] = 1
+
+                elif whiteboard.startswith("rca:"):
+                    if re.sub("]", "", whiteboard).split(":")[1] in self.RCA_LIST:
+                        target[
+                            self.RCA_LIST.index(
+                                re.sub("]", "", whiteboard).split(":")[1]
+                            )
+                        ] = 1
+
```
marco-c: Print a warning if `rca` is in whiteboard but we were not able to find a class for it

bugbug/models/rcatype.py
```diff
@@ -0,0 +1,167 @@
...
+    def get_labels(self):
+        classes = {}
+
+        for bug_data in bugzilla.get_bugs():
+            target = np.zeros(len(self.RCA_LIST))
+            whiteboard_data = re.sub(" ", "", bug_data["whiteboard"]).lower()
+            for whiteboard in whiteboard_data.split("["):
+                if whiteboard.startswith("rca-"):
+                    rca_whiteboard = re.sub("]", "", whiteboard)
+                    if rca_whiteboard.split("-", 1)[1] not in self.RCA_LIST:
```
marco-c: How will this work when you have the category-level model and you have `rca-codingerror-unhandledexceptions` in the whiteboard?

ayush-1506:
Thanks for the review. I've made quite a few revisions to ensure we don't have redundant code. Some highlights are:
1. Warning when a category is not found. (when using categories)
2. Warning saying that a subcategory has been added to the rca_list (when using subcategories)
3. get_rca function returns a list of rca whiteboards from a list of bugs. It is now used inside get_labels and overwrite_classes.
4. Fixed the issue with https://github.com/mozilla/bugbug/pull/1283#discussion_r397812462.

bugbug/models/rcatype.py
```diff
@@ -0,0 +1,193 @@
...
+        self.RCA_TYPES = (
+            RCA_SUBCATEGORIES if rca_subcategories_enabled else RCA_CATEGORIES
+        )
+
+        self.RCA_LIST = sorted(set(self.RCA_TYPES))
+
+        feature_extractors = [
+            bug_features.has_str(),
+            bug_features.severity(),
+            # Ignore keywords that would make the ML completely skewed
```
marco-c, line 75: This comment should be rephrased and moved to before the whiteboard line

bugbug/models/rcatype.py
marco-c: I think this comment still applies. When `rca_subcategories_enabled` is True, you are only considering subcategories, we should instead consider both.

bugbug/models/rcatype.py
```diff
@@ -0,0 +1,193 @@
...
+
+        self.clf = OneVsRestClassifier(xgboost.XGBClassifier(n_jobs=16))
+
+    def get_rca(self, bugs):
+        all_rca = []
+        bug_ids = []
+        for bug in bugs:
+            bug_ids.append(int(bug["id"]))
+            rca = []
+            whiteboard_data = re.sub(" ", "", bug["whiteboard"]).lower()
```
marco-c, line 134: I'd put the code to get the RCA out of a whiteboard string in a separate function, so you can write some tests for it.

bugbug/models/rcatype.py
```diff
@@ -0,0 +1,193 @@
...
+                    whiteboard[3]
+                    if (whiteboard.startswith("rca-") or whiteboard.startswith("rca:"))
+                    else False
+                )
+                if delimiter:
+                    rca_whiteboard = re.sub("]", "", whiteboard)
+                    rca_whiteboard_split = (
+                        rca_whiteboard.split(delimiter, 1)
+                        if self.rca_subcategories_enabled
+                        else rca_whiteboard.split(delimiter)
```
marco-c, line 146: How does this work when you have something like: `rca: codingerror - compatibilityissue`? The delimiter you get earlier is `:`, but you should split by `-`.
(This is a good reason to have some unit tests like mentioned above)

marco-c:
Thanks @ayush-1506, and sorry for the delay, I've been super busy!
The code looks good overall, there are just a couple of details to fix (and I'd like to see some unit tests for the part where you parse the whiteboard, as that's the core of the model).

bugbug/models/rcatype.py
ayush-1506: Oh, I didn't know we could have such "hybrid" cases too. Will add support for these types also.

ayush-1506:
@marco-c Quick question: I don't understand why the build is failing. Any idea? I think it's probably because I added a new bug to bugs.json in `tests/fixtures`, but I don't get why it's failing.

marco-c:
> @marco-c Quick question: I don't understand why the build is failing. Any idea? I think it's probably because I added a new bug to bugs.json in tests/fixtures, but I don't get why it's failing.

You can look at the error on Taskcluster, just click on "Details", then select the task which failed and then click on "View Live Log".

tests/test_rcatype.py
```diff
@@ -0,0 +1,58 @@
...
+    ]
+    # Case 4: subcategories enabled, with rca already present in the list
+    model = RCATypeModel(rca_subcategories_enabled=True)
+    assert model.get_rca_from_whiteboard("[RCA: codingerror - syntaxerror]") == [
+        "codingerror-syntaxerror"
+    ]
+    assert model.get_rca_from_whiteboard(
+        "[RCA: codingerror - syntaxerror][rca: codingerror:logicalerror]"
+    ) == ["codingerror-syntaxerror", "codingerror-logicalerror"]
+    # Case 5: subcategories enables, with rca not present in list
```
marco-c, line 44:
```suggestion
    # Case 5: subcategories enabled, with rca not present in list
```

tests/test_rcatype.py
```diff
@@ -0,0 +1,58 @@
...
+    initial_rca_subcategories = len(model.RCA_LIST)
+    assert model.get_rca_from_whiteboard("[RCA: codingerror - newerror]") == [
+        "codingerror-newerror"
+    ]
+    assert len(model.RCA_LIST) == initial_rca_subcategories + 1
```
marco-c, line 49: As we said earlier, we should assume the RCA list is complete and just print a warning when we find a non-existing category/subcategory, not add it to the list

tests/test_rcatype.py
```diff
@@ -0,0 +1,58 @@
...
+        "cornercase",
+        "testingerror",
+    ]
+    assert model.get_rca_from_whiteboard("[rca : cornercase][rca : codingerror]") == [
+        "cornercase",
+        "codingerror",
+    ]
+    # Case 4: subcategories enabled, with rca already present in the list
+    model = RCATypeModel(rca_subcategories_enabled=True)
+    assert model.get_rca_from_whiteboard("[RCA: codingerror - syntaxerror]") == [
```
marco-c, line 38: You should also add a test where subcategories are disabled and the output for `[RCA: codingerror - syntaxerror]` is `codingerror`

bugbug/models/rcatype.py
```diff
@@ -0,0 +1,200 @@
...
+                    logger.warning("Adding " + rca_whiteboard_split[1] + " to RCA_LIST")
+                    self.RCA_TYPES.append(rca_whiteboard_split[1])
+                    bisect.insort(self.RCA_LIST, rca_whiteboard_split[1])
+                    rca.append(rca_whiteboard_split[1])
+                else:
+                    rca.append(rca_whiteboard_split[1])
+        return rca
+
+    # return rca for all bugs
+    def get_rca_from_bugs(self, bugs):
```
marco-c, line 168: Nit: you can just move this inside `get_labels`

bugbug/models/rcatype.py
```diff
@@ -0,0 +1,200 @@
...
+    # return rca from a whiteboard string
+    def get_rca_from_whiteboard(self, whiteboard_data):
+        rca = []
+        whiteboard_data = re.sub(" ", "", whiteboard_data).lower()
+        for whiteboard in whiteboard_data.split("["):
+            delimiter = any(
+                [whiteboard.startswith("rca-"), whiteboard.startswith("rca:")]
+            )
+
+            if delimiter:
```
marco-c, line 138: Nit: you can do:
```
if not any(whiteboard.startswith(s) for s in ("rca-", "rca:")):
    continue

...
```

This way:
- the rest of the code doesn't have to be indented;
- you avoid a variable that is just used for the `if`, making the code a bit more readable

bugbug/models/rcatype.py
```diff
@@ -0,0 +1,186 @@
...
+
+            if rca_whiteboard_split[1] not in self.RCA_LIST:
+                logger.warning(rca_whiteboard_split[1] + " not in RCA_LIST")
+            else:
+                rca.append(rca_whiteboard_split[1])
+        return rca
+
+    def get_labels(self):
+        # return rca for all bugs
+        def get_rca_from_bugs(bugs):
```
marco-c, line 155: Nit: I don't see the need for this function, why not making it inline directly in the outer function (`get_labels`)?

marco-c:
Looks good! Just one thing to fix (removing that function and just writing its code in `get_labels` directly)

ayush-1506:
@marco-c Done.

bugbug/models/rcatype.py
```diff
@@ -0,0 +1,181 @@
...
+        bug_ids = []
+        for bug in bugzilla.get_bugs():
+            bug_ids.append(int(bug["id"]))
+            all_rca.append(self.get_rca_from_whiteboard(bug["whiteboard"]))
+        classes = {}
+        for bug_id, bug_rca in zip(bug_ids, all_rca):
+            target = np.zeros(len(self.RCA_LIST))
+            for rca in bug_rca:
+                target[self.RCA_LIST.index(rca)] = 1
+            classes[bug_id] = target
```
marco-c, line 164: This code is still a bit overcomplex, why do you need to do two passes?

Something like this should work as well:
```Python
    target = np.zeros(len(self.RCA_LIST))
    for rca in self.get_rca_from_whiteboard(bug["whiteboard"]):
        target[self.RCA_LIST.index(rca)] = 1
    classes[bug["id"]] = target
```

ayush-1506:
Phew, finally! :)

Issue #1300, opened 2020-02-21
Add the scaffolding needed to add an API call to select tests


marco-c:
This issue is just about adding an API call (with a hardcoded result) to the http service, preparing the ground for the actual API.

It should be similar to the bug classification API call (`model_prediction`), but its input will be a repository and a commit hash.

As part of this issue, we should:
- add the API definition (in http_service/app.py);
- the implementation of queuing the classification from the web dyno to the worker dyno using rq (again, in http_service/app.py, like we do for bugs).
- a `select_tests` function in http_service/models.py which for now returns a hardcoded result.

Note: the implementation should be simpler than the one for bugs, as we don't need (actually, we don't want to!) cache results.

Blocks #632.

ahal:
@La0 would it help if I took a look at this issue while you worked on #1299 ? Or do you think you would have this finished in a couple days anyway?

La0:
I'm on PTO tommorow, will start looking into these issues on Wednesday.
I *should* have something by the end of the week, if that's too slow, feel free to re-assign the issues.

ahal:
Ok I don't have permission to re-assign, but I'm going to take a look at getting a head start here. If I'm not done by the time you're ready to work on it.. I can upload what I have and you can take over again if you like.

La0:
@ahal I've assigned the issue to you, as I see some commits from your fork linked to this issue.

PR #1319, opened 2020-02-25
Implement the outline of a push scheduler service

ahal:
Fixes issue #1300

### DIFF
--- a/http_service/bugbug_http/app.py
+++ b/http_service/bugbug_http/app.py
@@ -7,7 +7,9 @@ import json
 import logging
 import os
 import uuid
+from dataclasses import dataclass, field
 from datetime import datetime, timedelta
+from typing import Any, Callable, List
 
 from apispec import APISpec
 from apispec.ext.marshmallow import MarshmallowPlugin
@@ -22,7 +24,7 @@ from rq.exceptions import NoSuchJobError
 from rq.job import Job
 
 from bugbug import get_bugbug_version
-from bugbug_http.models import MODELS_NAMES, change_time_key, classify_bug, result_key
+from bugbug_http.models import MODELS_NAMES, classify_bug, schedule_tests
 from bugbug_http.utils import get_bugzilla_http_client
 
 API_TOKEN = "X-Api-Key"
@@ -73,7 +75,7 @@ class BugPrediction(Schema):
     extra_data = fields.Dict()
 
 
-class BugPredictionNotAvailableYet(Schema):
+class NotAvailableYet(Schema):
     ready = fields.Boolean(enum=[False])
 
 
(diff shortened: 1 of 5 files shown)


ahal:
This is super WIP and not ready for review. I pushed it in case anyone wanted to provide early feedback and to ensure someone else doesn't start on this from scratch in the meantime.

ahal:
I think it would be better if we implemented a `JobInfo` dataclass that stored all the job related state (rather than passing `func` and `*args` around all over the place). I started implementing it and then realized it was distracting me more than it should, so decided to leave it for a follow up commit.

http_service/bugbug_http/app.py
```diff
@@ -115,13 +170,12 @@ def schedule_bug_classification(model_name, bug_ids):
     job_id_mapping = {get_mapping_key(model_name, bug_id): job_id for bug_id in bug_ids}
     redis_conn.mset(job_id_mapping)
```
marco-c: This part is now in schedule_job and should be removed, right? (note: move the comment too)

http_service/bugbug_http/app.py
```diff
@@ -85,24 +91,73 @@ class UnauthorizedError(Schema):
...
-def get_mapping_key(model_name, bug_id):
-    return f"bugbug:mapping_{model_name}_{bug_id}"
+def get_mapping_key(*args):
+    """Return a key that can be used to map `args` to a job id.
+
+    If args aren't specified, a key that maps `request.path` to a job id will
+    be used instead.
+    """
+    func = get_target_func()
+    return f"bugbug:job_id:{func.__name__}:{'_'.join(args)}"
```
marco-c: What if we just use the args? Since the model name is in there, we shouldn't have problems

http_service/bugbug_http/app.py
```diff
@@ -85,24 +91,73 @@ class UnauthorizedError(Schema):
...
+
+
+def get_result_key(*args, func=None):
+    """Return a key that can be used to store the result of a job.
+
+    If args aren't specified, a key that maps `request.path` to a job result
+    will be used instead.
+    """
+    func = func or get_target_func()
```
marco-c: Same here

http_service/bugbug_http/app.py
```diff
@@ -85,24 +91,73 @@ class UnauthorizedError(Schema):
...
+    return f"bugbug:job_result:{func.__name__}:{'_'.join(args)}"
+
+
+def schedule_job(*args, job_id=None):
```
marco-c: And here I'd make `func` a parameter of `schedule_job`, so we can get rid of get_target_func and make it more explicit

http_service/bugbug_http/app.py
```diff
@@ -401,7 +449,7 @@ def batch_prediction(model_name):
     auth = headers.get(API_TOKEN)
 
     if not auth:
-        return jsonify(UnauthorizedError().dump({}).data), 401
+        return jsonify(UnauthorizedError().dump({})), 401
```
marco-c: Why this change?

http_service/bugbug_http/app.py
```diff
@@ -452,6 +500,81 @@ def batch_prediction(model_name):
...
+    else:
+        LOGGER.info("Request with API TOKEN %r", auth)
+
+    # Support the string 'autoland' for convenience.
+    if branch == "autoland":
+        branch = "integration/autoland"
+
+    url = f"https://hg.mozilla.org/{branch}/json-automationrelevance/{rev}"
+    r = requests.get(url)
```
marco-c: We might want to move this to the worker side to keep the web dyno always quickly responsive, as the request could be slow.

http_service/bugbug_http/models.py
```diff
@@ -84,6 +80,8 @@ def preload_models():
 def classify_bug(
     model_name, bug_ids, bugzilla_token, expiration=DEFAULT_EXPIRATION_TTL
 ):
+    from bugbug_http.app import get_result_key
```
marco-c: Is it a problem if we add this as a top-level import?

http_service/bugbug_http/models.py
```diff
@@ -145,3 +143,22 @@ def classify_bug(
...
     return "OK"
+
+
+def schedule_tests(revset, expiration=DEFAULT_EXPIRATION_TTL):
+    from bugbug_http.app import get_result_key
+
+    data = {
+        "tasks": ["test-macosx1014-64/debug-gtest-1proc"],
+        "groups": ["caps/test/unit/xpcshell.ini", "dom/indexedDB/test/mochitest.ini",],
+    }
```
marco-c: Not sure yet how we will handle configuring different shadow schedulers with different parameters.
I see two options:
- the configuration (confidence level) is passed to the HTTP service, and the HTTP service only returns tasks/groups with confidence higher than the one given as input;
- the HTTP service returns all tasks/groups with their associated confidence level, and the client (the shadow scheduler) does the filtering on the confidence level.

If we do the second, we're going to have to change a bit the output here (something like `["test-macosx1014-64/debug-gtest-1proc", 0.67]`).

http_service/tests/test_push_schedules.py
```diff
@@ -0,0 +1,155 @@
...
+import pytest
+from rq.exceptions import NoSuchJobError
+
+from bugbug_http import app
+
+# { <job id>: <result key> }
+JOBS = {}
+
+
+class RedisMock:
```
marco-c: There is a `fakeredis` package for mocking too, but this is good too for me.

http_service/bugbug_http/app.py
marco-c: > I think it would be better if we implemented a JobInfo dataclass that stored all the job related state (rather than passing func and *args around all over the place). I started implementing it and then realized it was distracting me more than it should, so decided to leave it for a follow up commit.

And yes, this would probably look better, but we can leave it for a follow-up.

http_service/bugbug_http/app.py
ahal: No, because `schedule_job` sets a mapping key that contains the entire list of bug ids, whereas this is setting mapping keys for each individual bug id. This is needed in case someone calls:
`/model/predict/batch` (while posting bug 123456)

and then later someone calls:
`/model/predict/123456`

This way a mapping key for the second call already exists and we don't do unnecessary work.

http_service/bugbug_http/app.py
ahal: It would work for now, but technically args aren't unique across jobs. The unique identifier would be function + args. Imagine we add a `/commit/<rev>/schedules` API in the future. Then a `<rev>` could be both a valid argument to the `/push` and the `/commit` endpoints. At that point, we'd need the function name to help uniquely identify those jobs. So I think it's best to leave it in.

I guess we could remove the `bugbug:job_id` prefix, but I figured it might help if we ever needed to inspect the redis store manually. I'm also not sure what other types of keys are being placed in there, so it helps avoid conflicts in general.

http_service/bugbug_http/app.py
ahal: Will do, I was trying to avoid needing to pass it around everywhere.. but agree it ended up being more fragile than it needed to.

http_service/bugbug_http/app.py
ahal: It is broken at the moment and raises `AttributeError: dict objects have no attribute 'data'`. This is because the `.data` is being called on the return value from `dump` (which is just a dict) as opposed to the return value from `jsonify` (which is a flask `Response` object).

In my testing, returning a `Response` object seems to work just as well and I thought it might be more appropriate than returning a dict here. Though maybe that's a bad assumption.

http_service/bugbug_http/models.py
ahal: Yeah, it's a circular import :/

This could definitely be implemented better.. maybe when we create a `JobInfo` class it can move to a utility file or something.

http_service/bugbug_http/models.py
ahal: I think I like the second approach better. Though it will be easy to adjust on the client side so I think it's fine to keep it simple for now and adjust as needed when we run into it.

http_service/tests/test_push_schedules.py
ahal: It's simple enough and avoids another dependency :). Feel free to fix in a follow-up if you like.

http_service/bugbug_http/app.py
marco-c: Oh, OK. I wonder if that could be a problem, is there a limit on the length of the key?

http_service/bugbug_http/models.py
marco-c: OK, let's file a follow-up so we don't forget

http_service/bugbug_http/app.py
ahal: Apparently the redis key size limit is 512MB.. so we're probably good :)

That said, we don't actually use the key that has all the bug ids in it.. we only use that to schedule the job, and then the worker adds results for each individual bug. So we could potentially not set that key if you are worried (though doing so makes the two endpoints diverge more in implementation).

ahal:
Please make sure the bug classification endpoint is working as expected before deploying live! I wasn't sure how to test it, and the `http_service/tests` aren't very thorough.

PR #1320, opened 2020-02-27
Clone MC & Download DBs in http workers

### LINKED ISSUE #1299: Add a setup step to the http service to setup the environment for classifying commits
This is partly similar to the what we are doing in the code-review backend.
We need to do:
- Clone mozilla-central;
- Download the test scheduling DB (https://github.com/mozilla/bugbug/blob/cb51deac6235477e7f2dbc13352a2627d129451c/scripts/commit_classifier.py#L179-L183);
- Update the commits DB (https://github.com/mozilla/bugbug/blob/cb51deac6235477e7f2dbc13352a2627d129451c/scripts/commit_classifier.py#L219-L229).

We should update the commits DB periodically, but I think doing it daily (whenever Heroku restarts) is a good enough approximation.

Blocks #632.

La0:
Should fix #1299

### DIFF
--- a/bugbug/repository.py
+++ b/bugbug/repository.py
@@ -12,6 +12,7 @@ import logging
 import math
 import os
 import pickle
+import subprocess
 import sys
 import threading
 from datetime import datetime
@@ -965,11 +966,7 @@ def clone(repo_dir):
 
     cmd.insert(0, hglib.HGPATH)
 
-    proc = hglib.util.popen(cmd)
-    out, err = proc.communicate()
-    if proc.returncode:
-        raise hglib.error.CommandError(cmd, proc.returncode, out, err)
-
+    subprocess.check_output(cmd)
     logger.info("mozilla-central cloned")
 
     # Remove pushlog DB to make sure it's regenerated.
--- a/http_service/Dockerfile.bg_worker
+++ b/http_service/Dockerfile.bg_worker
@@ -1,6 +1,6 @@
 ARG BUGBUG_VERSION=latest
 
-FROM mozilla/bugbug-base:$BUGBUG_VERSION
+FROM mozilla/bugbug-commit-retrieval:$BUGBUG_VERSION
 
 # Install dependencies first
 COPY requirements.txt /requirements-http.txt
--- a/infra/data-pipeline.yml
+++ b/infra/data-pipeline.yml
@@ -1333,6 +1333,7 @@ tasks:
       dependencies:
         - docker-build
         - integration-test
+        - test-label-scheduling-history-generator
       scopes:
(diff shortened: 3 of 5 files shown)


http_service/Dockerfile.bg_worker
```diff
@@ -14,6 +14,7 @@ WORKDIR /code/
 ARG CHECK_MODELS
 ENV CHECK_MODELS="${CHECK_MODELS}"
 
-RUN bash /code/http_service/ensure_models.sh
+RUN bash /code/http_service/tools/ensure_models.sh
+RUN bash /code/http_service/tools/bootstrap-mercurial.sh
```
marco-c: Maybe you could base this image off the mozilla/bugbug-commit-retrieval image instead, so it will be quicker to build

http_service/worker.py
```diff
@@ -19,10 +19,14 @@
...
 # `sys.modules['http_service.models'].MODEL_CACHE`
 import http_service.models  # noqa: E402 isort:skip
+import http_service.boot  # noqa: E402 isort:skip
 
 # Preload libraries
 http_service.models.preload_models()
 
+# Bootstrap the worker assets
+http_service.boot.boot_worker()
+
```
marco-c: Nit: we could possibly do preload_models and boot_worker in parallel

http_service/Dockerfile.bg_worker
La0: Done.

La0:
This PR is now rebased on top of #1327 and is only Python code (no more mercurial setup as the image is based on top of bugbug-commit-retrieval).

bugbug/repository.py
```diff
@@ -965,11 +966,7 @@ def clone(repo_dir):
 
     cmd.insert(0, hglib.HGPATH)
 
-    proc = hglib.util.popen(cmd)
-    out, err = proc.communicate()
-    if proc.returncode:
-        raise hglib.error.CommandError(cmd, proc.returncode, out, err)
-
+    subprocess.check_output(cmd)
```
marco-c: With this change we'd lose the stdout, previously it was included in the exception

marco-c: Sorry, the stderr

infra/data-pipeline.yml
```diff
@@ -1333,6 +1333,7 @@ tasks:
       dependencies:
         - docker-build
         - integration-test
+        - test-label-scheduling-history-generator
```
marco-c: When we will also download the model, we can remove this and just make the build task depend on the trainer task

marco-c:
Merged in cfd82a8ad3635ab0214818c461b5220dccf76aa8 (without the changes in clone).

Issue #1257, opened 2020-01-31
Use exponential backoff when retrying with tenacity


marco-c:
After switching to tenacity (#1234), we should start using exponential backoff instead of normal retries.

aditya-hari:
Where will this change have to be made?

menarulalam:
Can I work on this? I would like to contribute to the open source community.

marco-c:
This requires #1234 to be done first.

Luis-gd:
I am new in open source contributions and I want to work on this issue, where I can start?

marco-c:
@Luis-gd the first step is reading the documentation of the tenacity library to figure out how to use exponential backoff, then find all usages of tenacity in bugbug and apply the necessary changes.

Luis-gd:
All right, I am working on this.

marco-c:
Note that there are already two PRs open to fix this, the first opened PR has the priority.

PR #1321, opened 2020-02-27
wait_fixed() to wait_exponential()

IreneTenison:
Use exponential backoff when retrying with tenacity #1257

### DIFF
--- a/scripts/commit_classifier.py
+++ b/scripts/commit_classifier.py
@@ -192,7 +192,7 @@ class CommitClassifier(object):
                 lambda: subprocess.run(
                     ["git", "clone", "--quiet", repo_url, repo_dir], check=True
                 ),
-                wait=tenacity.wait_fixed(30),
+                wait=tenacity.wait_exponential(),
                 stop=tenacity.stop_after_attempt(5),
             )()
 
@@ -203,7 +203,7 @@ class CommitClassifier(object):
                 capture_output=True,
                 check=True,
             ),
-            wait=tenacity.wait_fixed(30),
+            wait=tenacity.wait_exponential(),
             stop=tenacity.stop_after_attempt(5),
         )()
 
@@ -211,7 +211,7 @@ class CommitClassifier(object):
             lambda: subprocess.run(
                 ["git", "checkout", rev], cwd=repo_dir, capture_output=True, check=True
             ),
-            wait=tenacity.wait_fixed(30),
+            wait=tenacity.wait_exponential(),
             stop=tenacity.stop_after_attempt(5),
         )()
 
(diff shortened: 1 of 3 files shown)


scripts/commit_classifier.py
```diff
@@ -192,7 +192,7 @@ def clone_git_repo(self, repo_url, repo_dir, rev="master"):
                 lambda: subprocess.run(
                     ["git", "clone", "--quiet", repo_url, repo_dir], check=True
                 ),
-                wait=tenacity.wait_fixed(30),
+                wait=tenacity.wait_exponential(),
                 stop=tenacity.stop_after_attempt(5),
             )()
```
marco-c: What is the total time we would wait with the new code vs the old code?

IreneTenison: wait_fixed(30) would be waiting for 30 seconds between retries.
wait_exponential() would wait for 2^x * 1 second between each retry starting with 0 seconds, then up to 1073741823 seconds. 
should the max time be limited to 30 seconds?

marco-c: So with wait_fixed we were waiting 30 seconds * 5 times = 150 seconds. We should ensure we have a similar total with wait_exponential. If it is 2^x * 1 up to x == 4 (if it starts from 0), it is 1 second, 2 seconds, 4 seconds, 8 seconds, 16 seconds -> in total 31 seconds.
So we should:
- start with waiting more than just 1 second (probably with 4);
- stop after we have waited in total more than just 31 seconds.

IreneTenison: wait_exponential(multiplier = 1, min = 4, max = 31) would wait for 4 seconds, 8 seconds, 16 seconds,  a total of 28 seconds (less than 30s). If max = 32, the total wait time would be 60 seconds, which would be too high. Is it okay if the total wait time is 28 seconds?

marco-c: Before your PR we were waiting 150 seconds in total, so 60 seconds is not too high, it's too low :)

IreneTenison: wait_exponential(multiplier = 1, min = 4, max = 150) would wait for a maximum of 150 seconds. Is this fine? I have done a PR with this change as well.

marco-c: This doesn't work as you expect, try it in a standalone script or read the documentation (https://tenacity.readthedocs.io/en/latest/#waiting-before-retrying). It is waiting forever, never stopping.

marco-c: E.g. you can try with this script:
```Python
import tenacity

@tenacity.retry(wait=tenacity.wait_exponential(multiplier=1, min=4, max=150))
def wait_test():
    print("qua")
    raise Exception

wait_test()
```

marco-c:
@IreneTenison are you still planning on working on this PR?

marco-c:
Closing for lack of response.

PR #1327, opened 2020-02-27
Setup http_service as Python package bugbug_http

La0:
With this PR, the bugbug API is available as a Python package named `bugbug_http`.
Its inner modules are accessible through `from bugbug_http import xxx`, which fixes all the import paths shenanigans.

- I move both Dockerfiles into `infra` in order to respect the current architecture.
- ~To share the same version number, i write it as `/etc/bugbug-version` in the docker base image.~ I simply use the installed bugbug version.
- The worker script is exposed as `bugbug-http-worker` console script.


infra/dockerfile.http_worker
```diff
@@ -2,11 +2,9 @@ ARG BUGBUG_VERSION=latest
...
 FROM mozilla/bugbug-base:$BUGBUG_VERSION
 
-COPY requirements.txt /code/http_service/
-
-RUN pip install --disable-pip-version-check --quiet --no-cache-dir -r /code/http_service/requirements.txt
-
-COPY . /code/http_service/
+# Setup http service as package
+COPY http_service /code/http_service
+RUN pip install --disable-pip-version-check --quiet --no-cache-dir /code/http_service
```
marco-c: This way we will reinstall dependencies even when we don't need to (e.g. if we change a python file).

.pre-commit-config.yaml
```diff
@@ -50,6 +50,15 @@ repos:
     rev: v0.761
     hooks:
     -   id: mypy
```
marco-c: Why are these changes necessary?

.dockerignore
```diff
@@ -97,5 +97,4 @@ venv/
 # Project-specific stuff
 cache/
 data/
-http_service/
```
marco-c: I guess doing this doesn't increase the build time by much

http_service/Dockerfile
```diff
@@ -1,14 +0,0 @@
-ARG BUGBUG_VERSION=latest
```
marco-c: I'd keep the Dockerfiles under http_service, so we can more easily move it to a separate repository if we want in the future.

http_service/bugbug_http/worker.py
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env python
```
marco-c: Can you make a first commit where you just move the file and one where you apply the modifications? This way maybe git will be able to tell that it isn't a new file and we won't lose the history

infra/dockerfile.http_worker
La0: We still need to trigger the http_service's setup.py. What would you prefer ? pip install requirements & then a pip install for http_service module ?

.pre-commit-config.yaml
La0: Unfortunately mypy does not support having twice the same module name ([failing lint task](https://community-tc.services.mozilla.com/tasks/JCk2ea2TStq4FDguMjZD6Q/runs/0/logs/https%3A%2F%2Fcommunity-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FJCk2ea2TStq4FDguMjZD6Q%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log#L110))

http_service/Dockerfile
La0: Done

http_service/bugbug_http/worker.py
La0: I just did that in 52f4eb4d125145e7a70bcdf094115e78938c6854 (moving files) and then 24fe16099a82458d3e3ddd7208093fad191cfe9b (patching files), but the result is the same.

La0: If you want to keep track of the history, you should merge without squashing

infra/dockerfile.http_worker
La0: Just a side note: that build is slow because of the models loading/checking, not really because of the deps setup

marco-c: > We still need to trigger the http_service's setup.py. What would you prefer ? pip install requirements & then a pip install for http_service module ?

Yeah, similarly to what we do for the bugbug-base

.pre-commit-config.yaml
marco-c: Painful :(

infra/dockerfile.http_worker
La0: Done

La0:
The PR is now setup with 2 commits only:
1. move files in `bugbug_http`
2. fix usage

marco-c:
Merged with 194c1845d66a56a8300ce40fde7648aff6c70cdd and 0eb7f91a232857eef8a6b578024ede569dfe85f1.

PR #1329, opened 2020-02-27
Use exponential backoff in tenacity calls

max-reborn:
proposed fix for [#1257](https://github.com/mozilla/bugbug/issues/1257)

I have set 16 seconds as the minimum wait time, 128 seconds as the maximum, while keeping 5 retries. I'd be great if someone could confirm that these parameters make sense.

I was also wondering if someone could suggest good parameters to these scripts to perform an end to end test of them, and verify my fixes didn't break anything. The scripts to not appear unit tested, so I think the best thing to do would be just to run them, but being new to the project I currently lack the context to know what parameters would be good, and help would be much appreciated!

Thanks :)


marco-c:
@max-reborn the first PR to fix #1257 has been closed, so I'll now consider this one.

marco-c:
Looks good, but let's use 64 as the max, so in total we wait 240 seconds.

max-reborn:
Sounds good, will do

PR #1332, opened 2020-02-28
Install bugbug_http module in integration tests

La0:


La0:
Testing on [taskcluster](https://community-tc.services.mozilla.com/tasks/MOZbOjRoQz-PSlyc6shv9g)

scripts/integration_test.sh
```diff
@@ -69,4 +69,4 @@ worker_pid=$!
 trap 'kill $gunicorn_pid && kill $worker_pid && kill $redis_pid' EXIT
 
 # Then check that we can correctly classify a bug
-sleep 10 && python http_service/tests/test_integration.py
+sleep 300 && python http_service/tests/test_integration.py
```
marco-c: Instead of doing this, let's increase the retry count in test_integration.py (so if the setup ends earlier, we don't needlessly sleep)

PR #1336, opened 2020-02-28
Make sure we run the 'http_service' tests in CI

ahal:
Running them alongside the `/tests` directory in the same pytest invocation results in:
```
pytest fixture "mock_data" not found
```

for each of the http_service tests. I guess `pytest` registers it from the /tests/conftest.py file first, and then later can't find it anywhere in `/http_service/tests`.

This gets around the issue in CI, but running:
```
pytest tests http_service/tests
```

still fails locally for me.


ahal:
Turns out these are already running in a separate task.

PR #1347, opened 2020-03-02
[http_service] Don't schedule new jobs when job_id has status 'queued'

### LINKED ISSUE #1340: Don't fill queue with requests while the worker is booting
Here's an example failure of the integration test:
[live_backing.txt](https://github.com/mozilla/bugbug/files/4268644/live_backing.txt)

I think what's happening is that we enqueue a lot of requests while the workers are starting, and so we spend a lot of time doing them all.

We should queue jobs only once while the workers are not running.

ahal:
Fixes #1340


ahal:
I don't think this was adequately tested. Though it worked for me while implementing https://github.com/mozilla/bugbug/pull/1319

PR #1359, opened 2020-03-03
Return 404 when model doesn't exist

### LINKED ISSUE #1265: Return 404 when a model which doesn't exist is requested
The HTTP service APIs (http_service/bugbug_http/app.py) should return 404 when the requested model is not supported (that is, not in the list at https://github.com/mozilla/bugbug/blob/e2a54a04ab69efd91c74a67805271009afb714a5/http_service/bugbug_http/models.py#L23).

kaushiksk:
Fixes #1265

Returns 404 with an error message when an unknown model is used.
Added test cases for the same.


kaushiksk:
@marco-c Please review and  let me  know if this is what you intended

PR #1369, opened 2020-03-05
Apply mercurial patches on local repo when scheduling tests

### LINKED ISSUE #1301: In the API to select tests, apply the requested patch
Depends on #1299 and #1300.

This is quite similar to what we're doing in the code-review bot, with the difference that we're not getting patches from Phabricator but from hgmo.

In the worker, we should retrieve the patch from the specified repository and at the specified revisions (with its non public parents) and apply it to the locally cloned mozilla-central repository.

Blocks #632.

La0:
Fixes #1301

It should probably have some unit tests around the stack application directly.


bugbug/utils.py
```diff
@@ -370,3 +370,11 @@ def __exit__(self, *args):
                 future.cancel()
             raise e
         return super(ThreadPoolExecutorResult, self).__exit__(*args)
+
+
+def hgmo_patch(branch: str, revision: str) -> str:
```
marco-c: Nit: ```suggestion
def get_hgmo_patch(branch: str, revision: str) -> str:
```

File: bugbug/repository.py [hunk 2]
```diff
@@ -1011,6 +1012,48 @@ def clone(repo_dir):
     clean(repo_dir)
 
 
+def apply_stack(repo_dir, stack, branch):
```
marco-c: We have to analyze all patches even if one was already available in the repo, so we should do something like:
- find parent of the first patch in the list returned by json-automationrelevance (the info is available in the data returned by json-automationrelevance);
- update mercurial repo to ^;
- apply the stack of patches.

marco-c: > update mercurial repo to ^;

Or "tip" if that's not available (but we should assert that it is when the repo is autoland).

http_service/bugbug_http/models.py
```diff
@@ -145,18 +143,22 @@ def classify_bug(model_name, bug_ids, bugzilla_token):
...
+    # Load the full stack of patches leading to that revision
+    try:
+        stack = hgmo_stack(branch, rev)
+    except requests.exceptions.RequestException:
+        LOGGER.warning(f"Push not found for {branch} @ {rev}!")
         return "NOK"
 
-    first_rev = r.json()["changesets"][0]["node"]
+    # Apply the stack on the local repository
+    apply_stack(REPO_DIR, stack, branch)
```
marco-c: We should return "NOK" when the stack fails to apply

http_service/bugbug_http/utils.py
```diff
@@ -22,3 +22,11 @@ def get_bugzilla_http_client():
     http_client.mount(bugzilla_url, HTTPAdapter(max_retries=retries))
 
     return http_client, bugzilla_api_url
+
+
+def hgmo_stack(branch: str, revision: str) -> list:
```
marco-c: Nit:
```suggestion
def hgmo_stack(branch: str, revision: str) -> list:
```

bugbug/repository.py
La0: Ok, so that's even simpler than what i thought (like in libmozevent)

http_service/tests/conftest.py
```diff
@@ -131,3 +156,78 @@ def inner(key, change_time):
...
+
+@pytest.fixture
+def mock_repo(tmpdir, monkeypatch):
+    """Create an empty mercurial repo"""
+    repo_dir = tmpdir / "repo"
+
+    # Setup the worker env to use that repo dir
+    monkeypatch.setattr(bugbug_http, "REPO_DIR", str(repo_dir))
+
+    # Silent the clean method
```
marco-c:
```suggestion
    # Silence the clean method
```

http_service/tests/test_schedule_tests.py
```diff
@@ -0,0 +1,89 @@
...
+                "Orphan 456",
+                "Base history 3",
+                "Base history 2",
+                "Base history 1",
+                "Base history 0",
+            ],
+        ),
+    ],
+)
+def test_schedule(
```
marco-c: It'd be nice to also have a test where the parent is not available in the local repo but is available on the remote repo and so we can update to it after pulling

marco-c: (where the patch wouldn't apply without the parent, but would apply with the parent)

marco-c: Can be done in a follow-up (as part of #1371).

PR #1372, opened 2020-03-05
Clone mozilla autoland in http worker

La0:


bugbug/repository.py
```diff
@@ -1000,7 +1000,7 @@ def clone(repo_dir):
     if proc.returncode:
         raise hglib.error.CommandError(cmd, proc.returncode, out, err)
 
-    logger.info("mozilla-central cloned")
+    logger.info(f"{url} cloned")
```
marco-c: Nit for consistency:
```suggestion
    logger.info(f"{repo_dir} cloned")
```

Issue #1342, opened 2020-02-29
Make the models cache an actual cache


marco-c:
Right now, it's a simple dict. We preload all models and keep them in memory forever.
We should instead only preload the ones we use often, and unload models from the cache when they haven't been used for a while.

marco-c:
Also, not cache the seldomly requested ones.

marco-c:
So we could do the following:
- cache a model only if it was loaded at least X times in two hours (maybe starting with a low value, like X = 2);
- set a TTL of two hours for cache entries;
- preload only the component model, which is used hourly by relman-auto-nag (and in the future the test selection models).

mvkski:
I'm interested in working on this, and curious as to the motivation for this enhancement. Are there any specific operational issues the service is currently experiencing, or that you foresee it experiencing, if this enhancement is not implemented and the service keeps loading all models into a dict?

marco-c:
> I'm interested in working on this

@mvkski better to start with an issue with the good-first-bug label: https://github.com/mozilla/bugbug/issues?q=is%3Aissue+is%3Aopen+label%3Agood-first-bug.
You can still work on this if you want of course, but it might not be as easy as those.
NOTE: The only thing missing here is removing items from the cache after two hours.

> , and curious as to the motivation for this enhancement. Are there any specific operational issues the service is currently experiencing, or that you foresee it experiencing, if this enhancement is not implemented and the service keeps loading all models into a dict?

The reason behind this change is reducing memory usage of the HTTP service. It is currently high, and will be too high if we add more models.

mvkski:
@marco-c Thanks for the suggestion! This issue looked like it might be a good fit for me because the enhancement seems straightforward (if I'm not missing anything) and I am interested in caching technologies, but I may pick up a good-first-bug issue seeing as you recommend doing so. Plan of action for this issue would be something like:

Replace the object at this line of code https://github.com/mozilla/bugbug/blob/22b27b97b0a7ec4d84c8b374cb56c2c5f3f9d734/http_service/bugbug_http/models.py#L41

with a cache that allows the kind of policy you suggest, such as a TTL cache from cachetools (or similar python caching library if cachetools for some reason doesn't work out):
https://github.com/tkem/cachetools/
If using cachetools, we would configured with the 'TTL' parameter specified above set to two hours, and we would make sure that when we get and item, we re-add it to the cache with the same TTL, since by default the TTL does not appear to get updated to be current time + TTL on a get.

[...]

PR #1375, opened 2020-03-06
Ttlcache

mvkski:
PR for https://github.com/mozilla/bugbug/issues/1342

I added unit tests and integration tests for the proposed cache, but haven't yet been able to get the http service running locally on my machine so have yet to verify end to end. I'll be asking questions on how to do so in matrix, since that seems like the most appropriate channel for the topic.


mvkski:
The PR appears to be failing due to linting. Is there any way to run the linting rules locally? I'll be fixing these later, as it is very late here in the US

http_service/bugbug_http/utils.py
```diff
@@ -30,3 +33,38 @@ def get_hgmo_stack(branch: str, revision: str) -> list:
...
+    self.items_last_touched[key] = datetime.datetime.now()
+    return item
+
+  def __contains__(self, key):
+    return key in self.items
+
+  def purge_expired_entries(self):
+    purge_entries_before = datetime.datetime.now() - self.ttl
+    for (key, time_last_touched) in list(self.items_last_touched.items()):
+      print(time_last_touched)
```
mvkski: Remove this! This was left in for testing, and I missed it before submitting PR

marco-c:
> The PR appears to be failing due to linting. Is there any way to run the linting rules locally? I'll be fixing these later, as it is very late here in the US

You can use pre-commit, that's the easiest way.
Just install the hooks with `pre-commit install` and then whenever you commit the linting will run automatically.

To fix the current issues, since you have already committed, you can use `pre-commit run --all-files`.

mvkski:
I have reposted this pr [here](https://github.com/mozilla/bugbug/pull/1386) using my other github account max-reborn, which I would prefer to use for non professional projects at this time

PR #1386, opened 2020-03-07
Make models cache an 'idle' TTL cache, use timedelta instead of relat…

max-reborn:
PR for #1342

repost of [this pr](https://github.com/mozilla/bugbug/pull/1375), but with another account as author. Note that I (as mvkski) made a comment in the other thread naming this account as his preferred account for mozilla contributions


max-reborn:
Its getting a lint error I somehow didn't notice. The lint error points out to me that I didn't add generic typing to the cache, and will now do that.

mvkski:
There appears to be an issue with the service tests. I can't seem to be able to access the logs for the test failures.

max-reborn:
There appears to be an issue with the service tests. I can't seem to be able to access the logs for the test failures

max-reborn:
If I'm not mistaken the failing task is just trying to calculate test coverage? Maybe this is an intermittent failure and would be fixed on a rerun?

max-reborn:
Finally ran `pre-commit install` so I can stop getting these linting errors

max-reborn:
I watched the above test run while it was happening - I saw `bugbug http service tests` finish succesfully, but now looking at the taskgroup the task is failed with `Exception: Deadline exceeded`. Not sure what the problem might be

max-reborn:
Im having this problem locally too now that I look at it - pytest runs, but then does not terminate after all tests pass. This is good news since it means I can investigate

max-reborn:
And I think I have the answer - if I comment out the line of code in models.py that I added which starts the TTL cache cleanup thread, pytest terminates after unit tests are completed

max-reborn:
Closing this PR in favor of [this one](https://github.com/mozilla/bugbug/pull/1389), where the git log has been squashed, to avoid polluting the repo with small commits to fix linting/other build checks.

marco-c:
> Closing this PR in favor of this one, where the git log has been squashed, to avoid polluting the repo with small commits to fix linting/other build checks.

A suggestion for the future: you can just force push to the same branch after you squash. Or, you can keep the multiple commits when they make sense, and I'll squash when merging.
I think it's preferable because this way a PR's history is not split into multiple PRs.

PR #1388, opened 2020-03-07
Commit classifier now applies phabricator patch just after repo clone

### LINKED ISSUE #1268: Make the commit classifier script apply the Phabricator patch as its first step
Most failures happen when applying patches, so that should be the first step (right after cloning). This way, we don't waste time doing other things if the patch fails to apply.

kaushiksk:
Fixes #1268


kaushiksk:
@marco-c The PR is failing due to linting errors, specifically "check-executables-have-shebangs". I ran pre-commit locally, it passes all except this check.

kaushiksk:
@marco-c I must have misunderstood the description then. Thanks for clarifying. Are you saying that the phabricator call should go inside the `update_commit_db` function right after `repository.clone()` ?

marco-c:
> @marco-c I must have misunderstood the description then. Thanks for clarifying.

Well, your patch doesn't work at all as-is, as you're trying to apply a patch on top of a repository which might not exist yet :D

> Are you saying that the phabricator call should go inside the update_commit_db function right after repository.clone() ?

Applying the patch should be done as early as possible, after cloning the repositories. You could move the `update_commit_db` call in `__init__` and then apply the patch right after that.

kaushiksk:
@marco-c I've moved `repository.clone` to `__init__` and not all of `update_commit_db`. I presume the downloading of the commit db is independent of cloning and applying a patch. Does this achieve what you intended?

scripts/commit_classifier.py
```diff
@@ -136,6 +144,12 @@ def __init__(self, model_name, repo_dir, git_repo_dir, method_defect_predictor_d
         if git_repo_dir:
             self.clone_git_repo("https://github.com/mozilla/gecko-dev", git_repo_dir)
 
+        repository.clone(self.repo_dir)
+
+        with hglib.open(self.repo_dir) as hg:
+            if phabricator_deployment is not None and diff_id is not None:
+                self.apply_phab(hg, phabricator_deployment, diff_id)
+
```
marco-c: We could move the `download_and_load_model` after this, so we don't download the models needlessly if we can't apply the patch.

scripts/commit_classifier.py
```diff
@@ -551,26 +565,12 @@ def feature_sort_key(f):
...
-        runnable_jobs_path=None,
+        self, revision=None, runnable_jobs_path=None,
     ):
-        if revision is not None:
-            assert phabricator_deployment is None
-            assert diff_id is None
-
-        if diff_id is not None:
-            assert phabricator_deployment is not None
-            assert revision is None
```
marco-c: Please move these checks in the `main` function rather than removing them

scripts/commit_classifier.py
kaushiksk: Yeah, makes sense.

scripts/commit_classifier.py
kaushiksk: Will do.

PR #1389, opened 2020-03-08
Make models cache an 'idle' TTL cache, use timedelta instead of relat…

max-reborn:
PR for #1342

This has the same content as [this PR](https://github.com/mozilla/bugbug/pull/1386) (which is itself a repost of [this pr](https://github.com/mozilla/bugbug/pull/1375)), but has a cleaner git log. All of the small commits made to satisfy checks were squashed.


http_service/bugbug_http/models.py
```diff
@@ -61,7 +61,7 @@ def get_model(model_name):
         # Cache the model only if it was last used less than two hours ago.
         if model_name in MODEL_LAST_LOADED and MODEL_LAST_LOADED[
```
kaushiksk: Is `MODEL_LAST_LOADED` required anymore now that TTLCache has a `items_last_touched`? 
You could just write an API in TTLCache 
```python
def last_touched_time(self, item):
    return self.items_last_touched[item]
```

and replace the above line with:
```python
if model_name in MODEL_CACHE and MODEL_CACHE.last_touched_time(model_name)
```

http_service/bugbug_http/utils.py
```diff
@@ -30,3 +35,44 @@ def get_hgmo_stack(branch: str, revision: str) -> list:
...
+        for (key, time_last_touched) in list(self.items_last_touched.items()):
+            if time_last_touched < purge_entries_before:
+                del self.items_last_touched[key]
+                del self.items[key]
+
+    def start_ttl_thread(self):
+        self.purge_expired_entries()
+        timer = threading.Timer(self.ttl.total_seconds(), self.start_ttl_thread)
+        timer.setDaemon(True)
+        timer.start()
```
marco-c: It'd probably be preferable to have a single thread with an infinite loop which waits for the TTL and then purges, rather than create a new thread for every period.

http_service/bugbug_http/models.py
```diff
@@ -61,7 +61,7 @@ def get_model(model_name):
         # Cache the model only if it was last used less than two hours ago.
```
marco-c: This should be handled directly by your class now, or we split the logic in two places.

http_service/bugbug_http/utils.py
max-reborn: Will do.

I think you're right - a single thread with a sleep seems like it'd run more efficiently as well as result in marginally more readable code.

I implemented this with a Timer assuming (without checking the assumption) that Timers give up the GIL while waiting to run, whereas threads do not.  Digging deeper after reading your comment, I do think Timers work this way. However, if I correctly understood this [article on the GIL by a core python developer](http://jessenoller.com/blog/2009/02/01/python-threads-and-the-global-interpreter-lock), so do Threads when they are executing a sleep.

Using a Thread with a sleep and an infinite loop therefore seems strictly superior to a Timer, since it looks like neither take up the GIL while waiting, but a Thread with a loop has less overhead than a Timer.

http_service/bugbug_http/models.py
max-reborn: Sounds good, will do.

I did it this way because I was able to abstract the cache as a general purpose 'IdleTTLCache' (as described in its comment) if I separated this logic from the cache. Im not sure if I can come up with as clean of an abstraction if I put all of this functionality in one place.

However, weather or not this idea of mine is 'better' or even 'good' seems very subjective. I'm happy to put the logic into one place and possibly compromise the quality of the abstraction (in my eyes) if you prefer.

marco-c: And I think it's still generic: a cache which expires entries after TTL and which only caches them if they're loaded twice within a TTL timespan. It's just a different kind of cache than just an expiring one :)

http_service/bugbug_http/models.py
```diff
@@ -32,41 +30,34 @@
...
 ]
+
 DEFAULT_EXPIRATION_TTL = 7 * 24 * 3600  # A week
+redis = Redis.from_url(os.environ.get("REDIS_URL", "redis://localhost/0"))
 
 
-MODEL_LAST_LOADED: Dict[str, datetime] = {}
-MODEL_CACHE: Dict[str, Model] = {}
+def load_model_for_service(model_name):
+    LOGGER.info("Recreating the %r model in cache" % model_name)
```
marco-c: Nit: use f-string `f"Recreating the {model_name} model in cache"`

http_service/bugbug_http/models.py
```diff
@@ -32,41 +30,34 @@
...
     "testgroupselect",
 ]
+
 DEFAULT_EXPIRATION_TTL = 7 * 24 * 3600  # A week
+redis = Redis.from_url(os.environ.get("REDIS_URL", "redis://localhost/0"))
 
 
-MODEL_LAST_LOADED: Dict[str, datetime] = {}
-MODEL_CACHE: Dict[str, Model] = {}
+def load_model_for_service(model_name):
```
marco-c: Actually, we can just get rid of this function completely, and add the try-except around https://github.com/mozilla/bugbug/blob/e00e951659e32d8a509801494f5b878529b2b52b/http_service/bugbug_http/check_models.py#L19

http_service/bugbug_http/models.py
```diff
@@ -32,41 +30,34 @@
...
-        if model_name in MODEL_LAST_LOADED and MODEL_LAST_LOADED[
-            model_name
-        ] > datetime.now() - relativedelta(hours=1):
-            MODEL_CACHE[model_name] = model
-    else:
-        model = MODEL_CACHE[model_name]
-
-    MODEL_LAST_LOADED[model_name] = datetime.now()
-    return model
+    return MODEL_CACHE.get(model_name)
```
marco-c: Since this function is a one-liner, we can just remove it and use `MODEL_CACHE.get(model_name)` directly

http_service/bugbug_http/utils.py
```diff
@@ -5,6 +5,12 @@
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
 
+import datetime
+import threading
+import time
+from datetime import timedelta
```
marco-c: Nit: `from datetime import datetime, timedelta` so you can use `datetime.now()` instead of `datetime.datetime.now()` in the rest of the file

http_service/bugbug_http/models.py
```diff
@@ -32,41 +30,34 @@
...
+                % model_name
+            )
+            return None
+        else:
+            raise
 
 
-redis = Redis.from_url(os.environ.get("REDIS_URL", "redis://localhost/0"))
+MODEL_CACHE: ReadthroughTTLCache[str, Model] = ReadthroughTTLCache(
+    timedelta(hours=2), load_model_for_service
```
marco-c: Let's use 1 hour, I've changed that recently

http_service/bugbug_http/utils.py
```diff
@@ -30,3 +36,91 @@ def get_hgmo_stack(branch: str, revision: str) -> list:
...
+        item = self.items[key]
+        self.items_last_touched[key] = datetime.datetime.now()
+        return item
+
+    def __contains__(self, key):
+        return key in self.items
+
+    def purge_expired_entries(self):
+        purge_entries_before = datetime.datetime.now() - self.ttl
+        for (key, time_last_touched) in list(self.items_last_touched.items()):
```
marco-c: Nit: no need for the parenthesis around key and time_last_touched

http_service/bugbug_http/utils.py
```diff
@@ -30,3 +36,91 @@ def get_hgmo_stack(branch: str, revision: str) -> list:
...
+# A simple TTL cache to use with models. Because we expect the number of models
+# in the service to not be very large, simplicity of implementation is
+# preferred to algorithmic efficiency of operations.
+#
+# Called an 'Idle' TTL cache because TTL of items is reset after every get
+Key = TypeVar("Key")
+Value = TypeVar("Value")
+
+
+class IdleTTLCache(Generic[Key, Value]):
```
marco-c: We don't use this anywhere, so better to just keep ReadthroughTTLCache and remove this

http_service/bugbug_http/utils.py
```diff
@@ -30,3 +36,91 @@ def get_hgmo_stack(branch: str, revision: str) -> list:
...
+            # Cache the item only if it was last accessed within the past TTL seconds
+            # Note that all entries in items_last_accessed are purged if item was not
+            # accessed in the last TTL seconds.
+            if key in self.items_last_accessed:
+                self.items_storage[key] = item
+
+        self.items_last_accessed[key] = datetime.datetime.now()
+        return item
+
+    def force_store(self, key):
```
marco-c: I'd say we can remove this function, in the tests you can just `get` twice to force a store

http_service/tests/test_utils_integration.py
```diff
@@ -0,0 +1,33 @@
...
+import time
+from datetime import timedelta
+
+from bugbug_http.utils import ReadthroughTTLCache
+
+# to load bugbug_http.utils, when running from project root
+sys.path.append(".")
+
+
+def integration_test_cache_thread():
```
marco-c: This doesn't have to be an integration test, it could be a normal test (so, you can move it to test_utils.py).

http_service/bugbug_http/models.py
```diff
@@ -32,41 +30,34 @@
...
+)
+MODEL_CACHE.start_ttl_thread()
```
marco-c: Nit: maybe we could just start the thread from the ReadthroughTTLCache constructor. I'll let you decide, I don't have a strong preference

http_service/tests/test_utils_integration.py
max-reborn: My thinking at the time was that its better to have this be separate script rather unit test. How does the below justification sound to you? I'm happy to re-evaluate if you'd like me to

The advantage of having this be a seperate script rather than a unit test is that this takes real time to run - as written, it'd make every single unit test run take an extra seven seconds, which means noticeably longer test runs for every test run for this project! But, as a separate script, we can exclude it from test runs and not have that problem.

Since the cache expects a value on the order of seconds for its TTL, I'm not sure if there is a way to have this test run in a negligible amount of time, which is what I expect would be ideal for a unit test.

http_service/bugbug_http/utils.py
max-reborn: Ack! Forgot to delete this. Will do!

http_service/bugbug_http/utils.py
max-reborn: Below is my reasoning for having force_store - its subjective, and I'm happy to switch to using two gets if you prefer, but at this time I'd like to try to convince you that the force_store function belongs here, because in my eyes its better to have it than use two gets for the following reasons:

1. This is more future proof. If in the future we want to change the internals of the cache to store models after, say. three accesses in the past hour, we won't need to update as many unit tests.

2. I think its more readable this way. I originally had two gets in my unit tests, but upon seeing those two gets sitting next to each other I felt like someone who wasn't familiar with the code might be confused by what was happening and be compelled to read the cache internals to understand the test, so I wrote a comment above each test. 

Thus, each place where we now have a force_store was replaced by a lot of characters - two calls to get, and a comment. The codebase ended up being as clear, and less verbose in my eyes when I had the force_store function.

http_service/bugbug_http/models.py
max-reborn: I think its better to keep it separate for the sake of our unit tests. 

We need nonzero values for the TTL to unit test this object, but if we try to unit test using a real thread rather that the datetime mocking I was doing, the tests will noticeably delay unit test runs, and also will open the door for non deterministic behavior since I dont think theres actually a guarantee that the sleep function sleeps for the amount of time it says its going to sleep for.

http_service/tests/test_utils_integration.py
max-reborn: Wait nevermind - I can write this as a unit test by mocking out sleep. Thanks!

max-reborn: On second thought, it might be possible to write this as a unit test with proper mocking - it'd be a bit tricky though

http_service/bugbug_http/models.py
max-reborn: on second thought, it might be possible to unit test despite this, but it'd be a bit tricky if it is

http_service/tests/test_utils_integration.py
marco-c: You could mock `time.sleep`, dividing the sleep time by 10.

http_service/bugbug_http/models.py
marco-c: Feel free to keep it as-is, I don't mind :)

http_service/bugbug_http/utils.py
marco-c: OK, you convinced me with the first reason ;)
Maybe we could add a `force_store` parameter to `get` though, so we don't need to have a separate function.

http_service/bugbug_http/utils.py
max-reborn: Unfortunatly, if I do as you advise I can no longer patch out the datetime object. See also https://stackoverflow.com/questions/31306080/pytest-monkeypatch-isnt-working-on-imported-function/31746577

marco-c: OK, np!

http_service/bugbug_http/models.py
```diff
@@ -137,13 +118,15 @@ def schedule_tests(branch, rev):
     try:
         stack = get_hgmo_stack(branch, rev)
     except requests.exceptions.RequestException:
+        traceback.print_exc(file=sys.stdout)
```
max-reborn: Ack! Forgot to delete this

http_service/bugbug_http/utils.py
```diff
@@ -99,19 +70,27 @@ def get(self, key):
...
         self.items_last_accessed[key] = datetime.datetime.now()
         return item
 
     def force_store(self, key):
+        LOGGER.info(f"Storing item with the following key in readthroughcache: {key}")
         self.items_storage[key] = self.load_item_function(key)
         self.items_last_accessed[key] = datetime.datetime.now()
 
     def purge_expired_entries(self):
+        print("thread did a thing")
```
max-reborn: ack! forgot to delete this too

pytest.ini
```diff
@@ -1,2 +0,0 @@
-[pytest]
```
max-reborn: Ack! Accidentally  forgot to un-delete this

http_service/bugbug_http/utils.py
max-reborn: Will update get to take a parameter

max-reborn:
Well, the unit tests pass locally - I can only assume some nondeterministic behavior is at play

http_service/bugbug_http/models.py
marco-c: Don't forget to update the `get_model` calls in download_models.py too (it used to be in check_models.py, it is now in download_models.py).

marco-c: I hadn't noticed you did modify check_models.py! So, you just need to apply that change to download_models.py instead ;)

http_service/bugbug_http/utils.py
max-reborn: and done!

http_service/bugbug_http/download_models.py
```diff
@@ -4,14 +4,14 @@
...
-from bugbug_http.models import MODELS_NAMES, get_model
+from bugbug_http.models import MODEL_CACHE, MODELS_NAMES
 
 
 def download_models():
     for model_name in MODELS_NAMES:
         utils.download_model(model_name)
         # Try loading the model
-        get_model(model_name)
+        MODEL_CACHE.get(model_name)
```
marco-c: Here we have to allow missing models like you were doing before in check_models.py

http_service/bugbug_http/readthrough_cache.py
```diff
@@ -0,0 +1,76 @@
...
+        self.ttl = ttl
+        self.load_item_function = load_item_function
+        self.items_last_accessed: Dict[Key, datetime.datetime] = {}
+        self.items_storage: Dict[Key, Value] = {}
+        self.last_purged_time = datetime.datetime.now()
+
+    def __contains__(self, key):
+        return key in self.items_storage
+
+    def get(self, key, **kwargs):
```
marco-c: I'd prefer `force_store` to be an explicit parameter rather than using kwargs

http_service/bugbug_http/readthrough_cache.py
```diff
@@ -0,0 +1,76 @@
...
+        store_item = kwargs.get("force_store", False)
+        item = None
+        if item in self.items_storage:
```
marco-c: Should this be `key in ...` :P

http_service/bugbug_http/readthrough_cache.py
```diff
@@ -0,0 +1,76 @@
...
+        self.items_last_accessed: Dict[Key, datetime.datetime] = {}
+        self.items_storage: Dict[Key, Value] = {}
+        self.last_purged_time = datetime.datetime.now()
+
+    def __contains__(self, key):
+        return key in self.items_storage
+
+    def get(self, key, **kwargs):
+        store_item = kwargs.get("force_store", False)
+        item = None
```
marco-c: Nit: No need to define it as None, as long as a value is given to item in both branches next

http_service/tests/test_readthrough_cache.py
```diff
@@ -0,0 +1,125 @@
...
+        self.mock_now = new_mock_now
+
+
+class MockSleep:
+    def __init__(self, mock_datetime):
+        self.mock_datetime = mock_datetime
+        self.wakeups_count = 0
+
+        # count variable used to ensure that context switch into
+        # thread claling sleep occurred
```
marco-c:
```suggestion
        # thread calling sleep occurred
```

http_service/tests/test_readthrough_cache.py
```diff
@@ -0,0 +1,125 @@
...
+
+            # after one hour
+            mockdatetime.set_now(datetime(2019, 4, 1, 11))
+            assert "key_a" in cache
+            assert mocksleep.wakeups_count == 0
+
+            # after two hours one minute
+            before_purge_timestamp = cache.last_purged_time
+            mockdatetime.set_now(datetime(2019, 4, 1, 12, 1))
+            while cache.last_purged_time == before_purge_timestamp:
```
marco-c: You could instead mock purge_expired_entries to wait until it is called, this way you can also remove the `last_purged_time` attribute.

http_service/tests/test_readthrough_cache.py
```diff
@@ -0,0 +1,125 @@
...
+        self.mock_datetime = mock_datetime
+        self.wakeups_count = 0
+
+        # count variable used to ensure that context switch into
+        # thread claling sleep occurred
+
+    def sleep(self, sleep_seconds):
+        self.wakeup_time = self.mock_datetime.now() + timedelta(seconds=sleep_seconds)
+        while self.mock_datetime.now() < self.wakeup_time:
+            pass
```
marco-c: If you use `time.sleep(0)`, you will yield control to the main thread, this might help avoiding hanging during the test

http_service/tests/test_readthrough_cache.py
marco-c: (and while waiting, you could use `time.sleep(0)` instead of `pass`, to yield control to the cleanup thread if needed and avoid possible hangs)

PR #1394, opened 2020-03-11
Only pull when needed to apply patches

### LINKED ISSUE #1371: Try to apply patches without pulling first
In most cases, we'd be able to apply patches without pulling first, so we should:
- If the parent is available, checkout to the parent
- If the parent is not available, checkout to "tip"
- Try to apply the patch
- If the patch fails to apply and we had updated to "tip", pull from the remote repo and restart from the first step (if the parent still doesn't exist, we can directly raise an exception).
- If the patch fails to apply and we had updated to the corrent parent, raise an exception.

La0:
Fixes #1371

### DIFF
--- a/bugbug/repository.py
+++ b/bugbug/repository.py
@@ -1011,9 +1011,10 @@ def clone(repo_dir, url="https://hg.mozilla.org/mozilla-central"):
     clean(repo_dir)
 
 
-def apply_stack(repo_dir, stack, branch, default_base):
+def apply_stack(repo_dir, stack, branch, default_base="tip"):
     """Apply a stack of patches on a repository"""
     assert len(stack) > 0, "Empty stack"
+    assert isinstance(default_base, str)
 
     def has_revision(revision):
         try:
@@ -1046,10 +1047,6 @@ def apply_stack(repo_dir, stack, branch, default_base):
         if all(map(has_revision, parents)):
             return parents[0]
 
-        # Some repositories need to have the exact parent to apply
-        if default_base is None:
-            raise Exception("Parents are not available, cannot apply this stack")
-
         return default_base
 
     # Start by cleaning the repo, without pulling
@@ -1073,7 +1070,7 @@ def apply_stack(repo_dir, stack, branch, default_base):
             raise Exception(f"Failed to apply on valid parent {base}")
 
         # We tried to apply on tip, let's try to find the valid parent after pulling
-        clean(repo_dir)
+        clean(repo_dir, pull=True)
 
         # Check if the valid base is now available
         new_base = get_base()
--- a/bugbug/repository.py
+++ b/bugbug/repository.py
@@ -1014,9 +1014,6 @@ def apply_stack(repo_dir, stack, branch, default_base):
     """Apply a stack of patches on a repository"""
     assert len(stack) > 0, "Empty stack"
 
-    # Start by updating the repository
-    clean(repo_dir)
-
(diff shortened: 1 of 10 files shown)


http_service/tests/conftest.py
```diff
@@ -230,4 +232,20 @@ def mock_repo(tmpdir, monkeypatch):
...
+        # Revert to initial state
+        repo.update(initial, clean=True)
+
+        # Then create a new "remote" file to emulate a pull
+        remote = repo_dir / "remote.txt"
+        remote.write_text("New remote file !", encoding="utf-8")
+        repo.add([str(remote).encode("utf-8")])
+        repo.commit("Pulled from remote", user="bugbug")
+
+    monkeypatch.setattr(bugbug.repository, "clean", _clean)
```
marco-c: What if we didn't monkeypatch it? We could have a local and remote like the tests in tests/test_repository.py.

bugbug/repository.py
```diff
@@ -1014,42 +1014,72 @@ def apply_stack(repo_dir, stack, branch, default_base):
...
+        # Some repositories need to have the exact parent to apply
+        if default_base is None:
+            raise Exception("Parents are not available, cannot apply this stack")
 
-        # Update to base revision
-        logger.info(f"Will apply stack on {base}")
-        hg.update(base, clean=True)
+        return default_base
+
+    with hglib.open(repo_dir) as hg:
```
marco-c: As a first step, we have to run part of `clean` (reverting and stripping), otherwise when we apply a stack after another one, we might fail.

bugbug/repository.py
```diff
@@ -1014,42 +1014,72 @@ def apply_stack(repo_dir, stack, branch, default_base):
...
-            base = parents[0]
-        else:
-            # Some repositories need to have the exact parent to apply
-            if default_base is None:
-                raise Exception("Parents are not available, cannot apply this stack")
+            return parents[0]
 
-            base = default_base
+        # Some repositories need to have the exact parent to apply
+        if default_base is None:
```
marco-c: We should now support "tip" for both autoland and try

bugbug/repository.py
```diff
@@ -1010,12 +1011,10 @@ def clone(repo_dir, url="https://hg.mozilla.org/mozilla-central"):
     clean(repo_dir)
 
 
-def apply_stack(repo_dir, stack, branch, default_base):
+def apply_stack(repo_dir, stack, branch, default_base="tip"):
```
marco-c: Since it's always "tip" now, we can just remove the parameter to simplify.

http_service/tests/test_schedule_tests.py
```diff
@@ -74,6 +79,20 @@ def test_simple_schedule(patch_resources, mock_hgmo, mock_repo):
             "NOK",
             ["Base history 2", "Base history 1", "Base history 0"],
         ),
+        # patch that only applies after a pull has been done
```
marco-c: Can you add another case with `autoland` instead of `try`

http_service/tests/conftest.py
La0: Done

bugbug/repository.py
La0: Done

http_service/tests/test_schedule_tests.py
La0: Done

Issue #632, opened 2019-06-26
Support classifying commits for test selection


marco-c:
- [x] Add a setup step to the http service to setup the environment for classifying commits - #1299
- [x] Add the scaffolding needed to add an API call to select tests - #1300
- [x] In the API to select tests, apply the requested patch - #1301 and #1372
- [x] Download test scheduling models in the HTTP service - #1330
- [x] In the API to select tests, apply the model to select tests - #1302 

------------------------------------------------------------

This is not going to be so easy.
1. We need to clone a mercurial repository on the service
2. We need to download the commits DB from the commit_retrieval task
3. We need to update the commits DB whenever there are new commits on mozilla-central
4. The client will send a patch, we will apply the commit, analyze it to get its features, revert it

Similar to https://github.com/mozilla/relman-auto-nag/blob/b7bda7f4c29e59d36684b39fec522a79bacefcbb/auto_nag/scripts/regressor.py#L1, but this one is easier as it's only analyzing new real commits, we will instead analyze proposed patches too (which might or might not be landed at some point).

marco-c:
For 1, we likely need batch checkout as pulselistener, as otherwise we can't clone mozilla-central on Heroku.

marco-c:
I'm starting to write this as a script, which we can then move to an API endpoint.

Lothiraldan:
Sounds good, the hardest part seems to be applying the patch locally as it involves the disk. Does `https://github.com/mozilla/bugbug/blob/master/bugbug/repository.py#L270` is retrieving all the information needed for the commits or other functions are retrieving other information?

marco-c:
There are other functions too, all called from https://github.com/mozilla/bugbug/blob/89bba8efcaa418a8d2a0bfcfea4713e758e521ac/bugbug/repository.py#L662 (log, calculate_experiences, _transform). I should improve the names of these functions.

Lothiraldan:
I take a look at what we are actually reading from the mercurial repository and I think that the data that requires to actually apply the patch is https://github.com/mozilla/bugbug/blob/master/bugbug/repository.py#L239.

I think we should be able to do all the work in-memory, maybe with a custom Mercurial extension so we don't need to fiddle with applying a patch then stripping it right after.

marco-c:
I've added a script to the repo: https://github.com/mozilla/bugbug/commit/96ebe8777a333d08105fb39d83cee2cb052bae2c.
[...]

PR #1395, opened 2020-03-11
Use 'tip' branch to support autoland but keep skipping relbranch commits

marco-c:
For example, 961f253985a4388008700a6a6fde80f4e17c0b4b wouldn't be skipped if
we simply removed the branch parameter.

This way, we can also stop using the firefoxtree extension.


marco-c:
This was also required for #632.

PR #1396, opened 2020-03-11
In the API to select tests, apply the model to select tests

### LINKED ISSUE #1302: In the API to select tests, apply the model to select tests
Depends on #1301.

We should:
1) Analyze the patches relevant for the try push (the non public ones? or the ones coming from json-automationrelevance?), as https://github.com/mozilla/bugbug/blob/cb51deac6235477e7f2dbc13352a2627d129451c/scripts/commit_classifier.py#L575-L580;
2) Apply the model to select tests, as https://github.com/mozilla/bugbug/blob/cb51deac6235477e7f2dbc13352a2627d129451c/scripts/commit_classifier.py#L614-L694.

Blocks #632.

marco-c:
Fixes #1302.

### DIFF
--- a/http_service/bugbug_http/models.py
+++ b/http_service/bugbug_http/models.py
@@ -9,14 +9,14 @@ import os
 from datetime import datetime
 from typing import Dict
 
+import numpy as np
 import requests
 from dateutil.relativedelta import relativedelta
 from redis import Redis
 
-from bugbug import bugzilla
+from bugbug import bugzilla, commit_features, repository, test_scheduling
 from bugbug.model import Model
 from bugbug.models import load_model
-from bugbug.repository import apply_stack
 from bugbug_http import ALLOW_MISSING_MODELS
 from bugbug_http.utils import get_hgmo_stack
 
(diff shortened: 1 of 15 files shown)


http_service/bugbug_http/models.py
```diff
@@ -159,29 +159,47 @@ def schedule_tests(branch, rev):
...
     try:
-        apply_stack(REPO_DIR, stack, branch, default_base)
+        first_rev = repository.apply_stack(REPO_DIR, stack, branch, default_base)
     except Exception as e:
         LOGGER.warning(f"Failed to apply stack {branch} @ {rev}: {e}")
         return "NOK"
 
-    first_rev = stack[0]["node"]
-    if first_rev != rev:
-        revset = f"{first_rev}::{rev}"
```
ahal: I still think it would be nice to support revsets here so we can more easily use the commit classifier in other contexts, but it's definitely something we can implement later on when/if we need it.

http_service/tests/conftest.py
```diff
@@ -230,4 +234,69 @@ def mock_repo(tmpdir, monkeypatch):
...
+    responses.add(
+        responses.GET,
+        "https://firefox-ci-tc.services.mozilla.com/api/index/v1/task/gecko.v2.mozilla-central.latest.source.source-bugzilla-info/artifacts/public/components.json",
+        status=200,
+        json={},
+    )
+
+
+@pytest.fixture
+def mock_schedule_tests_classify(tmpdir, monkeypatch):
```
ahal: This fixture looks like it would be better off in `test_schedule_tests.py`

.taskcluster.yml
```diff
@@ -124,7 +124,6 @@ tasks:
             - "/bin/bash"
             - "-lcx"
             - "apt-get -qq update &&
```
ahal: Looks like we can remove the update too.

marco-c: Right! Done in fe09616

http_service/tests/conftest.py
marco-c: Yeah, there are a couple more that might make sense to move in test_schedule_tests.py, we can do it in a follow-up as I don't want to cause too much conflicts with @La0's PR.

.taskcluster.yml
marco-c: Actually, in c5d92ae.

PR #1440, opened 2020-03-30
Use scientific literature names for Halstead

Luni-4:
Starting from https://github.com/mozilla/rust-code-analysis/pull/115, `rust-code-analysis` is using scientific literature names for the `Halstead` metric, so it would be better to change those names even here. 

Thanks in advance for your review! :)


marco-c:
Thanks!

PR #1468, opened 2020-04-07
Use a single API call to rust-code-analysis to get both functions and metrics

### LINKED ISSUE #1453: Use a single rust-code-analysis API call to get both functions and metrics
The function information is already available in the metrics API call result, so there's no need to do two API calls.

marco-c:
Fixes #1453

### DIFF
--- a/bugbug/repository.py
+++ b/bugbug/repository.py
@@ -365,7 +365,10 @@ def get_touched_functions(metrics_space, deleted_lines, added_lines):
 
 
 def get_metrics(commit, metrics_space):
-    if metrics_space["kind"] == "function" and metrics_space["name"]:
+    name = metrics_space["name"]
+    error = metrics_space["kind"] in {"unit", "function"} and name == ""
+
+    if metrics_space["kind"] == "function" and not error:
         metrics = metrics_space["metrics"]
         commit.total_cyclomatic += metrics["cyclomatic"]
         commit.total_halstead_n2 += metrics["halstead"]["n2"]
@@ -420,7 +423,9 @@ def get_metrics(commit, metrics_space):
         )
 
     for space in metrics_space["spaces"]:
-        get_metrics(commit, space)
+        error |= get_metrics(commit, space)
+
+    return error
 
 
 def transform(hg, repo_dir, commit):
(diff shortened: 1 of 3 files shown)


bugbug/repository.py
```diff
@@ -288,23 +288,29 @@ def hg_modified_files(hg, commit):
...
+    for space in metrics_space["spaces"]:
+        functions += get_functions_from_metrics(space)
 
-    function_data = code_analysis_server.function(path, content)
-    if not function_data:
-        return set()
+    return functions
 
+
+def get_touched_functions(metrics_space, deleted_lines, added_lines):
```
marco-c, line 309: NOTE: I'm assuming the spaces are sorted, is that a safe assumption @calixteman?

bugbug/repository.py
```diff
@@ -288,23 +288,29 @@ def hg_modified_files(hg, commit):
     )
 
 
-def get_touched_functions(path, deleted_lines, added_lines, content):
-    if content is None:
-        return set()
+def get_functions_from_metrics(metrics_space):
+    functions = []
+
+    if metrics_space["kind"] == "function" and metrics_space["name"] != "<anonymous>":
```
marco-c, line 294: Previously, we were skipping entries with `error` == `True`, are there such entries in the metrics results too?

calixteman: In case of error the space name is empty:
https://github.com/mozilla/rust-code-analysis/blob/master/src/metrics.rs#L105

bugbug/repository.py
calixteman: Yep, we make a DFS on the AST:
https://github.com/mozilla/rust-code-analysis/blob/master/src/metrics.rs#L470-L479

bugbug/repository.py
marco-c: OK, so I'll skip spaces with no name

marco-c: Done with d2e4bae.

bugbug/repository.py
```diff
@@ -365,7 +365,10 @@ def get_touched(functions, lines):
 
 
 def get_metrics(commit, metrics_space):
-    if metrics_space["kind"] == "function" and metrics_space["name"]:
+    name = metrics_space["name"]
+    error = metrics_space["kind"] in {"unit", "function"} and name == ""
```
calixteman, line 369: unit should always have a name (the file name in fact)

marco-c: Yep, so I'm going to return an error if the kind is either function or unit and the name is missing

PR #1474, opened 2020-04-08
Use struct instead of bytes conversion

### LINKED ISSUE #1462: Use struct.pack and struct.unpack instead of int.to_bytes and int.from_bytes
In bugbug/test_scheduling.py.

The changes can simply be tested by running the tests/test_test_scheduling.py tests.

yabirgb:
Fixes #1462 

What I did was to use [struct](https://docs.python.org/3.7/library/struct.html) as mentioned in the issue to make the conversion

### DIFF
--- a/bugbug/test_scheduling.py
+++ b/bugbug/test_scheduling.py
@@ -7,6 +7,7 @@ import itertools
 import os
 import pickle
 import shelve
+import struct
 import sys
 
 from bugbug import db, repository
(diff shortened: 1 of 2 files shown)


marco-c:
Thanks, the patch looks good overall, but linting is failing. Check the failure on Taskcluster by clicking the `Details` link.
You can also run linting locally by using pre-commit.

yabirgb:
@marco-c Done. Sorry but yesterday the service was under maintenance and I couldn't see what was happening

marco-c:
There is yet another usage to update in the file, in the `get_touched_together` function.

bugbug/test_scheduling.py
```diff
@@ -125,11 +126,11 @@ def set_touched_together(f1, f2):
     key = get_touched_together_key(f1, f2)
 
     if key not in touched_together:
-        touched_together[key] = (1).to_bytes(4, sys.byteorder)
+        touched_together[key] = struct.pack("i", 1)
```
marco-c, line 129: It's always positive, so you can use an unsigned int (`I`).

yabirgb: In the `else` is also always positive?

marco-c: Yep, it's the same number, in the else it's just increased by 1 :)

marco-c:
> There is yet another usage to update in the file, in the get_touched_together function.

Don't forget ^

PR #1509, opened 2020-04-22
End to end duplicate model script

ayush-1506:
Generate csv for performance of the similarity model + linearsvc classifier.


ayush-1506:
@marco-c r?

ayush-1506:
Upon further consideration, I think it's a better idea to add another method in `BaseSimilarity` to use the duplicatemodel directly within evaluation function. Thoughts?

scripts/duplicate_end_to_end.py
```diff
@@ -0,0 +1,127 @@
...
+def parse_args(args):
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "similaritymodel",
+        help="Similarity algorithm to test",
+        choices=similarity.model_name_to_class.keys(),
+    )
+
+    # At the moment, we use LinearSVC
+    parser.add_argument("duplicatemodel", help="Duplicate algorithm to test")
```
marco-c, line 48: I'd remove this argument, and just load the duplicate model from the file, whatever the underlying algorithm of the model is

scripts/duplicate_end_to_end.py
```diff
@@ -0,0 +1,127 @@
...
+        bug for bug in test_bugs.values() if not bug["creator"] in REPORTERS_TO_IGNORE
+    ]
+    with open("end_to_end_duplicate_predictions.csv", "w") as csvfile:
+        spamwriter = csv.writer(csvfile)
+
+        spamwriter.writerow(
+            ["bug 1 ID", "bug 1 summary", "bug 2 ID", "bug 2 summary", "prediction"]
+        )
+        for test_bug in test_bugs:
+            similar_bugs_id = similaritymodel.get_similar_bugs(test_bug)
```
marco-c, line 101:
```suggestion
            similar_bug_ids = similaritymodel.get_similar_bugs(test_bug)
```

marco-c:
> Upon further consideration, I think it's a better idea to add another method in BaseSimilarity to use the duplicatemodel directly within evaluation function. Thoughts?

Oh yes, good idea, this way it can be an option and we can compare the results with/without it

bugbug/similarity.py
```diff
@@ -139,6 +143,15 @@ def evaluation(self):
                 num_hits = 0
                 queries += 1
                 similar_bugs = self.get_similar_bugs(bug)[:10]
+                if end_to_end:
+                    sim_bugs = bugzilla.get(similar_bugs)
```
marco-c, line 147: We should have these bugs in the DB, so we shouldn't need to retrieve them from bugzilla

scripts/duplicate_end_to_end.py
```diff
@@ -0,0 +1,122 @@
+# -*- coding: utf-8 -*-
```
marco-c, line 1: Could you merge this script into scripts/generate_duplicate_sheet.py (using a parameter to choose between using only the duplicate model or using similarity first then duplicate model)?

scripts/generate_duplicate_sheet.py
```diff
@@ -1,53 +1,116 @@
...
+                except requests.HTTPError:
+                    logger.error(
+                        f"A pre-trained model is not available, you will need to train it yourself using the trainer script"
+                    )
+                    raise SystemExit(1)
+
+                zstd_decompress(model_file_name)
+                assert os.path.exists(
+                    model_file_name
+                ), "Decompressed file doesn't exist"
```
marco-c, line 53: Could you try using the `download_model` or `download_and_load_model` helper functions from bugbug/utils.py?

scripts/generate_duplicate_sheet.py
```diff
@@ -1,53 +1,116 @@
...
+                ]
+                probs = duplicate_model.classify(test_tuple, probabilities=True)
+            else:
+                test_tuple = [
+                    (test_bug, other_bug)
+                    for other_bug in test_bugs
+                    if other_bug != test_bug
+                ]
+                probs = duplicate_model.classify(test_tuple, probabilities=True)
+
```
marco-c, line 101: If you put the if before the for, you can keep classifying all test bugs at the same time in the case where the similarity is not used

ayush-1506:
@marco-c `download_model` currently doesn't expect similarity models as an argument. Do you want me to add support for these in this PR or another PR?

marco-c:
> @marco-c download_model currently doesn't expect similarity models as an argument. Do you want me to add support for these in this PR or another PR?

Ah, right! As you prefer, we can do it either here or in another PR.

scripts/generate_duplicate_sheet.py
marco-c: This comment still applies

bugbug/similarity.py
```diff
@@ -169,6 +173,17 @@ def evaluation(self):
...
+                if end_to_end:
+                    sim_bugs = [
+                        bug for bug in bugzilla.get_bugs() if bug["id"] in similar_bugs
+                    ]
+                    bug_couples = [(bug, sim_bugs[bug_id]) for bug_id in sim_bugs]
+                    probs = duplicatemodel.classify(bug_couples, probabilities=True)
+                    similar_bugs = [
+                        similar_bugs[idx]
+                        for idx, prob in enumerate(probs)
+                        if prob[1] > 0.8
```
marco-c, line 185: The 0.8 value should be configurable with a parameter (add the parameter to the constructor and save it in an attribute)

scripts/generate_duplicate_sheet.py
```diff
@@ -1,53 +1,82 @@
...
+        spamwriter.writerow(
+            ["bug 1 ID", "bug 1 summary", "bug 2 ID", "bug 2 summary", "prediction"]
+        )
+        if similarity_model:
+            test_tuple = []
+            for test_bug in test_bugs:
+                similar_bug_ids = similarity_model.get_similar_bugs(test_bug)
+                sim_bugs = bugzilla.get(similar_bug_ids)
+                test_tuple += [
+                    (sim_bugs[similar_bug_id], test_bug) for similar_bug_id in sim_bugs
```
marco-c, line 61: mmmh, shouldn't this be `for similar_bug_id in similar_bug_ids`? Make sure you test the script

scripts/generate_duplicate_sheet.py
```diff
@@ -1,53 +1,82 @@
...
+    with open("duplicate_predictions.csv", "w") as csvfile:
+        spamwriter = csv.writer(csvfile)
 
-# Warning: Classifying all the test bugs takes a while
-probs = m.classify(bug_tuples, probabilities=True)
+        spamwriter.writerow(
+            ["bug 1 ID", "bug 1 summary", "bug 2 ID", "bug 2 summary", "prediction"]
+        )
+        if similarity_model:
+            test_tuple = []
```
marco-c, line 56: Rename it as before, `bug_tuples`

scripts/generate_duplicate_sheet.py
```diff
@@ -1,53 +1,82 @@
...
+            for test_bug in test_bugs:
+                similar_bug_ids = similarity_model.get_similar_bugs(test_bug)
+                sim_bugs = bugzilla.get(similar_bug_ids)
```
marco-c, line 59: For consistency, rename this `similar_bugs`

bugbug/similarity.py
```diff
@@ -93,6 +94,10 @@ def __init__(self, cleanup_urls=True, nltk_tokenizer=False):
             self.cleanup_functions.append(feature_cleanup.url())
 
         self.nltk_tokenizer = nltk_tokenizer
+        self.threshold = threshold
+
+    def get_threshold(self):
+        return self.threshold
```
marco-c, line 100: You can remove this function, user of the class can just do model.threshold.

Also, rename it as `confidence_threshold`

scripts/generate_duplicate_sheet.py
ayush-1506: That won't make a difference, since it'll give the same result. sim_bugs is a dictionary mapping bug id to bug, and similar_bug_ids is a list of ids, so iterating through either of them finally just iterates through bug ids.

marco-c: Right, sorry!
You can rewrite it as this though, which is a bit nicer:
```
                bug_tuples += [
                    (similar_bug, test_bug)
                    for similar_bug in similar_bugs.values()
                ]
```

marco-c: Also, the opposite order (`(test_bug, similar_bug)`) should make the CSV output look nicer

bugbug/similarity.py
```diff
@@ -63,6 +64,9 @@ def get_nlp():
 
 def download_similarity_model(model_name):
     path = f"{model_name_to_class[model_name].__name__.lower()}.similaritymodel"
+
+    if path in os.listdir():
+        return path
```
marco-c, line 69: ?

bugbug/similarity.py
```diff
@@ -140,6 +147,9 @@ def evaluation(self):
             and "dupeme" not in bug["keywords"]
         )
 
+        if end_to_end:
+            duplicatemodel = DuplicateModel.load("duplicatemodel")
```
marco-c, line 151: In a follow-up, we should remove this from `evaluation` and make it run (optionally) as part of `get_similar_bugs`.

ayush-1506: Yeah, I'll open a new PR for that.

PR #1524, opened 2020-04-30
Download similarity models via download_model

ayush-1506:
Added an option to download similarity models via download_model.

### DIFF
--- a/bugbug/similarity.py
+++ b/bugbug/similarity.py
@@ -5,9 +5,12 @@
 
 import abc
 import bisect
+import logging
+import os
 import random
 import re
 from collections import defaultdict
+from functools import lru_cache
 from itertools import chain
 
 import joblib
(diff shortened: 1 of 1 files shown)


marco-c:
The tests are failing because https://github.com/mozilla/bugbug/blob/b83f1fbcbc1b1ccdc183cea0be707dd215a60944/bugbug/similarity.py#L47 can't be loaded (since it's not installed during tests).

We should make it load on-demand. You can do something like:
```Python
@lru_cache(maxsize=None)
def get_nlp():
    return spacy.load("en_core_web_sm")
```

And then replace the usage of `nlp` in https://github.com/mozilla/bugbug/blob/b83f1fbcbc1b1ccdc183cea0be707dd215a60944/bugbug/similarity.py#L83 with something like:
```Python
            text = [word.lemma_ for word in get_nlp()(text)]
```

ayush-1506:
I think adding similarity in utils is doing more harm than good, since it breaks a lot of other things. Maybe it's a better idea to use path as an argument in `download_model` @marco-c ? Might involve some changes but should make things simpler.

PR #1531, opened 2020-05-03
Return a list of tasks known by the test selection model

### LINKED ISSUE #1528: Return a list of known tasks together with the selection results
This way, when a new task is added, the scheduler in mozilla-central can schedule it.

marco-c:
The list of known tasks is generated during the HTTP service's boot.
This way, since the HTTP service is restarted ~daily, we will schedule newly added
tasks for all pushes in approximately a whole day since their introduction (useful
to gather enough data about them for the next training run, and to help sheriffs
identify if they are intermittently failing).

Fixes #1528

### DIFF
--- a/http_service/bugbug_http/boot.py
+++ b/http_service/bugbug_http/boot.py
@@ -7,6 +7,9 @@ import concurrent.futures
 import logging
 import os
 
+import mozci.push
+import requests
+
 from bugbug import repository, test_scheduling, utils
 from bugbug_http import ALLOW_MISSING_MODELS, REPO_DIR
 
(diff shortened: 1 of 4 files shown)


http_service/bugbug_http/boot.py
```diff
@@ -87,9 +90,23 @@ def extract_commit_experiences():
...
+        # Store in a file the list of tasks in the latest autoland push.
+        r = requests.get(
+            "https://hg.mozilla.org/integration/autoland/json-pushes?version=2"
+        )
+        r.raise_for_status()
+        data = r.json()
+        last_push = data["pushes"][str(data["lastpushid"])]
+        last_push_revs = last_push["changesets"][::-1]
+        with open("known_tasks", "w") as f:
+            f.write("\n".join(mozci.push.Push(last_push_revs).target_task_labels))
```
ahal, line 103: This is fine, so treat this comment as optional..

But it's just as easy to grab this from the index. I.e:
https://firefox-ci-tc.services.mozilla.com/api/index/v1/task/gecko.v2.autoland.latest.taskgraph.decision/artifacts/public/target-tasks.json

That way you don't need to query hgmo to find the latest push. Maybe `mozci.Push` should accept an index path to a decision task and then use the `parameters.yml` to instantiate itself.

marco-c: Right, it doesn't change much as it's just done once per day basically, but it's less code, so I've done it in cbfe61c.

PR #1548, opened 2020-05-06
Bug 1622336 - Trigger schedules API as soon as a push happens

armenzg:


armenzg:
@marco-c could you please give me some initial feedback?

http_service/Dockerfile
```diff
@@ -10,4 +10,5 @@ RUN pip install --disable-pip-version-check --quiet --no-cache-dir -r /requireme
 COPY . /code/http_service
 RUN pip install --disable-pip-version-check --quiet --no-cache-dir /code/http_service
 
-CMD gunicorn -b 0.0.0.0:$PORT bugbug_http.app --preload --timeout 30 -w 3
+# Run the Pulse listener in the background
+CMD (/code/http_service/listener.py &> /code/http_service/pulse.log &) && gunicorn -b 0.0.0.0:$PORT bugbug_http.app --preload --timeout 30 -w 3
```
marco-c, line 14: We should find a way to log to the foreground rather than a file

http_service/listener.py
```diff
@@ -0,0 +1,62 @@
...
+BUGBUG_HTTP_SERVER = os.environ.get("BUGBUG_HTTP_SERVER", "http://localhost:8000")
+
+
+def _on_message(body, message):
+    try:
+        branch = body["payload"]["data"]["repo_url"].split("/")[-1]
+        rev = body["payload"]["data"]["heads"][0]
+        if branch in ["autoland", "try"]:
+            url = "{}/push/{}/{}/schedules".format(BUGBUG_HTTP_SERVER, branch, rev)
+            response = requests.get(url, headers={"X-Api-Key": "pulse_listener"})
```
marco-c, line 33: I was thinking we'd directly put the job in the queue, but this should work exactly as well.

http_service/pulse/consumers.py
```diff
@@ -0,0 +1,48 @@
...
+_CONNECTION_URL = "amqp://{}:{}@pulse.mozilla.org:5671/?ssl=1"
+
+
+def _generate_hg_pushes_queue(user):
+    return Queue(
+        name="queue/{}/pushes".format(user),
+        exchange=Exchange("exchange/hgpushes/v2", type="topic", no_declare=True,),
+        routing_key="#",
+        durable=True,
+        # XXX: This should not be auto delete
```
marco-c, line 27: Maybe it makes sense for it to be auto delete? So e.g. if the service is down we don't keep getting messages in the queue

http_service/pulse/consumers.py
```diff
@@ -0,0 +1,48 @@
...
+        self.queues = queues
+        self.callback = callback
+
+    def get_consumers(self, Consumer, channel):
+        return [Consumer(queues=self.queues, callbacks=[self.callback])]
+
+
+class ConsumerFactory:
+    @staticmethod
+    def hg_pushes(user, password, callback):
```
marco-c, line 44: I'd make this actually a contextmanager, so the user can do:
```
with ConsumerFactory.hg_pushes(...) as consumer:
    consumer.run()
```

http_service/pulse/consumers.py
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env python3
```
marco-c, line 1: Nit: since the amount of code is relatively limited, could you merge the two files?

http_service/listener.py
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env python3
```
marco-c, line 1: Could you move the code into `http_service/bugbug_http`?
Then we can add the script to `console_scripts` in `http_service/setup.py` so we can run it directly as `bugbug-http-pulse-listener` instead of `/code/http_service/listener.py`

http_service/requirements.txt
```diff
@@ -5,6 +5,7 @@ Flask==1.1.2
 flask-apispec==0.8.8
 flask-cors==3.0.8
 gunicorn==20.0.4
-marshmallow==3.5.2
+kombu==4.6.8
+marshmallow==3.5.1
```
marco-c, line 9: I assume this was unintended

http_service/listener.py
marco-c: (This would also probably avoid the `pulse` addition in `.isort.cfg`).

http_service/pulse/consumers.py
```diff
@@ -0,0 +1,48 @@
...
+        consumer.run()
+```
+"""
+from kombu import Connection, Exchange, Queue
+from kombu.mixins import ConsumerMixin
+
+_CONNECTION_URL = "amqp://{}:{}@pulse.mozilla.org:5671/?ssl=1"
+
+
+def _generate_hg_pushes_queue(user):
```
marco-c, line 21: Nit: you could just inline this function in the `hg_pushes` function below, it's just a line after all (split into multiple because of its length, but still not much code)

http_service/Dockerfile
armenzg: The whole problem I was facing had to do that in order for new code to be executed I needed to call `docker-compose build bugbug-http-service`.

Redirection to disk was left by mistake. Fixed locally.

http_service/pulse/consumers.py
armenzg: If we don't delete the queue then we will keep adding pushes to the Pulse queue and once our Pulse consumer goes back up we will be able to process and catch up.

marco-c: Yep, but we don't want to do that, as those would be old pushes and so probably test selection for them is not needed anymore (and if it is, we'd already have got the request directly via the API).

armenzg:
@marco-c I've rebased and added a commit with a description of the change in case you want to review the [commit message](https://github.com/mozilla/bugbug/pull/1548/commits/c7bed970771efda0f86ba9bfa6f1398a923b4c71).

```
 Bug 1622336 - Initialize schedules service when Hg push happens

This change adds a Pulse listener for Hg pushes as a background process
to `bugbug-http-service`.

This listener works if `PULSE_USER` and `PULSE_PASSWORD` are set before starting
up the Docker containers.

The listener only pays attention to autoland and try pushes. When these happen
we do a GET request from the bugbug deployment. This will get the schedules service
ready for when the Gecko decision task will query it.
```

Issue #1506, opened 2020-04-21
Instead of applying patch in git, fetch it from the local Mercurial repository


marco-c:
Using git-cinnabar, we can replace https://github.com/mozilla/bugbug/blob/e60a51b6b744ae1c1bbd65cbc28fc7f796f851e0/scripts/commit_classifier.py#L358-L380 and just fetch the commit from the local Mercurial repository where we already applied the patch.

Then, we can stop installing `patch` in our Docker image: https://github.com/mozilla/bugbug/blob/e60a51b6b744ae1c1bbd65cbc28fc7f796f851e0/infra/dockerfile.commit_retrieval#L10.

Dcoder99:
Hi @marco-c, I'd like to try and fix this issue! I hope that's okay :)

marco-c:
Sure, feel free! Any issue is open for contribution by anyone. As a first step, read https://github.com/mozilla/bugbug/issues/1092.

Dcoder99:
May I know where the Mercurial repo for bugbug is?

marco-c:
https://hg.mozilla.org/mozilla-central/

Dcoder99:
So what worked on my terminal is this:
$git cinnabar fetch hg::https://hg.mozilla.org/mozilla-central/ 50244579252a229131ae04a1d2c60682611e5702

I replaced lines 358-380 with:
![bugbug1](https://user-images.githubusercontent.com/28946131/81383012-f4362880-912c-11ea-8530-ad0ecfe5198a.png)

But I am not sure how to test this file, without having to run trainer.py as mentioned in the README.

marco-c:
You will need to fetch the commit that was just applied to the local mercurial repository, not a specific commit which already exists in https://hg.mozilla.org/mozilla-central/.

Dcoder99:
I think I've not understood what a "local mercurial repo" means. 
If I clone the link https://hg.mozilla.org/mozilla-central/ ,is it referred to as a local .hg repo? 
How or Where do I find the 'commit that was just applied to the local mercurial repository'?
I'm really sorry if its a silly question, I'm a complete beginner :P

marco-c:
If you look at the code of the commit classifier script mentioned in the first comment, it is cloning a Mercurial repository, then cloning a git repository using git cinnabar. Then it applies a set of patches to the locally cloned Mercurial repository and tries to apply the same set of patches to the locally cloned git repository.
[...]

PR #1552, opened 2020-05-09
Patches fetched from local mercurial

anubhabsen:
Issue #1506 
I think this applies the patches from the local mercurial repository using a fetch. Please do let me know if there is something I am missing, and how I should go about testing it.

### DIFF
--- a/scripts/commit_classifier.py
+++ b/scripts/commit_classifier.py
@@ -361,7 +361,7 @@ class CommitClassifier(object):
                         "cinnabar",
                         "fetch",
                         f"hg::{self.git_repo_dir}",
-                        f"{revision['id']}"
+                        f"{revision['id']}",
                     ],
                     check=True,
                     cwd=self.git_repo_dir,
--- a/scripts/commit_classifier.py
+++ b/scripts/commit_classifier.py
@@ -355,28 +355,18 @@ class CommitClassifier(object):
             )
 
             if self.git_repo_dir:
-                patch_proc = subprocess.Popen(
-                    ["patch", "-p1", "--no-backup-if-mismatch", "--force"],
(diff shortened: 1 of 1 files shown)


scripts/commit_classifier.py
```diff
@@ -355,28 +355,18 @@ def load_user(phid):
...
-                        f"user.name={author_name}",
-                        "-c",
-                        f"user.email={author_email}",
-                        "commit",
-                        "-am",
-                        message,
+                        "cinnabar",
+                        "fetch",
+                        f"hg::{self.git_repo_dir}",
+                        f"{revision['id']}",
```
marco-c, line 364: This is a Phabricator revision ID, it's not the hash of the commit we just applied to the local mercurial repository.

anubhabsen: Sorry about that. If I understand correctly, ```needed_stack``` contains the list of patches to be applied. So ```path.base_revision``` where patch is an element of needed_stack is the hash of a commit that needs to be applied?

marco-c: No, base_revision is the revision in the Mercurial repository on top of which the patch must be applied (https://github.com/mozilla/bugbug/blob/78e5b5a961f1985d7510a74de4c5741111a58d79/scripts/commit_classifier.py#L278-L282).

The process should be:
1) Get patches from Phabricator;
2) Apply each patch to the local Mercurial repository at the specified base revision;
3) Fetch the commits from the local Mercurial repository in the local git repository.

For 3, the commits come from the local Mercurial repository, Phabricator is already out of the picture.

marco-c:
Closing for lack of updates, feel free to reopen if you still intend to work on this.

PR #1560, opened 2020-05-14
[WIP] Fetched local mercurial repo's commit hash

### LINKED ISSUE #1506: Instead of applying patch in git, fetch it from the local Mercurial repository

Dcoder99:
closes #1506 
I have currently only fetched the local mercurial repo's commit hash where the patches are applied, yet to apply that commit to the git repo and replace lines 358-380.
I hope I'm in the right direction :)


scripts/commit_classifier.py
```diff
@@ -355,6 +355,31 @@ def load_user(phid):
...
+                
+                # Finding 1st occurrence of ',' 
+                substr = ","
+                occurrence = 1
+                inilist = [i for i in range(0, len(ini_str)) 
+                            if ini_str[i:].startswith(substr)] 
+                end = inilist[occurrence - 1]
+
+                # hash of the local commit
+                local_commit_hash = res[start:end]
```
marco-c, line 381: You should be able to replace all this by using hg.log instead, taking the last applied revisions. You can use https://github.com/mozilla/bugbug/blob/c8ad32b2d6480f65132208448b21cf3eda3a46c8/bugbug/repository.py#L687.

So, first apply all patches to the local mercurial repository, then after the end of the `for patch in needed_stack:` loop, you can get the applied revisions with `get_revs(hg, f"-{len(stack)}")`.

Dcoder99: Thanks, will do.
But the first comment of #1506 issue says **_"just fetch the commit from the local Mercurial repository where we already applied the patch."_**
So I am assuming that the patches are already applied to the local hg repo. So I will just need to use get_revs to get the hash of latest revision. Is that correct?

marco-c: The code which is applying the patch is the `for patch in needed_stack:` loop I mentioned above :)

scripts/commit_classifier.py
```diff
@@ -315,10 +315,12 @@ def load_user(phid):
                 revision["fields"]["title"], revision["fields"]["summary"]
             )
 
+            latest_rev = repository.get_revs(hg, f"-{len(stack)}")
```
marco-c, line 318: We haven't applied the patches yet here, so this will return wrong revisions.

marco-c: The code should moved after the end of the for loop.

Dcoder99: Done!

scripts/commit_classifier.py
```diff
@@ -377,6 +373,8 @@ def load_user(phid):
                     check=True,
                     cwd=self.git_repo_dir,
                 )
+        latest_rev = repository.get_revs(hg, f"-{len(stack)}")
```
marco-c, line 376: You're setting this variable here, but trying to use it earlier :smile: 
Maybe this is not an easy enough task to start with, I'd suggest working on some personal projects first to get some initial experience with programming.

Issue #573, opened 2019-06-09
Get bug data all at once instead of attachments and comments separately


marco-c:
Using the feature from https://github.com/mozilla-bteam/bmo/pull/1148.

sladyn98:
@marco-c This looks like a good issue to work on :)
Could I try it ?

marco-c:
Sure! The function to modify is https://github.com/mozilla/bugbug/blob/master/bugbug/bugzilla.py#L107.
You need to make sure it behaves the same before/after your change. Choose a bug number, e.g. 1405567, try to download it before and after your change and make sure the data is the same.

sladyn98:
If I understand correctly I need to make a request to the api 
`requests.get("https://bugzilla.mozilla.org/rest/bug/[ID]?include_fields=attachments,comments", params=bug_query_params)`
and then return the list of bugs

sladyn98:
cc @marco-c

marco-c:
Yes, making sure the same data is returned as the current `bugzilla.get` function.

sladyn98:
https://github.com/mozilla/bugbug/pull/596
Opened a draft pr

marco-c:
Unfortunately it seems we can't do this. Bugzilla would return the attachment's data too, which is really big and would slow down our requests.

PR #1575, opened 2020-05-15
Added Code Fix for #573 Get bug data all at once instead of attachments and comments separately

rajeshkumargp:
Code update for the fixing the issue #573


bugbug/bugzilla.py
```diff
@@ -93,50 +94,78 @@ def bughandler(bug):
...
-        comment_include_fields=COMMENT_INCLUDE_FIELDS,
-        attachmenthandler=attachmenthandler,
-        attachment_include_fields=ATTACHMENT_INCLUDE_FIELDS,
-        historyhandler=historyhandler,
-    ).get_data().wait()
+    if isinstance(bug_ids, list):
+        # Expected Format
+        bug_ids_list = list(set(map(int, bug_ids)))
+
+    elif isinstance(bug_ids, six.string_types):
```
marco-c, line 119: Let's not use six, we don't care about compatibility with Python 2.

bugbug/bugzilla.py
```diff
@@ -93,50 +94,78 @@ def bughandler(bug):
...
+        bug_ids_list = [int(bug_ids)]
+
+    elif isinstance(bug_ids, int):
+        bug_ids_list = [bug_ids]
```
marco-c, line 123: Let's just overwrite `bug_ids`

bugbug/bugzilla.py
```diff
@@ -93,50 +94,78 @@ def bughandler(bug):
...
+    else:
+        raise ValueError(
+            "Invalid datatype for the parameter-bug_ids with value- {bug_ids} and datatype-{type(bug_ids)}"
+        )
```
marco-c, line 127: Instead of the runtime check on type, let's use typing (enforcing the parameter to be a list of IDs and updating all callers to pass a list of IDs).

bugbug/bugzilla.py
```diff
@@ -93,50 +94,78 @@ def bughandler(bug):
...
+    new_bugs = dict()
+    for a_bug_id in bug_ids_list:
+        params_for_custom_fields = {
+            "id": a_bug_id,
+            "include_fields": "_default,history,comments,attachments",
+        }
+        response = utils.get_session("bugzilla").get(
+            "https://bugzilla.mozilla.org/rest/bug", params=params_for_custom_fields
+        )
+        response.raise_for_status()
```
marco-c, line 138: This would be super slow, as we'd make one request per bug, while previously we were doing requests in batches.

bugbug/bugzilla.py
```diff
@@ -93,50 +94,78 @@ def bughandler(bug):
...
+        # Realignment of results to cope with previous versions of data manipulation
+        bug_id_of_a_bug_info = int(a_bug_info["id"])
+
+        # Delete other fields from comments
+        current_comments_array = a_bug_info["comments"]
+        current_comments_array = filter_keys(
+            array_of_dict_values=current_comments_array,
+            required_fields=COMMENT_INCLUDE_FIELDS,
+        )
+        a_bug_info["comments"] = current_comments_array
```
marco-c, line 156: Can't we pass the include fields to Bugzilla instead of manually removing the fields after the fact?

bugbug/bugzilla.py
rajeshkumargp: Changed six.string_types to str

bugbug/bugzilla.py
rajeshkumargp: `bug_ids` is overwritten, and is alone used. Fixed

bugbug/bugzilla.py
rajeshkumargp: Added  typing enforcing to function - get - Since Set, int, str, are also used in other calls, the typing is enforced to use all these including list.

bugbug/bugzilla.py
rajeshkumargp: include_fields is assigned `"_default,history,comments,attachments"`. We don`t have fine grained control on sub-fields of `comments`, 'attachments`.  So, the manual deletion is implemented here.

bugbug/bugzilla.py
rajeshkumargp: Fixed with this :   https://github.com/mozilla/bugbug/pull/1575/commits/cfba88fa234eb3d4d3fcc91eea8a640faec937bc

bugbug/bugzilla.py
rajeshkumargp: Fixed here: https://github.com/mozilla/bugbug/pull/1575/commits/cfba88fa234eb3d4d3fcc91eea8a640faec937bc

bugbug/bugzilla.py
rajeshkumargp: Added typing - enforcing here https://github.com/mozilla/bugbug/pull/1575/commits/cfba88fa234eb3d4d3fcc91eea8a640faec937bc

bugbug/bugzilla.py
rajeshkumargp: Fixed in this commit : https://github.com/mozilla/bugbug/pull/1575/commits/15b460fe9eb94af905461d114a45860c47c92be8

bugbug/bugzilla.py
rajeshkumargp: Fixed in this commit : https://github.com/mozilla/bugbug/pull/1575/commits/15b460fe9eb94af905461d114a45860c47c92be8

bugbug/bugzilla.py
rajeshkumargp: Fixed in this commit : https://github.com/mozilla/bugbug/pull/1575/commits/15b460fe9eb94af905461d114a45860c47c92be8

bugbug/bugzilla.py
rajeshkumargp: Fixed in this commit : https://github.com/mozilla/bugbug/pull/1575/commits/15b460fe9eb94af905461d114a45860c47c92be8

bugbug/bugzilla.py
rajeshkumargp: Fixed in this commit : https://github.com/mozilla/bugbug/pull/1575/commits/15b460fe9eb94af905461d114a45860c47c92be8

bugbug/bugzilla.py
marco-c: No, this is super-ugly :)
I was suggesting enforcing it is a list and updating all callers to use a list.

bugbug/bugzilla.py
marco-c: Oh, if that's the case, then unfortunately we cannot do this, as we'd spend too much time and resources downloading the attachments' contents.

bugbug/bugzilla.py
rajeshkumargp: We are making one request for `Bugzilla.BUGZILLA_CHUNK_SIZE` size of bugs

rajeshkumargp:
In a single request, 
- We can get _DEFAULT, HISTORY, COMMENTS, ATTACHMENTS  information for mutliple bug ids, such that we can include or exclude some fields, **which can not be a sub-field of main-field such as data field of ATTACHMENTS** in /bug API.
- However,  We can select fields  /bug/{bug_id}/attachment api, which should be per bug level call.

bugbug/bugzilla.py
rajeshkumargp: 1) Using "include_fields": "_default,history,comments,attachments", We are getting HISTORY, COMMENTS and ATTACHMENTS of a bug source: https://bmo.readthedocs.io/en/latest/api/core/v1/bug.html#get-bug
2) The HISTORY is an array, COMMENTS is an array, ATTACHMENTS is an array. 
3) ATTACHMENTS will have 15 fields, But we need only 7 fields. Even if we dont delete them, there will be no impact in running the successive codes.
4) The fields deletion is done to sync with previous version of code.
5) Similarly, COMMENTS will have 11 fields, but we need 5 fields.
6) Alternatively, For a BATCH of Bugs, 
    - We can get BUGS  _default fields, history fields,comments fields using REST API.
    - Using separate call for each bug, we can get Custom Attachment Fields which will save from getting attachment data.
    - Example https://bugzilla.mozilla.org/rest/bug/1633278/attachment?include_fields=id,is_obsolete,flags,is_patch,creator,content_type,creation_time 
    - Here we have to make request for each bug.
    - Getting attachment meta-information  can be retrieved per bug, instead of batches of bugs. 

7) Please give suggestions on how to proceed further.

marco-c:
OK, so unfortunately we can't do this. The attachment data can be really big, so this would slow things down rather than improve.

PR #1588, opened 2020-05-21
Refactor similarity classes

ayush-1506:
Addresses : https://github.com/mozilla/bugbug/pull/1509#discussion_r426522938 .

### DIFF
--- a/bugbug/similarity.py
+++ b/bugbug/similarity.py
@@ -82,7 +82,11 @@ def download_and_load_similarity_model(model_name):
 
 class BaseSimilarity(abc.ABC):
     def __init__(
-        self, cleanup_urls=True, nltk_tokenizer=False, confidence_threshold=0.8
+        self,
+        cleanup_urls=True,
+        nltk_tokenizer=False,
+        confidence_threshold=0.8,
+        end_to_end=False,
     ):
         self.cleanup_functions = [
             feature_cleanup.responses(),
@@ -98,6 +102,10 @@ class BaseSimilarity(abc.ABC):
         self.nltk_tokenizer = nltk_tokenizer
         self.confidence_threshold = confidence_threshold
 
+        self.duplicatemodel = (
+            DuplicateModel.load("duplicatemodel") if end_to_end else None
+        )
+
     def get_text(self, bug, all_comments=False):
         if all_comments:
             comments = " ".join(comment["text"] for comment in bug["comments"])
@@ -105,6 +113,29 @@ class BaseSimilarity(abc.ABC):
             comments = bug["comments"][0]["text"]
 
         return "{} {}".format(bug["summary"], comments)
+
+    def get_similar_bugs(self, bug):
+        similar_bug_ids = self.search_similar_bugs(bug)
+        if self.duplicatemodel:
+            similar_bugs = [
+                bug for bug in bugzilla.get_bugs() if bug["id"] in similar_bug_ids
+            ]
+            bug_couples = [(bug, similar_bug) for similar_bug in similar_bugs]
+            probs_bugs_couples = sorted(
+                zip(
+                    self.duplicatemodel.classify(bug_couples, probabilities=True),
+                    bug_couples,
+                ),
+                key=lambda v: -v[0][1],
+            )
+
+            similar_bug_ids = [
+                similar_bug["id"]
+                for prob, (bug, similar_bug) in probs_bugs_couples
+                if prob[1] > self.confidence_threshold
+            ]
+
+        return similar_bug_ids
 
     def text_preprocess(self, text, stemming=True, lemmatization=False, join=False):
 
@@ -134,7 +165,7 @@ class BaseSimilarity(abc.ABC):
             return " ".join(word for word in text)
         return text
 
-    def evaluation(self, end_to_end=False):
+    def evaluation(self):
         # A map from bug ID to its duplicate IDs
         duplicates = defaultdict(set)
         all_ids = set(
@@ -143,9 +174,6 @@ class BaseSimilarity(abc.ABC):
             if bug["creator"] not in REPORTERS_TO_IGNORE
             and "dupeme" not in bug["keywords"]
         )
-
-        if end_to_end:
-            duplicatemodel = DuplicateModel.load("duplicatemodel")
 
         for bug in bugzilla.get_bugs():
             dupes = [entry for entry in bug["duplicates"] if entry in all_ids]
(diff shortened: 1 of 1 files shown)


bugbug/similarity.py
```diff
@@ -106,6 +111,21 @@ def get_text(self, bug, all_comments=False):
 
         return "{} {}".format(bug["summary"], comments)
 
+    def get_similar_bugs(self, bug):
+        similar_bug_ids = self.get_similar_bugs_child(bug)
+        similar_bugs = bugzilla.get(similar_bug_ids)
```
marco-c, line 116: If `end_to_end` is disabled, there is no need to get the bugs.

bugbug/similarity.py
```diff
@@ -106,6 +111,21 @@ def get_text(self, bug, all_comments=False):
...
+        if self.end_to_end:
+            duplicatemodel = DuplicateModel.load("duplicatemodel")
+            bug_couples = [(bug, similar_bugs[bug_id]) for bug_id in similar_bugs]
+            probs = duplicatemodel.classify(bug_couples, probabilities=True)
+            similar_bug_ids = [
+                similar_bug_ids[idx]
+                for idx, prob in enumerate(probs)
+                if prob[1] > self.confidence_threshold
+            ]
```
marco-c, line 125: As a side note, maybe we should also sort them by prob (higher prob should come first).

bugbug/similarity.py
```diff
@@ -106,6 +111,21 @@ def get_text(self, bug, all_comments=False):
 
         return "{} {}".format(bug["summary"], comments)
 
+    def get_similar_bugs(self, bug):
+        similar_bug_ids = self.get_similar_bugs_child(bug)
+        similar_bugs = bugzilla.get(similar_bug_ids)
+        if self.end_to_end:
+            duplicatemodel = DuplicateModel.load("duplicatemodel")
```
marco-c, line 118: How fast is this? If it's slow, maybe we should load it just once in the constructor.

bugbug/similarity.py
marco-c, line 116: We should avoid making a query here and use the bugs from the DB (`get_similar_bugs_child` will return bugs from the DB anyway, so we are sure they are available).

bugbug/similarity.py
```diff
@@ -280,7 +286,7 @@ def __init__(
             output_prefix="simdata.shdat", corpus=corpus_lsi, num_features=300
         )
 
-    def get_similar_bugs(self, query, k=10):
+    def get_similar_bugs_child(self, query, k=10):
```
marco-c, line 289: Let's try to find a nicer name than `get_similar_bugs_child`, maybe `search_similar_bugs`.

marco-c: So `get_similar_bugs` would be the function in the base class, `search_similar_bugs` the function in the children classes.

bugbug/similarity.py
ayush-1506: Yeah that's probably more efficient.

bugbug/similarity.py
```diff
@@ -97,6 +101,10 @@ def __init__(
 
         self.nltk_tokenizer = nltk_tokenizer
         self.confidence_threshold = confidence_threshold
+        self.end_to_end = end_to_end
```
marco-c, line 104: Instead of storing this, just do:
```
self.duplicate_model = DuplicateModel.load("duplicatemodel") if end_to_end else None
```

And in `get_similar_bugs` check `if self.duplicate_model is not None` instead of `if self.end_to_end`.

bugbug/similarity.py
```diff
@@ -106,6 +114,28 @@ def get_text(self, bug, all_comments=False):
...
+                bug for bug in bugzilla.get_bugs() if bug["id"] in similar_bug_ids
+            ]
+            bug_couples = [(bug, similar_bug) for similar_bug in similar_bugs]
+            probs = list(
+                zip(
+                    self.duplicatemodel.classify(bug_couples, probabilities=True),
+                    bug_couples,
+                )
+            )
+            probs.sort(key=lambda v: -v[0][1])
```
marco-c, line 130: Nit: instead of converting to list and then using sort, you can use sorted:
```
probs_bugs = zip(...)
probs_bugs = sorted(probs_bugs, ...)
```
or even:
```
probs_bugs = sorted(zip(...), ...)
```

bugbug/similarity.py
```diff
@@ -106,6 +114,28 @@ def get_text(self, bug, all_comments=False):
...
         return "{} {}".format(bug["summary"], comments)
 
+    def get_similar_bugs(self, bug):
+        similar_bug_ids = self.search_similar_bugs(bug)
+        if self.end_to_end:
+            similar_bugs = [
+                bug for bug in bugzilla.get_bugs() if bug["id"] in similar_bug_ids
+            ]
+            bug_couples = [(bug, similar_bug) for similar_bug in similar_bugs]
+            probs = list(
```
marco-c, line 124: Rename it `probs_bugs`, as `probs` is misleading

bugbug/similarity.py
```diff
@@ -106,6 +114,28 @@ def get_text(self, bug, all_comments=False):
...
+                    self.duplicatemodel.classify(bug_couples, probabilities=True),
+                    bug_couples,
+                )
+            )
+            probs.sort(key=lambda v: -v[0][1])
+            similar_bug_ids = [
+                prob[1][1]["id"]
+                for prob in probs
+                if prob[0][1] > self.confidence_threshold
+            ]
```
marco-c, line 135: Nit: the following is more readable:
```Python
similar_bug["id"]
for prob, (bug, similar_bug)
in probs_bugs
if prob[1] > self.confidence_threshold
```

bugbug/similarity.py
marco-c: Or `probs_bug_couples`

ayush-1506:
@marco-c Made the required changes. Apologies for the delay, got caught up in some stuff.

bugbug/similarity.py
```diff
@@ -106,6 +114,29 @@ def get_text(self, bug, all_comments=False):
...
         return "{} {}".format(bug["summary"], comments)
 
+    def get_similar_bugs(self, bug):
+        similar_bug_ids = self.search_similar_bugs(bug)
+        if self.duplicatemodel:
+            similar_bugs = [
+                bug for bug in bugzilla.get_bugs() if bug["id"] in similar_bug_ids
+            ]
+            bug_couples = [(bug, similar_bug) for similar_bug in similar_bugs]
+            probs_bugs_couples = sorted(
```
marco-c, line 124:
```suggestion
            probs_bug_couples = sorted(
```

bugbug/similarity.py
```diff
@@ -106,6 +114,29 @@ def get_text(self, bug, all_comments=False):
...
+                zip(
+                    self.duplicatemodel.classify(bug_couples, probabilities=True),
+                    bug_couples,
+                ),
+                key=lambda v: -v[0][1],
+            )
+
+            similar_bug_ids = [
+                similar_bug["id"]
+                for prob, (bug, similar_bug) in probs_bugs_couples
```
marco-c, line 134:
```suggestion
                for prob, (bug, similar_bug) in probs_bug_couples
```

marco-c:
Thanks!

PR #1589, opened 2020-05-22
Load target tasks from more than one autoland push to make sure we didn't get them from a broken one

marco-c:


http_service/bugbug_http/boot.py
```diff
@@ -95,13 +95,39 @@ def extract_commit_experiences():
...
+        ]
+
+        # Store in a file the list of tasks in the latest autoland pushes.
+        # We use more than one to protect ourselves from broken decision tasks.
+        known_tasks = set()
+        for rev in revs:
+            r = requests.get(
+                f"https://firefox-ci-tc.services.mozilla.com/api/index/v1/task/gecko.v2.autoland.revision.{rev}.taskgraph.decision/artifacts/public/target-tasks.json"
+            )
+            r.raise_for_status()
```
marco-c, line 126: I guess I should allow some to not be present? (e.g. if a push totally broke the decision task)

marco-c: Done.

PR #1597, opened 2020-05-27
Reduced number of requests for getting bug information from 4 to 2

rajeshkumargp:
- Reduced number of requests  for getting  bug information from 4 to 2
- For a batch of bugs, 1 request for getting bug -> history, comments, and generic info using bugzilla REST API.
- For a batch of bugs, 1 more request for getting bug -> attachment info using Earlier BugZilla module.
- Added typing enforcement for parameters of  bugzilla.get() function
- Accordingly the parameters in the usage of bugzilla.get(), is also updated.


bugbug/bugzilla.py
```diff
@@ -93,50 +94,84 @@ def bughandler(bug):
...
+        for a_ind, a_value in enumerate(array_of_dict_values):
+            a_value = {
+                a_field: a_value[a_field]
+                for a_field in list(a_value.keys())
+                if a_field in required_fields
+            }
+            array_of_dict_values[a_ind] = a_value
+        return array_of_dict_values
+
+    def cleanup_fields_in_batch_of_bugs(batch_of_bugs):
```
marco-c, line 117: Can you give me a list of the attributes that are cleaned up?

bugbug/bugzilla.py
```diff
@@ -93,50 +94,84 @@ def bughandler(bug):
...
+            )
+            a_bug_info["comments"] = current_comments_array
+        return batch_of_bugs
 
     def attachmenthandler(bug, bug_id):
         bug_id = int(bug_id)
-
         if bug_id not in new_bugs:
             new_bugs[bug_id] = dict()
-
```
marco-c, line 120: Nit: Don't remove unrelated white lines

scripts/similarity_query.py
```diff
@@ -54,7 +54,7 @@ def main(args):
             f"{similarity.model_name_to_class[args.algorithm].__name__.lower()}.similaritymodel"
         )
 
-    bug_ids = model.get_similar_bugs(bugzilla.get(args.bug_id)[args.bug_id])
+    bug_ids = model.get_similar_bugs(bugzilla.get([int(args.bug_id)])[args.bug_id])
```
marco-c, line 57: It should already be an int, as we have `parser.add_argument("bug_id", help="Bug to test", type=int)`

scripts/bug_classifier.py
```diff
@@ -46,8 +47,10 @@ def classify_bugs(model_name, classifier, bug_id):
     model_class = get_model_class(model_name)
     model = model_class.load(model_file_name)
 
+    bug_id = int(bug_id)
```
marco-c, line 50: Add `type=int` to the `--bug-id` argument instead.

scripts/bug_classifier.py
```diff
@@ -18,6 +18,7 @@
 
 
 def classify_bugs(model_name, classifier, bug_id):
+
```
marco-c, line 21: Nit: don't add an unrelated newline here

bugbug/bugzilla.py
```diff
@@ -93,50 +94,84 @@ def bughandler(bug):
     return all_ids
 
 
-def get(ids_or_query):
-    new_bugs = {}
+def get(bug_ids: Union[List[int], Set[int]]):
```
marco-c, line 97: You just need something that is iterable, so use `Iterable[int]` instead of this `Union`

marco-c, line 97: You need to add the return type too

bugbug/bugzilla.py
```diff
@@ -93,50 +94,84 @@ def bughandler(bug):
...
+
+        Bugzilla(
+            batch,
+            bughandler=None,
+            commenthandler=None,
+            comment_include_fields=None,
+            attachmenthandler=attachmenthandler,
+            attachment_include_fields=ATTACHMENT_INCLUDE_FIELDS,
+            historyhandler=None,
+        ).get_data().wait()
```
marco-c, line 174: This means we'll query the bugs/history/comment first, and then query the attachments.
Without the patch, we were querying all of them in parallel.

bugbug/bugzilla.py
rajeshkumargp: The Comments sub fields - bug_id, attachment_id, raw_text, creator, time, creation_time, is_private, is_markdown are Cleaned-Up.

bugbug/bugzilla.py
rajeshkumargp: we'll query the bugs/history/comment first, and then query the attachments. - yes.

"Without the patch, we were querying all of them in parallel." -  If the handlers are present, then the bug information is retrieved, followed by history info, comment info, and then attachment info in order. 
I meant  1) retrieve bug info, 2) retrieve history info 3) retrieve comment info  4) retrieve attachment info are done Sequentially one after another.

bugbug/bugzilla.py
rajeshkumargp: Updated the source code by replacing `Union` with `Iterable`

bugbug/bugzilla.py
rajeshkumargp: Fixed

scripts/bug_classifier.py
rajeshkumargp: Fixed

scripts/bug_classifier.py
rajeshkumargp: Fixed

scripts/similarity_query.py
rajeshkumargp: Removed unnecessary type conversion here

bugbug/bugzilla.py
rajeshkumargp: Resolved by adding `Dict[int, dict]` as return type

bugbug/bugzilla.py
marco-c: If those are the only ones, I think we can avoid the cleanup altogether (so the code will be simpler).

bugbug/bugzilla.py
marco-c: I don't think that's the case, we were retrieving 1, 2, 3, and 4 in parallel. Why do you say they were retrieved in sequence?

rajeshkumargp: Bugzilla(..).get_data().wait() - call is propagated as below:

https://github.com/mozilla/libmozdata/blob/master/libmozdata/bugzilla.py#L152-L169

In this function we can find sequential retrieval of

1. Bug Information - https://github.com/mozilla/libmozdata/blob/master/libmozdata/bugzilla.py#L159
which calls
https://github.com/mozilla/libmozdata/blob/master/libmozdata/bugzilla.py#L556

2. History Information ->           https://github.com/mozilla/libmozdata/blob/master/libmozdata/bugzilla.py#L161
3. Comments -> https://github.com/mozilla/libmozdata/blob/master/libmozdata/bugzilla.py#L163
4. Attachments -> https://github.com/mozilla/libmozdata/blob/master/libmozdata/bugzilla.py#L165

Please let me know if my understanding is incorrect.

bugbug/bugzilla.py
rajeshkumargp: Removed the codes for cleanup activity of Comments fields. Fixed

bugbug/bugzilla.py
marco-c: Those functions are triggering requests, but not waiting for them to finish (they add futures to `self.bugs_results`, `self.history_results`, and so on). Then in `wait` all of the futures are awaited.

rajeshkumargp: Yes ,  Futuresession is used. https://github.com/mozilla/libmozdata/blob/b08f2cb144e1db97ce9113160695b6477df28b45/libmozdata/connection.py#L73 .

Thanks for clarification.
The retrieval is being done asynchrously + parallely.

marco-c:
The patch was reducing the parallelism, so it isn't clear if it would improve things.

PR #1631, opened 2020-06-10
Make spawn_pipeline not depend on the order of tasks in the yaml file

### LINKED ISSUE #1282: Make spawn_pipeline not depend on the order of tasks in the yaml file
Currently, if a task is defined in the yaml file before its dependencies, the spawn_pipeline script fails with:
```
  File "/code/spawn_pipeline.py", line 132, in <module>
    main()
  File "/code/spawn_pipeline.py", line 110, in main
    new_dependencies.append(id_mapping[dependency])
KeyError: 'regressor-finder'
```

So things like https://github.com/mozilla/bugbug/commit/aaa67b3b0a1db7530cbf88df644aff076fcd2e4e are needed.

We should make the spawn_pipeline script not depend on the order of definition of tasks in the yaml file.

armsarun:
Added id_mapping dictionary creation before checking the dependencies of the task

Fixes #1282


infra/spawn_pipeline.py
```diff
@@ -85,13 +85,19 @@ def main():
         # mapping between an internal ID and the generate ID
 
         task_id = taskcluster.utils.slugId()
-        task_internal_id = task.pop("ID")
-
+        task_internal_id = task.get("ID")
```
marco-c, line 88:
```suggestion
        task_internal_id = task["ID"]
```

Also nit: don't remove the newline

infra/spawn_pipeline.py
```diff
@@ -85,13 +85,19 @@ def main():
...
         if task_internal_id in id_mapping:
             raise ValueError(f"Conflicting IDs {task_internal_id}")
 
+        # add each task id to id_mapping dictionary
+        # before process the task dependencies
```
marco-c, line 93:
```suggestion
        # Store each task ID in the id_mapping dictionary before processing dependencies.
        # This way, tasks can be defined in any order.
```

infra/spawn_pipeline.py
```diff
@@ -85,13 +85,19 @@ def main():
...
         id_mapping[task_internal_id] = task_id
 
+    for task in rendered["tasks"]:
+
```
marco-c, line 97: Nit: no newline between `for` and the first line of the block (`task_internal_id = ...`).

infra/spawn_pipeline.py
armsarun: I'm not sure ID  key will be available always in task dictionary. So I used get instead of direct key access. I'll update this

infra/spawn_pipeline.py
```diff
@@ -85,13 +85,20 @@ def main():
         # mapping between an internal ID and the generate ID
 
         task_id = taskcluster.utils.slugId()
-        task_internal_id = task.pop("ID")
 
+        task_internal_id = task["ID"]
```
marco-c, line 89: Nit: put this in the same line where it was before, so the diff is smaller

Issue #1616, opened 2020-06-04
Store compressed results in Redis


marco-c:
This would reduce the size of the data in Redis, and allow us to cache more results.

See also #1532.

marco-c:
This would be nice to have for https://phabricator.services.mozilla.com/D85275, to ensure we have as many results as possible in the cache.

PR #1632, opened 2020-06-10
Compress HTTP service responses with gzip

### LINKED ISSUE #1532: Compress HTTP service responses
The JSON responses could be gzipped or brotlified.

rajeshkumargp:
Compress reponses whose status_code is 2XX

Fixes #1532


http_service/bugbug_http/app.py
```diff
@@ -284,6 +285,28 @@ def get_result(job):
...
+    :return: response with gzip compressed data
+    :rtype: Response
+    """
+
+    gzip_buffer = gzip.compress(orjson.dumps(data), compresslevel=9)
+
+    response = Response(status=status_code)
+    response.set_data(gzip_buffer)
+    response.headers["Content-Encoding"] = "gzip"
+    response.headers["Content-Length"] = len(response.get_data())
```
marco-c, line 304: Nit: You can probably just do:
```suggestion
    response.headers["Content-Length"] = len(gzip_buffer)
```

rajeshkumargp: Fixed

rajeshkumargp:
While running `python -m pytest http_service/tests/`,  we are getting responses (as in https://github.com/mozilla/bugbug/blob/master/http_service/tests/test_bug_classification.py#L29), which belongs to <class 'flask.wrappers.Response'>.  
When using `requests` for getting reponses, we will get `<class 'requests.models.Response'>` object. 

The object of `<class 'requests.models.Response'>` decompresses automatically, and retrieval of  gzipped responses is fine. Please refer https://github.com/mozilla/bugbug/pull/1632#discussion_r438336116 

Accordingly, the test cases referring `json` conversion is changed.

https://github.com/mozilla/bugbug/blob/master/http_service/tests/test_bug_classification.py#L31
https://github.com/mozilla/bugbug/blob/master/http_service/tests/test_bug_classification.py#L36
https://github.com/mozilla/bugbug/blob/master/http_service/tests/test_bug_classification.py#L45

https://github.com/mozilla/bugbug/blob/master/http_service/tests/test_push_schedules.py#L32


In other test responses, gzip compression is not done, So the test cases were failed earlier because they are different from responses of `<class 'flask.wrappers.Response'>` is different from <class 'requests.models.Response'>` object, when `response.json`  is called.

The code is updated now.

http_service/bugbug_http/app.py
rajeshkumargp: Resolved

marco-c:
@rajeshkumargp you might be interested in #1616, it is kind of similar ;)

rajeshkumargp:
> @rajeshkumargp you might be interested in #1616, it is kind of similar ;)

Sure, @marco-c

PR #1674, opened 2020-07-05
Add tests for get commit function and add additional mock commits

### LINKED ISSUE #383: Add tests for repository.get_commits

luckyvs1:
- Add tests for get commit function 
- Add additional mock commit in test fixtures to test parameters in get commit
- Closes #383


tests/test_repository.py
```diff
@@ -1770,3 +1770,53 @@ def test_get_touched_functions():
...
+    #   include_ignored: bool = False
+
+    # 10 mock commits, 1 ignored, 1 backouts, 1 no_bug, 1 no_bug and ignored, 1 no_bug, ignored, and backouts
+    retrieved_commits = list(repository.get_commits())
+    assert len(retrieved_commits) == 5
+    assert retrieved_commits[0]["node"] == "9d576871fd33bed006dcdccfba880a4ed591f870"
+
+    retrieved_commits = list(repository.get_commits(include_backouts=True))
+    assert len(retrieved_commits) == 6
+    assert retrieved_commits[5]["node"] == "ec01c146f756b74d18e4892b4fd3aecba00da93e"
```
marco-c, line 1788: This (and similar) assertions might start failing if we change the order of the commits in the mock file, you could do:
```
assert "ec01c146f756b74d18e4892b4fd3aecba00da93e" in {c["node"] for c in retrieved_commits}
```

tests/test_repository.py
```diff
@@ -1770,3 +1770,53 @@ def test_get_touched_functions():
...
+def test_get_commits():
+    # By default get_commits utilizes the following parameters:
+    #   include_no_bug: bool = False
+    #   include_backouts: bool = False
+    #   include_ignored: bool = False
+
+    # 10 mock commits, 1 ignored, 1 backouts, 1 no_bug, 1 no_bug and ignored, 1 no_bug, ignored, and backouts
+    retrieved_commits = list(repository.get_commits())
+    assert len(retrieved_commits) == 5
+    assert retrieved_commits[0]["node"] == "9d576871fd33bed006dcdccfba880a4ed591f870"
```
marco-c, line 1784: Assert that the commits which should not be there are not there, e.g.:
```Python
assert "ec01c146f756b74d18e4892b4fd3aecba00da93e" not in {c["node"] for c in retrieved_commits}
```

luckyvs1:
Thanks for the feedback @marco-c I have updated the PR based on the feedback

tests/test_repository.py
```diff
@@ -1770,3 +1770,93 @@ def test_get_touched_functions():
...
+        "ec01c146f756b74d18e4892b4fd3aecba00da93e",
+        "75276e64701bfde7cf2dd1f851adfea6a92d5747",
+        "10c9e57fb4f8c70d4772a7809ba3d652536b984d",
+    ]
+
+    # 10 mock commits, 1 ignored, 1 backouts, 1 no_bug, 1 no_bug and ignored, 1 no_bug, ignored, and backouts
+    assert len(retrieved_commits) == 5
+
+    for commit in excluded_commits:
+        assert commit not in {c["node"] for c in retrieved_commits}
```
marco-c, line 1796: Nit: you can rewrite this as:
```Python
assert not any(excluded_commit in {c["node"] for c in retrieved_commits} for excluded_commit in excluded_commits)
```

tests/test_repository.py
```diff
@@ -1770,3 +1770,93 @@ def test_get_touched_functions():
...
+    retrieved_commits = list(
+        repository.get_commits(include_ignored=True, include_backouts=True)
+    )
+
+    assert len(retrieved_commits) == 7
+
+    assert "7f27080ffee35521c42fe9d4025caabef7b6258c" in {
+        c["node"] for c in retrieved_commits
+    }
+
```
marco-c, line 1831: Also assert the backout commit is in there too

tests/test_repository.py
```diff
@@ -1770,3 +1770,93 @@ def test_get_touched_functions():
...
+    assert len(retrieved_commits) == 5
+
+    for commit in excluded_commits:
+        assert commit not in {c["node"] for c in retrieved_commits}
+
+    retrieved_commits = list(repository.get_commits(include_backouts=True))
+
+    assert len(retrieved_commits) == 6
+
+    assert "ec01c146f756b74d18e4892b4fd3aecba00da93e" in {
```
marco-c, line 1802: Add a comment on all assertions like this mentioning what commit it is, e.g.:
```Python
# Ensure the backout commit was retrieved.
assert "ec01c146f756b74d18e4892b4fd3aecba00da93e" in {
    c["node"] for c in retrieved_commits
}
```

tests/test_repository.py
marco-c: Similarly with all other cases

tests/test_repository.py
marco-c: Another option, at the top of the function define the commits as:
```
BACKOUT_COMMIT = "ec01c146f756b74d18e4892b4fd3aecba00da93e"
```

And then change the assertions to be:
```Python
assert BACKOUT_COMMIT in {
    c["node"] for c in retrieved_commits
}
```

This will make things clearer

luckyvs1:
Thank you for the feedback @marco-c, I just had a chance to address them and re-requested a review

tests/test_repository.py
```diff
@@ -1770,3 +1770,87 @@ def test_get_touched_functions():
...
+    assert NOBUG_COMMIT in (c["node"] for c in retrieved_commits)
+
+    retrieved_commits = list(
+        repository.get_commits(include_ignored=True, include_backouts=True)
+    )
+    included_commits = [IGNORED_COMMIT, BACKOUT_COMMIT]
+    assert len(retrieved_commits) == 7
+    assert all(
+        {c["node"] for c in retrieved_commits} for included_commit in included_commits
+    )
```
marco-c, line 1821: This assertion is not right, it is just asserting that `{c["node"] for c in retrieved_commits}` is not empty.
It should instead be:
```Python
assert all(included_commit in {c["node"] for c in retrieved_commits} for included_commit in included_commits)
```

Or better, if you make `included_commits` a set instead of a list (`included_commits = {IGNORED_COMMIT, BACKOUT_COMMIT}`), you can just do:
```Python
assert included_commits.issubset({c["node"] for c in retrieved_commits})
```

The same applies to the other assertions below.

luckyvs1: I've updated this and the following assertions to use `issubset` and thank you for the correction on the assert using the all syntax.

PR #1678, opened 2020-07-06
Improvements to the bug type model

marco-c:


bugbug/models/bug_type_classification.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+        self.calculate_importance = False
+
+        feature_extractors = [
+            bug_features.has_str(),
+            bug_features.severity(),
+            # Ignore keywords that would make the ML completely skewed
+            # (we are going to use them as 100% rules in the evaluation phase).
+            bug_features.keywords(set(self.label_names)),
```
marco-c, line 88: We need to exclude the keywords from https://github.com/mozilla/bugbug/blob/72659c4e38255b2b7bc31cf2525080b4e6cfbb15/bugbug/models/bugtype.py#L16-L31.
The strings in `self.label_names` don't actually exist as keywords

bugbug/models/bug_type_classification.py
```diff
@@ -0,0 +1,191 @@
...
+            bug_features.is_coverity_issue(),
+            bug_features.has_crash_signature(),
+            bug_features.has_url(),
+            bug_features.has_w3c_url(),
+            bug_features.has_github_url(),
+            bug_features.whiteboard(),
+            bug_features.patches(),
+            bug_features.landings(),
```
marco-c, line 96: Let's remove `patches` and `landings`, they can't possibly be useful.

bugbug/models/bug_type_classification.py
```diff
@@ -0,0 +1,191 @@
...
+            bug_features.blocked_bugs_number(),
+            bug_features.ever_affected(),
+            bug_features.affected_then_unaffected(),
```
marco-c, line 99: `ever_affected` and `affected_then_unaffected` too

bugbug/models/bug_type_classification.py
```diff
@@ -0,0 +1,191 @@
...
+                assert category in [self.single_class_name, "Other"], f"unexpected category {category}"
+                classes[int(bug_id)] = 1 if category == self.single_class_name else 0
+            return classes, [0, 1]
+
+
+    def get_feature_names(self):
+        return self.extraction_pipeline.named_steps["union"].get_feature_names()
+
+
+    def overwrite_classes(self, bugs, classes, probabilities):
```
marco-c, line 173: In this function we'll have to do something similar to https://github.com/mozilla/bugbug/blob/72659c4e38255b2b7bc31cf2525080b4e6cfbb15/bugbug/models/bugtype.py#L115 (again, the strings we use as labels are not available as keywords).
For now, let's just remove this function completely.

bugbug/models/bug_type_classification.py
```diff
@@ -0,0 +1,191 @@
...
+        if self.single_class is not None:
+            assert self.single_class is not None, f"unexpected class\n\tUse the args --single_class with one of the following {self.label_names}"
+            assert self.single_class in self.label_names, f"unexpected class {self.single_class}\n\tUse the args --single_class with one of the following {self.label_names}"
+            self.single_class_name = self.single_class
+            self.label_names = [self.single_class_name, "Other"]
+
+
+        self.target_size = len(self.label_names)
+
+        self.list_labels = self.label_names
```
marco-c, line 73: Nit: no need to define a variable that is simply mirroring another

bugbug/models/bug_type_classification.py
```diff
@@ -0,0 +1,191 @@
...
+        # Set the class mode ( Single class VS All the category )
+        self.single_class = single_class
+
+        # Add the sub-classes to the classification list
+        if self.mode == True:
+            self.label_names = self.label_names + label_names_sub_classes
+
+        # Check if the element is based over the single class
+        if self.single_class is not None:
+            assert self.single_class is not None, f"unexpected class\n\tUse the args --single_class with one of the following {self.label_names}"
```
marco-c, line 65: Nit: this assert always passes, since you are asserting the same condition of the if

bugbug/models/bug_type_classification.py
```diff
@@ -0,0 +1,208 @@
...
+                classes[int(bug_id)] = 0
+                if self.mode:
+                    for keyword in bug_labels[3:]:
+                        if keyword != '' and keyword == self.single_class:
+                            classes[int(bug_id)] = 1
+                else:
+                    for keyword in bug_labels[:3]:
+                        if keyword != '' and keyword == self.single_class:
+                            classes[int(bug_id)] = 1
+
```
marco-c, line 182: Let's add some logging as in https://github.com/mozilla/bugbug/blob/0b78e000e1d5493a88ed06afb40267ed3c6ef6d4/bugbug/models/defect_enhancement_task.py#L18-L30 so we have an idea of the number of examples for each class.

bugbug/models/bug_type_classification.py
```diff
@@ -0,0 +1,208 @@
...
+                            ("data", DictVectorizer(), "data"),
+                            ("title", self.text_vectorizer(min_df=0.001), "title"),
+                            (
+                                "first_comment",
+                                self.text_vectorizer(min_df=0.001),
+                                "first_comment",
+                            ),
+                            (
+                                "comments",
+                                self.text_vectorizer(min_df=0.001),
```
marco-c, line 129: Try to experiment with different values of min_df (0.01 or 0.0001) and see how the results change

bugbug/models/bug_type_classification.py
```diff
@@ -0,0 +1,208 @@
...
+                    bug_features.BugExtractor(feature_extractors, cleanup_functions),
+                ),
+                (
+                    "union",
+                    ColumnTransformer(
+                        [
+                            ("data", DictVectorizer(), "data"),
+                            ("title", self.text_vectorizer(min_df=0.001), "title"),
+                            (
+                                "first_comment",
```
marco-c, line 123: Try removing the first_comment from the features, I'm not sure for this model it makes sense to consider it separately than the rest of the comments.

bugbug/models/bug_type_classification.py
```diff
@@ -0,0 +1,223 @@
...
+                        elif keyword != '' and keyword != self.single_class:
+                            categories["Other"] = categories["Other"] + 1
+
+        # Some logging so we have an idea of the number of examples for each class
+        if self.single_class is None:
+            for lbl in self.label_names:
+                print(str(categories[lbl]) + " " + lbl)
+        else:
+            print(str(categories[self.single_class]) + " " + self.single_class)
+            print(str(categories["Other"]) + " Other")
```
marco-c, line 198: Instead of introducing `categories` to keep track of the number, you can do something like:
```
        for value in set(classes.values()):
            print(
                "{} {}".format(
                    sum(1 for label in classes.values() if label == value), value
                )
            )
```

(this also avoids a bug you currently have that makes you overcount the "Other": when there are more than one class for a bug, you add 1 both to `categories[self.single_class]` and to `categories["Other"]`)

requirements.txt
```diff
@@ -1,13 +1,13 @@
 adr==0.19.2
-boto3==1.16.10
+boto3==1.16.11
```
marco-c, line 2: These changes should not be there

scripts/trainer.py
```diff
@@ -37,6 +37,7 @@ def go(self, args):
             model_name = args.model
 
         model_class = get_model_class(model_name)
+
```
marco-c, line 40: Nit: remove extraneous new line

bugbug/models/__init__.py
```diff
@@ -29,6 +29,7 @@
     "testfailure": "bugbug.models.testfailure.TestFailureModel",
     "tracking": "bugbug.models.tracking.TrackingModel",
     "uplift": "bugbug.models.uplift.UpliftModel",
+    "bugtypeclassification": "bugbug.models.bug_type_classification.BugTypeClassificationModel",
```
marco-c, line 32: We should merge this with the pre-existing bugtype model (just overwrite bugbug/models/bugtype.py with the contents from bugbug/models/bug_type_classification.py, and then remove bugbug/models/bug_type_classification.py).

bugbug/model.py
```diff
@@ -455,16 +718,32 @@ def train(self, importance_cutoff=0.15, limit=None):
...
+            for num, array in enumerate(confusion_matrix):
+                df_cm = pd.DataFrame(
+                    array,
+                    index=[self.class_names[num], "Other"],
+                    columns=[self.class_names[num], "Other"],
+                )
+                sn.set(font_scale=1)
+                sn.heatmap(df_cm, annot=True, annot_kws={"size": 16})  # font size
+                plt.title(self.class_names[num] + " Confusion Matrix")
+                plt.show()
```
marco-c, line 745: Can you show me the output of this? Is it that different from the command line output from `print_labeled_confusion_matrix`?
If it isn't, we can remove this code and remove the seaborn dependency

bugbug/model.py
```diff
@@ -340,7 +354,239 @@ def get_labels(self) -> Tuple[Dict[Any, Any], List[Any]]:
...
+        # GridSearchCV
+        model_tunning = GridSearchCV(self.clf, param_grid=self.parameters)
+
+        model_tunning.fit(X.toarray(), y)
+
+        print(
+            f"GridSearch for CLASSIFIER {self.type_classifier}\n best score: {model_tunning.best_score_}\nbest params: {model_tunning.best_params_}"
+        )
+
+    def train_compact_statistics(
```
marco-c, line 382: Why did you introduce this alternative?

bugbug/model.py
```diff
@@ -340,7 +354,239 @@ def get_labels(self) -> Tuple[Dict[Any, Any], List[Any]]:
         """Subclasses implement their own function to gather labels."""
         pass
 
-    def train(self, importance_cutoff=0.15, limit=None):
+    def train_with_gridserach(self):
```
marco-c, line 357: I like the idea of adding a method to do grid search, but it should be part of a separate PR.
Please remove all the grid search related code, and let's add it with a separate PR (and branch).

PR #1713, opened 2020-07-21
Set up Sentry for bugbug's background worker

### LINKED ISSUE #1698: Set up Sentry for bugbug
Adding it will help investigate issues (like the Mercurial woes).

armenzg:
Fixes #1698

### DIFF
--- a/http_service/bugbug_http/worker.py
+++ b/http_service/bugbug_http/worker.py
@@ -7,10 +7,20 @@
 import os
 import sys
 
+import sentry_sdk
 from redis import Redis
 from rq import Connection, Worker
+from sentry_sdk.integrations.rq import RqIntegration
 
 import bugbug_http.boot
+from bugbug import get_bugbug_version
+
+if os.environ.get("SENTRY_DSN"):
+    sentry_sdk.init(
+        os.environ.get("SENTRY_DSN"),
+        integrations=[RqIntegration()],
+        release=get_bugbug_version(),
+    )
 
 
 def main():
--- a/http_service/requirements.txt
+++ b/http_service/requirements.txt
@@ -10,3 +10,4 @@ marshmallow==3.7.1
 requests==2.24.0
 rq==1.4.3
 rq-dashboard==0.6.1
+sentry-sdk==0.16.1
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -3,4 +3,4 @@ multi_line_output=3
 include_trailing_comma=True
 line_length=88
 known_first_party = bugbug,bugbug_http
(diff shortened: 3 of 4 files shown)


armenzg:
@marco-c r?

http_service/bugbug_http/worker.py
```diff
@@ -7,11 +7,15 @@
...
 import sys
 
+import sentry_sdk
 from redis import Redis
 from rq import Connection, Worker
 
 import bugbug_http.boot
 
+if os.environ.get("SENTRY_DSN"):
+    sentry_sdk.init(os.environ.get("SENTRY_DSN"))
```
marco-c, line 17: In code-review bot, we set `release=VERSION`, so we can associate errors to versions, it would be beneficial here too

armenzg:
@marco-c r?

This can be tested locally by following these steps:
```shell
export SENTRY_DSN=<value_from_sentry>
// Comment out RUN bash /code/http_service/ensure_models.sh
// This will help build the image fast and upon execution there will be an AssertionError
docker-compose build
docker-compose run -e SENTRY_DSN bugbug-http-service-bg-worker
```

image

http_service/requirements.txt
```diff
@@ -10,3 +10,4 @@ marshmallow==3.7.1
 requests==2.24.0
 rq==1.4.3
 rq-dashboard==0.6.1
+sentry-sdk==0.10.2
```
marco-c, line 13: Why not the latest version of sentry-sdk?

marco-c:
For the RQ integration, should we use the `--sentry-dsn=""` option? From https://docs.sentry.io/platforms/python/rq/, it seems it is needed for RQ < 1.0, but still suggested in general.

http_service/requirements.txt
armenzg: When setting up a new project via IT's deployment of Sentry it showed that version. I've fixed it. Thanks for catching it!

armenzg:
> For the RQ integration, should we use the `--sentry-dsn=""` option? From https://docs.sentry.io/platforms/python/rq/, it seems it is needed for RQ < 1.0, but still suggested in general.

It's a bit unclear from the documentation. `--sentry-dsn=""` is to prevent RQ to report to Sentry according to their original implementation and it is mostly for the CLI invokation which we don't do (We import the library and call Worker() directly).

The CLI does this `sentry_dsn = sentry_dsn or settings.get('SENTRY_DSN')` which then does:
```python
       if sentry_dsn:
            from rq.contrib.sentry import register_sentry
            register_sentry(sentry_dsn)
```
which then does:
```python
def register_sentry(sentry_dsn):
    """Given a Raven client and an RQ worker, registers exception handlers
    with the worker so exceptions are logged to Sentry.
    """
    import sentry_sdk
    from sentry_sdk.integrations.rq import RqIntegration
    sentry_sdk.init(sentry_dsn, integrations=[RqIntegration()])
```

In other words, it is not necessary since we don't call the `rq` binary (as far as I can tell).

armenzg:
@marco-c do you believe my answer helps?

marco-c:
> @marco-c do you believe my answer helps?

Yes, thanks!

PR #1716, opened 2020-07-23
Raise exceptions to track in Sentry

armenzg:
This fixes reporting of autoland cloning failures.

### DIFF
--- a/http_service/bugbug_http/models.py
+++ b/http_service/bugbug_http/models.py
@@ -129,7 +129,7 @@ def schedule_tests(branch: str, rev: str) -> str:
         repository.pull(REPO_DIR, branch, rev)
     except Exception as e:
         LOGGER.warning(f"Failed to pull {branch} @ {rev}: {e}")
-        return "NOK"
+        raise
 
     test_selection_threshold = float(
         os.environ.get("TEST_SELECTION_CONFIDENCE_THRESHOLD", 0.5)


http_service/bugbug_http/models.py
```diff
@@ -121,15 +121,15 @@ def schedule_tests(branch: str, rev: str) -> str:
         revs = get_hgmo_stack(branch, rev)
     except requests.exceptions.RequestException:
         LOGGER.warning(f"Push not found for {branch} @ {rev}!")
-        return "NOK"
+        raise
```
marco-c, line 124: In this case we should keep returning "NOK", if the push can't be found it's not useful to retry

http_service/bugbug_http/boot.py
```diff
@@ -97,10 +97,10 @@ def extract_commit_experiences() -> None:
             utils.extract_file(os.path.join("data", repository.COMMIT_EXPERIENCES_DB))
             logger.info("Commit experiences DB extracted.")
         except FileNotFoundError:
+            assert ALLOW_MISSING_MODELS
```
marco-c, line 100: Can you make this change in a separate PR?

http_service/bugbug_http/models.py
armenzg: This was left unstaged. Fixed now.

http_service/bugbug_http/boot.py
armenzg: Done https://github.com/mozilla/bugbug/pull/1720

PR #1722, opened 2020-07-24
Submit logged warnings to Sentry

### LINKED ISSUE #1721: Submit logged warnings to Sentry
Currently, we only report exceptions to Sentry.
It'd be nice to report warnings too, so we can get a sense of how often they happen.
IIRC in the code-review bot (https://github.com/mozilla/code-review), we are doing that kind of automatically.

armenzg:
Fixes #1721

Sentry by default logs `ERROR` messages as events, however, we would
like to track `WARNING` messages as well.

### DIFF
--- a/http_service/bugbug_http/worker.py
+++ b/http_service/bugbug_http/worker.py
@@ -4,21 +4,29 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
+import logging
 import os
 import sys
 
 import sentry_sdk
 from redis import Redis
 from rq import Connection, Worker
+from sentry_sdk.integrations.logging import LoggingIntegration
 from sentry_sdk.integrations.rq import RqIntegration
 
 import bugbug_http.boot
 from bugbug import get_bugbug_version
 
 if os.environ.get("SENTRY_DSN"):
+    logging_integration = LoggingIntegration(
+        # Default behaviour: INFO messages will be included as breadcrumbs
+        level=logging.INFO,
+        # Change default behaviour (ERROR messages events)
+        event_level=logging.WARNING,
+    )
     sentry_sdk.init(
-        os.environ.get("SENTRY_DSN"),
-        integrations=[RqIntegration()],
+        dsn=os.environ.get("SENTRY_DSN"),
+        integrations=[RqIntegration(), logging_integration],
         release=get_bugbug_version(),
     )
 


armenzg:
@marco-c have a look at [Sentry](https://sentry.prod.mozaws.net/operations/bugbug) and this comment to see if you're satisfied with the results. 

[The documentation](https://docs.sentry.io/platforms/python/logging/) says that some logging integration happens already by default.

> event_level (default ERROR): The Sentry Python SDK will report log records with a level higher than or equal to event_level as events. If a value of None occurs, the SDK won’t send log records as events.

```python
# All of this is already happening by default!
sentry_logging = LoggingIntegration(
    level=logging.INFO,        # Capture info and above as breadcrumbs
    event_level=logging.ERROR  # Send errors as events
)
sentry_sdk.init(
    dsn="https://<key>@<organization>.ingest.sentry.io/<project>",
    integrations=[sentry_logging]
)
```

I assume you don't want the default behavior where `error` level messages get reported but you want `warning` level messages reported. The pull request reflects this change.

In short:

- Warning messages *will* create events
  - I assume **error** messages will also be included
- Exceptions will include stack traces

You can also include extra metadata with warning messages like this:
```python
logging.warning("I am an event", extra=dict(bar=43))
```

I tested this change locally with these steps and following diff:

```shell
export SENTRY_DSN=<same value from Heroku>
docker-compose build && docker-compose run bugbug-http-service-bg-worker
```

```diff
diff --git a/http_service/Dockerfile.bg_worker b/http_service/Dockerfile.bg_worker
index df4ab3d..62a1a35 100644
--- a/http_service/Dockerfile.bg_worker
+++ b/http_service/Dockerfile.bg_worker
@@ -19,6 +19,6 @@ ENV CHECK_MODELS="${CHECK_MODELS}"
 ARG TAG
 ENV TAG="${TAG}"

-RUN bash /code/http_service/ensure_models.sh
+# RUN bash /code/http_service/ensure_models.sh

 CMD bugbug-http-worker high default low
diff --git a/http_service/bugbug_http/worker.py b/http_service/bugbug_http/worker.py
index bb77c08..06f3bd8 100644
--- a/http_service/bugbug_http/worker.py
+++ b/http_service/bugbug_http/worker.py
@@ -17,6 +17,8 @@ from sentry_sdk.integrations.rq import RqIntegration
 import bugbug_http.boot
 from bugbug import get_bugbug_version

+logger = logging.getLogger(__name__)
+
 if os.environ.get("SENTRY_DSN"):
     logging_integration = LoggingIntegration(
         # Default behaviour: INFO messages will be included as breadcrumbs
@@ -29,6 +31,10 @@ if os.environ.get("SENTRY_DSN"):
         integrations=[RqIntegration(), logging_integration],
         release=get_bugbug_version(),
     )
+    sentry_sdk.capture_exception(Exception("This is an example of an error message."))
+    logger.info("This is a breadcrumb")
+    logger.warning('Foo', extra=dict(bar=43))
+    sys.exit(1)


 def main():
```

You can see the `extra` data in Sentry:
image

You can see a `breadcrumb` (info messages) in here:
image

Issue #945, opened 2019-09-12
For binary models, plot and check the classes separately


marco-c:
For binary models, we are particularly interested in seeing the metrics per class and not only on average.

scripts/analyze_training_metrics.py is currently plotting and checking the "average", but for binary classes we want to plot and check the two classes separately.

To test the script, first run scripts/retrieve_training_metrics.py to retrieve the metrics files.

kaushiksk:
Can you elaborate a bit on this, and direct me towards which files I need to be looking into?

datavizard:
Is anybody working on this? If not, I'd be happy to help, given some more context of the problem.

marco-c:
@datavizard read the contributing rules, if there is no PR attached to an issue, it is unassigned. So, feel free to work on this (@kaushiksk feel free to work on this too, I had missed your message).
I've added more details in the first comment.

rishishounak:
I would like to work on this but I dont have any previous experience so please guide me

marco-c:
Hello @rishishounak, you can follow the steps in the README to set the repository up, then you can run the scripts/retrieve_training_metrics.py script to make sure it works for you. Once everything is set up correctly, you can try to apply the required change (explained in the first comment of this issue) to the scripts/retrieve_training_metrics.py file.

PR #1824, opened 2020-09-27
Plot and check classes separately for binary models. #945

yuyan9:
This analyzes the two classes separately for binary models rather than just checking the "average".

edit: #945 
Sorry, I didn't realize the issue number had to be in the comment.

### DIFF
--- a/scripts/analyze_training_metrics.py
+++ b/scripts/analyze_training_metrics.py
@@ -125,6 +125,15 @@ def analyze_metrics(
         date, model_name, metric = parse_metric_file(metric_file_path)
 
         # Then process the report
+        is_binary = len(metric["report"]["targets"]) == 2
+        if is_binary:
+            for target, target_metrics in metric["report"]["targets"].items():
+                for key, value in target_metrics.items():
+                    if key not in REPORT_METRICS:
+                        continue
+
+                    metrics[model_name][f"{key}_class_{target}"][date] = value
+
         for key, value in metric["report"]["average"].items():
             if key not in REPORT_METRICS:
                 continue


scripts/analyze_training_metrics.py
```diff
@@ -85,12 +85,12 @@ def parse_metric_file(metric_file_path: Path) -> Tuple[datetime, str, Dict[str,
...
 
-    assert file_path_parts[:5] == ["metric", "project", "bugbug", "train"]
-    model_name = file_path_parts[5]
-    assert file_path_parts[6:8] == ["per", "date"]
-    date_parts = list(map(int, file_path_parts[8:14]))
+    file_path_parts = metric_file_path.stem.split("_")
+    assert file_path_parts[:4] == ["metric", "project", "bugbug", "train"]
+    model_name = file_path_parts[4]
+    assert file_path_parts[5:7] == ["per", "date"]
+    date_parts = list(map(int, file_path_parts[7:13]))
```
marco-c, line 93: Could you make a separate PR with this fix? I prefer to have self-contained PRs.

scripts/analyze_training_metrics.py
```diff
@@ -124,7 +124,16 @@ def analyze_metrics(
 
         date, model_name, metric = parse_metric_file(metric_file_path)
 
+        is_binary = len(metric["report"]["targets"]) == 2
         # Then process the report
+        if is_binary:
+            for target in range(2):
```
marco-c, line 130: Instead of "hardcoding" the class names like this, you could do `for target in metric["report"]["targets"]:` or even better `for target, target_metrics in metric["report"]["targets"].items():`, so in the next line you can do `for key, value in target_metrics.items():` instead of `for key, value in metric["report"]["targets"][str(target)].items():`.

marco-c:
Thanks, looks good overall, just a couple of minor things :)

scripts/analyze_training_metrics.py
yuyan9: Thank you for the tip!

scripts/analyze_training_metrics.py
yuyan9: Ok done.

marco-c: You'll now need to rebase this PR on top of current master (`git pull upstream master --rebase`).

marco-c: It looks like it didn't go well, https://github.com/mozilla/bugbug/pull/1824/files still shows the changes from the other PR.

yuyan9: Ok, got it now I think!

marco-c (approved):
Thanks!

PR #1826, opened 2020-09-27
Fix index error to accept metric files

yuyan9:
This fixes an index error in the parse_metric_file function which prevented it from accepting the metric files as mentioned in #1824.


marco-c (approved):

marco-c:
Thanks!

PR #1958, opened 2020-12-01
Fix code quality issues

withshubh:
This pull request fixes some of the code quality issues raised by DeepSource on my fork of this repository. I have already fixed some issues using DeepSource's Autofix.

Take a quick look at all the issues caught by DeepSource for this repository [here](https://deepsource.io/gh/withshubh/bugbug/issues/?category=recommended).

### Summary of fixes

- Refactor `if` expression
- Remove redundant `None` default
- Merge `isinstance` calls
- Iterate dictionary directly
- ~Pass string format arguments as logging method parameters~
- ~Remove length check in favour of truthiness of the object~
- ~Use literal syntax instead of function calls to create data structure~
- Remove unnecessary use of comprehension

You can also have a look at the [configuration file](https://github.com/withshubh/bugbug/blob/deepsource/.deepsource.toml) I used for DeepSource Analysis.


withshubh:
Hi :wave: @bgrins @marco-c 
I'd request you to please review the PR.

marco-c:
Thanks @withshubh!

I agree with most of the changes, except:
- Use literal syntax instead of function calls to create data structure
- Remove length check in favour of truthiness of the object
- Pass string format arguments as logging method parameters (I'd prefer f-strings instead, or `.format` when it's not possible to use f-strings)
- All of our checkers are defined in a pre-commit configuration so they run at commit time. Is it possible to add the DeepSource checkers as part of pre-commit?

Are the checkers used in DeepSource open source?

withshubh:
Hi @marco-c :wave: 

> Use literal syntax instead of function calls to create data structure

- This fix helps in performance as the interpreter doesn't have to call a function.
> Remove length check in favour of truthiness of the object

- Because we are removing an unnecessary step of finding string length, an empty string has the same truthy value as len:0.
>Pass string format arguments as logging method parameters (I'd prefer f-strings instead, or .format when it's not possible to use f-strings)

- When string interpolation methods are used, they are always carried out, even when the message is not to be logged. Hence the issue was raised.
Example: If the severity of the logger is set to `warning` : this means we want the logger to log messages with severity >= `warning`. If there are `info` messages, they shouldn't be logged.
Now, if `f-string` or `.format` is used, string interpolation would be performed, and then the decision would be taken on what happens to the message. i.e. Interpolation would be performed for `info` messages here but would be thrown away because the logger's severity was set to `warning`.
> All of our checkers are defined in a pre-commit configuration so they run at commit time. Is it possible to add the DeepSource checkers as part of pre-commit?

- Not at the moment. We'll be bringing CLI soon that will run on uncommitted code.

I hope this answers all your questions. If you still don't want these changes then I can revert these fixes and update the PR. ✨

marco-c:
> This fix helps in performance as the interpreter doesn't have to call a function.
> Because we are removing an unnecessary step of finding string length, an empty string has the same truthy value as len:0.
> When string interpolation methods are used, they are always carried out, even when the message is not to be logged. Hence the issue was raised.

I think these a very minor differences in terms of performance, I prefer the readability in these cases.

> Not at the moment. We'll be bringing CLI soon that will run analysis on uncommitted code.

Looking forward for trying that out.

withshubh:
> > This fix helps in performance as the interpreter doesn't have to call a function.
> > Because we are removing an unnecessary step of finding string length, an empty string has the same truthy value as len:0.
> > When string interpolation methods are used, they are always carried out, even when the message is not to be logged. Hence the issue was raised.
> 
> I think these a very minor differences in terms of performance, I prefer the readability in these cases.
> 
> > Not at the moment. We'll be bringing CLI soon that will run analysis on uncommitted code.
> 
> Looking forward to trying that out.

Hey @marco-c 

If you prefer readability then I'll revert these changes.

withshubh:
Hey @marco-c :wave: 

Done with the changes!

marco-c (approved):
Thanks!

I've removed the `.deepsource.toml` configuration file as I don't intend to use it outside of pre-commit for now, the rest of the fixes looks good.

withshubh:
Hey, @marco-c I hope you're doing good :sparkles: 

Just wanted to share that DeepSource released [CLI](https://deepsource.io/docs/config/cli.html) last month. I would really appreciate it if you could try it out and share your feedback with us. : )

PR #1982, opened 2020-12-02
Update strings on testing page.

### LINKED ISSUE #1981: fix help text in testing graph ui for backeds and regressions
backouts: from "Note that this includes _all_ commits in a bug that gets backed out, so if there are a stack of commits this can overestimate tags." to  "Note that many times _all_ commits in a bug get backed out at once, so if there are a stack of commits this can overestimate tags. For instance, a commit might have been appropriately marked as `unchanged` but still get backed out."

regressions: from "Note that this includes _all_ commits in a bug that gets backed out, so if there are a stack of commits this can overestimate tags." to "Note that this will include _all_ commits in a bug that caused a regression, so if there are a stack of commits this can overestimate tags. For instance, a commit might have been appropriately marked as `unchanged` but still be marked as causing a regression."

jrbenny35:
Closes #1981

### DIFF
--- a/ui/changes/src/testing.html
+++ b/ui/changes/src/testing.html
@@ -49,8 +49,8 @@
           <h3>Backed Out Commits</h3>
         </summary>
         <div class="arewe-update arewe-update-warn">
-          Note that this includes _all_ commits in a bug that gets backed out,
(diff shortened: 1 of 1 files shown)


bgrins (approved):
Thanks!

Issue #1866, opened 2020-10-21
Classify multiple bugs at a time in the code to find bug-fixing commits


marco-c:
We are classifying one bug at a time in https://github.com/mozilla/bugbug/blob/22222328ad277b94f08ea4f4634cee87db0ad013/scripts/regressor_finder.py#L264, we could instead collect the bugs in a list and then classify them in one go.

The script can be easily tested by running `python -m scripts.regressor_finder bug_fixing`.

marco-c:
Hello! We don't need to sort them, we can classify them in any order.
You have an example classification of multiple items (commits in this case, but you can apply the same to bugs) here: https://github.com/mozilla/bugbug/blob/4bd73121bb86afe3d9a045ad97aea9a92fc71e21/scripts/generate_landings_risk_report.py#L277.

Abhishek5002443:
Hey this is my first time on Github and I am looking for my first contribution. Can you please tell me, what will I need to do exactly in this request?

marco-c:
Hello @Abhishek5002443, the description of what needs to be done is above. First of all follow https://github.com/mozilla/bugbug/issues/1092. Then, once everything is set-up, follow the pointers I gave above, try to understand that code, and then try to implement what was requested above. If you get stuck, feel free to ask questions.

LuiGiovanni:
@marco-c Hello is this issue being worked on? I was starting to set up everything and read the guidelines because I am interested in this being my first contribution, just wanted to make sure it wouldn't conflict with someone else.

willT97:
Hi I had a go fixing this issue, making the defect model classify in one go. Should the regression model also classify bugs at the same time?

marco-c:
> Hi I had a go fixing this issue, making the defect model classify in one go. Should the regression model also classify bugs at the same time?

If possible yes, can you file a new issue for that?

willT97:
> If possible yes, can you file a new issue for that?

Ok will do, created the new issue #2007 .

5aumy4:
Hello everyone!, I have a little past experience with open source contribution. Can I work on this bug?

marco-c:
@5aumy4 please read the introduction first; there is already a PR open to fix this issue, so you'd need to choose another issue to work on.

PR #2001, opened 2020-12-10
Classify multiple bugs at the same time #1866

willT97:
Hello there this is my first contribution.

I have modified the code to classify the bugs at the same time by appending them to a list.
I then iterate over that list and apply the same logic as before.


willT97:
For the issue #1866

scripts/regressor_finder.py
```diff
@@ -262,7 +263,15 @@ def append_bug_fixing_commits(bug_id: int, type_: str) -> None:
...
                 continue
 
-            if defect_model.classify(bug)[0] == "defect":
+            bugs_to_classify.append(bug)
+
+        classified_bugs = []
+        if bugs_to_classify:
+            classified_bugs = defect_model.classify(bugs_to_classify)
+
+        for i, label in enumerate(classified_bugs):
```
marco-c, line 272: Nit: you can do `for i, (bug, label) in enumerate(zip(bugs_to_classify, classified_bugs)):`
so you can skip `bug = bugs_to_classify[i]`

willT97: Thanks, I only use i for getting the bug so just removed the enumerate and kept the zip.

marco-c: Good idea

marco-c (approved):

marco-c:
Thanks!

Issue #2000, opened 2020-12-10
Improve the summary section of the changes view


marco-c:
A few ideas:
- [ ] Graph showing number of landings in each release or over time
- [x] Graph showing number of high/low/medium risk patches landing in each release or over time (#2002)
- [x] Graph showing number of bugs over time by type -> crashes / security bugs / etc in each release or over time (f9aeafdc9625c7dc5d5879326c3a9f4aa99ae7e9)
- [x] Graph showing number of regressions in each release or over time (64985d9fad726957c7046474f7c5d6fca7ff0511)

marco-c:
I'll call this fixed for now, we can file individual issues for further enhancements to the summary section.

PR #2002, opened 2020-12-11
Add a line chart to show the evolution of high/medium/low risk changes

marco-c:
Part of #2000

Screenshot_2020-12-11 mc-changes

### DIFF
--- a/ui/changes/src/common.js
+++ b/ui/changes/src/common.js
@@ -140,53 +140,53 @@ export let landingsData = (async function () {
   return returnedObject;
 })();
 
-export function getNewTestingTagCountObject() {
-  let obj = {};
-  for (let tag in TESTING_TAGS) {
-    obj[tag] = 0;
+export class Counter {
+  constructor() {
+    return new Proxy(
+      {},
+      {
+        get: (target, name) => (name in target ? target[name] : 0)
+      }
+    );
   }
-  return obj;
 }
 
-export async function getTestingPolicySummaryData(grouping = "daily", filter) {
-  let data = await landingsData;
+export function getSummaryData(
+  bugSummaries,
+  grouping = "daily",
+  startDate,
+  counter,
+  filter
+) {
+  let dates = new Set(bugSummaries.map(summary => summary.date));
 
-  // console.log(data);
-  // let startDate = grouping == "daily" ? "2020-09-15" : "2020-08-16";
-  let startDate = "2020-09-01";
   let dailyData = {};
-  for (let date in data) {
-    // Ignore data before the testing policy took place.
+  for (let date of dates) {
     if (
-      Temporal.PlainDate.compare(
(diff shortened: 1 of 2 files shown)


bgrins:
@marco-c can you clarify the meaning of low/medium/high here? I want to make sure I understand what this chart is telling me.

marco-c:
> @marco-c can you clarify the meaning of low/medium/high here? I want to make sure I understand what this chart is telling me.

Low -> Half than average risk of causing regressions (around 4%)
Medium -> Average risk of causing regressions (around 9%)
High -> Higher than average risk of causing regressions (around 18%)

It's probably not so useful at the level of the whole repo, it is more useful at the level of an individual team or an individual metabug.

bgrins:
Screen Shot 2020-12-11 at 8 06 51 PM

Do I need to do anything besides pull this down and `npm run dev`? I see an error when loading it locally (see screenshot)

bgrins:
Oh it looks like I have an empty "teams" filter section (maybe due to an artifact that hasn't been updated yet?) which is causing no bugs to show up in the search. Then on top of that I'm assuming the JS error about no initial value is an exception hit with the new code when there are 0 bugs in the search (i.e. it would throw even without this artifact problem if you just put normal filters on to get to 0 bugs).

ui/changes/src/common.js
```diff
@@ -140,53 +140,53 @@ export let landingsData = (async function () {
...
-  return obj;
 }
 
-export async function getTestingPolicySummaryData(grouping = "daily", filter) {
-  let data = await landingsData;
+export function getSummaryData(
+  bugSummaries,
+  grouping = "daily",
+  startDate,
+  filter,
```
bgrins, line 158: Nit: I'd probably flip the order of `filter` and `counter` since filter is unused by one of the consumers (so it can just be dropped from the parameters in that case instead of having to pass null)

ui/changes/src/index.js
```diff
@@ -334,13 +316,154 @@ function renderTestingSummary(bugSummaries) {
...
-  let chartEl = document.createElement("div");
-  resultSummary.append(chartEl);
-
   var chart = new ApexCharts(chartEl, options);
   chart.render();
 }
 
+function renderRiskChart(chartEl, bugSummaries) {
+  let minDate = Temporal.PlainDate.from(
+    bugSummaries.reduce((minSummary, summary) =>
```
bgrins, line 325: We need to handle empty array somehow (see screenshot of js error earlier). Maybe just early return at the start if bugSummaries.length is 0.

bgrins (dismissed):
Feel free to land if this empty teams thing is an unrelated artifact issue, and after the fix for empty summaries.

marco-c:
Yeah, I haven't released the latest version (which contains teams) yet.

PR #2028, opened 2020-12-15
Add ability to filter by release version.

### LINKED ISSUE #1969: Add ability to filter changes by target release

jrbenny35:
Closes #1969

### DIFF
--- a/ui/changes/src/index.html
+++ b/ui/changes/src/index.html
@@ -60,7 +60,7 @@
         </label>
         <label>
           Version<br />
-          <select name="releaseVersion" id=releaseVersion multiple></select>
+          <select name="releaseVersions" id=releaseVersions multiple></select>
         </label>
       </div>
     </aside>
--- a/ui/changes/src/index.html
+++ b/ui/changes/src/index.html
@@ -60,7 +60,7 @@
         </label>
         <label>
           Version<br />
-          <select name="releaseVersion" id=releaseVersion></select>
+          <select name="releaseVersion" id=releaseVersion multiple></select>
         </label>
       </div>
     </aside>
--- a/ui/changes/src/index.html
+++ b/ui/changes/src/index.html
@@ -58,6 +58,10 @@
         <label>White Board
           <input type="text" name="whiteboard" id="whiteBoard">
         </label>
+        <label>
+          Version<br />
+          <select name="releaseVersion" id=releaseVersion></select>
+        </label>
       </div>
     </aside>
     <main>
--- a/ui/changes/src/index.js
+++ b/ui/changes/src/index.js
@@ -558,7 +558,6 @@ async function buildTable(rerender = true) {
     );
   }
   if (releaseVersion) {
-    console.log(releaseVersion)
     bugSummaries = bugSummaries.filter((bugSummary) => {
       for (let item of bugSummary.versions) {
         for(let version of releaseVersion) {
--- a/ui/changes/src/index.js
+++ b/ui/changes/src/index.js
@@ -18,8 +18,6 @@ const HIGH_RISK_COLOR = "rgb(255, 13, 87)";
 const MEDIUM_RISK_COLOR = "darkkhaki";
 const LOW_RISK_COLOR = "green";
 
-var LATEST_VERSION = 0;
-
 let options = {
   metaBugID: {
     value: null,
@@ -243,11 +241,9 @@ function addRow(bugSummary) {
 
 async function populateVersions() {
   var versionSelector = document.getElementById("releaseVersion")
-  versionSelector.addEventListener("change", () => {
-    setOption("releaseVersion", versionSelector.value);
-    rebuildTable();
-  });
+
   let data = await landingsData;
+
   var allVersions = new Set();
     for (let date in data) {
       for (let bug of data[date]) {
@@ -258,19 +254,14 @@ async function populateVersions() {
     }
   var versions = [...allVersions]
   versions.sort()
-  
-  function addOption(item, index, array) {
-    let opt = document.getElementById("releaseVersion")
+
+  for (let version of versions) {
     let el = document.createElement("option")
-    if (item == array[array.length - 1]){
-      el.setAttribute("selected", "")
-    }
-    el.textContent = item
-    el.value = item
-    opt.appendChild(el)
+    el.setAttribute("value", version)
+    el.textContent = version
+    el.selected = true
+    versionSelector.appendChild(el)
   }
-  versions.forEach(addOption)
-  LATEST_VERSION = versions[versions.length - 1]
 }
 
 function renderTestingSummary(bugSummaries) {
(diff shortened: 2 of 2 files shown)


jrbenny35:
Screenshot_2020-12-15 bugbug ui

jrbenny35:
Can someone help me understand the lint failures?

ui/changes/src/index.html
```diff
@@ -58,6 +58,10 @@ <h3>Filters</h3>
         <label>White Board
           <input type="text" name="whiteboard" id="whiteBoard">
         </label>
+        <label>
+          Version<br />
+          <select name="releaseVersion" id=releaseVersion></select>
```
bgrins, line 63: I think we might want to make this a <select multiple> and then handle checking across multiple releases in the filtering code. That way by default we'll continue to show everything that matches other filters (i.e. in the date range) and not also require changing to the appropriate release.

We do this for others like testing tags etc so there are some examples of this in buildTable.

ui/changes/src/index.js
```diff
@@ -233,6 +239,42 @@ function addRow(bugSummary) {
...
+        }
+      }
+    }
+  var versions = [...allVersions]
+  versions.sort()
+  
+  function addOption(item, index, array) {
+    let opt = document.getElementById("releaseVersion")
+    let el = document.createElement("option")
+    if (item == array[array.length - 1]){
```
bgrins, line 265: If we switch to a multiple select, we can drop this code, along with all the LATEST_VERSION tracking as well since we can always just read the selected values off of the option. @marco-c does this sound right to you?

marco-c:
> Can someone help me understand the lint failures?

You can see the logs on Taskcluster (click on "Details", then click on the "bugbug lint" task, then click on "View Live Log"). It looks like there are some trailing whitespaces.
You can also reproduce this locally by installing `pre-commit` and running `pre-commit run --all-files`.

ui/changes/src/index.js
marco-c: Agreed!

ui/changes/src/index.html
jrbenny35: Make this a what?

bgrins: Sorry, I don't know what happened with that comment. Add the `multiple` attribute to this select tag i.e. `<select name="releaseVersion" id="releaseVersion" multiple>`

marco-c: You can take a look at testing tags, teams and components for examples of what @bgrins is suggesting

ui/changes/src/index.js
```diff
@@ -523,6 +557,16 @@ async function buildTable(rerender = true) {
...
     );
   }
+  if (releaseVersion) {
+    bugSummaries = bugSummaries.filter((bugSummary) => {
+      for (let item of bugSummary.versions) {
+        for(let version of releaseVersion) {
+          if (item == version)
+            return bugSummary
+        }
+      }
```
marco-c, line 567: You could use `some` and `includes` to make this super readable:
```suggestion
      return bugSummary.versions.some(version => releaseVersions.includes(version));
```

This would mean: at least one version of the bug is in the selected versions.

ui/changes/src/index.js
```diff
@@ -617,8 +661,10 @@ function setTableHeaderHandlers() {
   buildMetabugsDropdown();
   await buildComponentsSelect();
   await buildTeamsSelect();
+  await populateVersions();
 
   setTableHeaderHandlers();
+
```
marco-c, line 667: Nit: remove extraneous newline

ui/changes/src/index.html
```diff
@@ -58,6 +58,10 @@ <h3>Filters</h3>
         <label>White Board
           <input type="text" name="whiteboard" id="whiteBoard">
         </label>
+        <label>
+          Version<br />
+          <select name="releaseVersion" id=releaseVersion multiple></select>
```
marco-c, line 63: Nit: since it's multiple, maybe call it `releaseVersions`

ui/changes/src/index.js
```diff
@@ -233,6 +237,35 @@ function addRow(bugSummary) {
...
+}
+
+async function populateVersions() {
+  var versionSelector = document.getElementById("releaseVersion")
+
+  let data = await landingsData;
+
+  var allVersions = new Set();
+    for (let date in data) {
+      for (let bug of data[date]) {
```
marco-c, line 249: Nit: you could avoid the outer for loop by doing:
```
for (let bugSummary of Object.values(data)) {
```

ui/changes/src/index.js
```diff
@@ -233,6 +237,35 @@ function addRow(bugSummary) {
...
+    el.selected = true
+    versionSelector.appendChild(el)
+  }
+}
+
+function renderTestingSummary(bugSummaries) {
+  let metaBugID = getOption("metaBugID");
 
   risk_column.append(risk_text);
 }
```
marco-c, line 271: There is a problem here, maybe a bad rebase?

ui/changes/src/index.js
jrbenny35: Oh that's nice. I just defaulted to basics as my JS is still rusty 😅

ui/changes/src/index.js
jrbenny35: Seems so, will fix.

ui/changes/src/index.js
jrbenny35: I could only get this to work like this: 
```releaseVersion.some(version => bugSummary.versions.some(item => item == version)));```

Comment (marco-c) [reply ui/changes/src/index.js:567]: Mmmh, maybe it's because `bugSummary.versions` is a list of ints, while `releaseVersions` is a list of strings.
You could try:
```
return bugSummary.versions.some(version => releaseVersions.includes(version.toString()));
```
or
```
return releaseVersions.some(version => bugSummary.versions.includes(Number(version)));
```

File: ui/changes/src/index.js (hunk 6 above)
Comment (marco-c) [reply ui/changes/src/index.js:249]: You could even use `reduce`, something like this:
```JavaScript
let versions = [
  ...Object.values(data)
    .filter((bugSummary) => bugSummary.versions.length > 0)
    .reduce((set, bugSummary) => set.add(...bug.versions), new Set())
].sort();
```

It doesn't matter much in this case, but sometimes I find the functional style more readable.

ui/changes/src/index.js
```diff
@@ -523,6 +553,13 @@ async function buildTable(rerender = true) {
       bugSummary["whiteboard"].includes(whiteBoard)
     );
   }
+  if (releaseVersions) {
+    bugSummaries = bugSummaries.filter((bugSummary) => {
+      return releaseVersions.some(
```
marco-c, line 558: Nit: you can avoid the `return` statement if you remove the `{` parenthesis from the arrow function body

ui/changes/src/index.js
```diff
@@ -617,9 +654,11 @@ function setTableHeaderHandlers() {
   buildMetabugsDropdown();
   await buildComponentsSelect();
   await buildTeamsSelect();
+  await populateVersions();
 
   setTableHeaderHandlers();
 
+
```
marco-c, line 661: Nit: remove extraneous newline

ui/changes/src/index.js
```diff
@@ -237,6 +241,31 @@ function addRow(bugSummary) {
...
+  let data = await landingsData;
+
+  var allVersions = new Set();
+    for (let bug of Object.values(data)) {
+      bug.some(item => {
+        if (item.versions.length) {
+          allVersions.add(...item.versions)
+        }
+      })
```
marco-c, line 255: `some` is a function to check if at least one element of an array matches a given condition.
In this case, `forEach`, or a plain for loop, would be better.

Note: `some` and `forEach` are equivalent functionally here because your function is never returning a truthy value, but using `some` is a bit of a stretch.

ui/changes/src/index.js
```diff
@@ -237,6 +241,31 @@ function addRow(bugSummary) {
   risk_column.append(risk_text);
 }
 
+async function populateVersions() {
+  var versionSelector = document.getElementById("releaseVersions")
+
+  let data = await landingsData;
+
+  var allVersions = new Set();
+    for (let bug of Object.values(data)) {
```
marco-c, line 250:
```suggestion
    for (let bugs of Object.values(data)) {
```

marco-c:
LGTM, but there is still a linting failure to fix before landing

marco-c (approved):

Issue #2007, opened 2020-12-11
Classify multiple defect bugs at a time in the code to find bug-fixing commits


willT97:
Following on from #1866.

When a bug is classified as a defect, it is then classified by the regression model one bug at a time in

https://github.com/mozilla/bugbug/blob/1a584d62dc6c7f97f0e306596b45addefada8491/scripts/regressor_finder.py#L266

we could collect all the bugs that are classified as defect and classify them in one go.

Tested by running `python -m scripts.regressor_finder bug_fixing`

aadarsh10c:
hey guys i want to work on this issue how do i get started on this, i am a rookie

marco-c:
@aadarsh10c the first comment contains a good description of what needs to be done. You can also take a look at the recent PR by @willT97 which made a similar improvement: https://github.com/mozilla/bugbug/pull/2001.

PR #2043, opened 2020-12-18
Classify multiple defect bugs at a time

cjlh:
Similar to PR [#2001](https://github.com/mozilla/bugbug/pull/2001), this commit adds the bugs to be classified to a list and then classifies the list in a single call to the regression model.


cjlh:
Issue is #2007

marco-c (approved):
Thanks!

PR #2073, opened 2021-01-07
Add function-level complexity metrics

### LINKED ISSUE #1314: Add function-level complexity metrics
Similar to what was done in #1159, but at the level of modified functions.

marco-c:
Fixes #1314

### DIFF
--- a/bugbug/commit_features.py
+++ b/bugbug/commit_features.py
@@ -118,7 +118,9 @@ class functions_touched_num(object):
 class functions_touched_size(object):
     def __call__(self, commit, **kwargs):
         function_sizes = [
-            f[2] - f[1] + 1 for f_group in commit["functions"].values() for f in f_group
+            f["end"] - f["start"] + 1
+            for f_group in commit["functions"].values()
+            for f in f_group
         ]
 
         return {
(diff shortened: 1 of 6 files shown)


bugbug/commit_features.py
```diff
@@ -136,54 +138,291 @@ class source_code_file_metrics(object):
...
+            "Total file number of function arguments": commit["metrics"]["nargs_total"],
+            "Total file number of function exit points": commit["metrics"][
+                "nexits_total"
+            ],
+            "Total file cognitive": commit["metrics"]["cognitive_total"],
+        }
+
+
+def merge_function_metrics(objects):
+    return {
```
calixteman, line 217: nit: create a temp `metrics = map(lambda o: o["metrics"], objects)`
nit: use a `map` instead of `for...`

bugbug/commit_features.py
```diff
@@ -483,6 +722,74 @@ def __call__(self, commit, **kwargs):
         return commit["types"]
 
 
+def merge_metrics(objects):
```
calixteman, line 725: same comment as the previous one

bugbug/repository.py
```diff
@@ -682,35 +676,15 @@ def transform(hg: hglib.client, repo_dir: str, commit: Commit):
...
-            commit.total_instruction_loc / metrics_file_count
-        )
-        commit.average_logical_loc = commit.total_logical_loc / metrics_file_count
-        commit.average_comment_loc = commit.total_comment_loc / metrics_file_count
-        commit.average_nargs = commit.total_nargs / metrics_file_count
-        commit.average_nexits = commit.total_nexits / metrics_file_count
-        commit.average_cognitive = commit.total_cognitive / metrics_file_count
+        for key, value in commit.metrics.items():
+            if "total" in key:
+                commit.metrics[f"{key[:-5]}avg"] = value / metrics_file_count
```
calixteman, line 681: nit: if 5 is for `len("total")` then create a constant equals to `len...`

bugbug/commit_features.py
marco-c: Fixed in another way by storing the list of metrics in the repository module and building the dict without having to specificy all the metrics

PR #2075, opened 2021-01-07
change default filter to be based on previous two releases instead of a hardcoded date

bgrins:
Two functional changes here that we'll need to decide if we want:

1) This will get reset every time you reload the page (i.e. if you've selected all releases manually and then reload). This is an existing issue but currently it will reselect _all_ releases every time.
2) It deselects N/A which makes some of the reporting like "number of bugs by type" less useful (since it now doesn't include unfixed bugs). I wonder if we really should have a separate checkbox for "include unfixed bugs" rather than putting it into the release dropdown anyway.

### DIFF
--- a/ui/changes/src/index.html
+++ b/ui/changes/src/index.html
@@ -36,7 +36,7 @@
           <hr />
           <label
             >Start date<br />
-            <input id="startDate" type="date" value="2020-10-01" />
+            <input id="startDate" type="date" />
           </label>
           <label
             >End date<br />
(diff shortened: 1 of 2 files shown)


bgrins:
Do you think we should hold off on landing until we make a change to 1 or 2 above? I did 2 previous releases to avoid any confusion for the default filtering for QA teams (i.e. if all metabug work drops off right after a merge)

marco-c (approved):
> Do you think we should hold off on landing until we make a change to 1 or 2 above? I did 2 previous releases to avoid any confusion for the default filtering for QA teams (i.e. if all metabug work drops off right after a merge)

I'd say we can merge this and fix the caveats later.

marco-c:
I filed #2079 and #2081 as follow-ups.

PR #2077, opened 2021-01-07
dont rerender the table if the details section is closed, and close the details section by default. Fixes #2076

### LINKED ISSUE #2076: Don't render the bug table when it's collapsed and maybe collapse the bug table by default
This would help with perf and just clean up the page if people just wanted to see the summary graphs and don't care about the whole list. Should make sure to keep the state in localstorage so it remains expanded/collapsed based on user's preference

bgrins:
I got a little carried away as I was in here when I realized the functions weren't accurately named anymore (referring to building the table but it also rendered the summary) and overall the separation of responsibilities between "rebuild" and "build" weren't necessary AFAICT. I believe this still has the equivalent behavior but if you prefer I can split this out into multiple changes.

### DIFF
--- a/ui/changes/src/index.js
+++ b/ui/changes/src/index.js
@@ -1,10 +1,9 @@
 // TODO: On click, show previous components affected by similar patches.
 // TODO: On click, show previous bugs caused by similar patches.
 
+import localForage from "localforage";
 import { Temporal } from "proposal-temporal/lib/index.mjs";
-
 import ApexCharts from "apexcharts";
-
 import {
   TESTING_TAGS,
   featureMetabugs,
@@ -13,6 +12,11 @@ import {
   getSummaryData,
   summarizeCoverage,
 } from "./common.js";
+
+localForage.config({
+  driver: localForage.INDEXEDDB,
+  name: "bugbug-index",
+});
 
 const HIGH_RISK_COLOR = "rgb(255, 13, 87)";
 const MEDIUM_RISK_COLOR = "darkkhaki";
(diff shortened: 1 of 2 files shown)


ui/changes/src/index.js
```diff
@@ -1081,11 +1078,20 @@ function setTableHeaderHandlers() {
...
+  if (localStorage.getItem("detailsToggle") !== null) {
+    bugDetails.open = !!JSON.parse(localStorage.getItem("detailsToggle"));
+  }
+  bugDetails.addEventListener("toggle", () => {
+    localStorage.setItem("detailsToggle", bugDetails.open);
+    renderUI(false);
+  });
+
+  renderUI();
 })();
```
marco-c, line 1097: Nit: you could use localForage to avoid synchronous IO.

bgrins: Yeah I started on that and then realized we were setting the config to indexeddb in common.js and didn't know if that would race with a different driver here. Guess we could used indexeddb here, although in that case if we went async we'd have to wait to render until we get the result anyway. I don't feel super strongly either way, fine to update it.

marco-c (dismissed):

ui/changes/src/index.js
bgrins: Also I believe indexeddb doesn't work in private browsing windows so we'd lose the UI state there (again, we could use localStorage backend for localForage it just seemed like this was a global state based on https://github.com/mozilla/bugbug/blob/27512d1a8f25a602abd3aaa100e2bf0f10ef4647/ui/changes/src/common.js#L19 without digging into the docs

marco-c: Yeah I was thinking of using IndexedDB. I always try to avoid localStorage because of its sync IO (even with a small object to store like this one, it can always cause jank).

bgrins: Found a way to namespace the storage in the config (https://localforage.github.io/localForage/#settings-api-config) just to be safe, and updated the code to use localForage / indexeddb

marco-c (approved):

PR #2080, opened 2021-01-07
move the grouping options above the graphs and below the summary text

bgrins:
Just a visual improvement to move the options above the graph instead of buried at the bottom of the section


marco-c (approved):
Makes sense, I don't know why I added them at the end :P

PR #2088, opened 2021-01-11
Add diff in complexity metrics to the features

### LINKED ISSUE #1190: Add diff in complexity metrics to the features
We have added complexity metrics *after* the patch. It might be useful to also have the diff of complexity metrics, to check e.g. if the cognitive complexity increased with the patch, or if we have more function parameters than before, etc., etc..

See also #1159 and #1314.

marco-c:
Fixes #1190

### DIFF
--- a/bugbug/commit_features.py
+++ b/bugbug/commit_features.py
@@ -332,6 +332,34 @@ class source_code_function_metrics(object):
         }
 
 
+class source_code_metrics_diff(object):
+    name = "diff in metrics on source code"
+
+    def __call__(self, commit, **kwargs):
+        return {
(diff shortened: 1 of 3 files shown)


bugbug/commit_features.py
```diff
@@ -332,6 +332,34 @@ def __call__(self, commit, **kwargs):
         }
 
 
+class source_code_metrics_diff(object):
+    name = "diff in metrics on source code"
+
+    def __call__(self, commit, **kwargs):
+        return {
```
calixteman, line 339: nit: factorize in setting a var for `commit["metrics_diff]`

calixteman (approved):

PR #2089, opened 2021-01-13
share getComponentRegressionMap to common.js. This can be a building block for a dependency map structure

bgrins:


bgrins:
@marco-c this creates a flattened map of `<Component A, Map<Component B>, value>>` where Component B is likely to be regressed by Component A. It doesn't do any nesting or traversal of dependencies, that could be added separately. Does this look like the right logic / property to build this map?

bgrins:
Also could you clarify what the number means? .05, .125, etc

Screen Shot 2021-01-13 at 10 40 53 AM

ui/changes/src/common.js
```diff
@@ -299,3 +299,28 @@ export function summarizeCoverage(bugSummary) {
 
   return [lines_added, lines_covered, lines_unknown];
 }
+
+export async function getComponentRegressionMap(threshold = 0.05) {
```
marco-c, line 303: Later on, we could also add an option to take into accout test failures too (we'd need to come up with some sort of weighting to consider regression components more/less important than failing test components).

marco-c:
> @marco-c this creates a flattened map of <Component A, Map<Component B>, value>> where Component B is likely to be regressed by Component A. 

The code in getComponentRegressionMap is actually building a map where Component A is likely to be regressed by Component B, right?

> It doesn't do any nesting or traversal of dependencies, that could be added separately.

Agreed!

> Does this look like the right logic / property to build this map?

Yes, I think so. With this map we should be able to build a heatmap (either by component or by team).

marco-c:
> Also could you clarify what the number means? .05, .125, etc

They are percentages (past changes in the component caused regressions 5% of the times in that component, 12,5% in that other component, and so on). All components under `most_common_regression_components` should sum up to 1.

marco-c (approved):

Issue #2090, opened 2021-01-14
Add a function to generate a heatmap with component connections


marco-c:
We should add a function in common.js to draw a heatmap. The function will later be used in multiple views (at least the team-level view).

The function should use the data coming from `getComponentRegressionMap` (introduced in https://github.com/mozilla/bugbug/pull/2089).
The function should draw the heatmap either at the team level (showing dependencies between teams), or at the component level (showing dependencies between components).
The function should accept a list of source and target teams/components to filter the teams/components to draw in the heatmap (so, for example, in the team level view we will only show dependencies where the target of the dependency is the selected team).

bgrins:
A few notes from our conversation

* New parameter to getComponentRegressionMap to include "subdependency levels" 
* Take over graph.html for UI. Start with a single component

jrbenny35:
Is there an idea of what this UI would look like?

marco-c:
I'd start with a heatmap, like https://apexcharts.com/javascript-chart-demos/heatmap-charts/basic/. Or like the one that is currently present in graph.html.

jrbenny35:
Is the idea of the graph supposed to be one huge graph will all the components?

jrbenny35:
Trying to understand this a bit more. If we have Component 1 and Component 2. We enter those into the function and they both have similar components. What is the heatmap actually supposed to display? Is it the amount the components have caused regressions? So if Component 1 had a component that caused a regression by 10%, and component 2 had that same regression,  but it was 5%, would the heatmap display that?

marco-c:
Yes, the heatmap would basically be something like:
![immagine](https://user-images.githubusercontent.com/1616846/105750998-bb054900-5f45-11eb-8ac9-05c7dd9075bf.png)
where both rows ("metric1", "metric2", etc.) and columns ("w1", "w2", etc.) would actually be components, and the color would be stronger or lighter according to the connection between the two components.

The "connection" between the components would be the number which is in the map returned by the function @bgrins introduced in #2089 (e.g. `connectionsMap["Component 1"]["Component 2"]`). It is currently the percentage you mentioned, but in the future it could be more complex and be generated by some kind of formula which takes into account multiple data points. It doesn't matter for this issue though, as you can "blindly" use `getComponentRegressionMap` and assume its map is the right one to draw.

jrbenny35:
It seems like this will just create 1 row. 

Component 1 as the row, and the components that are children of it within the JSON object as columns. Is that correct?

[...]

PR #2120, opened 2021-01-29
Add function to generate heatmap based off of regression data.

jrbenny35:
Works on #2090

### DIFF
--- a/ui/changes/src/graph.js
+++ b/ui/changes/src/graph.js
@@ -24,24 +24,26 @@ function _generateData(count, yrange) {
 async function generateData() {
   let map = await getComponentRegressionMap();
   let return_map = []
-  Object.keys(map).slice(0, 10).forEach(element => {
-    let values = {}
-    values.data = []
-    values.name = element
-    for (const value in map[element]){
-      let obj = {}
-      obj.x = value
-      obj.y = Math.trunc(map[element][value] * 100)
-      values.data.push(obj)
-    }
-    return_map.push(values)
-  });
+  Object.keys(map)
+    .slice(0, 10)
+    .forEach(element => {
+      let values = {}
+      values.data = []
+      values.name = element
+      for (const value in map[element]){
+        let obj = {}
+        obj.x = value
+        obj.y = Math.trunc(map[element][value] * 100)
+        values.data.push(obj)
+      }
+      return_map.push(values)
+    });
   return return_map;
 }
 
 async function rerender(connections, teamComponentMapping) {
   var options = {
-    series: [...await generateData()],
+    series: [...(await generateData())],
     chart: {
       height: 700,
       width: 1200,
--- a/ui/changes/src/graph.js
+++ b/ui/changes/src/graph.js
@@ -4,29 +4,11 @@ import ApexCharts from "apexcharts";
 import teamComponentMapping from "./teams.json";
 import { getComponentRegressionMap } from "./common";
 
-function _generateData(count, yrange) {
-  var i = 0;
-  var series = [];
-  while (i < count) {
-    var x = (i + 1).toString();
-    var y =
-      Math.floor(Math.random() * (yrange.max - yrange.min + 1)) + yrange.min;
-
-    series.push({
-      x: x,
-      y: y,
-    });
-    i++;
-  }
-  return series;
-}
 
 async function generateData() {
   let map = await getComponentRegressionMap();
-  let return_map = [];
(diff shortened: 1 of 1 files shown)


ui/changes/src/graph.js
```diff
@@ -21,80 +21,36 @@ function generateData(count, yrange) {
...
+  Object.keys(map)
+    .slice(0, 10)
+    .forEach(element => {
+      let values = {}
+      values.data = []
+      values.name = element
+      for (const value in map[element]){
+        let obj = {}
+        obj.x = value
+        obj.y = Math.trunc(map[element][value] * 100)
```
pdehaan, line 36: TIL: [`Math.trunc()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/trunc) is a thing!!

```
/giphy "the more you know"
```

ui/changes/src/graph.js
```diff
@@ -4,7 +4,7 @@ import ApexCharts from "apexcharts";
 import teamComponentMapping from "./teams.json";
 import { getComponentRegressionMap } from "./common";
 
-function generateData(count, yrange) {
+function _generateData(count, yrange) {
```
marco-c, line 7: You could drop this function at this point.

ui/changes/src/graph.js
```diff
@@ -21,80 +21,36 @@ function generateData(count, yrange) {
   return series;
 }
 
-function rerender(connections, teamComponentMapping) {
+async function generateData() {
```
marco-c, line 24: Are you planning on implementing the additional features (like team-based vs component-based) in follow-up PRs?

ui/changes/src/graph.js
```diff
@@ -21,80 +21,36 @@ function generateData(count, yrange) {
...
+  let map = await getComponentRegressionMap();
+  let return_map = [];
+  Object.keys(map)
+    .slice(0, 10)
```
marco-c, line 28: I know this is just for testing, but instead of limiting the number of components, you could hardcode a source component/team.

ui/changes/src/graph.js
```diff
@@ -21,80 +21,36 @@ function generateData(count, yrange) {
   return series;
 }
 
-function rerender(connections, teamComponentMapping) {
+async function generateData() {
+  let map = await getComponentRegressionMap();
+  let return_map = [];
+  Object.keys(map)
```
marco-c, line 27: Nit: instead of having a list which you fill up in `.forEach`, you could use `.map`:
```
return Object.keys(map).map((element) => {
    ...
    return values;
});
```

Simple example:
```
return [1, 2, 3, 4].map((element) => element * 2);
```
instead of:
```
let array = [];
[1, 2, 3, 4].forEach((element) => array.push(element * 2));
return array;
```

ui/changes/src/graph.js
```diff
@@ -21,80 +21,36 @@ function generateData(count, yrange) {
...
+        obj.x = value;
+        obj.y = Math.trunc(map[element][value] * 100);
+        values.data.push(obj);
+      }
+      return_map.push(values);
+    });
+  return return_map;
+}
+
+async function rerender(connections, teamComponentMapping) {
```
marco-c, line 44: A chart in the style of https://github.com/mozilla/bugbug/issues/2090#issuecomment-767033208 would be nicer to read (the lightness of the colors would show the importance of a link, while with the current UI there are just two levels)

ui/changes/src/graph.js
jrbenny35: Yes!

ui/changes/src/graph.js
jrbenny35: I don't see any teams within the Component Regression Map. Is this something I should bring in from outside of that function?

ui/changes/src/graph.js
marco-c: If you do something like https://apexcharts.com/javascript-chart-demos/heatmap-charts/basic/, you shouldn't need to choose the increments beforehand, but the lightness of the colors will show the strength of the link (the lighter the color, the smaller the link).

ui/changes/src/graph.js
marco-c: Let's hardcode a component for now then, I'll see how to expose the team information (maybe a separate artifact similar to the current one but at the team level rather than the component level).

marco-c (approved):
Thanks!

PR #2167, opened 2021-02-19
Add eslint deps to package.json

pdehaan:
Some initial work on #2005.

This will at least make it easy to run ESLint via <kbd>npm run lint</kbd>; but there are still several (421 problems (412 errors, 9 warnings)) issues to resolve. Running <kbd>npm run lint -- --fix</kbd> will autofix 293 of the issues, leaving us with 106 problems (106 errors, 0 warnings) to manually resolve.

### DIFF
--- a/ui/changes/package.json
+++ b/ui/changes/package.json
@@ -5,6 +5,7 @@
   "private": true,
   "scripts": {
     "dev": "npx snowpack dev",
+    "lint": "eslint .",
     "release": "npx snowpack build",
     "test": "echo \"Error: no test specified\" && exit 1"
   },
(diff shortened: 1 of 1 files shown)


pdehaan:
When/if this lands, it might also be nice to set up Circle-CI or GitHub Actions to run the linter for each PR.

pdehaan:
Interesting, I failed CI because it was "unable to resolve dependency tree".
I think when installing locally I had to use `--force`.

Not sure if I should try including the package-lock.json file, or see if I have to downgrade to an older version of **eslint-config-prettier** to be more compatible with **eslint-plugin-mozilla**.

marco-c:
> When/if this lands, it might also be nice to set up Circle-CI or GitHub Actions to run the linter for each PR.

We have a linting job as part of the current Taskcluster CI, the linters are run through `pre-commit`.
Can you add eslint to the pre-commit configuration file too?

> Interesting, I failed CI because it was "unable to resolve dependency tree".

Maybe it'd be preferable to actually only add it to the pre-commit configuration file (and also remove prettier from package.json and only keep it in the pre-commit configuration file). Otherwise we have to make sure we keep the same version of the libraries in .pre-commit-config.yaml and package.json).

marco-c:
@pdehaan ping

marco-c:
@pdehaan feel free to reopen if you still intend to work on this.

Issue #1448, opened 2020-04-01
Add Mi metric


Luni-4:
https://github.com/mozilla/rust-code-analysis/commit/3e84e7172e8a482bb7275428a41a3d65e11cbb14

rajpratyush:
@Luni-4 So we have to implement the mi metric just like we dig in that repo of rust code analysis. can you guide me more on this ?? I would like to work on this issue

marco-c:
@rajpratyush this would be similar to 5119534.

PR #2207, opened 2021-03-01
Added mi metric

1Gokul:
Attempt to add the Mi metric (#1448 )

Please review and tell me if I'm doing it right.


marco-c:
@1Gokul the tests are failing, you can see the results by clicking on "Details" on the right of "Community-TC" and then navigate to the failure and see the logs.

You can also run tests locally with `python -m pytest tests/test_repository.py`.

1Gokul:
> 
> 
> @1Gokul the tests are failing, you can see the results by clicking on "Details" on the right of "Community-TC" and then navigate to the failure and see the logs.
> 
> You can also run tests locally with `python -m pytest tests/test_repository.py`.

Yeah, the logs are showing a difference in the values of `'mi_original_total'`, `'mi_sei_max'` and a few other values in `test_repository.py`.

Can you please guide me on how to find the correct values to enter?

1Gokul:
Also, if I can run tests locally, how to I pass in the location of `rust-code-analysis`?

I get a **FileNotFoundError** due to it when running the command you provided above.

marco-c:
> Also, if I can run tests locally, how to I pass in the location of rust-code-analysis?

You need to install rust-code-analysis and make sure it is in your path when you run the tests.
You can install it using `cargo` (see https://doc.rust-lang.org/cargo/commands/cargo-install.html).

marco-c:
> Yeah, the logs are showing a difference in the values of 'mi_original_total', 'mi_sei_max' and a few other values in test_repository.py.

It is because you set the expected values of those values to 0 in the tests, but they shouldn't be 0.

> Can you please guide me on how to find the correct values to enter?

You can log the values for the mi_metrics in repository.py, then manually compute the total, average, min and max and set them as the expected values in the test.

1Gokul:
Checks have passed!

I am a bit unsure about the values of the mi metrics though. A few values seem to be going negative. 
Can you take a look?

marco-c (approved):
Thanks!

Issue #259, opened 2019-04-09
Add support for building models on GitHub issues data


daxmc99:
Awesome project! 
I was curious if there were any plans for github integration in the future?
If not, what might be a way to begin adding that?

marco-c:
Yes, we were thinking about that. Some models might make sense for GitHub too (e.g. the assignee model), some might make sense only for very complex projects (e.g. the component assignment will not be so useful for small projects, as they have no components).

The first step would be adding support to retrieve issues' data from GitHub. There are likely already libraries to do that.

PR #2242, opened 2021-03-22
Add support for github issue retrieval (#259)

ksy36:
This PR adds and ability to fetch issues from a specified owner/repository


scripts/github_issue_retriever.py
```diff
@@ -0,0 +1,56 @@
...
+from bugbug import db, github
+from bugbug.utils import zstd_compress
+
+
+class Retriever(object):
+    def retrieve_issues(
+        self, owner=str, repo=str, state=str, retrieve_events=bool
+    ) -> None:
+        db.download(github.GITHUB_ISSUES_DB)
+        github.download_issues(owner, repo, state, retrieve_events)
```
ksy36, line 17: This script fetches all issues sorted by creation date (earliest -> latest).

For fetching new issues and updates I was thinking to use a similar approach as in https://github.com/mozilla/bugbug/blob/master/scripts/bug_retriever.py#L24 (get ids of issues that were modified since last run, delete them from the database and add them again during download process).

Github has `since` parameter, allowing to get updates after a given time https://docs.github.com/en/rest/reference/issues#list-issues-for-a-repository 

Wonder if I should address this in a separate PR (and after the db artifact is created) since `db.last_modified` is using the artifact url to get `Last-Modified` header.

bugbug/github.py
```diff
@@ -0,0 +1,109 @@
...
+        return data, response.links
+    except Exception as e:
+        print(e)
+        raise
+
+
+def get_start_page() -> int:
+    # Determine next page to fetch based on number of downloaded issues
+    issues = get_issues()
+    count = len(list(issues))
```
marco-c, line 81: Nit: `sum(1 for _ in issues)`

This way you don't have to collect all issues in a list.

scripts/github_issue_retriever.py
```diff
@@ -0,0 +1,56 @@
...
+
+import argparse
+
+from bugbug import db, github
+from bugbug.utils import zstd_compress
+
+
+class Retriever(object):
+    def retrieve_issues(
+        self, owner=str, repo=str, state=str, retrieve_events=bool
```
marco-c, line 14: Should be `owner: str, repo: str`, etc.

bugbug/github.py
```diff
@@ -0,0 +1,109 @@
...
+    return int(count / PER_PAGE) + 1
+
+
+def download_issues(owner, repo, state, retrieve_events) -> None:
```
marco-c, line 85: Can you add types to the arguments?

marco-c: (same for other functions in the module)

bugbug/github.py
```diff
@@ -0,0 +1,109 @@
...
+                    events = fetch_events(item["events_url"])
+                    item.update({"events": events})
+                except Exception as e:
+                    print(e)
+                    raise
+
+        return data, response.links
+    except Exception as e:
+        print(e)
+        raise
```
marco-c, line 75: Nit: no need to print and reraise, you can remove the try-except and let the exception propagate

bugbug/github.py
```diff
@@ -0,0 +1,109 @@
...
+# Rate limit period in seconds
+RATE_LIMIT_PERIOD = 900
+
+
+def get_issues() -> Iterator[IssueDict]:
+    yield from db.read(GITHUB_ISSUES_DB)
+
+
+@sleep_and_retry
+@limits(calls=1200, period=RATE_LIMIT_PERIOD)
```
marco-c, line 37: Isn't the rate limit higher than this?

Note: it'd be nice to use the API to get the rate limit, so we don't wait needlessly (https://docs.github.com/en/rest/reference/rate-limit), but we can do that in a follow-up.

scripts/github_issue_retriever.py
marco-c: > Wonder if I should address this in a separate PR (and after the db artifact is created) since db.last_modified is using the artifact url to get Last-Modified header.

As you prefer, you can do it here or in a follow-up PR.
The script should work also when db.last_modified is unavailable (it is a mistake that bug_retriever doesn't handle that case, I will fix that), because in some cases we need to regenerate it (e.g. if we change the format) and when we do we start from scratch.

marco-c:
@ksy36 it'd be great if you could also add some tests, but we can also do that in a follow-up.

bugbug/github.py
ksy36: ah ok, that's good to know, thanks!

bugbug/github.py
ksy36: Yeah, it's a bit higher (5000/hour so 1250 per 15 minutes). I went with 1200 per 15 minutes, to make sure we never hit the github rate limit exception, and after these 1200 the script sleeps for 15 minutes. 

I've tried another approach, where the script just fetches issues until github throws an exception, then makes a request to rate-limit api to fetch time until the reset and waits for it to reset. In this case wait time can be quite long, but we get more data between waits. 

But yeah, I can experiment and improve this in a follow-up :)

ksy36:
Thanks for the feedback! Yeah, I'll add tests to this PR

scripts/github_issue_retriever.py
ksy36: Ok, thanks, I'll create a follow-up PR for that

ksy36:
I've added some tests and made changes from the code review, could you please take another look @marco-c ?

marco-c (approved):
Thanks!

Issue #2240, opened 2021-03-17
Update the components select when a product is selected


marco-c:
Just like we do when a team is selected.

Ayushsunny:
Hello @marco-c I would like to work on this issue

marco-c:
Hello @Ayushsunny, this will need to be implemented in ui/changes/src/common.js.
You can run the ui project by doing: `cd ui/changes` and then `npm run dev`.

You will need to update `async function buildComponentsSelect(teams = null) {` by adding a `product` argument, and using it like the `teams` component to filter the bug summaries.
Next, you will need to something like https://github.com/mozilla/bugbug/blob/7077c0c3b0310f8c1a25ea9ff144f53e80595a32/ui/changes/src/common.js#L1722-L1725, in the `buildProductsSelect` function.

Ayushsunny:
> Hello @Ayushsunny, this will need to be implemented in ui/changes/src/common.js.
> You can run the ui project by doing: `cd ui/changes` and then `npm run dev`.
> 
> You will need to update `async function buildComponentsSelect(teams = null) {` by adding a `product` argument, and using it like the `teams` component to filter the bug summaries.
> Next, you will need to something like
> 
> https://github.com/mozilla/bugbug/blob/7077c0c3b0310f8c1a25ea9ff144f53e80595a32/ui/changes/src/common.js#L1722-L1725
> , in the `buildProductsSelect` function.

Hello @marco-c I'm getting this error while running this project by "npm run dev" on my terminal

![Screenshot from 2021-03-20 01-05-45](https://user-images.githubusercontent.com/73520373/111833997-f8200300-8918-11eb-9969-dab62ee5a98b.png)

marco-c:
Did you run `npm install` first?

Ayushsunny:
> Did you run `npm install` first?

Yes sir.

marco-c:
@Ayushsunny, the error messages says "node_modules" doesn't exist, so it seems you didn't run `npm install`.

Ayushsunny:
> @Ayushsunny, the error messages says "node_modules" doesn't exist, so it seems you didn't run `npm install`.

Thank you so much, It's working

Ayushsunny:
@marco-c Sir can you please review my code that I'm going right or not?
![Screenshot from 2021-03-22 22-11-53](https://user-images.githubusercontent.com/73520373/112026092-de223280-8b5b-11eb-98df-078b6d2bdbba.png)

[...]

PR #2256, opened 2021-03-22
Update the components select when a product is selected

Ayushsunny:
---

**Description**
Updated  the components select when a product is selected in #2240 

**I have tested this PR on**

  

 Linux/Ubuntu

**code?**
``` JavaScript
  options["products"].callback = async function () { 
    await buildProductsSelect(new Set(getOption("products"))); 
    return "components"; 
  }; ```

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,15 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "type": "pwa-chrome",
+            "request": "launch",
+            "name": "Launch Chrome against localhost",
+            "url": "http://localhost:5500",
+            "webRoot": "${workspaceFolder}"
+        }
+    ]
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,3 @@
+{
+  "lockfileVersion": 1
+}
--- a/ui/changes/src/common.js
+++ b/ui/changes/src/common.js
@@ -1691,6 +1691,10 @@ async function buildComponentsSelect(teams = null) {
     option.selected = true;
     componentSelect.append(option);
   }
+  options["products"].callback = async function () { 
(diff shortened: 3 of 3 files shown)


Ayushsunny:
> @Ayushsunny this won't work (did you test it?). You're re-building the products select when a component is selected, but we actually want to do the opposite (also you're calling `buildProductsSelect` with an argument, but `buildProductsSelect` doesn't take any argument as input).
> 
> Try applying the changes I suggested in [#2240 (comment)](https://github.com/mozilla/bugbug/issues/2240#issuecomment-802912548).

sorry I'm not able to get, what do you trying to explain in this ? https://github.com/mozilla/bugbug/issues/2240#issuecomment-802912548 
could you explain bit more or like exact what changes I've to do, please

marco-c:
What parts of the comment are unclear for you?

Ayushsunny:
> What parts of the comment are unclear for you?

This line sir : "by adding a product argument, and using it like the teams argument to filter the bug summaries.
Next, you will need to something like"

marco-c:
You need to add an additional argument to the `buildComponentsSelect` function I mentioned, the argument will be called `products`, and you will use it similarly to the already existing `teams` argument.

Ayushsunny:
> You need to add an additional argument to the `buildComponentsSelect` function I mentioned, the argument will be called `products`, and you will use it similarly to the already existing `teams` argument.

Yes but did right in this code?
``` Javascript 
 options["products"].callback = async function () { 
    await buildProductsSelect(new Set(getOption("products"))); 
    return "components"; 
  }; ```

what changes do I have to perform in this code like changing this "buildPRoductselect" to "buildComponentselect" ?

marco-c:
No @Ayushsunny, sorry. You didn't do any of the things I said in the comment above :)

Ayushsunny:
> No @Ayushsunny, sorry. You didn't do any of the things I said in the comment above :)

Sorry, but I really not able to get you? either explain me clearly or assign anyone else

marco-c:
@Ayushsunny unfortunately I'm not able to explain it more clearly than that.

Ayushsunny:
> @Ayushsunny unfortunately I'm not able to explain it more clearly than that.

Okay, and Thank you so much for your earlier help sir.
sorry

PR #2281, opened 2021-04-05
Get updates for github issues based on last_modified (#259)

ksy36:
This is a follow-up for https://github.com/mozilla/bugbug/pull/2242 (fetching and saving updates if `last_modified` is available and if not, download all issues)

### DIFF
--- a/bugbug/github.py
+++ b/bugbug/github.py
@@ -4,7 +4,7 @@
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import logging
-from typing import Iterator, NewType, Tuple
+from typing import Callable, Iterator, List, NewType, Tuple
 
 import requests
 from ratelimit import limits, sleep_and_retry
@@ -32,6 +32,10 @@ def get_issues() -> Iterator[IssueDict]:
     yield from db.read(GITHUB_ISSUES_DB)
 
 
+def delete_issues(match: Callable[[IssueDict], bool]) -> None:
+    db.delete(GITHUB_ISSUES_DB, match)
+
+
 @sleep_and_retry
 @limits(calls=1200, period=RATE_LIMIT_PERIOD)
 def api_limit():
@@ -43,7 +47,7 @@ def get_token() -> str:
     return get_secret("GITHUB_TOKEN")
 
 
-def fetch_events(events_url: str) -> str:
+def fetch_events(events_url: str) -> list:
     api_limit()
     logger.info(f"Fetching {events_url}")
     headers = {"Authorization": "token {}".format(get_token())}
@@ -55,7 +59,7 @@ def fetch_events(events_url: str) -> str:
 
 def fetch_issues(
     url: str, retrieve_events: bool, params: dict = None
-) -> Tuple[str, dict]:
+) -> Tuple[List[IssueDict], dict]:
     api_limit()
     headers = {"Authorization": "token {}".format(get_token())}
     response = requests.get(url, params=params, headers=headers)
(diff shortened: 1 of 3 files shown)


bugbug/github.py
```diff
@@ -32,6 +32,10 @@ def get_issues() -> Iterator[IssueDict]:
     yield from db.read(GITHUB_ISSUES_DB)
 
 
+def delete_issues(match):
```
marco-c, line 35: Can you add type annotations here? It should be something like `def delete_issues(match: Callable[IssueDict, bool]) -> None`.

(you'd need to define IssueDict like https://github.com/mozilla/bugbug/blob/e9106d8b2e1889b49fd0177d8f4537c83383d853/bugbug/bugzilla.py#L19).

bugbug/github.py
```diff
@@ -79,9 +83,34 @@ def get_start_page() -> int:
     return int(count / PER_PAGE) + 1
 
 
+def fetch_issues_updated_since_timestamp(
+    owner: str, repo: str, state: str, since: str, retrieve_events: bool = False
+):
```
marco-c, line 88: This is missing the return type annotation

tests/test_github.py
```diff
@@ -82,3 +82,58 @@ def test_download_issues():
     )
 
     github.download_issues("webcompat", "web-bugs", "all")
+
+
+def test_download_issues_updated_since_timestamp():
```
marco-c, line 87: Nit: `def test_download_issues_updated_since_timestamp() -> None:` so the function gets type checked.

marco-c (commented):
Looks great overall!

bugbug/github.py
ksy36: Oh right, thanks 👍

ksy36:
Thanks! I've pushed the fixes, could you please take another look @marco-c ?

tests/test_github.py
```diff
@@ -46,15 +48,15 @@ def test_fetch_issues_with_events():
     # Mock events request
     responses.add(responses.GET, TEST_EVENTS_URL, json=expected_events, status=200)
 
-    # Assert that response with events has expected format
-    response_with_events = github.fetch_issues(TEST_URL, True)
     expected_with_events = expected
     expected_with_events[0]["events"] = expected_events
 
+    # Assert that response with events has expected format
+    response_with_events = github.fetch_issues(TEST_URL, True)
```
marco-c, line 55: Nit: why did you move this?

bugbug/github.py
```diff
@@ -79,9 +83,34 @@ def get_start_page() -> int:
     return int(count / PER_PAGE) + 1
 
 
+def fetch_issues_updated_since_timestamp(
+    owner: str, repo: str, state: str, since: str, retrieve_events: bool = False
+) -> list:
```
marco-c, line 88: This can be `list[IssueDict]`

bugbug/github.py
```diff
@@ -55,7 +59,7 @@ def fetch_events(events_url: str) -> str:
 
 def fetch_issues(
     url: str, retrieve_events: bool, params: dict = None
-) -> Tuple[str, dict]:
+) -> Tuple[list, dict]:
```
marco-c, line 62: This can be list[IssueDict]

marco-c (dismissed):
Thanks!

tests/test_github.py
ksy36: I've separated building expected data from the actual response to make it a bit more readable

marco-c: Ok, WFM.

bugbug/github.py
ksy36: Getting an error with using `list`, `Class 'type' doesn't define '__getItem__', so the '[]' operator cannot be used on its instances`, so going to use `typing.List` instead of `list`

marco-c (approved):

marco-c:
The test failures are not this PR's fault, but a problem introduced with 0094619b67c74a8d5d188918a009b21e47c9c0b2.

ksy36:
Thanks!

PR #2313, opened 2021-04-21
Add needsdiagnosis model to classify webcompat issues

ksy36:


tests/test_needsdiagnosis.py
```diff
@@ -0,0 +1,15 @@
...
+from bugbug.models.needsdiagnosis import NeedsDiagnosisModel
+
+
+def test_get_needsdiagnosis_labels():
+    model = NeedsDiagnosisModel()
+    classes, _ = model.get_labels()
+    assert classes[71052]
+    assert classes[71011]
+    assert not classes[71012]
+    assert not classes[70962]
```
ksy36, line 15: While adding this test I have noticed that the test fixture file I've created didn't contain issues that have been moved to "needsdiagnosis" milestone. It only contained first 100 issues and it made me realize that we haven't been using this milestone in the very beginning (used a label instead). So I have updated the fixture file with some recent issues that have the milestone change in the events list.

ksy36: As a follow-up on this:

As I've noticed that we haven't used milestones in the beginning, and used a label `status-needsdiagnosis` instead, I've tried to experiment with the model in the hopes of getting more issues that would get classified as "needsdiagnosis". Because right now all these issues are attributing to `needsdiagnosis=false`, despite the fact that they were labeled as `status-needsdiagnosis` at some point in time. I've counted these issues and there is 1181 of them with the label (and without the milestone).

So I decided to mark these issues as `classes[issue["number"]] = 1` as well. However in this case the model performed a bit worse and would classify issues as `needsdiagnosis=false` with probability >0.95, while the current model would have probability 0.90 for the same issue (we use 0.95 threshold for closing the issue). 

Because of this I decided not to mark these issues as `classes[issue["number"]] = 1`, but I'm a bit puzzled on why it affected it in such way :)

marco-c: > However in this case the model performed a bit worse and would classify issues as needsdiagnosis=false with probability >0.95, while the current model would have probability 0.90 for the same issue (we use 0.95 threshold for closing the issue).

@ksy36 it is possible the model was actually better and you just needed to move the threshold. The best way to check is by comparing the cross validation results (they are outputted right after training the model, after `Cross Validation scores:`) before/after your change. NOTE: If you mainly care about needsdiagnosis=false, then it's better to make the label 1 for these cases rather than 0, as the cross validation considers 1 to be your target.
I'd suggest looking at the precision/recall values for different confidence thresholds, and then choosing the threshold that works for you the best.

For example, in the model from the PR, at confidence threshold 0.9, you'd have 96% precision and 83% recall for the 0 class (which corresponds to needsdiagnosis=false). Which means the model is able to find 83% of issues where needsdiagnosis is False and in 4% of the times assigns a needsdiagnosis=false to a issue where it should actually be true.
At confidence threshold 0.8, you'd have 94% precision and 94% recall, which means the model is able to find 94% of issues where needsdiagnosis is False and is wrong 6% of the times.
The lower the confidence threshold, the lower the precision (so you make more mistakes), but the higher the recall (so you miss fewer cases).

bugbug/github.py
```diff
@@ -135,3 +139,16 @@ def download_issues(
         db.append(GITHUB_ISSUES_DB, next_page_data)
 
     logger.info("Done downloading")
+
+
+def fetch_issue_by_number(
+    owner: str, repo: str, issue_number: int, retrieve_events: bool = False
+) -> List[IssueDict]:
```
marco-c, line 146: It should be a `IssueDict` here instead of a `List` of `IssueDict`s, right?

bugbug/issue_snapshot.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+
+def rollback(issue, when=None):
+    assert when is None
```
marco-c, line 8: Nit: add an assertion message that says rollbacking to a specific point in history is not supported yet.

infra/data-pipeline.yml
```diff
@@ -1210,6 +1210,48 @@ tasks:
...
+            path: /feature_importance.png
+            type: file
+
+      routes:
+        - notify.email.release-mgmt-analysis@mozilla.com.on-failed
+        - notify.irc-channel.#bugbug.on-failed
+        - index.project.bugbug.train-needsdiagnosis.${version}
+        - index.project.bugbug.train-needsdiagnosis.per_version.${version}.${year}.${month}.${day}.${hour}.${minute}.${second}
+        - index.project.bugbug.train-needsdiagnosis.per_date.${year}.${month}.${day}.${hour}.${minute}.${second}.${version}
+        - index.project.bugbug.train-needsdiagnosis.latest
```
marco-c, line 1248: Nit: not a big deal, but can you keep the same format as the other index paths? ("train_needsdiagnosis" instead of "train-needsdiagnosis")

infra/data-pipeline.yml
```diff
@@ -1210,6 +1210,48 @@ tasks:
     #        owner: release-mgmt-analysis@mozilla.com
     #        source: ${repository}/raw/master/data-pipeline.yml
 
+    - ID: train-needsdiagnosis
```
marco-c, line 1213: Can you add the task to the dependencies of the `check_metrics` task (https://github.com/mozilla/bugbug/blob/fd8bddb0e1d0a4e8a298105d34158d2940d33f3c/infra/data-pipeline.yml#L1255)?

The `check_metrics` task will download the metrics.json file from the task and generate a chart to show the evolution of the metrics (accuracy, precision and recall) over time, so you can see if there are regressions.

bugbug/models/needsdiagnosis.py
```diff
@@ -0,0 +1,85 @@
...
+                    event["event"] == "milestoned"
+                    and event["milestone"]["title"] == "needsdiagnosis"
+                ):
+                    classes[issue["number"]] = 1
+
+            if issue["number"] not in classes:
+                classes[issue["number"]] = 0
+
+        logger.info(
+            f"{sum(1 for label in classes.values() if label == 1)} issues has been moved to needsdiagnosis"
```
marco-c, line 76:
```suggestion
            f"{sum(1 for label in classes.values() if label == 1)} issues have been moved to needsdiagnosis"
```

bugbug/models/needsdiagnosis.py
```diff
@@ -0,0 +1,85 @@
...
+        )
+        logger.info(
+            f"{sum(1 for label in classes.values() if label == 0)} issues has not been moved to needsdiagnosis"
```
marco-c, line 79:
```suggestion
            f"{sum(1 for label in classes.values() if label == 0)} issues have not been moved to needsdiagnosis"
```

infra/data-pipeline.yml
ksy36: oh yeah right, thanks for noticing this!

tests/test_needsdiagnosis.py
ksy36: Thanks for the explanation. I have swapped the label to be 1 for needsdiagnosis=false and compared two models (the original one and one with taking into account the additional label). 

**Original**
```
Cross Validation scores:
Accuracy: f0.9243783099465173 (+/- 0.0015920386355566356)
Precision: f0.9295439183930381 (+/- 0.0010610813620656812)
Recall: f0.9923733608341815 (+/- 0.0009308004618768267)
X_train: (63778, 11124), y_train: (63778,)
X_test: (7087, 11124), y_test: (7087,)

Confidence threshold > 0.9 - 5668 classified

1       0.96      0.83      0.63      0.89      0.72      0.53      6454
0       0.97      0.12      1.00      0.22      0.35      0.11       633
```

**With labels:**

```
Cross Validation scores:
Accuracy: f0.9088243532163371 (+/- 0.0021256433728581436)
Precision: f0.916425381088529 (+/- 0.0009203475364896565)
Recall: f0.9884392095291099 (+/- 0.003011899305650207)
X_train: (63778, 11124), y_train: (63778,)
X_test: (7087, 11124), y_test: (7087,)

Confidence threshold > 0.9 - 5254 classified

1       0.95      0.78      0.69      0.86      0.73      0.54      6346
0       0.94      0.11      1.00      0.19      0.33      0.10       741
```
So looks like the original model is better by looking at the cross validation scores. 

As for the confidence threshold, in our case mistakenly labeling an issue with `needsdiagnoisis=False` is worse than not labeling it at all (that way we may miss a valid issue and close it), so I think the metrics for the threshold > 0.9 are acceptable:
> Which means the model is able to find 83% of issues where needsdiagnosis is False and in 4% of the times assigns a needsdiagnosis=false to a issue where it should actually be true.

As we have a very low number of issues where `needsdiagnos=True` every such issue is important, so labeling `needsdiagnos=False` and closing issues only when confidence threshold is >0.95 seems to be the right approach (this is how current model is working).

bugbug/github.py
ksy36: yeah, true :)

ksy36:
I've made the fixes, could you please take another look @marco-c?

tests/test_needsdiagnosis.py
marco-c: Interesting! Maybe it's because those issues are old and have some other different characteristic than the more recent ones...
If you want to test additional possible values of the confidence threshold, you can modify this line: https://github.com/mozilla/bugbug/blob/253ef6461144e151165fb17f7c84b60e1c59f56a/bugbug/model.py#L487 (I guess we could make it configurable by model, but for now if you want to do a quick test you can just hardcode `0.95` temporarily and rerun the training).

marco-c:
> I've made the fixes, could you please take another look @marco-c?

I'll re-review it first thing tomorrow ;)

marco-c (approved):
LGTM!

ksy36:
Thanks!

tests/test_needsdiagnosis.py
ksy36: Ok, thanks! I'll test the 0.95 threshold as well.

Issue #1883, opened 2020-10-26
Add other Halstead metrics


marco-c:
We are using N1, n1, N2 and n2, but others are available in rust-code-analysis now.

```
            |- halstead
            |  |- n1: 6
            |  |- N1: 86
            |  |- n2: 69
            |  |- N2: 147
            |  |- length: 233
            |  |- estimated program length: 436.9979625220206
            |  |- purity ratio: 1.8755277361460112
            |  |- vocabulary: 75
            |  |- volume: 1451.3147548855402
            |  |- difficulty: 6.391304347826087
            |  |- level: 0.1564625850340136
            |  |- effort: 9275.794302964105
            |  |- time: 515.3219057202281
            |  `- bugs: 0.14715651181813164
```

sayantan1413:
Hi, I am new to this contribution and I would like to work on this issue so can you help me to start with it? @marco-c

devkumar24:
hey, I want to work on the issue, can you assign me for this issue?

marco-c:
@devkumar24 you can work on any issue which has no PR linked to it, no need to be assigned ;)

Rishabhc711:
Hey! Where exactly does the code lie for the computed metrics? How do we look for it?

marco-c:
@Rishabhc711 in bugbug/repository.py and in bugbug/commit_features.py (e.g. search for `halstead_N1` in those files).

Amotul-raheem:
Hi, I'm one of outreachy's participants. I want to work on this issue.

suhaibmujahid:
Resolved by #2372

PR #2372, opened 2021-05-29
Add more Halstead metrics

tuncbkose:
Hi!

This is related to issue #1883. I just copied what was happening with the previous metrics, let me know if I misunderstood what had to be done (This is my first contribution and I am not very clear on what the rest of the code does).


tuncbkose:
Noticed the failing tests. Will work on it and if appropriate add more tests.

marco-c (approved):
I won't merge it yet as it requires a DB version bump (which I don't want to do at the moment), but it LGTM, thanks!

PR #2387, opened 2021-06-08
Fetch private issue contents to replace automatically closed public issue body and title

ksy36:
In this PR I've adjusted github data fetching step to make a request to fetch original issue content from the private repository and use it in the dataset instead of the public issue content.

This is to account for the moderation workflow in webcompat repository. Issues that are labelled needsdiagnosis=False with high probability by bugbug are now closed automatically in private and public repositories. Public issues that are closed that way are receiving placeholder text instead of the actual content (as we don't want to accidentally post abusive content because it is no longer moderated). With this PR the dataset will have the original content of such issues instead of placeholder body and title.

### DIFF
--- a/bugbug/issue_snapshot.py
+++ b/bugbug/issue_snapshot.py
@@ -13,6 +13,7 @@ def rollback(issue, when=None):
             if (
                 event["event"] == "renamed"
                 and event["rename"]["from"] != "In the moderation queue."
+                and event["rename"]["from"] != "Issue closed."
             ):
                 issue["title"] = event["rename"]["from"]
 
--- a/bugbug/utils.py
+++ b/bugbug/utils.py
@@ -8,9 +8,11 @@ import errno
 import json
 import logging
 import os
+import re
 import socket
 import subprocess
 import tarfile
+import urllib.parse
 from collections import deque
 from contextlib import contextmanager
 from datetime import datetime
@@ -468,3 +470,29 @@ def get_hgmo_stack(branch: str, revision: str) -> List[bytes]:
 
 def get_physical_cpu_count() -> int:
     return psutil.cpu_count(logical=False)
+
+
+def extract_metadata(body: str) -> dict:
+    """Extract metadata as dict from github issue body.
+
+    Extract all metadata items and return a dictionary
+    Example metadata format: <!-- @public_url: *** -->
+    """
+    match_list = re.findall(r"<!--\s@(\w+):\s(.+)\s-->", body)
+    return dict(match_list)
+
+
+def extract_private(issue_body: str) -> Optional[tuple]:
+    """Extract private issue information from public issue body
+
+    Parse public issue body and extract private issue number and
+    its owner/repository (webcompat repository usecase)
+    """
+    private_url = extract_metadata(issue_body).get("private_url", "").strip()
+    private_issue_path = urllib.parse.urlparse(private_url).path
+
+    if private_issue_path:
+        owner, repo, _, number = tuple(private_issue_path.split("/")[1:])
+        return owner, repo, number
+
+    return None
--- a/scripts/github_issue_retriever.py
+++ b/scripts/github_issue_retriever.py
@@ -5,16 +5,48 @@
 
 import argparse
 from logging import getLogger
+from typing import List, Tuple
 
 from bugbug import db, github
-from bugbug.utils import zstd_compress
+from bugbug.github import IssueDict
+from bugbug.utils import extract_private, zstd_compress
 
 logger = getLogger(__name__)
 
 
+def replace_with_private(original_data: List[IssueDict]) -> Tuple[List[IssueDict], set]:
+    """Replace title and body of automatically closed public issues.
+
+    Replace them with title and body of a corresponding private issue
+    to account for moderation workflow in webcompat repository
+    """
+    updated_ids = set()
+    updated_issues = []
+    for item in original_data:
+        if item["title"] == "Issue closed.":
+            extracted = extract_private(item["body"])
+            if extracted is None:
+                continue
+
+            owner, repo, issue_number = extracted
+            private_issue = github.fetch_issue_by_number(owner, repo, issue_number)
+            if private_issue:
+                item["title"] = private_issue["title"]
+                item["body"] = private_issue["body"]
+                updated_ids.add(item["id"])
+                updated_issues.append(item)
+
+    return updated_issues, updated_ids
+
+
 class Retriever(object):
     def retrieve_issues(
-        self, owner: str, repo: str, state: str, retrieve_events: bool
+        self,
+        owner: str,
+        repo: str,
+        state: str,
+        retrieve_events: bool,
+        retrieve_private: bool,
     ) -> None:
 
         last_modified = None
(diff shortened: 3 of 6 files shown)


bugbug/issue_snapshot.py
```diff
@@ -13,6 +13,7 @@ def rollback(issue, when=None):
             if (
                 event["event"] == "renamed"
                 and event["rename"]["from"] != "In the moderation queue."
+                and event["rename"]["from"] != "Issue closed."
```
ksy36, line 16: (this is to avoid double renaming)

ksy36:
Hi @marco-c , could you please take a look?

README.md
```diff
@@ -125,8 +127,9 @@ Note: the script will take a long time to run (on my laptop more than 7 hours).
 - `bugbug/nlp` contains utility functions for NLP;
 - `bugbug/labels.py` contains utility functions for handling labels;
 - `bugbug/bug_snapshot.py` contains a module to play back the history of a bug.
+- `bugbug/github.py` contains the functions to retrieve issues from github based on specified owner/repository;
 
 ## Using bugbug for non-Mozilla projects
 
-Bugbug is focussing on Mozilla use-cases for Firefox and Bugzilla.
+Bugbug is focussing on Mozilla use-cases for Firefox, Bugzilla and github.
 However, we will be happy to accept pull requests adding support for other projects or bug trackers.
```
marco-c, line 135: Could you move the documentation changes to a separate PR?

README.md
```diff
@@ -125,8 +127,9 @@ Note: the script will take a long time to run (on my laptop more than 7 hours).
 - `bugbug/nlp` contains utility functions for NLP;
 - `bugbug/labels.py` contains utility functions for handling labels;
 - `bugbug/bug_snapshot.py` contains a module to play back the history of a bug.
+- `bugbug/github.py` contains the functions to retrieve issues from github based on specified owner/repository;
 
 ## Using bugbug for non-Mozilla projects
 
-Bugbug is focussing on Mozilla use-cases for Firefox and Bugzilla.
+Bugbug is focussing on Mozilla use-cases for Firefox, Bugzilla and github.
```
marco-c, line 134: Nit:
```suggestion
Bugbug is focussing on Mozilla use-cases for Firefox, Bugzilla and GitHub.
```

scripts/github_issue_retriever.py
```diff
@@ -44,7 +54,9 @@ def retrieve_issues(
             logger.info("Updating finished")
         else:
             logger.info("Retrieving all issues since last_modified is not available")
-            github.download_issues(owner, repo, state, retrieve_events)
+            github.download_issues(
+                owner, repo, state, retrieve_events, retrieve_private
+            )
```
marco-c, line 59: Instead of adding support for this in the `github` module, what if you add support for it as part of the github_issue_retriever script or even a separate script?
So the `github` module is "clean" and doesn't have any webcompat-specific code.

You could download all issues, then parse all of them and then download all the ones you need to download from the private repo.
Or, you could download all issues from the public and the private repos (assuming you will need almost all the ones inthe private repo anyway), and then replace the ones from the public that you need to replace with the ones from the private.

ksy36: Sounds good, I'll move it to a separate script

ksy36: > You could download all issues, then parse all of them and then download all the ones you need to download from the private repo.

Decided to go with this approach, as I only need about 1/3 of the issues from the private repo on average (for example, last week we got 655 issues and the bot closed 262 out of them, so I will only need to make a request to get content of these 262 issues from the private repo). 

I've moved the retrieval to `github_issue_retriever` script, based on `--retrieve_private` argument.

ksy36:
I've added the changes, could you please take another look @marco-c?

bugbug/utils.py
```diff
@@ -468,3 +470,20 @@ def should_skip(changeset):
 
 def get_physical_cpu_count() -> int:
     return psutil.cpu_count(logical=False)
+
+
+def extract_metadata(body) -> dict:
```
marco-c, line 475: Could you add a short comment in this function and `extract_private` to explain what they do exactly?

Nit: please also add type information to the arguments of the two functions

bugbug/utils.py
```diff
@@ -468,3 +470,20 @@ def should_skip(changeset):
...
+    match_list = re.findall(r"<!--\s@(\w+):\s(.+)\s-->", body)
+    metadata_dict = dict(match_list)
+    return metadata_dict
```
marco-c, line 478: Nit:
```suggestion
    return dict(match_list)
```

bugbug/utils.py
```diff
@@ -468,3 +470,20 @@ def should_skip(changeset):
...
+
+def extract_private(issue_body) -> tuple:
+    private_url = extract_metadata(issue_body).get("private_url", "").strip()
+    private_issue_path = urllib.parse.urlparse(private_url).path
+
+    if private_issue_path:
+        owner, repo, _, number = tuple(private_issue_path.split("/")[1:])
+        return owner, repo, number
+
+    return ()
```
marco-c, line 489: Nit: `return None`, I prefer to use None when a value is to be considered as missing (like in this case). Then in the code using the function, you can do `if XXX is None` instead of `if not XXX` to be more precise

marco-c (dismissed):

ksy36:
I've added the changes from the code review, could you please one more look @marco-c ?

scripts/github_issue_retriever.py
```diff
@@ -5,16 +5,50 @@
...
+
+            owner, repo, issue_number = extracted
+            private_issue = github.fetch_issue_by_number(owner, repo, issue_number)
+            if private_issue:
+                item["title"] = private_issue["title"]
+                item["body"] = private_issue["body"]
+                updated_ids.add(item["id"])
+                updated_issues.append(item)
+
+    return original_data, updated_issues, updated_ids
```
marco-c, line 41: Sorry, one last nit: since the `original_data` dict is not changed (only the elements inside it are), you can avoid returning it from the function.

scripts/github_issue_retriever.py
```diff
@@ -33,6 +67,12 @@ def retrieve_issues(
                 owner, repo, state, last_modified.isoformat(), retrieve_events
             )
 
+            if retrieve_private:
+                logger.info(
+                    "Replacing contents of auto closed public issues with private issues content"
+                )
+                data, _, _ = replace_with_private(data)
```
marco-c, line 74: If you apply the nit suggested above, this becomes simply `replace_with_private(data)`

scripts/github_issue_retriever.py
```diff
@@ -46,6 +86,20 @@ def retrieve_issues(
             logger.info("Retrieving all issues since last_modified is not available")
             github.download_issues(owner, repo, state, retrieve_events)
 
+            if retrieve_private:
+                logger.info(
+                    "Replacing contents of auto closed public issues with private issues content"
+                )
+
+                all_issues = list(github.get_issues())
+                _, updated_issues, updated_ids = replace_with_private(all_issues)
```
marco-c, line 95: If you apply the nit suggested above, this becomes `updated_issues, updated_ids = replace_with_private(data)`

scripts/github_issue_retriever.py
ksy36: Ah right, thanks :)

ksy36:
Thanks, I've fixed that and adjusted the tests

marco-c (approved):

ksy36:
Thanks!

Issue #674, opened 2019-07-05
Try using doc2vec as similarity algorithm


marco-c:
After #637, we should try with doc2vec too and see what works better.

Goku2699:
@marco-c   is this issue resolved or still on? I'm new to open source so need your help to make some contributions.

marco-c:
No, it's not resolved yet. It depends on #637, so that one must be fixed first before fixing this one.
Once that is fixed, something similar needs to be done but using doc2vec instead of word2vec.

ayush-1506:
@Goku2699 #637 is done. Feel free to give this a try.

probaku1234:
I'm trying to run the word2vec code, but `bugzilla.get_bugs()` in init function returns nothing and because of that error pops up.

marco-c:
You should train a model first, so that the bugs DB will be downloaded.
```
python3 run.py --goal defect --train
```

probaku1234:
I downloaded db with that code but still it gets nothing.

marco-c:
Can you post the output of `ls -al data/`?
What error pops up?

Try to give as much info as possible, otherwise it'll be hard to help you :)

probaku1234:
This is the result of ls
```
total 6278256
drwxr-xr-x  10 yunseoblee  staff         320 Jul  2 14:22 .
drwxr-xr-x  34 yunseoblee  staff        1088 Jul 12 22:25 ..
-rw-r--r--   1 yunseoblee  staff  1410490733 Jul  2 14:22 bugs.json
-rw-r--r--   1 yunseoblee  staff           1 Jun 21 15:53 bugs.json.version
-rw-r--r--   1 yunseoblee  staff   235073932 Jul  2 14:22 bugs.json.zst
-rw-r--r--   1 yunseoblee  staff          34 Jul  2 14:22 bugs.json.zst.etag
-rw-r--r--   1 yunseoblee  staff  1398900302 Jul  2 14:22 commits.json
-rw-r--r--   1 yunseoblee  staff           1 Jun 21 15:53 commits.json.version
-rw-r--r--   1 yunseoblee  staff   150652308 Jul  2 14:22 commits.json.zst
-rw-r--r--   1 yunseoblee  staff          34 Jul  2 14:22 commits.json.zst.etag
```
This is the error traceback
```
Traceback (most recent call last):
  File "/Users/yunseoblee/Desktop/bugbug/scripts/evaluate_similarity.py", line 38, in <module>
[...]

PR #2406, opened 2021-07-02
[WIP] Add doc2vec similarity algorithm

rock420:
Fixes: #674 

Implemented the **PV-DM**  based `doc2vec` similarity algorithm with cosine similarity as similarity metrics between two documents. 

Todo:
- [x] Try different other `doc2vec` algorithms
- [ ] Try different other similarity metrics
- [x] Tuning different hyperparameters
- [ ] Discussion about the [class structure](https://github.com/mozilla/bugbug/issues/674#issuecomment-871333753)

### DIFF
--- a/bugbug/similarity.py
+++ b/bugbug/similarity.py
@@ -397,9 +397,15 @@ class Word2VecSimilarityBase(BaseSimilarity):
                 for doc, bug_id in zip(self.corpus, self.bug_ids)
             ]
             self.d2vmodel = Doc2Vec(
-                taggedCorpus, vector_size=100, window=5, min_count=5, epochs=20, dm=1
+                taggedCorpus,
+                vector_size=150,
+                window=5,
+                min_count=5,
+                epochs=20,
+                dm=0,
+                alpha=0.03,
+                negative=6,
             )
-            self.d2vmodel.init_sims(replace=True)
         else:
             raise ValueError('model must be "word2vec" or "doc2vec"')
 
--- a/bugbug/similarity.py
+++ b/bugbug/similarity.py
@@ -38,6 +38,7 @@ try:
     from gensim.corpora import Dictionary
     from gensim.matutils import sparse2full
     from gensim.models import TfidfModel, Word2Vec, WordEmbeddingSimilarityIndex
+    from gensim.models.doc2vec import Doc2Vec, TaggedDocument
     from gensim.models.ldamodel import LdaModel
     from gensim.similarities import SoftCosineSimilarity, SparseTermSimilarityMatrix
     from gensim.summarization.bm25 import BM25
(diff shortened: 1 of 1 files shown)


rock420:
@marco-c can you guide me on why the docker build is failing?

rock420:
Tested both the `PV-DM` and `PV-DBOW` doc2vec algorithm on the bug data (tuned other hyperparameters also). PV-DBOW is working better compare to PV-DM in terms of both precision and recall. 
Comparison of the evaluation result (for the same set of hyperparameters and same number of data) -
|                             |  PV-DM         |     PV-DBOW   |
|    :-----:                 |    :-----:          |        :-----:          |
| Recall @ 1:         |   9.739%       |    14.285%        |
| Recall @ 5:         |   10.922%     |    16.041%        |
| Recall @ 10:       |   11.005%     |     16.178%       |
| Precision @ 1:    |   7.373%       |     9.896%         |
| Precision @ 5:    |   2.376%       |     3.397%         | 
| Precision @ 10:  |   1.339%       |     1.969%         |
| Recall:                 |  11.038%      |     16.232%       |
| Precision:            |   1.983%       |     2.302%         |
| MAP@k :             |   8.404%       |     11.673%       |

bugbug/similarity.py
```diff
@@ -368,6 +369,7 @@ def __init__(
         cleanup_urls=True,
         nltk_tokenizer=False,
         confidence_threshold=0.8,
+        model="word2vec",
```
marco-c, line 372: If `Word2VecWmdSimilarity`, `Word2VecWmdRelaxSimilarity` and `Word2VecSoftCosSimilarity` can only work when `model` is doc2vec, I would remove this option and totally separate the `Word2VecSimilarityBase` and `Doc2VecCosSimilarity` classes (maybe create a `Doc2VecCosSimilarity` class and make `Doc2VecCosSimilarity` inherit from it, so we can also introduce other kinds of Doc2Vec classes like for the Word2Vec case).

Otherwise, if we can make it so `Word2VecWmdSimilarity`, `Word2VecWmdRelaxSimilarity` and `Word2VecSoftCosSimilarity` work no matter what the underlying model is (word2vec or doc2vec), then we can have a single base class like you started doing.

rock420: Yeah, If we are using word embeddings from doc2vec, then we can do that.

But, I just noticed that gensim by default doesn't train word embeddings while using the `PV-DBOW` algorithm. Word embeddings are trained along with doc in the case of `PV-DM` only. Though we can set one argument ( [doc_link](https://radimrehurek.com/gensim/models/doc2vec.html#:~:text=using%20dm_concat%20mode.-,dbow_words,-(%7B1%2C0%7D%2C%20optional)%20%E2%80%93%20If) ) to train word embedding in the case of `PV-DBOW`, but it will make the training very slow.

Issue #2460, opened 2021-08-05
Add support for multiple DBs for github


ksy36:
We currently only support one DB in github module:

https://github.com/mozilla/bugbug/blob/bd9b593f163f82c36ffe68fd764e03db946f3cb0/bugbug/github.py#L17-L22

We need to change DB registration to support more than one project

PR #2465, opened 2021-08-05
Issue #2460 - Add support for multiple DBs for github

ksy36:


bugbug/model.py
```diff
@@ -756,12 +756,3 @@ def items_gen(self, classes):
 class IssueModel(Model):
     def __init__(self, lemmatization=False):
         Model.__init__(self, lemmatization)
-        self.training_dbs = [github.GITHUB_ISSUES_DB]
-
-    def items_gen(self, classes):
-        for issue in github.get_issues():
-            issue_number = issue["number"]
-            if issue_number not in classes:
-                continue
```
ksy36, line 765: Moved `self.training_dbs` and `items_gen` to needsdiagnosis model

bugbug/github_dbs_config.py
```diff
@@ -0,0 +1,12 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+config = {
+    "web-bugs": {
+        "path": "data/github_issues.json",
+        "url": "https://community-tc.services.mozilla.com/api/index/v1/task/project.bugbug.data_github_issues.latest/artifacts/public/github_issues.json.zst",
+        "version": 1,
+    }
+}
```
ksy36, line 12: created this config with a key as repository name, so we could add multiple DBs

requirements.txt
```diff
@@ -18,7 +18,7 @@ python-dateutil==2.8.2
 python-hglib==2.6.2
 ratelimit==2.2.1
 requests==2.26.0
-rs_parsepatch==0.3.5
+rs_parsepatch==0.3.6
```
marco-c, line 21: Unrelated change (maybe leftover from merge?)

scripts/github_issue_retriever.py
```diff
@@ -7,98 +7,111 @@
...
                     "Deleting public issues that were updated and saving updates"
                 )
-                github.delete_issues(lambda issue: issue["id"] in updated_ids)
-                db.append(github.GITHUB_ISSUES_DB, updated_issues)
+                self.github.delete_issues(lambda issue: issue["id"] in updated_ids)
+                db.append(self.github.db_path, updated_issues)
+
+        zstd_compress(self.github.db_path)
+
+    def replace_with_private(
```
marco-c, line 88: Could you keep this function where it was, to minimize the diff?

bugbug/model.py
marco-c: Could we instead keep them here, adding `owner` and `repo` as parameters to `IssueModel` so it can create a `GitHub` instance and store it in `self.github`?

bugbug/github_dbs_config.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
```
marco-c, line 1: I'd like to find another way than having a module listing all GitHub DBs.
What if we remove this module and call `db.register` in the `GitHub` constructor? The path for the DB could be programmatically built like this:
```
https://community-tc.services.mozilla.com/api/index/v1/task/project.bugbug.data_github_{owner}_{repo}_issues.latest/artifacts/public/github_{owner}_{repo}_issues.json.zst
```

ksy36: Sounds good, I'll move it to the same file where GitHub class is, but probably not inside the constructor since `db.register` will be called each time Github instance is created?

ksy36: I could probably check whether it's registered with `db.is_registered()` and not register it again. However there is also db version, which could be different for every DB, so it seems that some sort of config is still needed 🤔

marco-c: The version could be a constant in the `GitHub` module, since the version only changes when we make changes to the way we store data for GitHub repos, so it sill be the same for any GitHub DB.

requirements.txt
ksy36: Right, thanks, I've reverted the change

bugbug/github_dbs_config.py
ksy36: Thanks! I've made it a constant value.

bugbug/model.py
ksy36: Yeah, added it back

scripts/github_issue_retriever.py
ksy36: I've moved it inside the Retriever class because it's using its github instance in `self.github.fetch_issue_by_number`

ksy36:
I've made the changes, could you please take another look @marco-c?

marco-c:
@ksy36 thanks, LGTM!
We'll also need to update the paths and URLs in the pipeline definition file at infra/data-pipeline.yml, that is https://github.com/mozilla/bugbug/blob/028bcf19e7f4b941c9e0cca14f78968e99d25886/infra/data-pipeline.yml#L252-L257 and https://github.com/mozilla/bugbug/blob/028bcf19e7f4b941c9e0cca14f78968e99d25886/infra/data-pipeline.yml#L266-L267.

ksy36:
Oops, I've missed that! Will change it as well :)

marco-c (approved):
Thanks! I'll try to release this today so we can check if everything is working as expected.

marco-c:
CI is misbehaving, I'll contact the Taskcluster team to figure it out.

ksy36:
Thanks!

Issue #2634, opened 2021-11-22
Replace typing.Dict with dict


marco-c:
It is now possible to use `dict` directly instead of `typing.Dict` in type definitions.

anooptp:
@marco-c I would like to work on this.

Few Questions:
- Is this specific to any specific file or the entire project? 
- Any reason as to why all the ```typing.Dict``` needs to be replaced with ```dict```?

marco-c:
> Is this specific to any specific file or the entire project?

The entire project.

> Any reason as to why all the ```typing.Dict``` needs to be replaced with ```dict```?

Simply because of code cleanup.

webbdays:
I working on this issue.
Please look at my pull request #2640 .
Some checks are failing , please help.

PR #2640, opened 2021-11-28
Used dict instead of typing.Dict in most of the files.

webbdays:
Fix for issue #2634 .

### DIFF
--- a/bugbug/github.py
+++ b/bugbug/github.py
@@ -4,7 +4,7 @@
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import logging
-from typing import Callable, Iterator, List, NewType, Tuple
+from typing import Callable, Iterator, NewType
 
 from ratelimit import limits, sleep_and_retry
 
@@ -67,7 +67,7 @@ class Github:
 
     def fetch_issues(
         self, url: str, retrieve_events: bool, params: dict = None
-    ) -> Tuple[List[IssueDict], dict]:
+    ) -> tuple[list[IssueDict], dict]:
         self.api_limit()
         headers = {"Authorization": "token {}".format(self.get_token())}
         response = get_session("github").get(url, params=params, headers=headers)
@@ -93,7 +93,7 @@ class Github:
         count = sum(1 for _ in issues)
         return int(count / PER_PAGE) + 1
 
-    def fetch_issues_updated_since_timestamp(self, since: str) -> List[IssueDict]:
+    def fetch_issues_updated_since_timestamp(self, since: str) -> list[IssueDict]:
         # Fetches changed and new issues since a specified timestamp
         url = "https://api.github.com/repos/{}/{}/issues".format(self.owner, self.repo)
 
--- a/scripts/past_bugs_by_unit.py
+++ b/scripts/past_bugs_by_unit.py
@@ -7,7 +7,6 @@ import argparse
 import json
 import logging
 from collections import defaultdict
-from typing import Dict, List
 
 from tqdm import tqdm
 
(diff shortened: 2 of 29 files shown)


marco-c:
Thanks, the change looks good. Could you also apply the same change for `list` and `tuple` in separate commits of this PR?

webbdays:
Ok

webbdays:
@marco-c 

Is there any way in python to perform in-place replace.
Suppose ,
File 1 has these lines
Line1a
Line2b
Line3a

I want to just change Line2b to Line2a in that file.

In python we read file to a string
Modify the string and then write to the file.
But for git diff it is shown as deleting whole file and writing to the file.
-Line1a
-Line2b
-Line3a
+Line1a
+Line2a
+Line3a

I want to just
-Line2b
+Line2a

Which is possible to do in nano text editor.

But I want to do the same for multiple files. Like write some python program or any other way.

Please help.

webbdays:
i get to know this on stack overflow,
which performs replace recursively in dir and it's sub dirs.
`find . -type f -readable -writable -exec sed   
-i "s/string/nstring/g" {} \;`

webbdays:
Ok

webbdays:
I cannot understand what is the error in lines 46,47,48 of bugbug/test_scheduling.py file.

marco-c:
There is an unexpected empty line

marco-c:
So you just need to remove those unneeded empty lines

tests/test_test_scheduling.py
```diff
@@ -4,7 +4,7 @@
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
 from datetime import datetime
-from typing import List
+
```
marco-c, line 7: Here's an empty line

tests/test_phabricator.py
```diff
@@ -4,14 +4,14 @@
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
 from datetime import timedelta
-from typing import List
+
```
marco-c, line 7: Here's an empty line

scripts/past_bugs_by_unit.py
```diff
@@ -7,7 +7,7 @@
 import json
 import logging
 from collections import defaultdict
-from typing import Dict, List
+
```
marco-c, line 10: Here's an empty line

marco-c:
Other than the empty lines, it seems the error is on:
```
PushResult = tuple[
    tuple[Revision],
    Revision,
    tuple[Runnable, ...],
    tuple[Runnable, ...],
    tuple[Runnable, ...],
]
```

For some reason, mypy doesn't understand the "...".
I've updated mypy in the pre-commit configuration file. Could you try to rebase your patch on top of master?

webbdays:
Yes, that is the problem

webbdays:
Pull from upsteam to update master and merge to working branch and push to fork. ?

marco-c:
Yes, or you can directly rebase your branch:
```
git pull upstream master --rebase
git push origin dict --force
```

(assuming `upstream` is this repo and `origin` is your fork)

webbdays:
Ok

scripts/past_bugs_by_unit.py
webbdays: I think the pre-commit removes that.?

marco-c: Yes, it should

webbdays:
its again the same problem with "..."

marco-c:
> its again the same problem with "..."

:(
Let's keep typing.Tuple only for that one then. We'll change it later on.

webbdays:
Ok

webbdays:
Please look at this

https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html

webbdays:
... with tuple is for python 3.9+

marco-c:
We are using 3.9, so it should work. There must be a bug in mypy.
Anyway, let's keep typing.Tuple for those three specific cases for now.

webbdays:
Ok

marco-c:
Linting is still failing because of some unnecessary empty lines and white spaces. You can see for yourself like I described in this earlier comment: https://github.com/mozilla/bugbug/pull/2640#issuecomment-981377790.

webbdays:
Ok

webbdays:
Will Lint errors corrected by the pre-commit in that check(running in browser) added to the latest commit ?

webbdays:
In pre-commit check, black check is failing.
I am using Ubuntu on Termux.
I came to know that there is problem with multiprocessing.synchronize on Termux, it says named semaphores unavailable.

marco-c:
You can apply the fixes manually for now.
For next time, if you install pre-commit in the repo (pre-commit install), it will automatically run *before* you commit and apply some changes automatically.

webbdays:
I have done that and pre-commit is running when I try to commit.
There black check is failing, so no commit is performed.

marco-c:
When pre-commit fails, it either suggests you what fixes to apply, or it applies them automatically.
After you have applied your fixes, you should `git add` them and then try to commit again.

marco-c:
There are still some linting failures

webbdays:
Locally, all passed for pre-commit before commit.

marco-c:
> Locally, all passed for pre-commit before commit.

It isn't possible, maybe you forgot to `git add` some of your changes before committing again?

bugbug/github.py
```diff
@@ -67,7 +67,7 @@ def fetch_events(self, events_url: str) -> list:
 
     def fetch_issues(
         self, url: str, retrieve_events: bool, params: dict = None
-    ) -> Tuple[List[IssueDict], dict]:
+    ) -> tuple[ list[IssueDict], dict]:
```
marco-c, line 70: For example here there is an unnecessary whitespace.

webbdays:
I have patched dict to dict.patch (git diff master dict > dict.patch)
Old dict removed.
New dict is created.
Applied patch in dict.
All modified files are staged.
And commit is performed.
So, pre-commit is automatically performed.
Import error when black is running.
Error is related to import from _multiprocessing .

marco-c:
If you have a problem while using pre-commit, you can apply the suggested fixes from https://community-tc.services.mozilla.com/tasks/SvHj7rG5RoqAj2wn4_Nddw/runs/0/logs/public/logs/live.log manually.

webbdays:
Ok. How to do that.
Sign up in that website?

marco-c:
No, the fixes are suggested in the textual logs directly.
For example, https://community-tc.services.mozilla.com/tasks/SvHj7rG5RoqAj2wn4_Nddw/runs/0/logs/public/logs/live.log#L350 means you have to replace `    ) -> tuple[ list[IssueDict], set]:` with `    ) -> tuple[list[IssueDict], set]:` (that is, remove the unneeded whitespace).

webbdays:
Ok

marco-c (approved):
Thanks!

webbdays:
Thanks for the merge.
I think this can be marked as my first open source contribution.
My contribution in this pull request is very basic replace task.
But in this process I have learned lot of things under your @marco-c guidance.
I have improved my knowledge of git and some Linux terminal commands.
Thanks for your guidance.

marco-c:
Thank you for your contribution and the willpower to finish it!

PR #2641, opened 2021-11-28
Replace typing.Dict with dict

### LINKED ISSUE #2634: Replace typing.Dict with dict

anooptp:
Fixes #2634


marco-c:
Thanks for the PR, but there is an older PR that is fixing the same issue (#2640).

Issue #2459, opened 2021-08-05
db.last_modified should return None if the DB version changed


marco-c:
When we change the version of a DB, we want to re-download everything.

E.g. say we bump the GitHub issues DB to a newer version, the next run of the retriever will get the last modified of the DB in the old version but without downloading the old version DB, and then will only download issues starting from that last modified date. We'll end up with a small subset of issues.

webbdays:
How do we know, if the db version is changed.
How do we know the present db versions ?
Is it saved in any file for each db ?
I was not able to find.
If no such file is there,  we can create a db_versions file to store the present versions of all db.

Then have a check if ever the version of any db is changed in the register or last_modified function and perform what we needed when db version is changed. 

In db_versions file...
COMMIT_DB => 22
BUGS_DB => 7

webbdays:
According to this commit c7087d76462a2c6fa2b8fa1e39be7724f176d7d0

db version is changed manually.

so, I am gathering all the at present db version of each db into a db_versions.json file for doing version check.

webbdays:
In the first commit of my pull request #2653 , I have not used any json file.
I have created another function named is_db_version_changed  after looking at is_old_schema function.

For is_db_version_changed function, I have used the same code of is_old_schema function but small change in return statement.

webbdays:
> When we change the version of a DB, we want to re-download everything.
> 
> E.g. say we bump the GitHub issues DB to a newer version, the next run of the retriever will get the last modified of the DB in the old version but without downloading the old version DB, and then will only download issues starting from that last modified date. We'll end up with a small subset of issues.

But you mentioned here, we need to download

marco-c:
By "re-download everything" here I meant regenerate the DB from scratch.

PR #2653, opened 2021-12-02
Added database version change checking.

webbdays:
Fix for the issue #2459 .

### DIFF
--- a/bugbug/db.py
+++ b/bugbug/db.py
@@ -58,6 +58,23 @@ def is_old_schema(path):
     return DATABASES[path]["version"] > prev_version
 
 
+# used almost same code as is_old_schema function with small change at return statement.
+# added function docstring.
+def is_db_version_changed(path):
+    """This function checks if db version chnaged ? return 1 if version is changed and 0 if not changed"""
+
+    url = urljoin(DATABASES[path]["url"], f"{os.path.basename(path)}.version")
+    r = requests.get(url)
+
+    if not r.ok:
+        logger.info(f"Version file is not yet available to download for {path}")
+        return True
+
+    prev_version = int(r.text)
+
+    return DATABASES[path]["version"] != prev_version
+
+
 def download_support_file(path, file_name, extract=True):
     # If a DB with the current schema is not available yet, we can't download.
     if is_old_schema(path):
@@ -119,6 +136,22 @@ def upload(path):
 
 
 def last_modified(path):
+
+    # if database version changed then download the database again
+    if is_db_version_changed(path):
+        print("\nDatabase version changed")
+        download_status = download(path)
+
+        if download_status:
+            print(
+                f"\nDownload_status = {download_status}\n"
+                + "Database download is successful."
+            )
+        else:
+            return (
+                f"\nDownload_status = {download_status}\n" + "Database download failed."
+            )
+
     url = DATABASES[path]["url"]
     last_modified = utils.get_last_modified(url)
 
--- a/bugbug/db.py
+++ b/bugbug/db.py
@@ -45,24 +45,7 @@ def exists(path):
     return os.path.exists(path)
 
 
-def is_old_schema(path):
-    url = urljoin(DATABASES[path]["url"], f"{os.path.basename(path)}.version")
-    r = requests.get(url)
-
-    if not r.ok:
-        logger.info(f"Version file is not yet available to download for {path}")
-        return True
-
-    prev_version = int(r.text)
-
-    return DATABASES[path]["version"] > prev_version
-
-
-# used almost same code as is_old_schema function with small change at return statement.
-# added function docstring.
-def is_db_version_changed(path):
-    """This function checks if db version chnaged ? return 1 if version is changed and 0 if not changed"""
-
+def is_different_schema(path):
     url = urljoin(DATABASES[path]["url"], f"{os.path.basename(path)}.version")
     r = requests.get(url)
 
@@ -77,7 +60,7 @@ def is_db_version_changed(path):
 
 def download_support_file(path, file_name, extract=True):
     # If a DB with the current schema is not available yet, we can't download.
-    if is_old_schema(path):
+    if is_different_schema(path):
         return False
 
     try:
@@ -102,7 +85,7 @@ def download_support_file(path, file_name, extract=True):
 # Download and extract databases.
 def download(path, support_files_too=False, extract=True):
     # If a DB with the current schema is not available yet, we can't download.
-    if is_old_schema(path):
+    if is_different_schema(path):
         return False
 
     zst_path = f"{path}.zst"
@@ -136,27 +119,12 @@ def upload(path):
 
 
 def last_modified(path):
-
-    # if database version changed then download the database again
-    if is_db_version_changed(path):
-        print("\nDatabase version changed")
-        download_status = download(path)
-
-        if download_status:
-            print(
-                f"\nDownload_status = {download_status}\n"
-                + "Database download is successful."
-            )
-        else:
-            return (
-                f"\nDownload_status = {download_status}\n" + "Database download failed."
-            )
+    if is_different_schema(path):
+        raise LastModifiedNotAvailable()
 
     url = DATABASES[path]["url"]
     last_modified = utils.get_last_modified(url)
-
-    if last_modified is None:
-        raise LastModifiedNotAvailable()
+    # if no Last-Modified header then that needs to be handled. (in another PR)
 
     return last_modified
 
--- a/bugbug/db.py
+++ b/bugbug/db.py
@@ -119,13 +119,14 @@ def upload(path):
 
 
 def last_modified(path):
-
     if is_different_schema(path):
         raise LastModifiedNotAvailable()
 
     url = DATABASES[path]["url"]
     last_modified = utils.get_last_modified(url)
-    # if no Last-Modified header then that needs to be handled. (in another PR)
+
+    if last_modified is None:
+        raise LastModifiedNotAvailable()
 
     return last_modified
 
--- a/bugbug/db.py
+++ b/bugbug/db.py
@@ -119,6 +119,7 @@ def upload(path):
 
 
 def last_modified(path):
+
     if is_different_schema(path):
         raise LastModifiedNotAvailable()
 
--- a/tests/test_db.py
+++ b/tests/test_db.py
@@ -399,7 +399,7 @@ def test_download_support_file_missing(tmp_path, caplog):
     assert expected_message in caplog.text
 
 
-def test_is_old_schema(tmp_path):
+def test_is_different_schema(tmp_path):
     url_zst = "https://community-tc.services.mozilla.com/api/index/v1/task/project.bugbug.data_commits.latest/artifacts/public/prova.json.zst"
     url_version = "https://community-tc.services.mozilla.com/api/index/v1/task/project.bugbug.data_commits.latest/artifacts/public/prova.json.version"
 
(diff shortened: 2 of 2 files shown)


webbdays:
Some tests are failing.
Please help.

webbdays:
Is this the correct way to check database version change ?

bugbug/db.py
```diff
@@ -115,6 +132,22 @@ def upload(path):
 
 
 def last_modified(path):
+
+    # if database version changed then download the database again
```
marco-c, line 136: If the version changed, `last_modified` should return None.

bugbug/db.py
```diff
@@ -54,6 +54,23 @@ def is_old_schema(path):
     return DATABASES[path]["version"] > prev_version
 
 
+# used almost same code as is_old_schema function with small change at return statement.
+# added function docstring.
+def is_db_version_changed(path):
```
marco-c, line 59: Instead of introducing a new function, I think we can just use `is_old_schema` in `last_modified`

webbdays: But, it only checks if db is old or not.
If db is new then only it return true.
But here we need to check if db is changed to old or new version.

bugbug/db.py
webbdays: Returned None in the next commit.

bugbug/db.py
marco-c: It should basically never happen as we always increase version numbers and never decrease them, but to account for the remote possibility we can simply change the `is_old_schema` check to use `!=` instead of `>`.

webbdays: If we replace > with != It will not the same what function name is_old_schema implies(meaning).

marco-c: We can change its name to something like `is_different_schema`

webbdays: Ok

webbdays:
Is there a way to avoid entering PAT every time for push to origin ?
Please let me know.
It will be very helpful.

marco-c:
> Is there a way to avoid entering PAT every time for push to origin ?

What do you mean? What is "PAT"?

webbdays:
PAT => Personal Access Token for GitHub authentication.

marco-c:
You can use SSH, see https://docs.github.com/en/authentication/connecting-to-github-with-ssh/about-ssh.

bugbug/db.py
webbdays: But look at at first 3 lines of download function.
If we use is_different_schema with != in there, 
then there is no download for both old and new db versions.

webbdays:
Ok

bugbug/db.py
marco-c: If the schema version is newer than the schema of the database that is available at the URL, we don't want to download (this is the same behavior as before).
If the schema version is older than the schema of the database that is available at the URL, we also don't want to download (this is not the same behavior as before, but we do want this change of behavior).

webbdays: I didn't understand the second point

marco-c: We have two schema versions, the version that is the local one (the one bugbug is currently using) and the version that is the remote one (the one bugbug used to generate the DB and upload it at the URL).

When we change the version, there will be a mismatch between the local and the remote schema versions and bugbug will have to regenerate the DB from scratch (since the schema is different, bugbug won't be able to use the previous DB that was uploaded at the URL).

webbdays: Ok.
What I understand is :
1. The version with which bugbug previously run is stored in url location and is the old version.
2. When the bugbug  run again locally with same old version then the schema will get downloaded from that previous url.(according to download function.)
3. If used even older versions ,then no download.(download function return false.)
4. Now again when bugbug is run locally with new version, download function is returning false.

So, please clarify what to do if new version is used when bugbug is run again locally ?
Just return last modified as none if new version is used ?

webbdays: Is this is what actually happening ?

marco-c: Let me make an example with numbers so it will hopefully be clearer.

There is a run which generates the DB with the current schema (let's say version 7).

Now, there can be three cases for the following run:
- The schema hasn't changed, so the version is still 7. We can reuse the previous DB, so we download it.
- The schema has been updated, so the version is 8. We can't reuse the previous DB, so we don't download it.
- The schema has been reverted to a previous state, so the version is 6. As above, we can't reuse the previous DB, so we don't download it.

webbdays: Ok, now I understand clearly.
Thank you.

webbdays: ok then
Just return last modified as None if new version ? To fix #2459

marco-c: Return None if different (so either new or old)

webbdays: Ok

webbdays:
Please look at the commit

bugbug/db.py
```diff
@@ -115,13 +115,17 @@ def upload(path):
 
 
 def last_modified(path):
+    if is_different_schema(path):
+        raise Exception("Last-Modified is not available")
+        return None
```
marco-c, line 120: Since we are raising an exception, the `return None` will never be executed. You can remove it

tests/test_db.py
```diff
@@ -412,19 +412,13 @@ def test_is_old_schema(tmp_path):
...
+    assert db.is_different_schema(db_path)
 
-    # When the remote version file doesn't exist, we consider the db as being old.
-    assert db.is_old_schema(db_path)
-
-    # When the remote version file exists and returns the same version as the current db, we consider the remote db as not being old.
-    assert not db.is_old_schema(db_path)
-
-    # When the remote version file exists and returns a newer version than the current db, we consider the remote db as not being old.
-    assert not db.is_old_schema(db_path)
```
marco-c, line 425: You should keep the same assertions as before here, nothing should have changed in these results

tests/test_db.py
```diff
@@ -412,19 +412,13 @@ def test_is_old_schema(tmp_path):
...
+    # When the remote version file exists and returns a newer version than the current db, we consider the remote db as not being old.(different version)
+    assert not db.is_different_schema(db_path)
 
     db.register(db_path, url_zst, 43, support_files=[])
 
-    # When the remote version file exists and returns an older version than the current db, we consider the remote db as being old.
-    assert db.is_old_schema(db_path)
```
marco-c, line 430: You should also add a test with a version that is older than the one returned by requests (you can call db.register with 21 and then assert db.is_different_schema)

tests/test_db.py
webbdays: Both lines are same ..

marco-c: They are the same, you are executing multiple times the same code but with different mocked responses (see the `responses.get` above).

The first response is an error with status code 404, so the DB doesn't exist.
The second response is an error with status code 424, so the DB doesn't exist (we want to make sure not only 404 results in not existing).
The third response is successful with result "1", so the DB exists and has version 1.
The fourth response is successful with result "42", so the DB exists and has version 42.

bugbug/db.py
```diff
@@ -115,13 +115,21 @@ def upload(path):
 
 
 def last_modified(path):
```
marco-c, line 117: I know the issue title was saying to return None in case the schema was different, but let's actually implement the same behavior as when Last-Modified is not available.
So, to summarize, raise the Exception("Last-Modified is not available") instead of returning None.

Also, it would be good to create a custom exception, but we can do that in a follow-up PR.

marco-c: I filed #2459 for the follow-up.

webbdays: Ok

webbdays: What filed #2459 means ?

marco-c: I created the issue #2459 to track the future improvement after your PR is merged.

webbdays:
I didn't understand why this time a lot of tests failed.
Previously there are only two.
I resolved those two.
What happened ?

webbdays:
I think the problem is related to path.

bugbug/db.py
```diff
@@ -115,13 +115,20 @@ def upload(path):
...
-    url = DATABASES[path]["url"]
-    last_modified = utils.get_last_modified(url)
 
-    if last_modified is None:
+    if is_different_schema(path):
         raise Exception("Last-Modified is not available")
 
-    return last_modified
+    else:
```
marco-c, line 122: You can leave the rest of the code as it was, no need to add an `else` branch (when we enter in the `if` branch, we raise, so we never execute anything else).

webbdays: Ok

bugbug/db.py
```diff
@@ -115,13 +115,19 @@ def upload(path):
 
 
 def last_modified(path):
+
+    if is_different_schema(path):
+        raise Exception("Last-Modified is not available")
+
     url = DATABASES[path]["url"]
-    last_modified = utils.get_last_modified(url)
+    last__modified = utils.get_last_modified(url)
```
marco-c, line 123: Why are you changing the names of the variables?

bugbug/db.py
```diff
@@ -115,13 +115,19 @@ def upload(path):
 
 
 def last_modified(path):
+
```
marco-c, line 118: Nit: please remove this white line

tests/test_db.py
```diff
@@ -120,6 +120,28 @@ def test_exists_db(tmp_path):
...
+        responses.HEAD,
+        url_zst,
+        status=200,
+        headers={},
+    )
+
+    # when current db and remote db versions are same, Last-Modified header is not present at the remote db.
+    # then last_modified function should raise Exception.
+    with pytest.raises(Exception, match="Last-Modified is not available"):
+        db.last_modified(db_path)
```
marco-c, line 142: It should *not* raise an exception when the versions are the same

webbdays: If versions are same, utils.get_last_modified function is called.
But in utils.get_last_modified function if "Last-Modified" is not present in headers tuple/list then it is returning None.

And last_modified function raises Exception if None.

So, we need to remove that Exception if None.

bugbug/db.py
marco-c: Let's keep it the same name as before. We could change it, but it should be done in a separate PR (each PR should be self-contained)

webbdays: Ok

bugbug/db.py
webbdays: Done

bugbug/db.py
webbdays: Done

bugbug/db.py
webbdays: Done

tests/test_db.py
webbdays: Should we remove/keep Exception if header not found

webbdays: I think we should raise different exception for it.

marco-c: Ah ok, it makes sense then, let's keep this test but please rename it so it's clearer (`test_last_modified_header_not_present`).

tests/test_db.py
marco-c: @webbdays don't forget about this

tests/test_db.py
marco-c: Actually, it is worth having this test, but it should be added in a separate PR because it's not directly related to version change.
Can you remove it from this PR and add it in a separate PR?

webbdays: Ok

tests/test_db.py
webbdays: That is already included

webbdays:
rebased, removed test_last_modified function.

tests/test_db.py
webbdays: 4th assert

bugbug/db.py
```diff
@@ -119,12 +119,14 @@ def upload(path):
...
-    url = DATABASES[path]["url"]
-    last_modified = utils.get_last_modified(url)
 
-    if last_modified is None:
+    if is_different_schema(path):
         raise LastModifiedNotAvailable()
 
+    url = DATABASES[path]["url"]
+    last_modified = utils.get_last_modified(url)
+    # if no Last-Modified header then that needs to be handled. (in another PR)
```
marco-c, line 128: Please keep the same behavior as before, don't change it. The test can be added in a separate PR, but the behavior here should not change

tests/test_db.py
```diff
@@ -12,7 +12,6 @@
 import responses
 
 from bugbug import db
-from bugbug.db import LastModifiedNotAvailable
```
marco-c, line 15: Please don't change this as part of this PR, it makes the PR less readable (ideally each PR should be as self-contained as possible)

tests/test_db.py
marco-c: Good, I didn't notice.

bugbug/db.py
webbdays: Ok

tests/test_db.py
webbdays: Ok

marco-c (approved):
Thanks!

webbdays:
Thanks for your guidance.

I have learned Exceptions in python.
And also how to write tests.

PR #2654, opened 2021-12-03
Update python version in document

### LINKED ISSUE #2634: Replace typing.Dict with dict

anooptp:
Fixes #2634 

Replacing typing.Dict to dict requires Python 3.9+ version without using `from __future__ import annotations`

On using Python 3.7 or newer, we need to start the module with `from __future__ import annotations` to use `dict` as a generic type. And as of Python 3.9, `dict` supports being used as generic type.


marco-c (approved):
Thanks!

Issue #2664, opened 2021-12-07
Handle last_modified exception everywhere


marco-c:
Currently only bugbug/phabricator.py and bugbug/github_issue_retriever.py are handling `last_modified` exceptions. We should ensure all callers of `last_modified` are behaving well when `last_modified` raises.

PR #2665, opened 2021-12-09
to handle exception in last_modified

webbdays:


webbdays:
Fix for the issue #2664

marco-c (changes requested):
This is incomplete, there is another usage of last_modified in scrits/bug_retriever.py.

Also, the handling of the exception is wrong, we don't want to always just exit (return) in case of exception.

scripts/generate_landings_risk_report.py
```diff
@@ -846,7 +846,12 @@ def go(self, days: int) -> None:
 
         meta_bugs = self.get_meta_bugs(days)
 
-        last_modified = db.last_modified(bugzilla.BUGS_DB)
```
marco-c, line 849: Here we actually want to propagate the exception, as it should never happen here that the Last-Modified is not available.

webbdays:
Ok

scripts/github_issue_retriever.py
```diff
@@ -66,8 +66,9 @@ def retrieve_issues(self) -> None:
 
         try:
             last_modified = db.last_modified(self.github.db_path)
-        except Exception:
-            pass
+        except Exception as e:
+            if str(e) == "Last-Modified is not available":
```
marco-c, line 70: The correct behavior was the previous one. We are in the script which is generating the DB. If a DB with the current schema doesn't exist, we simply want to start from scratch (instead, by `return`ing, you are terminating the generation of the DB).

webbdays: Ok

webbdays: pass for any exception ?

marco-c: Good point. No, only if the exception is "Last-Modified is not available".

scripts/generate_landings_risk_report.py
```diff
@@ -846,7 +846,11 @@ def go(self, days: int) -> None:
...
+        try:
+            last_modified = db.last_modified(bugzilla.BUGS_DB)
+        except Exception:
+            raise
```
marco-c, line 852: No need to catch and re-raise, we can just rely on Python to propagate the exception.

webbdays: Ok

webbdays:
with git diff it shows really bad, 

although I just added try and except statement and a tab by selecting and making code to run only  if last_modified .

But git understand it in different way.
It is showing different - and + lines in its own way.

webbdays:
May be it's my fault not use tab in such a way ?

scripts/generate_landings_risk_report.py
```diff
@@ -847,6 +847,7 @@ def go(self, days: int) -> None:
         meta_bugs = self.get_meta_bugs(days)
 
         last_modified = db.last_modified(bugzilla.BUGS_DB)
+
```
marco-c, line 850: Please remove this unnecessary white line

scripts/bug_retriever.py
```diff
@@ -20,161 +20,172 @@ def retrieve_bugs(self, limit: int = None) -> None:
...
-        commit_bug_ids = list(
-            set(
-                commit["bug_id"]
-                for commit in repository.get_commits()
-                if commit["bug_id"]
-                and dateutil.parser.parse(commit["pushdate"]) >= start_date
+        try:
+            last_modified = db.last_modified(bugzilla.BUGS_DB)
+        except Exception as e:
+            if str(e) == "Last-Modified is not available":
```
marco-c, line 26: When the exception is thrown, last_modified is undefined and the next `if last_modified` will fail.
You should do the same as github_issue_retriever (define `last_modified` as None first), or assign a value to it here

scripts/bug_retriever.py
```diff
@@ -20,161 +20,172 @@ def retrieve_bugs(self, limit: int = None) -> None:
...
-        logger.info(
-            f"{len(regression_related_ids)} bugs which caused regressions fixed by commits."
-        )
-
-        # Get IDs of bugs linked to intermittent failures.
-        test_failure_bug_ids = [
-            item["bug_id"]
-            for item in test_scheduling.get_failure_bugs(
-                two_years_and_six_months_ago, datetime.utcnow()
+            logger.info(f"Retrieved {len(changed_ids)} IDs.")
```
marco-c, line 42: Only the part related to changed_ids should be behind the `if last_modified`, the rest of the code should be executed no matter if `last_modified` has a value or not

webbdays: Ok

scripts/bug_retriever.py
webbdays: Ok

scripts/generate_landings_risk_report.py
webbdays: Done

scripts/generate_landings_risk_report.py
webbdays: Done

scripts/bug_retriever.py
webbdays: Done

scripts/bug_retriever.py
webbdays: Done

scripts/bug_retriever.py
```diff
@@ -17,26 +17,39 @@ class Retriever(object):
...
+        except Exception as e:
+            if str(e) == "Last-Modified is not available":
+                pass
+
+        if last_modified:
+            logger.info(
+                f"Retrieving IDs of bugs modified since the last run on {last_modified}"
             )
-        )
+            changed_ids = set(
```
marco-c, line 34: Here you have the same problem you had before with `last_modified` being undefined.

If `last_modified` is not None, you enter this if and changed_ids has a value.
If `last_modified` is None, you don't enter this if and changed_ids doesn't have a value.
You should add an `else` to the `if last_modified` and set changed_ids to an empty set.

scripts/bug_retriever.py
```diff
@@ -17,26 +17,39 @@ class Retriever(object):
...
-        changed_ids = set(
-            bugzilla.get_ids(
-                {"f1": "delta_ts", "o1": "greaterthaneq", "v1": last_modified.date()}
+        try:
+            last_modified = db.last_modified(bugzilla.BUGS_DB)
+        except Exception as e:
+            if str(e) == "Last-Modified is not available":
+                pass
+
+        if last_modified:
```
marco-c, line 30:
```suggestion
        if last_modified is not None:
```

scripts/bug_retriever.py
```diff
@@ -17,26 +17,39 @@ class Retriever(object):
...
         logger.info(f"Retrieved {len(changed_ids)} IDs.")
 
         all_components = bugzilla.get_product_component_count(9999)
 
         deleted_component_ids = set(
             bug["id"]
             for bug in bugzilla.get_bugs()
-            if "{}::{}".format(bug["product"], bug["component"]) not in all_components
+            if "{}::{}".format(bug["product"], bug["component"])
+            not in all_components
```
marco-c, line 52: Please change this back to the old formatting

webbdays: Actually pre-commit changed it before.
black check in pre-commit is not working properly on my system.
It some times working. But not all the times.

Ok, I will change it manually.

scripts/bug_retriever.py
webbdays: Ok

scripts/bug_retriever.py
webbdays: Done

scripts/bug_retriever.py
webbdays: Done

scripts/bug_retriever.py
webbdays: Done

marco-c:
@webbdays this will need to be rebased on top of master, since #2667 landed.

webbdays:
Ok

scripts/bug_retriever.py
```diff
@@ -17,18 +17,32 @@ class Retriever(object):
...
-        last_modified = db.last_modified(bugzilla.BUGS_DB)
-        logger.info(
-            f"Retrieving IDs of bugs modified since the last run on {last_modified}"
-        )
-        changed_ids = set(
-            bugzilla.get_ids(
-                {"f1": "delta_ts", "o1": "greaterthaneq", "v1": last_modified.date()}
+        try:
+            last_modified = db.last_modified(bugzilla.BUGS_DB)
+        except Exception as e:
```
marco-c, line 26: This will need to change, as the exception thrown by `last_modified` has changed.

scripts/github_issue_retriever.py
```diff
@@ -66,8 +66,9 @@ def retrieve_issues(self) -> None:
 
         try:
             last_modified = db.last_modified(self.github.db_path)
-        except Exception:
-            pass
+        except Exception as e:
```
marco-c, line 69: Same here

scripts/bug_retriever.py
webbdays: Done

scripts/github_issue_retriever.py
webbdays: Done

marco-c (approved):
Thanks!

webbdays:
Thanks.

I think more tests need to be written.

marco-c:
> I think more tests need to be written.

What do you mean?

webbdays:
That codecov is saying that.
But 
I don't now about codecov

marco-c:
Ah, it means the code you changed isn't covered by tests. It's OK though, the code you touched is not super easy to cover by tests.

webbdays:
Ok

webbdays:
Thanks for the merge.

I have learned about Exceptions in python.

There are many logic errors this time, because I am not completely aware of what the bugbug is actually doing.

By next PR I hope I understand what bugbug code logic is, what it is actually doing. I will go through the code.

Thank you for your guidance in explaining what the code is actually doing.

PR #2666, opened 2021-12-09
last_modified exception handling

### LINKED ISSUE #2664: Handle last_modified exception everywhere

anooptp:
Fixes #2664

### DIFF
--- a/scripts/bug_retriever.py
+++ b/scripts/bug_retriever.py
@@ -20,7 +20,12 @@ class Retriever(object):
         db.download(bugzilla.BUGS_DB)
 
         # Get IDs of bugs changed since last run.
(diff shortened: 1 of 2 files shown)


marco-c (changes requested):
The handling of the exception is wrong, we don't want to always just exit (return) in case of exception.

marco-c:
There is an earlier PR to fix the same issue. Being older, the other PR takes priority over this one.

anooptp:
closing the PR as there is another PR for same issue.

PR #2667, opened 2021-12-09
custom-exception

### LINKED ISSUE #2663: Raise a custom exception when Last-Modified is not available
So we don't have to resort to https://github.com/mozilla/bugbug/blob/fa24337fc17c725c3017bc3ad303b4d8e43c7757/bugbug/phabricator.py#L150 but can directly catch the specific exception we want to catch.

anooptp:
Fixes #2663


bugbug/models/exception.py
```diff
@@ -0,0 +1,12 @@
+# -*- coding: utf-8 -*-
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+
+class CustomException(Exception):
```
marco-c, line 7: Let's define this directly in bugbug/db.py, and let's call it "LastModifiedNotAvailable"

marco-c: Doing this should also avoid the circular import problem that you're hitting.

anooptp: implemented the changes mentioned

bugbug/db.py
```diff
@@ -152,6 +152,14 @@ def read(self):
             pass
 
 
+class LastModifiedNotAvailable(Exception):
+    def __init__(self, message):
+        self.message = message
+
+    def __str__(self):
+        return self.message
```
marco-c, line 160: We can implement this as an empty class so it inherits everything from its parent `Exception`:
```
class LastModifiedNotAvailable(Exception):
    pass
```

bugbug/db.py
```diff
@@ -119,7 +119,7 @@ def last_modified(path):
     last_modified = utils.get_last_modified(url)
 
     if last_modified is None:
-        raise Exception("Last-Modified is not available")
+        raise LastModifiedNotAvailable("Last-Modified is not available")
```
marco-c, line 122: The name of the class says it all, we can pass an empty message now:
```suggestion
        raise LastModifiedNotAvailable()
```

bugbug/phabricator.py
```diff
@@ -146,9 +147,9 @@ def download_revisions(rev_ids: Collection[int]) -> None:
 def download_modified_revisions():
     try:
         last_modified = db.last_modified(REVISIONS_DB)
-    except Exception as e:
-        if str(e) == "Last-Modified is not available":
-            return
+    except LastModifiedNotAvailable as e:
+        logger.error("Exception:", e)
```
marco-c, line 151: Let's keep the same behavior as before: no logging

bugbug/db.py
anooptp: implemented

bugbug/phabricator.py
anooptp: excluded logging

bugbug/db.py
```diff
@@ -152,6 +152,10 @@ def read(self):
             pass
 
 
+class LastModifiedNotAvailable(Exception):
```
marco-c, line 155: Nit: can you move this to earlier than the `last_modified` function or at the top of the file? I prefer the definitions to be earlier than the usages.

anooptp: Moved to top of file.
Had kept it there so that all class definitions are together.

marco-c (approved):

marco-c:
Thanks!

anooptp:
Thank you!

webbdays (changes requested):
We can just use db.LastModifiedNotAvailable in test_db.py

webbdays:
@anooptp 
@marco-c 
Can I make that change ?

marco-c:
@webbdays sure, you can submit another PR for that. It's a minor cosmetic change though, not so important :)

webbdays:
Ok, I will do this in the current PR #2665

webbdays:
It's merged, I will do it in new PR.

Issue #2747, opened 2022-02-03
Fix an error in the issue retrieval for private github issues


ksy36:
```
Traceback (most recent call last):
  File "/usr/local/bin/bugbug-data-github", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.9/site-packages/scripts/github_issue_retriever.py", line 154, in main
    retriever.retrieve_issues()
  File "/usr/local/lib/python3.9/site-packages/scripts/github_issue_retriever.py", line 84, in retrieve_issues
    self.replace_with_private(data)
  File "/usr/local/lib/python3.9/site-packages/scripts/github_issue_retriever.py", line 51, in replace_with_private
    private_issue = self.github.fetch_issue_by_number(
  File "/usr/local/lib/python3.9/site-packages/bugbug/github.py", line 152, in fetch_issue_by_number
    data = self.fetch_issues(url=url, retrieve_events=retrieve_events)
  File "/usr/local/lib/python3.9/site-packages/bugbug/github.py", line 74, in fetch_issues
    response.raise_for_status()
  File "/usr/local/lib/python3.9/site-packages/requests/models.py", line 953, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 410 Client Error: Gone for url: https://api.github.com/repos/webcompat/web-bugs-private/issues/47074
[taskcluster 2022-02-02 11:49:11.545Z] === Task Finished ===
[taskcluster 2022-02-02 11:49:11.775Z] Artifact "public/github_webcompat_web-bugs_issues.json.zst" not found at "/data/github_webcompat_web-bugs_issues.json.zst": (HTTP code 404) no such container - Could not find the file /data/github_webcompat_web-bugs_issues.json.zst in container 4b64ef293246e33108d071eebd5f483783aa34153bd52d26bb59e30f9e6fe1ba 
[taskcluster 2022-02-02 11:49:12.338Z] Unsuccessful task run with exit code: 1 completed in 313.376 seconds
```

PR #2748, opened 2022-02-03
Issue #2747 - Fix an error in the issue retrieval for private github issues

ksy36:
The PR addresses the error while fetching private issue contents. The error happens in cases where a private issue is deleted, but a public issue still has a reference to it in the body.

### DIFF
--- a/scripts/github_issue_retriever.py
+++ b/scripts/github_issue_retriever.py
@@ -6,6 +6,8 @@
 import argparse
 from logging import getLogger
 
+import requests
+
 from bugbug import db
 from bugbug.github import Github, IssueDict
 from bugbug.utils import extract_private, zstd_compress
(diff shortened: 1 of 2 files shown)


scripts/github_issue_retriever.py
```diff
@@ -48,14 +50,20 @@ def replace_with_private(
...
+                    )
+
+                    if private_issue:
+                        item["title"] = private_issue["title"]
+                        item["body"] = private_issue["body"]
+                        updated_ids.add(item["id"])
+                        updated_issues.append(item)
+
+                except requests.HTTPError as e:
+                    logger.info(e)
```
marco-c, line 65: Maybe we should re-raise the exception in case of an error that is not 404?

scripts/github_issue_retriever.py
```diff
@@ -48,14 +50,20 @@ def replace_with_private(
...
+                    pass
```
marco-c, line 66: Nit: `pass` is not needed here because you already have at least one statement in this block (`logger.info(e)`).

scripts/github_issue_retriever.py
ksy36: Good point :) I've changed that. Also realized the error is actually "410 Gone" in this case, so I've adjusted the test as well

marco-c (approved):

ksy36:
Thanks!

PR #2749, opened 2022-02-04
Update README.md

MauryaRitesh:
Added "clone repository" in Readme


marco-c:
Thanks, but I think this step is obvious so there's no need to add it in the README.

PR #2806, opened 2022-03-10
Fixes #2805 - Change needsdiagnosis model to consider moved milestone as `needsdiagnosis = True`

### LINKED ISSUE #2805: Change needsdiagnosis model to consider moved milestone as `needsdiagnosis = True`
We have observed that needsdiagnosis model classifies certain issues that potentially need diagnosis as `needsdiagnosis = False`. While this is expected, I think it might be getting worse, as the issues data is unbalanced and has much more data points for `needsdiagnosis = False`. We've started a discussion in https://github.com/mozilla/webcompat-team-okrs/issues/256 

It's worth mentioning that in the [recent sample of 22 issues](https://docs.google.com/spreadsheets/d/1F9vcSpLQ_hNBeZinsytGXlfXpJLW6vh7C0BJYtd9hIY/edit?pli=1#gid=1640243023) most of the issues that looked like false negatives didn't end up needing diagnosis (not reproducible,  out of the scope of the project, etc.), so they're true negatives (prediction was correct for 21 of them). We'll continue tracking them to get a more representative sample.

As an example, issues that looked like they need diagnosis, but in the end, they didn't (prediction was correct):
https://github.com/webcompat/web-bugs/issues/100746
https://github.com/webcompat/web-bugs/issues/100676
https://github.com/webcompat/web-bugs/issues/100687

Issues that are false negatives (prediction was incorrect): 
https://github.com/webcompat/web-bugs/issues/100495
https://github.com/webcompat/web-bugs/issues/100645

I was thinking of including an additional set of issues that will contribute to `needsdiagnosis = True` pool. We have recently added a `moved` [milestone](https://github.com/webcompat/web-bugs/issues?q=is%3Aissue+milestone%3Amoved+is%3Aclosed). These issues often don't need diagnosis and are moved to bugzilla or elsewhere, but their content should be contributed to `needsdiagnosis = True`  rather than false.

ksy36:
I've described the change in detail in https://github.com/mozilla/bugbug/issues/2805

### DIFF
--- a/bugbug/models/needsdiagnosis.py
+++ b/bugbug/models/needsdiagnosis.py
@@ -71,9 +71,9 @@ class NeedsDiagnosisModel(IssueModel):
                 continue
 
             for event in issue["events"]:
-                if (
-                    event["event"] == "milestoned"
-                    and event["milestone"]["title"] == "needsdiagnosis"
+                if event["event"] == "milestoned" and (
+                    event["milestone"]["title"] == "needsdiagnosis"
+                    or event["milestone"]["title"] == "moved"
                 ):
                     classes[issue["number"]] = 0
 


bugbug/models/needsdiagnosis.py
```diff
@@ -71,9 +71,9 @@ def get_labels(self):
                 continue
 
             for event in issue["events"]:
-                if (
-                    event["event"] == "milestoned"
-                    and event["milestone"]["title"] == "needsdiagnosis"
+                if event["event"] == "milestoned" and (
+                    event["milestone"]["title"] == "needsdiagnosis"
+                    or event["milestone"]["title"] == "moved"
```
marco-c, line 76: Nit: `event["milestone"]["title"] in ("needsdiagnosis", "moved")`

marco-c (approved):

ksy36:
Thanks!

PR #2944, opened 2022-06-03
Fix test failures in defect models

suhaibmujahid:
In 816f351388a6de3577936df2a292b4821cdff3cf, we drop `meta` bugs. However, the test has an assertion that checks for Bug 447581, which is a meta bug.


tests/test_bug.py
```diff
@@ -13,7 +13,6 @@ def test_get_bug_labels():
     assert classes[1087488]
     assert not classes[1101825]
     # labels from regression_bug_nobug.csv
-    assert not classes[447581]  # nobug
```
marco-c, line 16: Could you get another one from regression_bug_nobug.csv that is "nobug"?

suhaibmujahid: I searched yesterday, and none of them overlaps with bug IDs in `classes`. We have only 37 bugs in `classes`.

marco-c: They're probably the bugs from the mock file. We might have to add a new one then as the `test_trainer` test is failing too

suhaibmujahid: Done in fb18f14c737c34d39e4340df4ad47f94bc67e18c

marco-c (approved):
Thanks!

PR #3102, opened 2022-10-10
Commit classifier script apply the phabricator patch as its first step

Amotul-raheem:


scripts/commit_classifier.py
```diff
@@ -135,10 +135,16 @@ def __init__(
...
         skip_feature_importance: bool,
+        phabricator_deployment: Optional[str] = None,
+        diff_id: Optional[str] = None,
     ):
-        self.model_name = model_name
         self.repo_dir = repo_dir
 
+        with hglib.open(self.repo_dir) as hg:
+            if phabricator_deployment is not None and diff_id is not None:
+                self.apply_phab(hg, phabricator_deployment, diff_id)
```
suhaibmujahid, line 145: I think we need `self.git_repo_dir` to be set before applying the patch. Having test coverage here would catch such bugs.

scripts/commit_classifier.py
```diff
@@ -135,10 +135,16 @@ def __init__(
         method_defect_predictor_dir: str,
         use_single_process: bool,
         skip_feature_importance: bool,
+        phabricator_deployment: Optional[str] = None,
+        diff_id: Optional[str] = None,
     ):
-        self.model_name = model_name
```
suhaibmujahid, line 139: I do not think it is neccesry to move down ` self.model_name`. It would be better for readability to have the attributes assigned at the top of the constructor function.

PR #3113, opened 2022-10-12
Update python image in task cluster

### LINKED ISSUE #629: Use python slim images where possible in .taskcluster.yml

Amotul-raheem:
fixes #629

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -39,8 +39,9 @@ tasks:
           maxRunTime: 3600
           image: python:3.9-slim
           command:
-            - "/bin/sh"
+            - "/bin/bash"
             - "-lcx"
+            - "apt-get update && apt-get install -y git"
             - "git clone --quiet ${repository} &&
               cd bugbug &&
               git -c advice.detachedHead=false checkout ${head_rev} &&
@@ -63,8 +64,9 @@ tasks:
             maxRunTime: 3600
             image: python:3.9-slim
             command:
-              - "/bin/sh"
+              - "/bin/bash"
               - "-lcx"
+              - "apt-get update && apt-get install -y git"
               - "git clone --quiet ${repository} &&
                 cd bugbug &&
                 git -c advice.detachedHead=false checkout ${head_rev} &&
@@ -86,10 +88,11 @@ tasks:
           env:
             CODECOV_TOKEN: 66162f89-a4d9-420c-84bd-d10f12a428d9
           command:
-            - "/bin/sh"
+            - "/bin/bash"
             - "-lcx"
             - "apt-get -qq update &&
               apt-get -qq install -y libhdf5-dev zstd &&
+              apt-get install -y git &&
               git clone --quiet ${repository} &&
               curl -L https://github.com/mozilla/rust-code-analysis/releases/download/v0.0.23/rust-code-analysis-linux-web-x86_64.tar.gz | tar -C /usr/bin -xzv &&
               cd bugbug &&
@@ -119,8 +122,9 @@ tasks:
           env:
             CODECOV_TOKEN: 66162f89-a4d9-420c-84bd-d10f12a428d9
           command:
-            - "/bin/sh"
+            - "/bin/bash"
             - "-lcx"
+            - "apt-get update && apt-get install -y git"
             - "git clone --quiet ${repository} &&
               curl -L https://github.com/mozilla/rust-code-analysis/releases/download/v0.0.23/rust-code-analysis-linux-web-x86_64.tar.gz | tar -C /usr/bin -xzv &&
               cd bugbug &&
@@ -149,9 +153,10 @@ tasks:
           env:
             CODECOV_TOKEN: 66162f89-a4d9-420c-84bd-d10f12a428d9
           command:
-            - "/bin/sh"
+            - "/bin/bash"
             - "-lcx"
             - "apt-get -qq update &&
+              apt-get install -y git &&
               apt-get -qq install -y libhdf5-dev &&
               git clone --quiet ${repository} &&
               cd bugbug &&
@@ -325,7 +330,7 @@ tasks:
               TC_SECRET_ID: project/bugbug/integration
               CACHE_DIR: "/cache"
             command:
-              - "/bin/sh"
+              - "/bin/bash"
               - "-lcx"
               - "apt-get -qq update &&
                 apt-get -qq install -y redis-server &&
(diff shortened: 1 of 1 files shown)


.taskcluster.yml
```diff
@@ -61,9 +61,9 @@ tasks:
           workerType: batch
           payload:
             maxRunTime: 3600
-            image: python:3.9
+            image: python:3.9-slim
             command:
-              - "/bin/bash"
+              - "/bin/sh"
```
suhaibmujahid, line 66: This seems causing issues. See: https://community-tc.services.mozilla.com/tasks/VKo0l4RKRPiPUo9EwvfFCQ/runs/0/logs/public/logs/live.log#L38

suhaibmujahid (commented):
@Amotul-raheem feel free to covert back the PR when you think it is ready for review.

.taskcluster.yml
Amotul-raheem: got it thank you. sorry for the trouble

suhaibmujahid: no trouble; thank you for the work!

suhaibmujahid:
@Amotul-raheem I would advise covert the images one by one. Also, you do not need to convert all of them, only where possible. Installing the missed tools should not significantly impact the build time compared to using the full Python image.

Note: It could be easier for you to experiment locally (create a Docker file that reproduces the same steps).

Amotul-raheem:
> @Amotul-raheem I would advise covert the images one by one. Also, you do not need to convert all of them, only where possible. Installing the missed tools should not significantly impact the build time compared to using the full Python image.
> 
> Note: It could be easier for you to experiment locally (create a Docker file that reproduces the same steps).

Yes, I did this exact same thing. I created a sample flask server running on docker and I **sh** into the running container using `docker exec it [container_id]  sh` to run test some of the commands to see what dependencies I need to install. 
I think everything is working fine now, but  bugbug tests status is [exception](https://community-tc.services.mozilla.com/tasks/JLwiYru7QtiOo_TapdNWSg) and I'm not sure of what to do. Maybe rerun the task again by making a dummy commit?

Amotul-raheem:
@suhaibmujahid [all tasks](https://community-tc.services.mozilla.com/tasks/groups/ZxwCbM_WRdKGibLwIU5IeA) passed successfully.However, codecov dropped as expected. Please, how do I go about resolving this?

.taskcluster.yml
```diff
@@ -153,8 +156,7 @@ tasks:
             - "-lcx"
             - "apt-get -qq update &&
               apt-get -qq install -y libhdf5-dev &&
-              git clone --quiet ${repository} &&
-              cd bugbug &&
```
marco-c, line 157: Nit: no need to move this to a single line

Amotul-raheem: @marco-c  and @suhaibmujahid Alright, will revert, thanks It's still failing code coverage. Can I add unnittest for [download_model.py ](https://github.com/mozilla/bugbug/blob/master/http_service/bugbug_http/download_models.py) file so as to bump the coverage? I'm thinking i should do it in another branch but it won't solve the problem as the branch lowers the code coverage.

.taskcluster.yml
```diff
@@ -82,14 +84,20 @@ tasks:
...
             - "/bin/bash"
             - "-lcx"
             - "apt-get -qq update &&
               apt-get -qq install -y libhdf5-dev zstd &&
+              apt-get install -y git &&
+              apt-get install -y curl &&
+              apt-get install -y gcc &&
+              apt-get install -y python3-dev &&
+              apt-get install -y g++ &&
+              apt-get install -y mercurial &&
```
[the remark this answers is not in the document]
Amotul-raheem: done, thank you

.taskcluster.yml
suhaibmujahid: > It's still failing code coverage. 

It is ok.

> Can I add unnittest for [download_model.py ](https://github.com/mozilla/bugbug/blob/master/http_service/bugbug_http/download_models.py) file so as to bump the coverage? I'm thinking i should do it in another branch but it won't solve the problem as the branch lowers the code coverage.

Yes, you could file an issue for that and open a PR to resolve it.

suhaibmujahid:
@Amotul-raheem could you evaluate if using slim + apt update + apt install git is more or less efficient than just using the full image?

Amotul-raheem:
> 

could you please explain what you mean by efficient here?

Amotul-raheem:
> @Amotul-raheem could you evaluate if using slim + apt update + apt install git is more or less efficient than just using the full image?

could you please explain what you mean by efficient here?

suhaibmujahid:
> could you please explain what you mean by efficient here?

I mean time to build (how fast)

Amotul-raheem:
@suhaibmujahid I did the analysis and my findings are in the table below. 

| image  | build time | size |
| ------------- | ------------- | ------------- |
| python full image  | 1m 28.463s  | 915MB
| python-slim + apt update + apt install git | 19.566s  | 225MB 

I think python-slim is better as it has less build time and smaller image size

suhaibmujahid:
Thank you, @Amotul-raheem! 

The time saving seems to be highly correlated with the image size. This indicates that the time saving is a reflection of the lower network traffic. I guess these results are based on a local test. I'm not sure if it will be the same on CI. If Taskcluster has any kind of caching or an on-prem image registry, traffic will have a smaller footprint on the overall result.

Amotul-raheem:
> Thank you, @Amotul-raheem!
> 
> The time saving seems to be highly correlated with the image size. This indicates that the time saving is a reflection of the lower network traffic. I guess these results are based on a local test. I'm not sure if it will be the same on CI. If Taskcluster has any kind of caching or an on-prem image registry, traffic will have a smaller footprint on the overall result.

Interesting, thinking about it now it makes sense. 
Should we close the PR and the issue then?

marco-c:
Agreed, timings on Taskcluster might be different. We can compare runs between master and the PR to see what the effect is.

Amotul-raheem:
It made sense to add python slim package to only two tasks; the `bugbug lint test`and `bugbug tag version check`.

After triggering the task cluster using python slim and without python slim, the `bugbug lint test` executed with the same time,(3 minutes).

With this, i don't think using the python slim image is going to have any significant effect on the run time of these processes.

Amotul-raheem:
@suhaibmujahid @marco-c , After these analysis, should the PR and issue be closed?

suhaibmujahid:
Thank you for valuable contrbution here, @Amotul-raheem! I lean toward using the full Python images. A smaller image size could still be slightly benfitule in case we are pushing the built images to the registry. @marco-c wdyt?

marco-c:
I agree, let's keep using the full Python image. Thank you anyway @Amotul-raheem, this was very useful investigation!

PR #3114, opened 2022-10-13
Use  tuple instead of typing.Tuple

### LINKED ISSUE #2652: Use `tuple` instead of `typing.Tuple` in bugbug/test_scheduling.py
We got rid of most typing.Tuple usage in #2640, except https://github.com/mozilla/bugbug/blob/cbaf0de92e96cdb287d461478312b32fc48ef4c1/bugbug/test_scheduling.py#L47-L49 where mypy complains when using `tuple` instead of `Tuple`.

Is it a mypy bug? If so, we should report it and, when it is fixed, we should switch to `tuple`.

Amotul-raheem:
Fixes #2652

### DIFF
--- a/bugbug/test_scheduling.py
+++ b/bugbug/test_scheduling.py
@@ -23,7 +23,6 @@ from typing import (
     NewType,
     Optional,
     Set,
-    Tuple,
     Union,
     cast,
 )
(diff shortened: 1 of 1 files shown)


suhaibmujahid (approved):
LGTM! Thank you @Amotul-raheem.

PR #1917, opened 2020-11-09
Added the feature requested at issue #38

RicardormDev:
Tbh this is my first try on solving an open source issue, I am not even sure if I understood what I needed to do but I don't have anything to lose. 

I just wrote a function to get the delta between the nightly landing and the uplift time. #38
https://github.com/mozilla/bugbug/issues/38

I hope this works.

Issue #38, opened 2018-12-16
Add some timing-related information about patches to the bug features


marco-c:
These might be useful for the uplift model:
- [x] #3118
- [x] Time delta between uplift request and merge day (we can approximate it with the release date of the next version at the time the uplift was requested) #97

marco-c:
The uplift request time can be retrieved from `when` inside `history`.

The landing of the patch in Nightly can be obtained by retrieving the date from Mercurial commits, or the date of the landing comments (https://github.com/mozilla/libmozdata/blob/0574613cc701eb10adff9ac3fcd422bc6efdd914/libmozdata/bugzilla.py#L313). This one is not so easy.

Merge day can be obtained by looking at the release dates, using https://github.com/mozilla/libmozdata/blob/0574613cc701eb10adff9ac3fcd422bc6efdd914/libmozdata/versions.py#L155.

marco-c:
> For the first, (correct me if I'm wrong), we'll have to retrieve comments from the get_landing_comments method, and use this in versions.getCloserRelease to get the date, right?

One option is to use the `get_landing_comments` method, the other is to use the date of the commit. I guess they are equivalent.

RicardormDev:
I made a PR for the first point on this issue (#1917 )

3qiu:
Hi @marco-c, I made a PR for the first task at #3118, let me know if this looks good.

mathewhodson:
@marco-c It looks like this can be closed since code for both tasks were merged.

suhaibmujahid:
Thank you, @mathewhodson!

PR #3118, opened 2022-10-16
Add feature: Time delta between landing of the patch in Nightly and uplift request

3qiu:
According to task 1 in issue #38 , Added time delta between uplift request and merge day, refactored code from (https://github.com/mozilla/bugbug/pull/1917).

The next step would be adding tests in test_bug_features.py, but since this part uses `bugzilla.get_landing_comments` it's not easy to create a local test json file though pytest.fixture.

Let me know what test data should I use for adding tests or current test is fine.


bugbug/bug_features.py
```diff
@@ -209,6 +209,46 @@ def __call__(self, bug, **kwargs):
...
+
+    def __call__(self, bug, **kwargs):
+        for history in bug["history"]:
+            for change in history["changes"]:
+                if change["added"].startswith("approval-mozilla"):
+
+                    if status == "+":
+                        uplift_request_datetime = parser.parse(history["when"])
+
+                        landing_comments = bugzilla.get_landing_comments(
```
suhaibmujahid, line 223:
```suggestion
                        landing_comments = Bugzilla.get_landing_comments(
```

bugbug/bug_features.py
```diff
@@ -11,7 +11,7 @@
 import dateutil.parser
 import pandas as pd
 from dateutil import parser
-from libmozdata import versions
+from libmozdata import bugzilla, versions
```
suhaibmujahid, line 14:
```suggestion
from libmozdata import versions
from libmozdata.bugzilla import Bugzilla
```

suhaibmujahid (commented):
> The next step would be adding tests in test_bug_features.py, but since this part uses `bugzilla.get_landing_comments` it's not easy to create a local test json file though pytest.fixture.

Regarding `get_landing_comments`, the method does not perform any request. So I don't think that you need pytest.fixture for that part.

> Let me know what test data should I use for adding tests or current test is fine.

Adding a test for this is impotent to catch bugs 😉. You could use data from `get_bugs()`:
https://github.com/mozilla/bugbug/blob/98f841de35efac61a91e3293405073e09ad17df5/bugbug/bugzilla.py#L80

bugbug/bug_features.py
3qiu: Corrected code

bugbug/bug_features.py
3qiu: corrected import

3qiu:
Thanks @suhaibmujahid ! I've updated the code to wrap the numbers with pytest.approx(), it should be good now.

bugbug/bug_features.py
```diff
@@ -209,6 +210,45 @@ def __call__(self, bug, **kwargs):
         return None
 
 
+class delta_nightly_request_merge(single_bug_feature):
+    name = "Time delta between landing of the patch in Nightly and uplift request"
+
+    def __call__(self, bug, **kwargs):
+        for history in bug["history"]:
+            for change in history["changes"]:
+                if change["added"].startswith("approval-mozilla"):
```
suhaibmujahid, line 219: This will match uplift requests and uplift approvals as well.
```suggestion
                if change["added"].startswith("approval-mozilla") and change["added"].endswith("?"):
```

Ending with  `?` means the uplift was requested; when the uplift gets approved, it gets changed to `+`. See the example below:
```json
{
    "when": "2019-09-11T19:45:37Z",
    "changes": [
        {
        "attachment_id": 9088191,
        "removed": "approval-mozilla-beta?",
        "added": "approval-mozilla-beta+",
        "field_name": "flagtypes.name"
        }
    ],
    "who": "lhenry@mozilla.com"
},
```

tests/test_bug_features.py
```diff
@@ -165,6 +166,14 @@ def test_comment_length(read):
     read("comment_length.json", comment_length, [566, 5291])
 
 
+def test_delta_nightly_request_merge(read):
+    read(
+        "nightly_uplift.json",
+        delta_nightly_request_merge,
+        [pytest.approx(6.4318), pytest.approx(0.87326389)],
```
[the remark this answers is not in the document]
3qiu: In general it is not a good practice to compare float values over "==", as python internal representation of fractions may be different. I have changed to precise values but still wrapped it in pytest.approx() in case client/developer use different python version or different platform/hardware.

bugbug/bug_features.py
```diff
@@ -209,6 +210,45 @@ def __call__(self, bug, **kwargs):
...
+
+                    if bug["status"] == "VERIFIED":
```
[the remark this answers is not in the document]
3qiu: unfinished code from previous developer included this check, I'm not sure what it represents (which I asked in prior conversations). I guess we don't have to check bug status so I removed this line.

bugbug/bug_features.py
3qiu: accepted changes

bugbug/bug_features.py
```diff
@@ -209,6 +210,45 @@ def __call__(self, bug, **kwargs):
...
+                            if uplift_request_datetime >= landing_time:
+                                curr_delta = uplift_request_datetime - landing_time
+                                if time_delta is None:
+                                    time_delta = curr_delta
+                                else:
+                                    time_delta = min(curr_delta, time_delta)
+                                found = True
+
+                        if found:
+                            return time_delta.days + time_delta.seconds / (24 * 60 * 60)
```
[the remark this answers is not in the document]
3qiu: flipped first if-statement in nested loop, removed bug status check.

bugbug/bug_features.py
```diff
@@ -209,6 +210,46 @@ def __call__(self, bug, **kwargs):
...
+                uplift_request_datetime = parser.parse(history["when"])
+
+                landing_comments = Bugzilla.get_landing_comments(
+                    bug["comments"], ["nightly"]
+                )
+
+                # This will help us to find the closest landing before the uplift request
+
+                time_delta = None
+                found = False
```
suhaibmujahid, line 234: Nit: I would refactor the following code to avoid the need to use the stop flag (i.e., `found`). Making a list of the qualified landings and then using `max()` to get the closest landing before the uplift request could simplify that.

tests/test_bug_features.py
```diff
@@ -165,6 +166,14 @@ def test_comment_length(read):
     read("comment_length.json", comment_length, [566, 5291])
 
 
+def test_delta_nightly_request_merge(read):
+    read(
+        "nightly_uplift.json",
+        delta_nightly_request_merge,
+        [pytest.approx(6.431805555555556), pytest.approx(0.8732638888888888)],
```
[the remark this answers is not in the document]
3qiu: added two test cases to cover cases above

bugbug/bug_features.py
3qiu: Refactored code to list search

bugbug/bug_features.py
```diff
@@ -209,6 +210,40 @@ def __call__(self, bug, **kwargs):
...
+
+                # This will help us to find the closest landing before the uplift request
+                landing_time_list = []
+                for landing in landing_comments:
+                    landing_time = parser.parse(landing["comment"]["creation_time"])
+
+                    # Only accept if the uplift is on the future and
+                    # if the landing_time is greater than the calculated now
+                    if uplift_request_datetime >= landing_time:
+                        landing_time_list.append(landing_time)
```
suhaibmujahid, line 239: It would be more pythonic to use list comprehension here :)

suhaibmujahid (approved):
Thank you, @3qiu! LGTM.

bugbug/bug_features.py
3qiu: Thanks for the advice, it would be same length but a bit more difficult to understand/debug complicated list comprehension; plus, calling parser.parse twice.
here's the code snippet in list comprehension format:


```
landing_time_list = [
                    parser.parse(landing["comment"]["creation_time"]) 
                        for landing in landing_comments 
                        if uplift_request_datetime >= parser.parse(
                            landing["comment"]["creation_time"]
                            ) 
                ]
```
So I would like to keep it as for-loop format.

suhaibmujahid: No need to parse the time at this point; you could keep the time in the string format and parse it at the end when you need to calculate the time delta. 
```python
landing_time_list = [
    landing["comment"]["creation_time"]
    for landing in landing_comments
    # Only accept if the uplift is on the future
    if history["when"] >= landing["comment"]["creation_time"]
]
```

In general, the usage of list comprehension is preferred where applicable, but I have no problem either way.

marco-c (approved):
LGTM! (but please change the title to `Time delta between landing of the patch in Nightly and uplift request`, as "merge day" is the day when a release is made)

PR #3129, opened 2022-10-17
Use wget to download files

marco-c:
It's more efficient and has built-in support for retries

### DIFF
--- a/bugbug/utils.py
+++ b/bugbug/utils.py
@@ -190,14 +190,7 @@ def download_check_etag(url, path=None):
     if old_etag == new_etag:
         return False
 
-    r = requests.get(url, stream=True)
-    r.raise_for_status()
-
-    with open(path, "wb") as f:
-        for chunk in r.iter_content(chunk_size=1048576):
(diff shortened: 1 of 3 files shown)


bugbug/utils.py
```diff
@@ -190,14 +190,7 @@ def download_check_etag(url, path=None):
...
 
-    r = requests.get(url, stream=True)
-    r.raise_for_status()
-
-    with open(path, "wb") as f:
-        for chunk in r.iter_content(chunk_size=1048576):
-            f.write(chunk)
-
-    r.raise_for_status()
+    subprocess.run(["wget", "--progress=dot:giga", url, "-O", path], check=True)
```
suhaibmujahid, line 193: The test fails here because `responses` does not mock requests made using `wget`.

suhaibmujahid (changes requested):

bugbug/utils.py
marco-c: Yeah, and that's going to be painful to change :(

PR #3321, opened 2023-03-06
Include `filed_via` field and bug feature.

jpangas:
This PR is a patch that includes the `filed_via` field. It also includes the field among the bug features in `bug_features.py`

### DIFF
--- a/bugbug/bug_features.py
+++ b/bugbug/bug_features.py
@@ -509,6 +509,11 @@ class op_sys(single_bug_feature):
         return bug["op_sys"]
 
 
+class filed_via(single_bug_feature):
+    def __call__(self, bug, **kwargs):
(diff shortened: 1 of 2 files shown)


jpangas:
I think some of the tests are expected to fail.

suhaibmujahid:
> I think some of the tests are expected to fail.

Tests should not fail because we did not use the feature yet.

suhaibmujahid (approved):
Thank you, @jpangas! LGTM.

jpangas:
> > I think some of the tests are expected to fail.
> 
> Tests should not fail because we did not use the feature yet.

Oh, nice! I will be eager to test the feature in `spam_bug.py` Thanks @suhaibmujahid

Issue #3322, opened 2023-03-06
Refactor print statments to use logger


suhaibmujahid:
Example of a print statement that we want to replace (not limited to the `duplicate.py` file):

https://github.com/mozilla/bugbug/blob/69972a1684f788319bf5c2944bbe8eeb79428c7d/bugbug/models/duplicate.py#L123

Example of a logger that we already use (to replace with):

https://github.com/mozilla/bugbug/blob/69972a1684f788319bf5c2944bbe8eeb79428c7d/scripts/regressor_finder.py#L521

Note: please use lazy % formatting in logging functions (more details can be found in the [pylint docs](https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/logging-fstring-interpolation.html))

WhiteWolf47:
This issue is similar to issue #3323 , i think i can help out with this too :blush:

suhaibmujahid:
The target is to refactor all cases. Most of the cases still need to be refactored.

ishaan-sangwan:
i think i can help out with this

skynette:
Hi @suhaibmujahid can I take on this issue?

suhaibmujahid:
> Hi @suhaibmujahid can I take on this issue?

Yes, you do not need to ask. Please start by reading https://github.com/mozilla/bugbug/issues/1092

BhaswatiRoy:
Hello @suhaibmujahid 
Can I work on refactoring the print statements in version_check.py?

suhaibmujahid:
> Hello @suhaibmujahid Can I work on refactoring the print statements in version_check.py? https://github.com/mozilla/bugbug/blob/0de8c327ad1ed275015231fd1657294eb32cf9b1/infra/version_check.py

@BhaswatiRoy This issue already has an open pull request (see the guidance for first contributors in https://github.com/mozilla/bugbug/issues/1092)



> > Hi @suhaibmujahid can I take on this issue?
> 
> Yes, you do not need to ask. Please start by reading #1092

@skynette Sorry, I just noticed that this issue has a linked pull request that is still open. Please consider working on another issue.

BhaswatiRoy:
> > Hello @suhaibmujahid Can I work on refactoring the print statements in version_check.py? https://github.com/mozilla/bugbug/blob/0de8c327ad1ed275015231fd1657294eb32cf9b1/infra/version_check.py
> 
[...]

PR #3325, opened 2023-03-06
Adding logger in duplicate.py

singla007:


suhaibmujahid:
Thank you @singla007 for submitting the PR. I think you are working on this as part of https://github.com/mozilla/bugbug/issues/3322.

Could you pleas take care of the the follwoing note from the issue:
> Note: please use lazy % formatting in logging functions (more details can be found in the [pylint docs](https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/logging-fstring-interpolation.html))

To avoid lint failures, it is recommended to setup `pre-commit`: https://github.com/mozilla/bugbug#auto-formatting

singla007:
Hi @suhaibmujahid , thanks for the review, I will update the code according to the suggestion.

bugbug/models/duplicate.py
```diff
@@ -120,7 +124,7 @@ def get_labels(self):
         # Store all remaining ids
         non_duplicate_ids = list(all_ids - set(duplicate_ids))
 
-        print(f"Number of duplicate labels is: {self.num_duplicates}")
+        logger.info("Number of duplicate labels is: %s", self.num_duplicates)
```
suhaibmujahid, line 127:
```suggestion
        logger.info("Number of duplicate labels is: %d", self.num_duplicates)
```


When the value is an integer, please use `%d`. Please apply the same change where applicable.

singla007: thanks for pointing this mistake out, i will do the changes

suhaibmujahid (approved):
Thank you, @singla007! LGTM.

PR #3341, opened 2023-03-08
Create a class to calibrate XGBoost models

### LINKED ISSUE #2828: Calibrate xgboost regressor model to make it isotonic
So its output can be interpreted as a confidence level.

WhiteWolf47:
fixes #2828 
This PR aims to calibrate the probabilities which the XGBoost model predicts.
For this, I've used the IsotonicRegression model from the sklearn.isotonic module.


bugbug/model.py
```diff
@@ -455,6 +456,19 @@ def train(self, importance_cutoff=0.15, limit=None):
         y_pred = self.clf.predict(X_test)
         y_pred = self.le.inverse_transform(y_pred)
 
+        # calibrating the model
+
+        # Fit isotonic regression model to the predicted probabilities
+        iso_reg = IsotonicRegression(out_of_bounds="clip")
+        iso_reg.fit(y_pred, y_test)
```
suhaibmujahid, line 463: The testing data should not be used to train the model.

bugbug/model.py
```diff
@@ -455,6 +456,19 @@ def train(self, importance_cutoff=0.15, limit=None):
...
+
+        # Use the isotonic regression model to transform the predicted probabilities
+        calibrated_y_pred = iso_reg.transform(y_pred)
```
suhaibmujahid, line 466: The calibrated probabilities are still not used

bugbug/model.py
```diff
@@ -455,6 +456,19 @@ def train(self, importance_cutoff=0.15, limit=None):
         y_pred = self.clf.predict(X_test)
         y_pred = self.le.inverse_transform(y_pred)
 
+        # calibrating the model
```
suhaibmujahid, line 459: We should have the calibrating process as optional. It could be enabled per model after checking the performance with and without enabling it and making sure that it improves the performance.

suhaibmujahid (commented):
Thank you, @WhiteWolf47! Please see my comments.

bugbug/model.py
WhiteWolf47: But we dont have any Validation data and the IsotonicRegression model needs to be trained on the predictions made by the XGBoost Model on the validation data

bugbug/model.py
WhiteWolf47: So everywhere i should replace the probabilities of the  XGBoost model by the calibrated probabilities?

scripts/regressor_finder.py
```diff
@@ -64,14 +64,16 @@ def __init__(
 
         with ThreadPoolExecutorResult(max_workers=3) as executor:
             if self.git_repo_url is not None:
-                logger.info(f"Cloning {self.git_repo_url} to {self.git_repo_dir}...")
+                logger.info("Cloning %s to %s...", self.git_repo_url, self.git_repo_dir)
```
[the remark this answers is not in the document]
WhiteWolf47: My bad !

bugbug/model.py
WhiteWolf47: I have no idea how to do that :sweat_smile: . Can you please guide me to do so.

WhiteWolf47: We can compare the Mean Square error of both without and with Calibration and then choose the better performing model, how bout this?

scripts/regressor_finder.py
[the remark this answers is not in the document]
suhaibmujahid: Still, there are unrelated changes in `scripts/regressor_finder.py`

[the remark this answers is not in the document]
suhaibmujahid: no, changing logs is not related to this work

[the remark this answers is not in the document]
WhiteWolf47: the logs are same as the master branch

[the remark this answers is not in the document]
WhiteWolf47: should i do the normal logs here or the lazy format ones, im confused rn :smiling_face_with_tear:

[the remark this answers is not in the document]
suhaibmujahid: you should not change the logs :)

bugbug/model.py
suhaibmujahid: > Using the training data to fit the IsotonicRegression model can lead to overfitting!

Yes, so we need to create validation data. Currently, we split the data into train/test. Instead, we could split it into train/validation/test.

bugbug/model.py
suhaibmujahid: We need to output the calibrated probabilities to be used as a confidence level. 

See: https://github.com/mozilla/bugbug/blob/f9906057a5281b8913fb5a92edbe73440953581b/bugbug/model.py#L584

bugbug/model.py
suhaibmujahid: >  have no idea how to do that 😅 . Can you please guide me to do so.

We could do something similar to the cross validation:

https://github.com/mozilla/bugbug/blob/f9906057a5281b8913fb5a92edbe73440953581b/bugbug/model.py#L368-L387

bugbug/model.py
WhiteWolf47: Great! I'll do that.

WhiteWolf47: @suhaibmujahid how much percentage of data should i have in train, val and test, is there a specific number you would recommend or should I do it as I see fit?

suhaibmujahid: 30% of the training set would be ok

bugbug/model.py
```diff
@@ -356,8 +357,10 @@ def train(self, importance_cutoff=0.15, limit=None):
         is_multilabel = isinstance(y[0], np.ndarray)
         is_binary = len(self.class_names) == 2
 
-        # Split dataset in training and test.
-        X_train, X_test, y_train, y_test = self.train_test_split(X, y)
```
suhaibmujahid, line 360: We should keep using `self.train_test_split()` - it is overridden in some models. 

You could create a similar method and call it `train_validation_split`.

bugbug/model.py
WhiteWolf47: i did 80% training, 10% validation and 10% test!

bugbug/model.py
WhiteWolf47: so I should use self.train_test_split() to split the data or I should create another method train_validation_split to split the data?

WhiteWolf47: Also the self.train_test_split() doesnt a parameter to specify the size of splitting

WhiteWolf47: A better solution imo can be to use Fast_ml's ‘train_valid_test_split’ , it splits the data into train/validation/test in a single line of code

suhaibmujahid: You can create a method called `train_validation_split`; then give it the training data from `self.train_test_split()`:

```py
X_train, X_test, y_train, y_test = self.train_test_split(X, y)
X_train, X_val, y_train, y_val = self.train_validation_split(X_train, y_train)
```

WhiteWolf47:
@suhaibmujahid can you take a look as to why the tests are failing?

WhiteWolf47:
@suhaibmujahid the checks are still failing :smiling_face_with_tear: . This time the logs are showing -
"FAILED tests/test_trainer.py::test_trainer - ZeroDivisionError: division by zero"

suhaibmujahid:
> @suhaibmujahid the checks are still failing 🥲 . The logs are showing - "FAILED tests/test_trainer.py::test_trainer - ZeroDivisionError: division by zero"

This could be due to the change in the split for the training data (having fewer data points).

WhiteWolf47:
> > @suhaibmujahid the checks are still failing smiling_face_with_tear . The logs are showing - "FAILED tests/test_trainer.py::test_trainer - ZeroDivisionError: division by zero"
> 
> This could be due to the change in the split for the training data (having fewer data points).

what could be a possible solution then? What if we only use self.train_test_split (like use it two times), we dont really need a train_validation_split method tbh, we can add a positional argument to the existing train_test_split method like test_size to specify the percentage of 

def train_test_split(self, X, y, test_size=0.1):
        return train_test_split(X, y, test_size= test_size, random_state=0)

this way we can have a default test_size of 0.1 and if we want we can pass a different test_size.

WhiteWolf47:
@suhaibmujahid the checks passed. Instead of splitting the train set into train and validation, I splitted the test set into test and validation :smile:

bugbug/model.py
WhiteWolf47: @suhaibmujahid im not really sure how to do that!

suhaibmujahid: You need to use the trained isotonic model here. Our goal is to use its output as a confidence level.

suhaibmujahid:
> @suhaibmujahid the checks passed. Instead of splitting the train set into train and validation, I splitted the test set into test and validation 😄

I do not think that in this way the test and validation data are enough. Instead, a possible solution to avoid the test failure is to apply https://github.com/mozilla/bugbug/pull/3341#discussion_r1129887877, then enable it gradually for models that does not cause a failure.

WhiteWolf47:
> > @suhaibmujahid the checks passed. Instead of splitting the train set into train and validation, I splitted the test set into test and validation smile
> 
> I do not think that in this way the test and validation data are enough. Instead, a possible solution to avoid the test failure is to apply [#3341 (comment)](https://github.com/mozilla/bugbug/pull/3341#discussion_r1129887877), then enable it gradually for models that does not cause a failure.

I dont get why the test and validation data would not be enough? If this is something related to the amount of test and validation data? If so,  then we can perhaps increase the partition size, because splitting the train into train and test is same as splitting the test into test and validation sets as long as the partition size is appropriate. And yes we should do something like [this](https://github.com/mozilla/bugbug/pull/3341#discussion_r1129887877).

bugbug/model.py
WhiteWolf47: can you tell me what exactly is the function "classify" doing and how can I use the calibrated probabilities as confidence level in this function :sweat_smile: :

WhiteWolf47: ok so I tried something, just take a look!

suhaibmujahid:
> I dont get why the test and validation data would not be enough? If this is something related to the amount of test and validation data, then we can perhaps increase the partition size, because splitting the train into train and test is same as splitting the test into test and validation sets as long as the partition size is appropriate. 

In the current split, the testing data is 10%, by taking 30% of that for the validation, we end up with 3% of the data for the validation and 7% for testing (which is low for both). We need to be careful before changing the split and make sure that changes do not heart the quality of any of the 24 models. 


> And yes we should do something like [this](https://github.com/mozilla/bugbug/pull/3341#discussion_r1129887877), also if we only enable it for models that does not cause a failure then it would leave some models without calibration!

Not all of our models are using XGBoost and some models even uses multi class classifiers. The way we do the calibration will be different from a model to other. Also, the calibration is not necessary needed for all models. For that, I think the calibration should be applied within the models that needed it. 

I suggest to do the calibration in wrapper class to have it independent from the `Model` class. Then the wrapper class can be used within the individual models to wrap the classifier. Assumenig that we name the wrapper class `IsotonicRegressionCalibrator`, then we will refactor the following line:
https://github.com/mozilla/bugbug/blob/f9906057a5281b8913fb5a92edbe73440953581b/bugbug/models/devdocneeded.py#L75

to be:
```py
        self.clf = IsotonicRegressionCalibrator(
            xgboost.XGBClassifier(n_jobs=utils.get_physical_cpu_count())
        )
```

By doing this, we will not need to change anything in the `bugbug/model.py` file. Spliting the training set to have part of it used for the calibration will be done in the wrapper class.



In all cases, we must evaluate the performance before and after calibrating the probabilities.

WhiteWolf47:
> > I dont get why the test and validation data would not be enough? If this is something related to the amount of test and validation data, then we can perhaps increase the partition size, because splitting the train into train and test is same as splitting the test into test and validation sets as long as the partition size is appropriate.
> 
> In the current split, the testing data is 10%, by taking 30% of that for the validation, we end up with 3% of the data for the validation and 7% for testing (which is low for both). We need to be careful before changing the split and make sure that changes do not heart the quality of any of the 24 models.
> 
> > And yes we should do something like [this](https://github.com/mozilla/bugbug/pull/3341#discussion_r1129887877), also if we only enable it for models that does not cause a failure then it would leave some models without calibration!
> 
> Not all of our models are using XGBoost and some models even uses multi class classifiers. The way we do the calibration will be different from a model to other. Also, the calibration is not necessary needed for all models. For that, I think the calibration should be applied within the models that needed it.
> 
> I suggest to do the calibration in wrapper class to have it independent from the `Model` class. Then the wrapper class can be used within the individual models to wrap the classifier. Assumenig that we name the wrapper class `IsotonicRegressionCalibrator`, then we will refactor the following line:
> 
> https://github.com/mozilla/bugbug/blob/f9906057a5281b8913fb5a92edbe73440953581b/bugbug/models/devdocneeded.py#L75
> 
> to be:
> 
> ```python
>         self.clf = IsotonicRegressionCalibrator(
>             xgboost.XGBClassifier(n_jobs=utils.get_physical_cpu_count())
>         )
> ```
> 
> By doing this, we will not need to change anything in the `bugbug/model.py` file. Spliting the training set to have part of it used for the calibration will be done in the wrapper class.
> 
> In all cases, we must evaluate the performance before and after calibrating the probabilities

I had already modified the partition! Firstly we are splitting the data into training and testing as 70%-training and 30%-testing, then we are splitting the testing data into testing and validation as 50%-testing and 50%-validation. So the partition comes out to be 70%-training, 15%-validation and 15%-testing.

As for your suggestion regarding creating a wrapper class, I think its a great idea and I'll start working in that direction!
Also just to be clear in the model.py , I should revert everything back to normal?

WhiteWolf47:
@suhaibmujahid so i reverted everything in model.py back to normal and will start working on a new wrapper class!

WhiteWolf47:
@suhaibmujahid so i did a rough attempt at the wrapper class, just take a look :sweat_smile:

suhaibmujahid:
> @suhaibmujahid so i did a rough attempt at the wrapper class, just take a look 😅

Good start. There are som issues that I'm sure you would solve if you run the code. Please, apply this to one of the models to run the code, solve the issues, and compare the performance before and after.

WhiteWolf47:
> > @suhaibmujahid so i did a rough attempt at the wrapper class, just take a look sweat_smile
> 
> Good start. There are som issues that I'm sure you would solve if you run the code. Please, apply this to one of the models to run the code, solve the issues, and compare the performance before and after.

you mean like refactor - " self.clf = xgboost.XGBClassifier(n_jobs=utils.get_physical_cpu_count()) " to "self.clf = IsotonicRegressionCalibrator(
            xgboost.XGBClassifier(n_jobs=utils.get_physical_cpu_count())
        )
" and then run the code?

WhiteWolf47:
@suhaibmujahid so i tried applying the wrapper to the xgboost model and resolved an error which was coming and it passed all the checks too!

suhaibmujahid (commented):
> by running the code, you mean running the pre-commit tests right?

No, I mean training a model to execute the new changes. 

BTW. The `regressor` model does not take much time to train (relatively).

bugbug/calibration_wrapper.py
```diff
@@ -0,0 +1,40 @@
...
+        self.y_train, self.y_val, self.y_test = None, None, None
+        self.ir = None
+
+    def fit(self, X, y):
+        self.X_train, self.X_test, self.y_train, self.y_test = train_test_split(
+            X, y, test_size=0.2, random_state=42
+        )
+        self.X_train, self.X_val, self.y_train, self.y_val = train_test_split(
+            self.X_train, self.y_train, test_size=0.2, random_state=42
+        )
```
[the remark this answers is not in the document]
WhiteWolf47: Wasn't I supposed to split the data into three parts? because we would be doing the calibration the validation set! 
Also i dont get wydm by this "What you get in the fit() method is always the training set."

bugbug/calibration_wrapper.py
```diff
@@ -0,0 +1,40 @@
...
+        if self.calibrated:
+            return self.ir.predict(self.model.predict(X))
+        else:
+            return self.model.predict(X)
+
+    def calibrate(self):
+        if not self.calibrated:
+            self.ir = IsotonicRegression()
+            self.ir.fit(self.model.predict(self.X_val), self.y_val)
+            self.calibrated = True
```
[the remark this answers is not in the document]
WhiteWolf47: I thought it would be clean to separate them in different methods!

bugbug/models/devdocneeded.py
```diff
@@ -10,6 +10,7 @@
 from sklearn.pipeline import Pipeline
 
 from bugbug import bug_features, bugzilla, feature_cleanup, utils
+from bugbug.calibration_wrapper import IsotonicRegressionCalibrator
```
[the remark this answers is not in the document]
WhiteWolf47: FIne by me :smile:

bugbug/model.py
```diff
@@ -451,7 +451,7 @@ def train(self, importance_cutoff=0.15, limit=None):
             )
 
         print("Test Set scores:")
-        # Evaluate results on the test set.
+        # Evaluating results on the test set.
```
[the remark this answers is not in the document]
WhiteWolf47: mb

bugbug/models/devdocneeded.py
```diff
@@ -72,8 +73,10 @@ def __init__(self, lemmatization=False):
             ]
         )
 
-        self.clf = xgboost.XGBClassifier(n_jobs=utils.get_physical_cpu_count())
-        self.clf.set_params(predictor="cpu_predictor")
+        # wrapping the model with IsotonicRegressionCalibrator
+        base_clf = xgboost.XGBClassifier(n_jobs=utils.get_physical_cpu_count())
+        base_clf.set_params(predictor="cpu_predictor")
+        self.clf = IsotonicRegressionCalibrator(base_clf)
```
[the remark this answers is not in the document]
WhiteWolf47: > > by running the code, you mean running the pre-commit tests right?
> 
> No, I mean training a model to execute the new changes.
> 
> BTW. The `regressor` model does not take much time to train (relatively).

Where is the data to train this on? Is there a script for that, im not really sure how to do that :sweat_smile:

bugbug/calibration_wrapper.py
```diff
@@ -0,0 +1,40 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from sklearn.isotonic import IsotonicRegression
+from sklearn.model_selection import train_test_split
+
+
+# Wrapper class to calibrate a model
+class IsotonicRegressionCalibrator:
+    def __init__(self, model_class):
```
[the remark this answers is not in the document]
WhiteWolf47: wdym? Cna you please suggest the change :sweat_smile: ?

bugbug/models/devdocneeded.py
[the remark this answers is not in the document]
suhaibmujahid: > Where is the data to train this on? 

The data will be downloaded automatically. 

> Is there a script for that, im not really sure how to do that 😅

It is explained in the readme file: https://github.com/mozilla/bugbug#usage

bugbug/calibration_wrapper.py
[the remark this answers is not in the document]
suhaibmujahid: The data is already splitted into train/test before being passed to the `fit()` method. So all the data that we have here is training data. We need to split it into train/val.

bugbug/model_calibration.py
```diff
@@ -0,0 +1,43 @@
...
+        self.X_train, self.X_test, self.y_train, self.y_test = train_test_split(
+            X, y, test_size=0.2, random_state=42
+        )
+        self.X_train, self.X_val, self.y_train, self.y_val = train_test_split(
+            self.X_train, self.y_train, test_size=0.2, random_state=42
+        )
+
+    def fit(self):
+        self.model.fit(self.X_train, self.y_train)
+        # invoking calibrate() to calibrate the model
```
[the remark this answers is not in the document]
WhiteWolf47: mb

bugbug/model_calibration.py
```diff
@@ -0,0 +1,43 @@
...
+
+from sklearn.isotonic import IsotonicRegression
+from sklearn.model_selection import train_test_split
+
+
+# Wrapper class to calibrate a model
+class IsotonicRegressionCalibrator:
+    def __init__(self, model):
+        self.model = model
+        self.calibrated = False
```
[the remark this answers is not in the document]
WhiteWolf47: one is used to store the model instance that is passed to it and self.calibrated is used to denote whether calibration is enabled for that model or not!

bugbug/model_calibration.py
```diff
@@ -0,0 +1,43 @@
...
+        self.X_train, self.X_val, self.X_test = None, None, None
+        self.y_train, self.y_val, self.y_test = None, None, None
+        self.ir = None
+
+    def split_data(self, X, y):
```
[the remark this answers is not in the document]
WhiteWolf47: so where should I split the data?

WhiteWolf47:
@suhaibmujahid okay so did some changes based on your reviews, so now I just have to train the model and check performance with and without calibration right?

bugbug/model_calibration.py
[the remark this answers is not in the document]
suhaibmujahid: I was asking about `self.calibrated`. We do not need to store that status. Every time the method `fit()` is called, we should call the `fit()` method on `IsotonicRegression`. Do we have a reason not to do so?

bugbug/model_calibration.py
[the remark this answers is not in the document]
suhaibmujahid: > This is not called anywhere.

This code is not used.

> why do we need to store the datasets in the class itself?

Why do we need to store the datasets since they will be used just once to fit?

suhaibmujahid:
> @suhaibmujahid okay so did some changes based on your reviews, 

Thank you!

> so now I just have to train the model and check performance with and without calibration right? 

This will allow you to catch some of the bugs in this PR by yourself.

> Also where should i use the MSE to compare the performance?, like i used the wrapper in the devdocneeded.py file so how I should I modify that file to compare the performance? 

Once you train the model it will show some reports.

> Also you mentioned that regressor model is easier to train, 

All models are easy to train ([it is just one command](https://github.com/mozilla/bugbug#usage)). The regressor model is known to be fast to be trained. 

> how should I wrap that to using IsotonicRegressionCalibrator cos im not really sure how to do that 😅 ?

The same way you did in `devdocneeded`.

bugbug/model_calibration.py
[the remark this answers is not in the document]
WhiteWolf47: split_data() is called in the train method!

bugbug/model_calibration.py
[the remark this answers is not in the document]
WhiteWolf47: okay so i removed the calibrated attribute. And you mean that the fit() method of the IsotonicRegressionCalibrator should directly fit the data on the IsotonicRegression method and not on the normal model?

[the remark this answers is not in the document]
suhaibmujahid: It should do both; that is the idea of the wrapper class. With this, we will not need to change the workflow, we will get calibrated probabilities out of the box by only using `IsotonicRegressionCalibrator` to wrap a classifier.

bugbug/model_calibration.py
```diff
@@ -0,0 +1,41 @@
...
+        self.calibrate(X_train, y_train)
+
+    def predict(self, X):
+        if self.calibrator is None:
+            return self.model.predict(X)
+        else:
+            return self.calibrator.predict(self.model.predict(X))
+
+    def calibrate(self, X_val, y_val):
+        self.calibrator = IsotonicRegression()
```
suhaibmujahid, line 35: It would be better to instantiate `IsotonicRegression` in the constructor.

bugbug/model_calibration.py
```diff
@@ -0,0 +1,41 @@
...
+            X_train_val, y_train_val, test_size=0.2, random_state=42
+        )
+        return X_train, X_val, X_test, y_train, y_val, y_test
+
+    def fit(self, X_train, y_train):
+        self.model.fit(X_train, y_train)
+        self.calibrate(X_train, y_train)
+
+    def predict(self, X):
+        if self.calibrator is None:
```
suhaibmujahid, line 29: We don't need to check this here. If the calibrator was instantiated in the constructor, the value will never be `None`. We could rely on `self.model.predict(X)` to raise an error if this gets called before calling `fit()`.

bugbug/model_calibration.py
```diff
@@ -0,0 +1,41 @@
...
+            X, y, test_size=0.2, random_state=42
+        )
+        X_train, X_val, y_train, y_val = train_test_split(
+            X_train_val, y_train_val, test_size=0.2, random_state=42
+        )
+        return X_train, X_val, X_test, y_train, y_val, y_test
+
+    def fit(self, X_train, y_train):
+        self.model.fit(X_train, y_train)
+        self.calibrate(X_train, y_train)
```
suhaibmujahid, line 26: The same data is used to train the classifier and the calibrator.

bugbug/model_calibration.py
WhiteWolf47: ok!

bugbug/model_calibration.py
WhiteWolf47: got it!

bugbug/model_calibration.py
WhiteWolf47: mb!

WhiteWolf47:
@suhaibmujahid so im trying to train and compare the regressor model's performance with and without calibration.
I'm wrapping the regressor model as follows - 
class RegressionModel(DefectModel):
    def __init__(self, lemmatization=False, historical=False):
        self.calculate_importance = False
        self.clf = IsotonicRegressionCalibrator(DefectModel.__init__(self, lemmatization, historical))

but running the training script gives me this error  - 
TypeError: fit() missing 2 required positional arguments: 'X_val' and 'y_val'
it traces the error back to model.py in the 374th line  - "scores = cross_validate(
                pipeline, X_train, self.le.transform(y_train), scoring=scorings, cv=5
            )"
I tried debugging it but couldnt find a soln!

WhiteWolf47:
@suhaibmujahid so im trying to train the regressor model to compare the performance with and without the calibration, but running the training script gives me the following error - 
"TypeError: fit() missing 2 required positional arguments: 'X_val' and 'y_val'"
it traces the error back to model.py at the 374th line i.e. 
scores = cross_validate(
                pipeline, X_train, self.le.transform(y_train), scoring=scorings, cv=5
            )
i tried debugging it but couldnt find a soln, do you have any idea?

bugbug/model_calibration.py
```diff
@@ -0,0 +1,42 @@
...
+    def split_data(self, X, y):
+        X_train_val, X_test, y_train_val, y_test = train_test_split(
+            X, y, test_size=0.2, random_state=42
+        )
+        X_train, X_val, y_train, y_val = train_test_split(
+            X_train_val, y_train_val, test_size=0.2, random_state=42
+        )
+        return X_train, X_val, X_test, y_train, y_val, y_test
+
+    def fit(self, X_train, y_train, X_val, y_val):
```
suhaibmujahid, line 25: > @suhaibmujahid so im trying to train the regressor model to compare the performance with and without the calibration, but running the training script gives me the following error -
"TypeError: fit() missing 2 required positional arguments: 'X_val' and 'y_val'"
it traces the error back to model.py at the 374th line i.e.
scores = cross_validate(
pipeline, X_train, self.le.transform(y_train), scoring=scorings, cv=5
)
i tried debugging it but couldnt find a soln, do you have any idea?

The signature of the `fit()` method here is not matching the scikit-learn API.

WhiteWolf47: which fit() method? the one in the IsotonicRegressionCalibrator class?

WhiteWolf47: Also i have to make two more methods predict_proba() and predict() only right?

WhiteWolf47: Also how to make the api for the class? just having  those three methods you mentioned fit(), predict_proba() and predict() will do the work right?

suhaibmujahid: > but the fit() method of the IsotonicRegressionCalibrator class has positional arguments 'X_val' and 'y_val'" , so why its showing "TypeError: fit() missing 2 required positional arguments: 'X_val' and 'y_val'"?

This is the problem, the `fit()` method on `IsotonicRegressionCalibrator` is not compatible, it asks for 'X_val' and 'y_val' where a scikit-learn compatible API should not.


> Also how to make the api for the class? just having those three methods you mentioned fit(), predict_proba() and predict() will do the work right?

The public methods on a class are the API. You should try and investigate more yourself.

WhiteWolf47:
@suhaibmujahid so im resolved those errors and i trained the model . The calibrated model is performing better than the normal one -
MSE of model before calibration: 0.1391
MSE of model after calibration: 0.1176
(MSE - Mean Squared Error)
But the scorings is giving error now - 
shap.utils._exceptions.InvalidModelError: Model type not yet supported by TreeExplainer: <class 'bugbug.model_calibration.IsotonicRegressionCalibrator'>

suhaibmujahid:
> @suhaibmujahid so I resolved those errors and i trained the model . The calibrated model is performing better than the normal one - MSE of model before calibration: 0.1391 MSE of model after calibration: 0.1176 (MSE - Mean Squared Error)

We need to compare the performance report for the model before and after the changes. This includes precision, recall, F1, etc...

> But the scorings is giving error now - shap.utils._exceptions.InvalidModelError: Model type not yet supported by TreeExplainer: <class 'bugbug.model_calibration.IsotonicRegressionCalibrator'>

When you have a problem you should not just post it here. You should investigate and try to solve it first.

For the problem that you mentioned, as a workaround, you could disable the part that is causing the error by adding the following to the constructor of the `regressor` model:

```
self.calculate_importance = False
```

WhiteWolf47:
@suhaibmujahid so i tried the workaround you suggested but it didnt work, so i searched more and the it turned out we need to transform the output probabilities back to class labels before calculating the classification metrics, so I did that in the predict() method of IsotonicRegressionCalibrator as -
def predict(self, X):
        p_pred = self.calibrator.predict(self.model.predict_proba(X)[:, 1])

        #transforming predictions back to class labels
        p_pred = p_pred.flatten()
        y_pred = np.where(p_pred > 0.5, 1, 0)

        return y_pred
 but now it traces a new error -                                                      
    IndexError: invalid index to scalar variable.
back to model.py in the 504th line i.e.
    argmax = 1 if y_pred_probas[i][1] > confidence_threshold else 0
i tried declaring argmax as a numpy array before storing values in it by doing -        
    argmax = 1 if y_pred_probas[i][1] > confidence_threshold else 0
but it still gives the same error!

bugbug/models/regression.py
suhaibmujahid, line 1: Unrelated changes!

bugbug/model_calibration.py
```diff
@@ -0,0 +1,50 @@
...
+    def predict(self, X):
+        p_pred = self.calibrator.predict(self.model.predict_proba(X)[:, 1])
+        # transforming predictions back to class labels
+        p_pred = p_pred.flatten()
+        y_pred = np.where(p_pred > 0.5, 1, 0)
+
+        return y_pred
+
+    def predict_proba(self, X_val):
+        return self.calibrator.transform(self.model.predict_proba(X_val)[:, 1])
```
suhaibmujahid, line 42: Are we returning the expected shape here?

bugbug/model_calibration.py
```diff
@@ -0,0 +1,50 @@
...
+        print(f"MSE of model before calibration : {mse_before:.4f}")
+
+        self.calibrate(X_val, y_val)
+        mse_after = mean_squared_error(y_val, self.predict(X_val))
+        print(f"MSE of model after calibration : {mse_after:.4f}")
+
+    def predict(self, X):
+        p_pred = self.calibrator.predict(self.model.predict_proba(X)[:, 1])
+        # transforming predictions back to class labels
+        p_pred = p_pred.flatten()
```
suhaibmujahid, line 36: Why do we need `flatten` here?

WhiteWolf47:
@suhaibmujahid so i used sklearn.calibration.CalibratedClassifierCV instead of IsotonicRegression model and for now set  self.calculate_importance = False  (i figured out that IsotonicRegressionCalibrator model type is not currently supported by the TreeExplainer function in the shap library thats why i was getting the "Model type not yet supported by TreeExplainer: <class 'bugbug.model_calibration.IsotonicRegressionCalibrator'>" error ). The training is now done without any errors!

bugbug/models/regression.py
suhaibmujahid, line 1: Unrelated changes!

bugbug/model.py
```diff
@@ -498,6 +498,7 @@ def train(self, importance_cutoff=0.15, limit=None):
             y_pred_filter = []
             classified_indices = []
             for i in range(0, len(y_test)):
+                argmax = np.zeros(len(y_pred_probas), dtype=float)
```
suhaibmujahid, line 501: Why is this needed?

suhaibmujahid (commented):
> @suhaibmujahid so i used sklearn.calibration.CalibratedClassifierCV instead of IsotonicRegression model and for now set  self.calculate_importance = False  (i figured out that IsotonicRegressionCalibrator model type is not currently supported by the TreeExplainer function in the shap library thats why i was getting the "Model type not yet supported by TreeExplainer: <class 'bugbug.model_calibration.IsotonicRegressionCalibrator'>" error ). The training is now done without any errors!

Did you check the following part of the Model class? Isn't this duplicating the cross-validation work?

https://github.com/mozilla/bugbug/blob/2075ddb4eafeddc558974bf910a7264a15d74f80/bugbug/model.py#L367-L382

bugbug/model_calibration.py
WhiteWolf47: > > @suhaibmujahid so i used sklearn.calibration.CalibratedClassifierCV instead of IsotonicRegression model and for now set  self.calculate_importance = False  (i figured out that IsotonicRegressionCalibrator model type is not currently supported by the TreeExplainer function in the shap library thats why i was getting the "Model type not yet supported by TreeExplainer: <class 'bugbug.model_calibration.IsotonicRegressionCalibrator'>" error ). The training is now done without any errors!
> 
> Did you check the following part of the Model class? Isn't this duplicating the cross-validation work?
> 
> https://github.com/mozilla/bugbug/blob/2075ddb4eafeddc558974bf910a7264a15d74f80/bugbug/model.py#L367-L382

thats something i didnt even touch so I didnt notice but let me check it out!

bugbug/model.py
WhiteWolf47: ohh thats something i tried when i was getting that "IndexError: invalid index to scalar variable." i will revert it back!

bugbug/models/regression.py
WhiteWolf47: oh mb!

WhiteWolf47:
@suhaibmujahid should i convert this pr back to an open pr?

bugbug/model_calibration.py
suhaibmujahid: But this applies to all models with the `cross_validation_enabled` flag enabled.

bugbug/model_calibration.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def predict_proba(self, X_val):
+        return self.calibrated_clf.predict_proba(X_val)
+
+    def calibrate(self, X_val, y_val):
+        self.calibrated_clf.cv = "prefit"
+        self.calibrated_clf.method = "isotonic"
+        self.calibrated_clf.fit(X_val, y_val)
+
+    def train(self, X, y):
```
suhaibmujahid, line 46: Why do we need this method?

bugbug/model_calibration.py
```diff
@@ -0,0 +1,48 @@
...
+        X_train, X_test, y_train, y_test = train_test_split(
+            X, y, test_size=test_size, random_state=random_state
+        )
+        return X_train, X_test, y_train, y_test
+
+    def fit(self, X_train, y_train):
+        X_train, X_val, y_train, y_val = self.train_test_split(X_train, y_train)
+        self.base_clf.fit(X_train, y_train)
+        mse_before = mean_squared_error(y_val, self.base_clf.predict(X_val))
+        print(f"MSE of model before calibration : {mse_before:.4f}")
```
suhaibmujahid, line 29: I do not think we need this (at least not here). When we do cross validation this we be noizy. Let us drop it for now.

bugbug/model_calibration.py
```diff
@@ -0,0 +1,48 @@
...
+        self.base_clf = base_clf
+        self.calibrated_clf = CalibratedClassifierCV(
+            base_clf, cv="prefit", method="isotonic"
+        )
+
+    def train_test_split(self, X, y, test_size=0.2, random_state=42):
+        X_train, X_test, y_train, y_test = train_test_split(
+            X, y, test_size=test_size, random_state=random_state
+        )
+        return X_train, X_test, y_train, y_test
```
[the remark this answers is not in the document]
WhiteWolf47: will remove it!

bugbug/models/regressor.py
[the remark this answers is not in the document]
WhiteWolf47: will do it asap but my local machine is not that powerful so it might take some time :smiling_face_with_tear: :

bugbug/model_calibration.py
```diff
@@ -0,0 +1,48 @@
...
+    def predict(self, X):
+        return self.calibrated_clf.predict(X)
+
+    def predict_proba(self, X_val):
+        return self.calibrated_clf.predict_proba(X_val)
+
+    def calibrate(self, X_val, y_val):
+        self.calibrated_clf.cv = "prefit"
+        self.calibrated_clf.method = "isotonic"
+        self.calibrated_clf.fit(X_val, y_val)
```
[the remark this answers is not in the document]
WhiteWolf47: to calibrate the model!

bugbug/model_calibration.py
WhiteWolf47: to train the model

bugbug/model_calibration.py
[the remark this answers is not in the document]
suhaibmujahid: The point is that I do not see why we need the following:
```py
        self.calibrated_clf.cv = "prefit"
        self.calibrated_clf.method = "isotonic"
```
If they are not necessary, then the method is only one line. It could be better for readability to use it directly.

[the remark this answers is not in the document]
WhiteWolf47: makes sense, i'll change it then!

bugbug/model_calibration.py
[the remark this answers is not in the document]
WhiteWolf47: i noticed that in the model.py file , we do the thing (creating a separate train_test_split method), so I thought it is some kind of convention, but you're right, we dont need a separate method for that!

bugbug/model_calibration.py
WhiteWolf47: dont we a need a train() method to make the class sklearn api compatible?

suhaibmujahid: I'm not aware of that! I think we need `fit()`, `predict()`, and `predict_proba()`.

See: https://scikit-learn.org/stable/developers/develop.html

bugbug/model_calibration.py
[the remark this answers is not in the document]
suhaibmujahid: We do that in `model.py` because we need to customize the behaviour for one of the models, which is not the case here (at least for now):

https://github.com/mozilla/bugbug/blob/e7b1ad46f3579f8f1a435be601f2b0f65500084d/bugbug/models/testselect.py#L493-L506

If customizing the behaviour becomes a requirement in future, we then implement a solution (e.g., creating a method for splitting or passing flags/options).

bugbug/models/regressor.py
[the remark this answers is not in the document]
WhiteWolf47: @suhaibmujahid training the model is just giving me this "268 out of 1358 patches with risk between 0.28145539164543165 and 0.9886175060272222 caused regressions (0.19734904270986744
2023-04-14 06:51:34,226:INFO:__main__:Training done" and not any metrics after the training is done?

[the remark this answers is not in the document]
suhaibmujahid: Could you please share the complete report? Precision? Recall? F1?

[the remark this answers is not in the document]
WhiteWolf47: @suhaibmujahid 
before calibration - 

pre      rec      spe     f1        geo    iba     sup
0.97    0.48    0.99    0.63    0.68   0.46   6292

after calibration - 

pre      rec      spe     f1        geo    iba     sup
0.97    0.52    0.99    0.66    0.71   0.50   6292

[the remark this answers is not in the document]
suhaibmujahid: Could you please share the complete reports/outputs?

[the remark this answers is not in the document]
suhaibmujahid: When you train the model, it will output some quality reports (I guess you get the numbers above from there). It will be great if you could copy/paste the full reports here.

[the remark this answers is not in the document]
WhiteWolf47: ahh got it! i now have to train it again cos i closed that terminal lol :)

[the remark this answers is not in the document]
WhiteWolf47: also after the training gets done i can only see  "268 out of 1358 patches with risk between 0.28145539164543165 and 0.9886175060272222 caused regressions (0.19734904270986744
2023-04-14 06:51:34,226:INFO:main:Training done"? The metrics are only visible when the training is going on! the terminal cannot access because a lot of output is there after the training is done!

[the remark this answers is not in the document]
suhaibmujahid: You could extend that from the terminal setting or log to a file:
```
python -m scripts.trainer regressor &> output.txt
```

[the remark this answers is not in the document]
WhiteWolf47: before clabiration - 
[output.txt](https://github.com/mozilla/bugbug/files/11241368/output.txt)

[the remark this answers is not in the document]
WhiteWolf47: after calibration - 
[output_cal.txt](https://github.com/mozilla/bugbug/files/11241376/output_cal.txt)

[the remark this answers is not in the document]
WhiteWolf47: @suhaibmujahid above are the reports for before and after calibration of the regressor model!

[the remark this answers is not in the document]
suhaibmujahid: Thank you @WhiteWolf47! Could you do the same to the `defect` model?

bugbug/models/regressor.py
```diff
@@ -121,9 +122,11 @@ def __init__(
                 ("union", ColumnTransformer(column_transformers)),
             ]
         )
+        base_clf = xgboost.XGBClassifier(n_jobs=utils.get_physical_cpu_count())
+        base_clf.set_params(predictor="cpu_predictor")
+        self.clf = IsotonicRegressionCalibrator(base_clf)
 
-        self.clf = xgboost.XGBClassifier(n_jobs=utils.get_physical_cpu_count())
-        self.clf.set_params(predictor="cpu_predictor")
+        self.calculate_importance = False
```
suhaibmujahid, line 129: Is `calculate_importance = False` still needed?

WhiteWolf47: lemme try it without that!

bugbug/models/regressor.py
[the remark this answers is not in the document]
suhaibmujahid: Doing each model in a separate PR will be better. Maybe you could drop the change for the `regressor` model from this PR. I will merge it then you could open two PRs, one for the `regressor` model and another for the  `defect` model.

WDYT?

bugbug/models/regressor.py
WhiteWolf47: its needed! otherwise it gives the precious error! maybe we can raise an issue regarding this?

bugbug/models/regressor.py
```diff
@@ -121,7 +121,6 @@ def __init__(
                 ("union", ColumnTransformer(column_transformers)),
             ]
         )
-
```
suhaibmujahid, line 124: unrelated change ;)

suhaibmujahid (approved):
Thank you, @WhiteWolf47! LGTM.

WhiteWolf47:
> Thank you, @WhiteWolf47! LGTM.

Welcome @suhaibmujahid , it was really educational working on this issue :smile:

Issue #3337, opened 2023-03-08
Replace general exceptions with a specific ones


suhaibmujahid:
More details here: https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/broad-exception-raised.html

Example of a general exception that should be replaced:
https://github.com/mozilla/bugbug/blob/69972a1684f788319bf5c2944bbe8eeb79428c7d/bugbug/model.py#L192

The goal is to replace all general exceptions, not only in the `model.py` file.

prernadabi23:
@suhaibmujahid Please assign this issue to me, I would like to work on it.

Emeka-Onwuepe:
@suhaibmujahid I am a bit curious, the get_feature_names method in the Model Class will always return an empty list. Is this the ideal implementation?

suhaibmujahid:
> @suhaibmujahid Please assign this issue to me, I would like to work on it.

You do not need to ask. Please start by reading https://github.com/mozilla/bugbug/issues/1092



> @suhaibmujahid I am a bit curious, the get_feature_names method in the Model Class will always return an empty list. Is this the ideal implementation?

The `get_feature_names()` will not be called directly; it is a placeholder that will be overridden. Alternatively, returning an empty list could be replaced by raising`NotImplementedError`.

However, this question is off-topic. For simmiler questions, you could open a new issue, use [Discussions](https://github.com/mozilla/bugbug/discussions) or chat with us in the [bugbug](https://chat.mozilla.org/#/room/#bugbug:mozilla.org) Matrix room.

Emeka-Onwuepe:
Thanks for the reply and the corrections are duly noted.

Munaelekwa:
https://github.com/mozilla/bugbug/pull/3343

Issue #1092, opened 2019-11-09
Guidance for first contributors


marco-c:
The first step is to read CONTRIBUTING.md.

Then, run the setup steps described in the README.

Then, you can pick up a good-first-issue to work on. As stated in CONTRIBUTING.md, **please don't ask to be assigned to an issue or if you can work on an issue**, you can work on it without asking, as long as there is no open PR linked to it.

If you need more help, feel free to ask in the issue, but make sure you have first thought about it and looked at the code a bit, especially when the first comment of the issue has a comprehensive description or pointers to the code.

PR #3343, opened 2023-03-09
Replaced general exception with specific in test-scheduling.py #3337

Munaelekwa:


suhaibmujahid:
Thank you, @Munaelekwa! Issue #3337 already has an open pull request. You can work on any issue, as long as there is no open pull request linked to it (see #1092)

PR #3344, opened 2023-03-09
Replace general exception with specific one #3337

fadebowaley:
Code refactoring  -  changing and replacing general exception with specific one  #3337


suhaibmujahid:
Thank you, @fadebowaley! Issue #3337 already has an open pull request. You can work on any issue, as long as there is no open pull request linked to it (see #1092)

PR #3345, opened 2023-03-09
Clearer error handling

Emeka-Onwuepe:
I refactored the classify_bug function in bugbug_http/model.py by removing unnecessary if-statement. The failed jobs were ignored and not stored in the Redis since the data are not needed. I observed that the words returned by the function doesn't matter provided it is of str, so I changed the returned str conditionally to either "Bug Found" or "Bug Not Found".

### DIFF
--- a/http_service/bugbug_http/models.py
+++ b/http_service/bugbug_http/models.py
@@ -55,61 +55,51 @@ def setkey(key: str, value: bytes, compress: bool = False) -> None:
 
 
 def classify_bug(model_name: str, bug_ids: Sequence[int], bugzilla_token: str) -> str:
-    from bugbug_http.app import JobInfo
+    from bugbug_http.app import JobInfo 
 
     # This should be called in a process worker so it should be safe to set
     # the token here
-    bug_ids_set = set(map(int, bug_ids))
+    fun_info = 'Bug Not Found'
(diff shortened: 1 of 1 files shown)


http_service/bugbug_http/models.py
```diff
@@ -59,55 +59,45 @@ def classify_bug(model_name: str, bug_ids: Sequence[int], bugzilla_token: str) -
...
 
-        # Save the bug last change
-        setkey(job.change_time_key, bugs[bug_id]["last_change_time"].encode())
+                # Save the bug last change
+                setkey(job.change_time_key, bugs[bug_id]["last_change_time"].encode())
+        else:
+            LOGGER.info("Missing model %r, aborting" % model_name)
 
-    return "OK"
+    return fun_info
```
suhaibmujahid, line 100: A general rule: having the code to the left as possible is better for readability.

suhaibmujahid (commented):
What is this PR fixing?

For first-time contributors, it is advised to start by reading https://github.com/mozilla/bugbug/issues/1092

http_service/bugbug_http/models.py
Emeka-Onwuepe: Ok noted with thanks

PR #3348, opened 2023-03-10
 Refactor print statments to use logger #3322 

singla007:
updated model.py

### DIFF
--- a/bugbug/model.py
+++ b/bugbug/model.py
@@ -5,8 +5,8 @@
 
 import pickle
 from collections import defaultdict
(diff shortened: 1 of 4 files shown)


suhaibmujahid:
Issue #3322 already has an open pull request (see https://github.com/mozilla/bugbug/issues/1092)

singla007:
I think these files are still pending

suhaibmujahid:
> I think these files are still pending

The linked PR will address all cases.

Thank you!

PR #3357, opened 2023-03-14
Recalculate 'Time to ___' to use business days

### LINKED ISSUE #2192: Calculate time to fix, time to confirm, etc. in business days instead of days?

ElusiveEllie:
Closes #2192. Using rrange to count only weekdays.


bugbug/bug_features.py
suhaibmujahid, line 1: The scope of the issue here is limited to the `Changes view`. Thus, changes to this file are out of the scope.

tests/test_business_day_counter.py
suhaibmujahid, line 1: The function could be moved to the `utils.py` file, and then the tests could live in the `test_utils.py` file.

bugbug/bug_features.py
ElusiveEllie: So keep the ones in the generate_landings_risk_report.py file but remove the bug_features.py changes?

tests/test_business_day_counter.py
ElusiveEllie: I've moved my test over to `test_utils.py`, but one test (unrelated to changes I've made) is failing. Should I push anyway, since it isn't being effected by my change?

tests/test_utils.py
suhaibmujahid, line 1: Maybe you could have all tests in one function for simplicity.

tests/test_utils.py
```diff
@@ -357,3 +358,57 @@ def test_extract_private_url_empty() -> None:
...
+    )
+    assert result == 19.0, f"Dates did not match, actual result: {result}"
+
+
+def test_work_days_start_on_weekend():
+    result = utils.business_day_range(
+        dateutil.parser.parse("March 10th 2023"),
+        dateutil.parser.parse("March 5th 2023"),
+    )
+    assert result == 5.0, f"Dates did not match, actual result: {result}"
```
suhaibmujahid, line 390: You do not need the report the actual result in the assertion message. If the assertion fails, the error will show the values. The message here could be something like "failed when the date starts on a weekend".

bugbug/utils.py
```diff
@@ -513,3 +514,26 @@ def escape_markdown(text: str) -> str:
...
+
+
+# Remove business days from date range
+def date_range(end_date, start_date):
+    return rrule(
+        DAILY, dtstart=start_date, until=end_date, byweekday=(MO, TU, WE, TH, FR)
+    )
+
+
+def business_day_range(end_date, start_date):
```
suhaibmujahid, line 526: Did you consider using `numpy`? We already have it as a dependency.

suhaibmujahid (commented):
Thank you, @ElusiveEllie!

bugbug/utils.py
```diff
@@ -513,3 +513,16 @@ def escape_markdown(text: str) -> str:
         .replace(")", "\\)")
         .replace("|", "\\|")
     )
+
+
+# Remove business days from date range
+
+
+def business_day_range(end_date, start_date):
```
suhaibmujahid, line 521: Why do we need this logic here? Why `np.busday_count` is not enough?

ElusiveEllie: If the final date falls on a weekend, it ignores that day entirely and returns a smaller number than I believe would be intended. I aimed to have it at least add those days to Monday, since that's when work on a bugfix would continue.

suhaibmujahid: I think using `np.busday_count()` as is would be OK. 

@marco-c wdyt?

suhaibmujahid, line 521: I would name the function something more descriptive like `get_business_days_count()`. Also, I expect to have the `start_date` as the first argument.

scripts/generate_landings_risk_report.py
```diff
@@ -621,10 +622,10 @@ def get_commit_data(commit_list: list[repository.CommitDict]) -> list[dict]:
                             and change["removed"] == "UNCONFIRMED"
                             and change["added"] in ("NEW", "ASSIGNED")
                         ):
-                            time_to_confirm = (
-                                dateutil.parser.parse(history["when"])
-                                - dateutil.parser.parse(bug["creation_time"])
-                            ).total_seconds() / 86400
+                            time_to_confirm = business_day_range(
+                                dateutil.parser.parse(history["when"]),
+                                dateutil.parser.parse(bug["creation_time"]),
```
suhaibmujahid, line 627: Is it necessary to parse the date here?

ElusiveEllie: I'm not quite sure why I did it this way to be honest, I had a reason but it might have been for before I switched to the numpy implementation.

bugbug/utils.py
ElusiveEllie: I'll go ahead and switch both of those. Initially I wanted to keep things the same order as the original subtraction but no, it doesn't really make sense here.

suhaibmujahid (commented):
@ElusiveEllie `time_to_assign` is still in days, not business days.

bugbug/utils.py
suhaibmujahid: @ElusiveEllie Marco agrees. So, please use `np.busday_count()` as is in the function.

bugbug/utils.py
```diff
@@ -513,3 +513,14 @@ def escape_markdown(text: str) -> str:
         .replace(")", "\\)")
         .replace("|", "\\|")
     )
+
+
+# Remove business days from date range
+def get_business_days_count(start_date, end_date):
+    np_start_date = np.datetime64(dateutil.parser.parse(start_date), "D")
```
suhaibmujahid, line 520: Why do we need `dateutil.parser.parse` here?

bugbug/utils.py
```diff
@@ -513,3 +513,14 @@ def escape_markdown(text: str) -> str:
         .replace(")", "\\)")
         .replace("|", "\\|")
     )
+
+
+# Remove business days from date range
```
suhaibmujahid, line 518: You could convert this to docstrings (please consider using Google Style Python Docstrings).

bugbug/utils.py
ElusiveEllie: Understood!

bugbug/utils.py
ElusiveEllie: You asked me to remove it from the function call here https://github.com/mozilla/bugbug/pull/3357/files/1afe81039520aad46d508071687ff1cee325b02f#r1148414434

suhaibmujahid: Yes, but is it needed at all?

ElusiveEllie: I'm not sure what format is being passed to the function call, so I wanted to keep it as close to the original code as I could. Depending on the format of the dates being passed in, it might not be necessary, but I'm not sure where to find an example of what's being passed in.

In my testing, without the parser, the function fails unless the date passed in is in ISO 8601 format

suhaibmujahid: Good point. The dates are in ISO 8601 format. Example: https://bugzilla.mozilla.org/rest/bug?id=36539&include_fields=creation_time,history.when

ElusiveEllie: Perfect! In that case, I don't believe it's necessary. I'll go ahead and remove it for now!

ElusiveEllie: Ah, when I tested the format used here, it gave me a warning: 
`DeprecationWarning: parsing timezone aware datetimes is deprecated; this will raise an error in the future`

Will this be a problem? It was able to work for now, but it looks like in the future, the "Z" after the date passed in may not work.

suhaibmujahid: Then it is needed :)  Could you please mention this in a comment?

ElusiveEllie: Will do!

ElusiveEllie: Would you like me to use the docstrings format, or just a regular comment? I'm unfamiliar with best practices with docstrings

suhaibmujahid: in this case just a regular comment

ElusiveEllie: Should I use the datetime.parser.parse `ignoretz` parameter? It looks like when you pass in a timezone-aware datetime through it, it keeps it as it passes it to numpy. If I set it to ignore, it won't pass through the timezone to numpy and there's no warning (and no error in the future, should the deprecated feature be removed completely)

suhaibmujahid: A work around could be dropping both `np.datetime64()` and `dateutil.parser.parse()`, instead slicing the first 10 characters from the datetime string (i.e., `start_date[:10]`).

ElusiveEllie: Aha, smart! All of the times are passed as strings?

ElusiveEllie: This runs into a different problem. I'm unable to add 1 to the start and end dates when passing them into `busday_count`, meaning the starting day is included and the ending day is excluded. I believe this is unintended, as when you start on a weekend and end on a weekday, it gives the wrong impression.

For example, I have one test set to start on March 5th, a Sunday, and end on March 10th, a Friday. Using the inclusive start date and exclusive end date, We get a count of 4 (Mon, Tue, Wed, Thu), though if you are told that something will take 4 business days on Sunday, you expect it to be completed on Thursday and not Friday.

I can use `datetime.date.fromisoformat` and `datetime.timedelta` to fix this, but in that case I feel I might as well just use `np.datetime64()` since I already have numpy as a dependency.

suhaibmujahid: > or example, I have one test set to start on March 5th, a Sunday, and end on March 10th, a Friday. Using the inclusive start date and exclusive end date, We get a count of 4 (Mon, Tue, Wed, Thu), though if you are told that something will take 4 business days on Sunday, you expect it to be completed on Thursday and not Friday.

I think that is OK. If it is 4 business days, then 4 full business days will end on Thursday and the response will be on Friday ;)

bugbug/utils.py
```diff
@@ -513,3 +513,10 @@ def escape_markdown(text: str) -> str:
         .replace(")", "\\)")
         .replace("|", "\\|")
     )
+
+
+def get_business_days_count(start_date, end_date):
+    """Removes weekends from date range."""
+    np_start_date = np.datetime64(start_date[:10], "D")
+    np_end_date = np.datetime64(end_date[:10], "D")
+    return np.busday_count(np_start_date + 1, np_end_date + 1) * 1.0
```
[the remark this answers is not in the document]
suhaibmujahid: If you add the following test it should pass:
```py

    start_on_saturday = utils.get_business_days_count(
        "2023-04-01T00:00:00Z", "2023-04-05T00:00:00Z"
    )
    start_on_sunday = utils.get_business_days_count(
        "2023-04-02T00:00:00Z", "2023-04-05T00:00:00Z"
    )
    start_on_monday = utils.get_business_days_count(
        "2023-04-03T00:00:00Z", "2023-04-05T00:00:00Z"
    )
    assert (
        start_on_saturday == start_on_sunday == start_on_monday
    ), "If the end date is the same, should not mater if the start date is on Saturday, Sunday, or Monday"
```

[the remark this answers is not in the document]
ElusiveEllie: I initially had it that way, since the original function, dividing with `.total_seconds()`, resulted in a float, and I wasn't sure if anything depended on it being stored that way. I didn't *think* it was necessary, though, so I'll go ahead and switch it.

bugbug/bug_features.py
```diff
@@ -558,11 +558,10 @@ def get_time_to_assign(bug):
...
             ):
-                return (
-                    dateutil.parser.parse(history["when"])
-                    - dateutil.parser.parse(bug["creation_time"])
-                ).total_seconds() / 86400
-
+                return utils.get_business_days_count(
+                    bug["creation_time"],
+                    history["when"],
+                )
```
suhaibmujahid, line 564: The `time_to_assign` feature uses this method; thus, we should not alter its behaviour to keep consistency with other time features.

I would suggest pulling up the logic that extracts the start and the end date in its own function and using it in a new function (e.g., `get_time_to_assign_busdays`) to count the number of business days. Also, we can use it in this existing function (i.e., `get_time_to_assign`) at the same time.

bugbug/utils.py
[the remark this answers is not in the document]
suhaibmujahid: @ElusiveEllie could you please add the test above to `test_business_day_counter()`?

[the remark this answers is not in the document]
ElusiveEllie: I'm so sorry for the delay, I had laptop issues. I'll get to work on this!

bugbug/bug_features.py
ElusiveEllie: > Also, we can use it in this existing function (i.e., get_time_to_assign) at the same time.
What do you mean by this? I've made a new function `get_time_to_assign_busdays` but I'm not understanding where to incorporate it in `get_time_to_assign`.

suhaibmujahid: I mean we should not change the behaviour of the `get_time_to_assign()` function since it is used somewhere else. Instead, creating a new function to do what we want.

ElusiveEllie: Okay, went ahead and put it as its own function, though nothing calls it at the moment. Should I add that somewhere?

PR #3376, opened 2023-03-22
Include patch to skip bugs filed by creators with `editbugs` permissions in `spambug` model.

### LINKED ISSUE #2787: Only evaluate spambug on bugs filed by people without "editbugs" permissions, then check if it's better to train on all or just non-editbugs
The spambug model is only applied to bugs filed by people without "editbugs" permissions, so it makes sense to only evaluate it on these kinds of bugs and not all bugs.

For training, we can keep using all bugs, but we should check if the performance improves or worsens in case we only use bugs filed by non-editbugs people.

jpangas:
#### Reference Issues
Resolves #2787

#### Explained Changes:
This PR includes small changes that allow us to skip bugs filed by creators with `editbugs` permissions. It also includes a change that skips bugs filed by creators with the `"@softvision"` domain.

#### Any other comments?
The PR could not be tested on my side because I lack the necessary permissions. i.e. It returns an `HTTPError` when the request is made. It needs to be tested by someone with the right permissions.


jpangas:
Ping @suhaibmujahid and @marco-c

suhaibmujahid:
> The PR could not be tested on my side because I lack the necessary permissions

Thank you, @jpangas. Unfortunately, I found out that I do not have permissions as well. Could you please work on another issue for now?

jpangas:
Sure! Thank you @suhaibmujahid, I am going to work on another in the mean time.

jpangas:
What had we concluded on this PR?

marco-c:
@jpangas I can run it for you.

The PR is excluding these bugs both from the training set and from the evaluation set.
It would be nice to only exclude them from the evaluation set, and then test the two alternatives (with or without the bugs) in the training set.

jpangas:
@marco-c , after discussing with @suhaibmujahid, there is some refactoring he is working on which will completion of this issue easier. We can wait and I will resume on this after.

suhaibmujahid:
> @marco-c , after discussing with @suhaibmujahid, there is some refactoring he is working on which will make completion of this issue easier. We can wait and I will resume on this after.

I filed an issue for that: https://github.com/mozilla/bugbug/issues/4407

PR #3392, opened 2023-03-28
applied Phabricator after cloning

### LINKED ISSUE #1268: Make the commit classifier script apply the Phabricator patch as its first step

Emeka-Onwuepe:
fixes  #1268

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ http_service/models/*model*
 
 data/
 sheets/
+git_repo-shared/
 
 .mypy_cache/
 .pytest_cache/
--- a/.gitignore
+++ b/.gitignore
@@ -17,7 +17,6 @@ http_service/models/*model*
 
 data/
 sheets/
-git_repo-shared/
 
 .mypy_cache/
 .pytest_cache/
--- a/scripts/commit_classifier.py
+++ b/scripts/commit_classifier.py
@@ -151,7 +151,8 @@ class CommitClassifier(object):
             )
 
         self.revision = None
-        if phabricator_deployment is not None and diff_id is not None:
+        if diff_id is not None:
+            assert phabricator_deployment is not None
             with hglib.open(self.repo_dir) as hg:
                 self.apply_phab(hg, phabricator_deployment, diff_id)
 
--- a/scripts/commit_classifier.py
+++ b/scripts/commit_classifier.py
@@ -135,8 +135,8 @@ class CommitClassifier(object):
         method_defect_predictor_dir: str,
         use_single_process: bool,
         skip_feature_importance: bool,
-        phabricator_deployment: str,
-        diff_id: int,
+        phabricator_deployment: Optional[str] = None,
+        diff_id: Optional[int] = None,
     ):
         self.model_name = model_name
         self.repo_dir = repo_dir
--- a/scripts/commit_classifier.py
+++ b/scripts/commit_classifier.py
@@ -135,6 +135,8 @@ class CommitClassifier(object):
         method_defect_predictor_dir: str,
         use_single_process: bool,
         skip_feature_importance: bool,
+        phabricator_deployment=None,
+        diff_id=None,
     ):
         self.model_name = model_name
         self.repo_dir = repo_dir
@@ -147,6 +149,13 @@ class CommitClassifier(object):
             self.clone_git_repo(
                 "hg::https://hg.mozilla.org/mozilla-central", git_repo_dir
             )
+
+        self.revision = None
+        if phabricator_deployment is not None and diff_id is not None:
+            with hglib.open(self.repo_dir) as hg:
+                self.apply_phab(hg, phabricator_deployment, diff_id)
+
+                self.revision = hg.log(revrange="not public()")[0].node.decode("utf-8")
 
         self.method_defect_predictor_dir = method_defect_predictor_dir
         if method_defect_predictor_dir:
@@ -597,10 +606,9 @@ class CommitClassifier(object):
         self.update_commit_db()
 
         if phabricator_deployment is not None and diff_id is not None:
-            with hglib.open(self.repo_dir) as hg:
-                self.apply_phab(hg, phabricator_deployment, diff_id)
-
-                revision = hg.log(revrange="not public()")[0].node.decode("utf-8")
+            
+            assert self.revision is not None
+            revision = self.revision
 
             commits = repository.download_commits(
                 self.repo_dir,
@@ -855,6 +863,8 @@ def main() -> None:
         args.method_defect_predictor_dir,
         args.use_single_process,
         args.skip_feature_importance,
+        args.phabricator_deployment, 
+        args.diff_id,
     )
     classifier.classify(
         args.revision, args.phabricator_deployment, args.diff_id, args.runnable_jobs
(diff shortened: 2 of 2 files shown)


Emeka-Onwuepe:
Hello sir, please kindly review this. Let me know if I am on the right track.

.gitignore
```diff
@@ -17,6 +17,7 @@ http_service/models/*model*
 
 data/
 sheets/
+git_repo-shared/
```
suhaibmujahid, line 20: Unrelated change!

scripts/commit_classifier.py
```diff
@@ -135,6 +135,8 @@ def __init__(
         method_defect_predictor_dir: str,
         use_single_process: bool,
         skip_feature_importance: bool,
+        phabricator_deployment=None,
+        diff_id=None,
```
suhaibmujahid, line 139: Could you please add the type hints?

scripts/commit_classifier.py
```diff
@@ -597,10 +606,8 @@ def classify(
         self.update_commit_db()
 
         if phabricator_deployment is not None and diff_id is not None:
```
suhaibmujahid, line 608: Now we could check if `self.revision` has a value instead of this. Also, we could drop `phabricator_deployment` and `diff_id` from the arguments of the `classify()` method since they are not needed anymore (the assertions could be moved to the constructor).

scripts/commit_classifier.py
suhaibmujahid, line 1: Did you run the script to make sure the it is still run as expected?

suhaibmujahid (commented):
Thank you, @Emeka-Onwuepe! Please see my comments.

scripts/commit_classifier.py
Emeka-Onwuepe: I only tested with revision;

scripts.commit_classifier regressor repo_dir --revision REVISION 

which I believe will not trigger self.apply_phab().

Emeka-Onwuepe: Can I get diff-id?

scripts/commit_classifier.py
Emeka-Onwuepe: Well noted

scripts/commit_classifier.py
Emeka-Onwuepe: Ok sir

.gitignore
Emeka-Onwuepe: The folder was created when I ran 

scripts.commit_classifier regressor repo_dir --revision REVISION 

I will remove it

scripts/commit_classifier.py
Emeka-Onwuepe: Tested with --diff-id=234724

scripts/commit_classifier.py
```diff
@@ -582,25 +592,12 @@ def feature_sort_key(f):
...
-        diff_id: Optional[int] = None,
         runnable_jobs_path: Optional[str] = None,
     ) -> None:
-        if revision is not None:
-            assert phabricator_deployment is None
-            assert diff_id is None
-
-        if diff_id is not None:
-            assert phabricator_deployment is not None
-            assert revision is None
```
suhaibmujahid, line 595: The assertions here got lost!

Emeka-Onwuepe: Phabricator_deployment and diff_id were no longer present in the method, so I didn't see the need to check for them.

Emeka-Onwuepe: The assertions are trying to avoid the presence of revision, diff_id, and phabricator_delopment while ensuring that diff_id is provided whenever phabricator_development is provided with revision. 

Since both phabricator_development and diff_id are absent, I think the best place to put the assertions will be in the main function.

Emeka-Onwuepe: I will fix it

Emeka-Onwuepe: I have resolved it somewhere else, please have a second look

Emeka-Onwuepe: I moved the assertions to main()

scripts/commit_classifier.py
```diff
@@ -582,25 +592,12 @@ def feature_sort_key(f):
...
-
         self.update_commit_db()
 
-        if phabricator_deployment is not None and diff_id is not None:
-            with hglib.open(self.repo_dir) as hg:
-                self.apply_phab(hg, phabricator_deployment, diff_id)
-
-                revision = hg.log(revrange="not public()")[0].node.decode("utf-8")
+        if self.revision is not None:
+            revision = self.revision
```
suhaibmujahid, line 600: I would add `assert revision is None`  here in case the method gets called outside the main with the wrong args.

scripts/commit_classifier.py
suhaibmujahid: Thank you!

scripts/commit_classifier.py
Emeka-Onwuepe: Ok sir, noted

scripts/commit_classifier.py
```diff
@@ -148,6 +150,14 @@ def __init__(
                 "hg::https://hg.mozilla.org/mozilla-central", git_repo_dir
             )
 
+        self.revision = None
+        if phabricator_deployment is not None and diff_id is not None:
+            with hglib.open(self.repo_dir) as hg:
+                self.apply_phab(hg, phabricator_deployment, diff_id)
+
+                self.revision = hg.log(revrange="not public()")[0].node.decode("utf-8")
+                assert self.revision is not None
```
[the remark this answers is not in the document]
Emeka-Onwuepe: Well noted sir

scripts/commit_classifier.py
```diff
@@ -135,6 +135,8 @@ def __init__(
         method_defect_predictor_dir: str,
         use_single_process: bool,
         skip_feature_importance: bool,
+        phabricator_deployment: str,
+        diff_id: int,
```
suhaibmujahid, line 139: The type hints here don't match the original ones:
https://github.com/mozilla/bugbug/blob/7a5e635999995b0ae603cdc960dfcdd06135358c/scripts/commit_classifier.py#L585-L586

Emeka-Onwuepe: well noted, Sir

scripts/commit_classifier.py
```diff
@@ -148,6 +150,13 @@ def __init__(
                 "hg::https://hg.mozilla.org/mozilla-central", git_repo_dir
             )
 
+        self.revision = None
+        if phabricator_deployment is not None and diff_id is not None:
```
suhaibmujahid, line 154: nit: Checking `phabricator_deployment` as an assertion will help catch missing argument if the this get called directly in the future:
```suggestion
        if diff_id is not None:
            assert phabricator_deployment is not None
```

suhaibmujahid (approved):
Thank you, @Emeka-Onwuepe! LGTM.

suhaibmujahid (approved):

marco-c (approved):

marco-c:
Thanks!

Issue #539, opened 2019-06-03
Make Regression model training set consist only of defects


marco-c:
Now that we finally have the bug type, we can make the Regressor model differentiate between defects which are regressions and defects which are not regressions, rather than between defects which are regressions and defects which are not regressions + non-defects.

This should make the problem easier and the results better.

Note, we probably have to wait for a bit in order to do this, as currently the bug type is still not extremely precise, as it's not required to fill it yet.

omegajudith:
maco, am outreachy applicant. am trying to find a python bug on to work on please assign

Amisha2778:
@marco-c I am an Outreachy applicant. Could you please assign any python related issue?

suhaibmujahid:
@omegajudith and @Amisha2778 please read #1092.

PR #3396, opened 2023-03-29
get_bugbug_labels no longer adds nobug type to regression training data

avinashselvam:
#539 

Modified `get_bugbug_labels` in `defect.py` to include only those data points that are labelled either `regression` or `bug_no_regression` in the training set.

Training the model without changes

72486 non-regression bugs

Cross Validation scores:
Accuracy: f0.9731263445549161 (+/- 0.0012810455820845609)
Precision: f0.9560802008310938 (+/- 0.006503421458310747)
Recall: f0.9316432362619518 (+/- 0.0042866900183067425)

Training the model after changes

71597 non-regression bugs (889 dropped)

Cross Validation scores:
Accuracy: f0.9739072259525028 (+/- 0.0019480324611321944)
Precision: f0.9561803892880535 (+/- 0.006928496874119621)
Recall: f0.9358629670750973 (+/- 0.0045683573571298)

Minor improvement in precision and recall.

Should categories `task`, `enhancement`, `feature` also be removed from the training data for regression?

Please let me know if I have misunderstood the task.

### DIFF
--- a/bugbug/models/defect.py
+++ b/bugbug/models/defect.py
@@ -98,9 +98,9 @@ class DefectModel(BugModel):
             assert category in ["True", "False"], f"unexpected category {category}"
             if kind == "bug":
                 classes[int(bug_id)] = 1 if category == "True" else 0
+            # do not include nobug in regression classifier training data 
             elif kind == "regression":
-                if category == "False":
-                    classes[int(bug_id)] = 0
+                continue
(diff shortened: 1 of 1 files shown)


marco-c:
@avinashselvam this is part of the request from #539. The other part is not to consider bugs with type "enhancement" or "task" as label 0 for the regression model.

PR #3397, opened 2023-03-30
creating a new YAML anchor that includes the common parts and referen…

rajdeep988:
…cing it using the * alias syntax

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -27,34 +27,33 @@ tasks:
       else: ${event.repository.html_url}
 
     taskboot_image: "mozilla/taskboot:0.3.6"
+    
   in:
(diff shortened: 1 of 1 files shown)


rajdeep988:
Hey , need to update yaml.load function to yaml.load_all to fix the error, can someone suggest relevant python program file.

suhaibmujahid:
> Hey , need to update yaml.load function to yaml.load_all to fix the error, can someone suggest relevant python program file.

I do not understand the context

rajdeep988:
> Hey , need to update yaml.load function to yaml.load_all to fix the error, can someone suggest relevant python program file.
I had found a way to resolve the error , error , [https://stackoverflow.com/questions/14359557/reading-yaml-file-with-python-results-in-yaml-composer-composererror-expected-a](url)

PR #3402, opened 2023-04-01
Avoid the unnecessary duplication

### LINKED ISSUE #970: Don't store /tmp/bugbug in Docker image

rajdeep988:
- fixes #970
- **changed the working directory to /bugbug and copied the entire contents of the current directory to that location**

### DIFF
--- a/infra/dockerfile.base
+++ b/infra/dockerfile.base
@@ -13,5 +13,7 @@ ENV LD_PRELOAD="libjemalloc.so.2"
 
 COPY infra/mozci_config.toml /root/.config/mozci/config.toml
 
-COPY . /tmp/bugbug
-RUN pip install --disable-pip-version-check --quiet --no-cache-dir /tmp/bugbug
+WORKDIR /bugbug
+COPY . .
+
+RUN pip install --disable-pip-version-check --no-cache-dir .


infra/dockerfile.base
```diff
@@ -13,5 +13,7 @@ ENV LD_PRELOAD="libjemalloc.so.2"
 
 COPY infra/mozci_config.toml /root/.config/mozci/config.toml
 
-COPY . /tmp/bugbug
-RUN pip install --disable-pip-version-check --quiet --no-cache-dir /tmp/bugbug
+WORKDIR /bugbug
+COPY . .
+
+RUN pip install --disable-pip-version-check --no-cache-dir .
```
suhaibmujahid, line 19: This does not solve the problem; we will still have two copies.

suhaibmujahid:
Thank you for the pull request. I'm closing for lack of response; please feel free to reopen if you still want to work on this.

Issue #1289, opened 2020-02-18
Investigate bad SpamBug classifications


marco-c:
E.g. https://bugzilla.mozilla.org/show_bug.cgi?id=1615701 was classified as spam with high confidence, but it isn't spam.
The first step is to check the feature importances for this classification, so we can see why the model chose the way it did.

gabbyprecious:
Can I work on this?

marco-c:
Sure, feel free.
@ayush-1506 might be interested to investigate too in order to complete the spambug-related project, but this issue can easily be worked on by multiple people since it's an investigation issue.

ayush-1506:
I tried generating the confidence scores this weekend but the terminal crashes with `Killed: 9`. (Which seems to be oom error). I then tried reducing the size of the dataset, but turns out the reduced dataset had only one label every time I tried, so that couldn't work too. Will look into the alternatives.

marco-c:
Probably best to reduce the features considered by the model (the best bet is using a `min_df` for text_vectorizer like we do in other models, you can try with a few different values to see which one doesn't decrease accuracy).

Nikhil1O1:
This one appears to be open, Is it? @marco-c 
If yes I would like to work.
Could you help me pinpoint it in the code base

tarunjarvis5:
Hi, I am new to the community would like to understand and learn to solve the issue. 
Can someone walk me through it?

marco-c:
This issue is not super-easy, as it requires some investigation and understanding of how the project works.
You can start with this one if you want, but you'll need to play around with the code a bit. You can start by training the spambug model (run the scripts/trainer.py script) and classifying a bug with the spambug model (run the scripts/bug_classifier.py script).
Then look at the source code of those scripts and try to understand how they work.

alrho007:
I'll like to work on this issue. 
Permit me to ask you any question if I get stuck or find anything strange.

suhaibmujahid:
> I'll like to work on this issue. Permit me to ask you any question if I get stuck or find anything strange.

You can work on any issue without asking, as long as there is no open PR linked to it (see #1092).

benjaminmah:
I've created a [spreadsheet](https://docs.google.com/spreadsheets/d/1-29_OXtjfVwWsF03DyH2Q3dMNtYwNtB7VO7MLxNtXOI/edit?usp=sharing) compiling the False and True positives made by the SpamBug model, as well as the features that are sent to the model.

[...]

PR #3451, opened 2023-04-28
Retrieve bugs for certain products and exclude them from other models except `spam_bug`

### LINKED ISSUE #2786: Train spambug model on components from all products
We apply the model to bugs from Thunderbird too for example, but we only train on Firefox, Toolkit, etc. bugs. This means Thunderbird bugs are more likely to be wrongly marked as spam.

To fix this, we have to change our bug mining script (https://github.com/mozilla/bugbug/blob/18d270df3cdaa639093a37bde96bed639fb7470b/bugbug/bugzilla.py#L53) to also get Thunderbird (and a few other products: "Calendar", "Chat Core", "MailNews Core") bugs, but then we have to exclude these products from all other models and only use them in the spambug model.

jpangas:
Fixes #2786 

This PR includes the bugs from `"Calendar"`,  `"Chat Core"`, `"MailNews Core"` and  `"Thunderbird"` products. As requested in the issue, this PR excludes the bugs from these products from all other models and only makes them available in the `spam_bug` model using the `include_mail_products` flag.

Comments:
I was guided by this [comment](https://github.com/mozilla/bugbug/issues/2786#issue-1158951552) from the issue.

Train on Taskcluster: spambug


jpangas:
This change is the first step into training the `spam_bug` model on bugs from the newly included products.

bugbug/bugzilla.py
```diff
@@ -82,12 +86,20 @@
...
+]
+
 
 def get_bugs(include_invalid: Optional[bool] = False) -> Iterator[BugDict]:
     yield from (
         bug
         for bug in db.read(BUGS_DB)
-        if include_invalid or bug["product"] != "Invalid Bugs"
+        if include_invalid or bug["product"] not in PRODUCTS_TO_EXCLUDE
     )
```
marco-c, line 103: Right now only the spam bug model is passing `include_invalid = True`, but we can't be sure in the future. There might be other models which want to use invalid bugs but no bugs from Thunderbird. I think it'd be better to keep the behavior of `include_invalid` as it is now, and add a new `products` argument that is `list(set(PRODUCTS) - set(MAIL_PRODUCTS))` by default.

jpangas: Thank you @marco-c . I will make the changes and submit them soon.

bugbug/models/spambug.py
```diff
@@ -130,10 +130,12 @@ def get_labels(self):
 
     def items_gen(self, classes):
         # Overwriting this method to add include_invalid=True to get_bugs to
```
marco-c, line 132:
```suggestion
        # Overwriting this method to add include_invalid=True and include_mail_products=True to get_bugs to
```

bugbug/models/spambug.py
```diff
@@ -130,10 +130,12 @@ def get_labels(self):
...
-        # include spam bugs.
+        # include spam bugs and mail products
```
marco-c, line 133:
```suggestion
        # include spam bugs and mail products.
```

bugbug/bugzilla.py
```diff
@@ -93,20 +100,16 @@
...
-    "Chat Core",
-    "Invalid Bugs",
-    "MailNews Core",
-    "Thunderbird",
-]
-
 
-def get_bugs(include_invalid: bool | None = False) -> Iterator[BugDict]:
+def get_bugs(
+    include_invalid: bool | None = False, include_mail_products: bool = False
```
marco-c, line 105: Instead of a specific "include_mail_products" which is either True or False, it would be more generic to have something like "include_additional_products" as a list, and the caller can specify the list.

marco-c:
From the spreadsheet at https://github.com/mozilla/bugbug/issues/1289 (https://docs.google.com/spreadsheets/d/1-29_OXtjfVwWsF03DyH2Q3dMNtYwNtB7VO7MLxNtXOI/edit), we should probably include additional products too ("CA Program" and "SeaMonkey").

bugbug/bugzilla.py
```diff
@@ -90,11 +100,20 @@
...
+        PRODUCTS + include_additional_products
+        if include_additional_products
+        else PRODUCTS
+    )
     yield from (
         bug
         for bug in db.read(BUGS_DB)
-        if include_invalid or bug["product"] != "Invalid Bugs"
+        if bug["product"] in products
+        and (include_invalid or bug["product"] != "Invalid Bugs")
```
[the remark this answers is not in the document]
jpangas: "Invalid Bugs" is in `PRODUCTS` by default, so we can do the opposite, if `include_invalid` is False, we remove "Invalid Bugs" from the `products`.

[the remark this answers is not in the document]
marco-c: Let's keep it this way then

jpangas:
Thanks!

Issue #3117, opened 2022-10-14
Upgrad to use the latest verion of `rust-code-analysis-web`


suhaibmujahid:
Upgrading the version from `v0.0.23` to a higher version includes breaking changes. As a result, this requires reflecting the changes where needed (e.g., [repository.py](https://github.com/mozilla/bugbug/blob/master/bugbug/repository.py)).

Amotul-raheem:
picking this up

suhaibmujahid:
@Amotul-raheem the latest version is `v0.024` and it does not have a downloadable pre-built binary: https://github.com/mozilla/rust-code-analysis/releases/tag/v0.0.24, thus we cannot use it in the CI (at least not in the same way that we are doing now).

Maybe you want to wait for @marco-c to attach the pre-built binaries for `v0.024` or release `v0.025` with the pre-built binaries attached.

Amotul-raheem:
Alright got it thanks!

marco-c:
We have released 0.0.25, we should upgrade to it.

Munaelekwa:
Hello, I would like to work on this, I would appreciate if you would give me a lead, on how to go about this. Thank you

suhaibmujahid:
> Hello, I would like to work on this, I would appreciate if you would give me a lead, on how to go about this. Thank you

Here is where we download `rust-code-analysis-web`:

https://github.com/mozilla/bugbug/blob/f9906057a5281b8913fb5a92edbe73440953581b/.taskcluster.yml#L94

Currently, we are using `v0.0.23` where the latest version is `v0.0.25`. We need to upgrade to the latest version; however, the latest version introduces some breakage changes, see https://github.com/mozilla/bugbug/issues/3117#issue-1409634850.

To work on this, you may want to start by installing/building [rust-code-analysis](https://github.com/mozilla/rust-code-analysis).

Once you have `v0.0.25` installed on your system, you could run the tests to check for breakage changes and fix them, for example:

```
python -m pytest tests/test_repository.py
```

Munaelekwa:
> > Hello, I would like to work on this, I would appreciate if you would give me a lead, on how to go about this. Thank you
> 
> Here is where we download `rust-code-analysis-web`:
> 
> https://github.com/mozilla/bugbug/blob/f9906057a5281b8913fb5a92edbe73440953581b/.taskcluster.yml#L94
> 
[...]

PR #3506, opened 2023-06-02
fix: metrics["nargs"]["sum"] into metrics["nargs"]["total"]

baolef:
metrics["nargs"]["sum"] is no longer there. From my understanding, it is renamed into metrics["nargs"]["total"].


marco-c:
Until we fix #3117, we are still on rust-code-analysis v0.0.23, so the key name is still "sum".

PR #3550, opened 2023-07-02
Only train component model on active components

### LINKED ISSUE #3549: Component check failing because of a component that is no longer active
The "component" check is failing with the following message:
> 2023-07-02 10:05:49,740:WARNING:bugbug.models.component:Component 'Graphics: Layers' of product 'Core' doesn't exists, failure

This is because, when training, we use all available components and when checking the model, we verify if there was at least one bug filed against that component in the past 12 months (https://github.com/mozilla/bugbug/blob/ed7c4460b41c6c0b6383e5532e6786dbaae27c27/bugbug/models/component.py#L237).

marco-c:
Fixes #3549

### DIFF
--- a/bugbug/bugzilla.py
+++ b/bugbug/bugzilla.py
@@ -364,8 +364,9 @@ def get_active_product_components(products=[]) -> set[tuple[str, str]]:
     return set(
         (product["name"], component["name"])
         for product in r.json()["products"]
+        if product["is_active"]
         for component in product["components"]
-        if product["is_active"] and component["is_active"]
+        if component["is_active"]
     )
 
 
--- a/bugbug/bugzilla.py
+++ b/bugbug/bugzilla.py
@@ -344,6 +344,31 @@ def get_product_component_count(months: int = 12) -> dict[str, int]:
     return bugs_number
 
 
+def get_active_product_components(products=[]) -> set[tuple[str, str]]:
+    r = utils.get_session("bugzilla").get(
+        "https://bugzilla.mozilla.org/rest/product",
+        params={
+            "type": "accessible",
+            "include_fields": [
+                "name",
+                "is_active",
+                "components.name",
+                "components.is_active",
+            ],
+            "names": products,
+        },
+        headers={"X-Bugzilla-API-Key": Bugzilla.TOKEN, "User-Agent": "bugbug"},
+    )
+    r.raise_for_status()
+
+    return set(
+        (product["name"], component["name"])
+        for product in r.json()["products"]
+        for component in product["components"]
+        if product["is_active"] and component["is_active"]
+    )
+
+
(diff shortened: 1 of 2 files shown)


bugbug/bugzilla.py
```diff
@@ -344,6 +344,31 @@ def get_product_component_count(months: int = 12) -> dict[str, int]:
...
+                "name",
+                "is_active",
+                "components.name",
+                "components.is_active",
+            ],
+            "names": products,
+        },
+        headers={"X-Bugzilla-API-Key": Bugzilla.TOKEN, "User-Agent": "bugbug"},
+    )
+    r.raise_for_status()
```
suhaibmujahid, line 362: Maybe we could at some point use the `BugzillaProduct` class to fetch components. We do not need to do it now since we already using the API directly in `get_component_team_mapping()`. We could file a follow-up issue to maintain consistency.

bugbug/bugzilla.py
```diff
@@ -344,6 +344,31 @@ def get_product_component_count(months: int = 12) -> dict[str, int]:
...
+
+    return set(
+        (product["name"], component["name"])
+        for product in r.json()["products"]
+        for component in product["components"]
+        if product["is_active"] and component["is_active"]
+    )
```
suhaibmujahid, line 369: nit: checking if the product is active before iterating over the components improves the readability and the performance: 
```suggestion
    return {
        (product["name"], component["name"])
        for product in r.json()["products"]
        if product["is_active"]
        for component in product["components"]
        if component["is_active"]
    }
```

bugbug/models/component.py
```diff
@@ -221,10 +221,14 @@ def get_meaningful_product_components(self, full_comp_tuples, threshold_ratio=10
         max_count = product_component_counts[0][1]
         threshold = max_count / threshold_ratio
 
+        active_product_components = bugzilla.get_active_product_components(
+            self.PRODUCTS + list(self.PRODUCT_COMPONENTS)
```
suhaibmujahid, line 225: `self.PRODUCTS` is a set, we cannot use `+` on it
```suggestion
            self.PRODUCTS | self.PRODUCT_COMPONENTS.keys()
```

bugbug/bugzilla.py
marco-c: Done in b9ea52cd87e0ebe2312d604cd5e84fca42f3d2b2.

bugbug/models/component.py
marco-c: Done in 53934f579c8e463189a3b150b02b8ebe5173cfe5.

bugbug/bugzilla.py
marco-c: Makes sense, filed #3551.

PR #3571, opened 2023-07-06
Run tests on Python 3.10

marco-c:


suhaibmujahid:
We may want to modify the readme file to require 3.10+ instead of 3.9+[^1]

[^1]: https://github.com/mozilla/bugbug#setup-and-prerequisites

PR #3585, opened 2023-07-19
change python-igraph to igraph

szhorvat:
The package was renamed on PyPI, and support was scheduled to be stopped on Sep 1st, 2022. See https://pypi.org/project/python-igraph/

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -1,9 +1,9 @@
 coverage==7.2.7
 hypothesis==6.81.2
+igraph==0.10.6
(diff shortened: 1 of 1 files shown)


marco-c:
Thanks for the PR @szhorvat! Could you fix the linting issues? You can use pre-commit to do that.

szhorvat:
Fixed. I hope you could use squash-and-merge with this PR so I can avoid a force push :-)

marco-c (approved):
Thanks!

PR #3598, opened 2023-07-24
Optimize synonyms feature cleanup by making a single re.sub call

marco-c:
I actually still need to test if this is an improvement or not.

### DIFF
--- a/bugbug/feature_cleanup.py
+++ b/bugbug/feature_cleanup.py
@@ -185,11 +185,11 @@ class synonyms(object):
             flags=re.IGNORECASE,
         )
 
-    def __call__(self, text):
-        def replacement(match):
-            return self.synonyms_dict[match.group(0).lower()]
+    def _replace(self, match):
+        return self.synonyms_dict[match.group(0).lower()]
 
-        return self.pattern.sub(replacement, text)
+    def __call__(self, text):
+        return self.pattern.sub(self._replace, text)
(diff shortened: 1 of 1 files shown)


marco-c:
On a subset of 7000 bugs, it takes around 5.5 seconds on my machine vs almost 8 seconds previously, so it's a 30% improvement.

bugbug/feature_cleanup.py
```diff
@@ -174,18 +174,22 @@ def __init__(self):
...
+            r"|".join(rf"\b{synonym}\b" for synonym in self.synonyms_dict.keys()),
+            flags=re.IGNORECASE,
+        )
 
     def __call__(self, text):
-        for synonym_group in self.pattern:
-            text = self.pattern[synonym_group].sub(synonym_group, text)
-        return text
+        def replacement(match):
+            return self.synonyms_dict[match.group(0).lower()]
```
suhaibmujahid, line 190: nit: Moving `replacement()` to be a private method could have a slight performance improvement.

suhaibmujahid (dismissed):
LGTM!

bugbug/feature_cleanup.py
marco-c: It did get to 4.2 seconds. I was expecting it to be slightly better but thought it would have been almost not visible!

PR #3668, opened 2023-09-20
Update to taskboot 0.4.2

marco-c:
CC @matt-boris @petemoore

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -26,7 +26,7 @@ tasks:
       then: ${event.pull_request.head.repo.html_url}
       else: ${event.repository.html_url}
 
-    taskboot_image: "mozilla/taskboot:0.3.10"
+    taskboot_image: "mozilla/taskboot:0.4.2"
   in:
     $if: 'tasks_for == "github-push" || (tasks_for == "github-pull-request" && event["action"] in ["opened", "reopened", "synchronize"])'
     then:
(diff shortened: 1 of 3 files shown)


matt-boris (approved):
Changes LGTM, but it looks like `proj-bugbug/generic-worker-ubuntu-22-04` is not an available worker pool at the moment.

marco-c:
I could have sworn to have found it in community-tc-config!

marco-c:
Oh, it looks like `ci` was already a generic worker worker.

matt-boris:
Ah ok perfect, so we can just switch it over to `ci` then 👍🏻

matt-boris (approved):

matt-boris (approved):

infra/data-pipeline.yml
```diff
@@ -1290,8 +1290,6 @@ tasks:
         - train-test-group-select
         - train-needsdiagnosis
       payload:
-        capabilities:
-          privileged: true
```
matt-boris, line 1294: You're going to have to keep these lines so that d2g adds in the `--privileged` flag to podman run.

.taskcluster.yml
```diff
@@ -175,8 +175,6 @@ tasks:
         provisionerId: proj-bugbug
         workerType: ci
         payload:
-          capabilities:
-            privileged: true
```
matt-boris, line 179: Same with this one.

matt-boris (commented):
Looks like docker builds are failing due to permission issues. I think you'll need to keep the docker worker privileged capability.

PR #3687, opened 2023-10-03
Refactored count_bugs to use libmozdata

### LINKED ISSUE #3676: Refactor `count_bugs` to use `libmozdata`
We should use the `Bugzilla` class from `libmozdata.bugzilla` instead of doing a direct request:

https://github.com/mozilla/bugbug/blob/082613fc3a1121a9ec6f235990e61b829f407267/bugbug/bugzilla.py#L282-L291

Example:

https://github.com/mozilla/bugbug/blob/082613fc3a1121a9ec6f235990e61b829f407267/bugbug/bugzilla.py#L110

rafey1104:
Fix #3676 @suhaibmujahid 

Used Bugzilla class from libmozdata.bugzilla instead of doing a direct request


suhaibmujahid:
The `libmozdata` API is not used correctly. You could have a look at the example in #3676 as guidance. 

Please test locally whether the function works correctly and make sure that your changes pass the CI tests.

I will cover the PR to draft; feel free to convert it back when it is ready for review.

rafey1104:
Before:-
def count_bugs(bug_query_params): 
  bug_query_params["count_only"] = 1 
  r = utils.get_session("bugzilla").get( 
         "https://bugzilla.mozilla.org/rest/bug", params=bug_query_params 
     ) 
     r.raise_for_status() 
     count = r.json()["bug_count"] 
   return count

After:-
def count_bugs(bug_query_params):
        count = Bugzilla(
            queries={"bug_query_parameter": "1"}, 
        bug_query_params={"count_only": 1}
    ).get_data().wait()     
    return count

I think now the API has been used correctly. How should I run the tests locally can you please guide me to some resources @suhaibmujahid ?

suhaibmujahid:
> I think now the API has been used correctly. 

Unfortunately, it is not :( See that example in the issue. Also, you could explore other places where that class is used. 


> How should I run the tests locally can you please guide me to some resources?

You do not need to run all existing tests locally to test the function. You could write a simple script/test to run it. Anyway, you could run the existing tests locally by running the following command:
```
python -m pytest tests/test_*.py
```
However, any time you push a commit to the PR, the tests will automatically run, and you can see the results. For example, the following are the results of your last comment: https://community-tc.services.mozilla.com/tasks/groups/Yf7oAmmcQJ-b4f7MOWFLGg


The commit failed on the lining task, not the tests. You could prevent that by using `pre-commit`. See https://github.com/mozilla/bugbug/tree/master#auto-formatting

PR #3690, opened 2023-10-04
fix: #3676 refactor count_bugs to use libmozdata

PromiseFru:


PromiseFru:
Hello @suhaibmujahid,

I've prepared a pull request addressing the issue at https://github.com/mozilla/bugbug/issues/3676.

I refactored the `count_bugs` method and conducted some testing by providing draft data. However, I observed that there is no corresponding test for this method in the `test_bugzilla.py` test file, and I couldn't locate any integration of this method into other functions.

I would appreciate it if you could confirm whether this method is integrated into another function or if there exists a unit test that utilizes it.

suhaibmujahid:
@PromiseFru thank you for your interest in the project. Unfortunately, the issue already has an open PR linked to it. You could pick up the issue again if the PR gets closed without fixing the issue.

There are many other issues you could work on. Issues do not need to be created by me in order to be able to work on them.

PR #3691, opened 2023-10-05
When retrieving past failure data for a TOML manifest, fallback on data from the homonymous INI manifest

marco-c:
If we don't, we would consider TOML manifests as new, which would negatively affect the test selection model

### DIFF
--- a/bugbug/test_scheduling.py
+++ b/bugbug/test_scheduling.py
@@ -310,21 +310,62 @@ def get_test_scheduling_history(granularity):
         yield obj["revs"], obj["data"]
 
 
-def get_past_failures(granularity, readonly):
-    if granularity == "label":
-        past_failures_db = os.path.join("data", PAST_FAILURES_LABEL_DB)
-    elif granularity == "group":
-        past_failures_db = os.path.join("data", PAST_FAILURES_GROUP_DB)
-    elif granularity == "config_group":
-        past_failures_db = os.path.join("data", PAST_FAILURES_CONFIG_GROUP_DB)
-    else:
-        raise UnexpectedGranularityError(granularity)
+class PastFailures:
+    def __init__(self, granularity, readonly):
+        if granularity == "label":
+            past_failures_db = os.path.join("data", PAST_FAILURES_LABEL_DB)
+        elif granularity == "group":
+            past_failures_db = os.path.join("data", PAST_FAILURES_GROUP_DB)
+        elif granularity == "config_group":
+            assert False, "config_group granularity not supported for past failures"
+        else:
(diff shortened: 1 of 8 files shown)


bugbug/test_scheduling.py
```diff
@@ -310,21 +310,62 @@ def get_test_scheduling_history(granularity):
...
+    @all_runnables.setter
+    def all_runnables(self, value) -> None:
+        self.db["all_runnables"] = value
+
+    def get(self, key: str) -> ExpQueue | None:
+        value = self.db.get(key, None)
+
+        # Fallback on INI if the group is now TOML.
+        if value is None and self.granularity == "group" and key.endswith(".toml"):
+            ini_key = f"{key[:-4]}ini"
```
jmaher, line 352: this logic is good, I think there is a case to consider where we have old data from .ini and new data from .toml, in this case we will only use the new data which might be less accurate.

jmaher (approved):

bugbug/test_scheduling.py
marco-c: This data is built push by push in order.
So once we get to the push which migrates from INI to TOML, we'll copy the data over, and from that push forward we will use the "INI copied" data for TOML and update it.

bugbug/test_scheduling.py
```diff
@@ -310,21 +310,62 @@ def get_test_scheduling_history(granularity):
...
+            writeback=not readonly,
+        )
+
+    @property
+    def push_num(self) -> int:
+        return self.db["push_num"]
+
+    @push_num.setter
+    def push_num(self, value: int) -> None:
+        self.db["push_num"] = value
```
suhaibmujahid, line 337: Why do we need setters? Can the values be passed to the constructor instead?

(same for `all_runnables`)

suhaibmujahid (approved):

bugbug/test_scheduling.py
marco-c: The values are being changed, for example here: https://github.com/mozilla/bugbug/pull/3691/files#diff-67e71ee19f5db00669e94c709cb6de177b6117efc0bfbcf96b46e365e7433ef5R340.

suhaibmujahid: In this specific case, the value didn't change based on any external factor. Which can be determined in the constructor.

suhaibmujahid: Never mind, I see it is modified in https://github.com/mozilla/bugbug/pull/3691/files#diff-f3557fe3a5b0819ec4d3a2f85a607a9450e4b9f18168838e95d3d2672564f8b4R566

PR #3696, opened 2023-10-07
Fix: #3695 Dropped definition of bug model

### LINKED ISSUE #3695: Drop the `bug` model
The model has been renamed to `defect` in https://github.com/mozilla/bugbug/pull/335/files#diff-74f32ab12cc9be314b1fead3a281590c768699c2e9883b8346066f6c4d6daa90 

However, the definition of `bug` model was not dropped.

We need to drop the following line: https://github.com/mozilla/bugbug/blob/c228c2a686b271f7ca24a0716cda4f9059229d9f/bugbug/models/__init__.py#L16

goldpanth3r:
Dropped the unnecessary definition of bug model as the implementations were deleted.
Closes #3695.

### DIFF
--- a/bugbug/models/__init__.py
+++ b/bugbug/models/__init__.py
@@ -13,7 +13,6 @@ MODELS = {
     "assignee": "bugbug.models.assignee.AssigneeModel",
     "backout": "bugbug.models.backout.BackoutModel",
     "browsername": "bugbug.models.browsername.BrowserNameModel",
-    "bug": "bugbug.model.BugModel",
     "bugtype": "bugbug.models.bugtype.BugTypeModel",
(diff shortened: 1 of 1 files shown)


suhaibmujahid:
@goldpanth3r thank you for the PR. Could you please link the PR correctly to the issue? The Issue number should be in the PR description, not the PR title. You need to use one of the linking keywords: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

goldpanth3r:
Sorry about that. I've updated the PR description above as you said. Please review.

suhaibmujahid (approved):

PR #3704, opened 2023-10-09
refactor count_bugs to use libmozdata (#3676)

### LINKED ISSUE #3676: Refactor `count_bugs` to use `libmozdata`

issamarabi:
Closes #3676

Refactored the `count_bugs` function in bugbug/bugzilla.py to utilize the `Bugzilla` class from `libmozdata.bugzilla` instead of making a direct request.

The previous approach involved sending a direct GET request to Bugzilla's REST API. This approach might not be sustainable and consistent in the long run, especially if there are updates or changes to the API.

By using the `libmozdata.bugzilla` library, we aim for a more maintainable and robust method of counting bugs. It also keeps our codebase consistent with other parts where the `Bugzilla` class is used.

### DIFF
--- a/bugbug/bugzilla.py
+++ b/bugbug/bugzilla.py
@@ -282,13 +282,14 @@ def delete_bugs(match):
 def count_bugs(bug_query_params):
     bug_query_params["count_only"] = 1
 
-    r = utils.get_session("bugzilla").get(
-        "https://bugzilla.mozilla.org/rest/bug", params=bug_query_params
-    )
-    r.raise_for_status()
-    count = r.json()["bug_count"]
+    counts = []
 
-    return count
+    def handler(bug):
+        counts.append(bug["bug_count"])
+
+    Bugzilla(bug_query_params, bughandler=handler).get_data().wait()
+
+    return sum(counts)
 
 
 def get_product_component_count(months: int = 12) -> dict[str, int]:


rafey1104:
Hi @issamarabi. Tq for the PR, but I have already solved this particular issue through #3687 PR.

bugbug/bugzilla.py
```diff
@@ -282,13 +282,14 @@ def delete_bugs(match):
...
-        "https://bugzilla.mozilla.org/rest/bug", params=bug_query_params
-    )
-    r.raise_for_status()
-    count = r.json()["bug_count"]
+    counts = []
+
+    def handler(bug):
+        counts.append(bug["bug_count"])
+
+    Bugzilla(bug_query_params, bughandler=handler).get_data().wait()
```
suhaibmujahid, line 290: Using `bughandler` expect to receive a response that has the property `bugs`: https://bugzilla.mozilla.org/rest/bug?id=1234567. However, passing the `count_only` returns JSON object that contains `bug_count`: https://bugzilla.mozilla.org/rest/bug?id=1234567&count_only=1

Thus, you need to pass the `Query` class to handle this, example: 

```
Bugzilla(queries=Query(Bugzilla.API_URL, bug_query_params, handler)).wait()
```

suhaibmujahid (commented):
Hi @issamarabi. Thank you for your interest in the project. See my feedback below and 
please test the function locally to make sure that things work as expected.

PR #3707, opened 2023-10-11
Use Booster.save_model when saving XGBoost models instead of pickle

### LINKED ISSUE #3505: Use `Booster.save_model` when saving XGBoost models instead of `pickle`
```
WARNING: /Users/runner/work/xgboost/xgboost/python-package/build/temp.macosx-11.0-arm64-cpython-38/xgboost/src/learner.cc:553: 
  If you are loading a serialized model (like pickle in Python, RDS in R) generated by
  older XGBoost, please export the model by calling `Booster.save_model` from that version
  first, then load it back in current version. See:

    https://xgboost.readthedocs.io/en/latest/tutorials/saving_model.html

  for more details about differences between saving model and serializing.
```

PromiseFru:
Closes #3505

### DIFF
--- a/bugbug/model.py
+++ b/bugbug/model.py
@@ -11,6 +11,7 @@ from typing import Any
 import matplotlib
 import numpy as np
 import shap
+from xgboost import Booster
 from imblearn.metrics import (
     classification_report_imbalanced,
     geometric_mean_score,
@@ -571,11 +572,8 @@ class Model:
             pickle.dump(self, f, protocol=pickle.HIGHEST_PROTOCOL)
 
         if self.store_dataset:
-            with open(f"{self.__class__.__name__.lower()}_data_X", "wb") as f:
-                pickle.dump(X, f, protocol=pickle.HIGHEST_PROTOCOL)
-
-            with open(f"{self.__class__.__name__.lower()}_data_y", "wb") as f:
-                pickle.dump(y, f, protocol=pickle.HIGHEST_PROTOCOL)
+            Booster.save_model(X, f"{self.__class__.__name__.lower()}_data_X")
+            Booster.save_model(y, f"{self.__class__.__name__.lower()}_data_y")
 
         return tracking_metrics
 
--- a/bugbug/model.py
+++ b/bugbug/model.py
@@ -572,8 +572,11 @@ class Model:
             pickle.dump(self, f, protocol=pickle.HIGHEST_PROTOCOL)
 
         if self.store_dataset:
-            Booster.save_model(X, f"{self.__class__.__name__.lower()}_data_X")
-            Booster.save_model(y, f"{self.__class__.__name__.lower()}_data_y")
+            with open(f"{self.__class__.__name__.lower()}_data_X", "wb") as f:
+                Booster.save_model(X, f)
+
+            with open(f"{self.__class__.__name__.lower()}_data_y", "wb") as f:
+                Booster.save_model(y, f)
 
         return tracking_metrics
 


jpangas:
Hey @PromiseFru , I was looking into this issue some day and I think it would be helpful I share with you some of the insights.
A little insight, the XGBoost model (classifier) object is `self.clf` 
e.g in Spambug:
https://github.com/mozilla/bugbug/blob/4d4e0d038a51d5439170f1c15c355694a2e1c379/bugbug/models/spambug.py#L83-L84

The XGBoost model is saved here:
https://github.com/mozilla/bugbug/blob/4d4e0d038a51d5439170f1c15c355694a2e1c379/bugbug/model.py#L570-L571

Note: The model object is the one "pickled."

Later, the model is loaded using the method here:
https://github.com/mozilla/bugbug/blob/4d4e0d038a51d5439170f1c15c355694a2e1c379/bugbug/model.py#L582-L585

The part you had modified was the storing the dataset instead.

I hope this helps.!

PromiseFru:
> Hey @PromiseFru , I was looking into this issue some day and I think it would be helpful I share with you some of the insights. A little insight, the XGBoost model (classifier) object is `self.clf` e.g in Spambug:
> 
> https://github.com/mozilla/bugbug/blob/4d4e0d038a51d5439170f1c15c355694a2e1c379/bugbug/models/spambug.py#L83-L84
> 
> The XGBoost model is saved here:
> 
> https://github.com/mozilla/bugbug/blob/4d4e0d038a51d5439170f1c15c355694a2e1c379/bugbug/model.py#L570-L571
> 
> Note: The model object is the one "pickled."
> 
> Later, the model is loaded using the method here:
> 
> https://github.com/mozilla/bugbug/blob/4d4e0d038a51d5439170f1c15c355694a2e1c379/bugbug/model.py#L582-L585
> 
> The part you had modified was the storing the dataset instead.
> 
> I hope this helps.!

Thank you very much, @jpangas 🚀. I will make the required adjustments.

PromiseFru:
Hello @suhaibmujahid, here are the changes I made: 

- I updated the `train` method to save the XGBoost model with "_xgb" added to the file name.
- I set `self.clf` to `None` to prevent the XGBoost model from being pickled along with the entire object.
- I updated the `load` method to load the XGBoost model from a file with "_xgb" and assign it to `clf`, and it also loads the previously pickled object.

I'd appreciate your feedback on these updates 🙏🏾 .

suhaibmujahid:
Thank you @PromiseFru!

We need to handle the following:
- A model could use something other than XGBoost. We could check if it supports `save_model()` before moving forward with that. We always could failback to use `pickle`.
- Once the model is trained, the workflow assumes that the model has only one file. The file gets compressed and later uploaded to Taskcluseter. Now we have two files for the same model, we need a way to handle that. One solution could be to have both files in a directory (that matches the name of the file that the workflow expects) and then when the directory gets compressed, we end up with one file.

PromiseFru:
> Thank you @PromiseFru!
> 
> We need to handle the following:
> 
>     * A model could use something other than XGBoost. We could check if it supports `save_model()` before moving forward with that. We always could failback to use `pickle`.
> 
>     * Once the model is trained, the workflow assumes that the model has only one file. The file gets compressed and later uploaded to Taskcluseter. Now we have two files for the same model, we need a way to handle that. One solution could be to have both files in a directory (that matches the name of the file that the workflow expects) and then when the directory gets compressed, we end up with one file.

Hello @suhaibmujahid, I appreciate your feedback.

I have updated the model saving logic to create a directory with the expected file name when it supports the `save_model()` function. The XGBoost model is saved as `model.xgb`, and the rest of the model is pickled as `model.pkl` within the same directory. Additionally, I have updated the model loading logic to handle two scenarios:

- When `model_file_name` is a directory, it loads the XGBoost model from "model.xgb" and the pickled model from "model.pkl" within that directory.
- When `model_file_name` is a file, it loads the model as a pickled object as usual.

Here is the updated code snippet:

*saving*
```python
if hasattr(self.clf, "save_model"):
    model_directory = self.__class__.__name__.lower()
    os.makedirs(model_directory, exist_ok=True)

    self.clf.save_model(os.path.join(model_directory, "model.xgb"))
    self.clf = None

    with open(os.path.join(model_directory, "model.pkl"), "wb") as f:
        pickle.dump(self, f, protocol=pickle.HIGHEST_PROTOCOL)

else:
    with open(self.__class__.__name__.lower(), "wb") as f:
        pickle.dump(self, f, protocol=pickle.HIGHEST_PROTOCOL)
```

*loading*
```python
@staticmethod
def load(model_file_name: str) -> "Model":
    if os.path.isdir(model_file_name):
        with open(os.path.join(model_file_name, "model.xgb"), "rb") as f:
            xgb_model = Booster()
            xgb_model.load_model(f)

        with open(os.path.join(model_file_name, "model.pkl"), "rb") as f:
            model = pickle.load(f)

        model.clf = xgb_model
    else:
        with open(model_file_name, "rb") as f:
            model = pickle.load(f)

    return model
```

However, I couldn't get `test_trainer` to pass. It has the following error:

```bash
/home/bugbug/scripts/trainer.py:62: in go
    zstd_compress(model_file_name)
/home/bugbug/bugbug/utils.py:240: in zstd_compress
    subprocess.run(["zstd", "-f", path], check=True)
```

Upon investigating the error, I noticed that the `Trainer.go()` method uses the `zstd_compress` function, which doesn't compress directories. 

*trainer.go*
https://github.com/mozilla/bugbug/blob/edc76735d7e8e95820926f5fcfa9c9407431a3a2/scripts/trainer.py#L62

*zstd_compress*
https://github.com/mozilla/bugbug/blob/edc76735d7e8e95820926f5fcfa9c9407431a3a2/bugbug/utils.py#L236-L240

So, I updated the `Trainer.go()` method to use the `create_tar_zst` function if the `model_file_name` is a directory, and the `test_trainer` passed:

```python
if os.path.isdir(model_file_name):
    create_tar_zst(f"{model_file_name}.tar.zst")
else:
    zstd_compress(model_file_name)
```

I would appreciate your feedback or suggestions on this before I proceed with making these changes. 🙏🏾

suhaibmujahid:
@PromiseFru looks good to me! There is some feedback that I will add as inline comments when you commit the changes.

In general, it will be great to consider the following:
- Add comments to explain "why" where necessary 
- Avoid redundant code
- Consider extracting some logic in a separate method/function where necessary

PromiseFru:
> @PromiseFru looks good to me! There is some feedback that I will add as inline comments when you commit the changes.
> 
> In general, it will be great to consider the following:
> 
>     * Add comments to explain "why" where necessary
> 
>     * Avoid redundant code
> 
>     * Consider extracting some logic in a separate method/function where necessary

Hello @suhaibmujahid, thank you for the feedback.
 
I made the updates I mentioned earlier and made some improvements based on your feedback. I couldn't implement every point in the feedback because I wasn't sure of the scope of the improvement. I would greatly appreciate your comments. 🙏

bugbug/model.py
```diff
@@ -567,22 +569,55 @@ def train(self, importance_cutoff=0.15, limit=None):
 
             self.clf.fit(X_train, self.le.transform(y_train))
 
-        with open(self.__class__.__name__.lower(), "wb") as f:
-            pickle.dump(self, f, protocol=pickle.HIGHEST_PROTOCOL)
+        model_name = self.__class__.__name__.lower()
+
+        if hasattr(self.clf, "save_model"):
```
suhaibmujahid, line 574: It is not robust to consider any classifier with `save_model()` method to be XGBoost. We could check the class instead.

bugbug/model.py
```diff
@@ -567,22 +569,55 @@ def train(self, importance_cutoff=0.15, limit=None):
...
+            # If the model has the "save_model" attribute (is an XGBoost model),
+            # To handle multiple model files as one, place them in a directory with the expected model file name.
+            # when compressing the directory, it results in a single file.
+            os.makedirs(model_name, exist_ok=True)
+
+            # Save the XGBoost model to a file named "model.xgb"
+            self.clf.save_model(os.path.join(model_name, "model.xgb"))
```
suhaibmujahid, line 581: To allow us in future to support saving different types of models in the same way, we could specify the model type in the file name (e.g., `xgboost.ubj`).

Using `.ubj` as a file extension here will instruct XGBoost to save the model in binary JSON format.

bugbug/model.py
```diff
@@ -567,22 +569,55 @@ def train(self, importance_cutoff=0.15, limit=None):
...
+            with open(model_name, "wb") as f:
+                pickle.dump(self, f, protocol=pickle.HIGHEST_PROTOCOL)
 
         if self.store_dataset:
-            with open(f"{self.__class__.__name__.lower()}_data_X", "wb") as f:
+            with open(f"{model_name}_data_X", "wb") as f:
                 pickle.dump(X, f, protocol=pickle.HIGHEST_PROTOCOL)
 
-            with open(f"{self.__class__.__name__.lower()}_data_y", "wb") as f:
+            with open(f"{model_name}_data_y", "wb") as f:
```
suhaibmujahid, line 598: This is outside the scope of this PR. We could do it later as a separate PR.

scripts/trainer.py
```diff
@@ -59,7 +59,11 @@ def go(self, args):
 
         model_file_name = f"{model_name}model"
         assert os.path.exists(model_file_name)
-        zstd_compress(model_file_name)
+
+        if os.path.isdir(model_file_name):
+            create_tar_zst(f"{model_file_name}.tar.zst")
```
suhaibmujahid, line 64: Why do we need to add `.tar.zst` here?

scripts/trainer.py
```diff
@@ -59,7 +59,11 @@ def go(self, args):
...
+        else:
+            zstd_compress(model_file_name)
```
suhaibmujahid, line 66: How do we handle this in the decompress process?

bugbug/model.py
```diff
@@ -567,22 +569,55 @@ def train(self, importance_cutoff=0.15, limit=None):
...
             self.clf.fit(X_train, self.le.transform(y_train))
 
-        with open(self.__class__.__name__.lower(), "wb") as f:
-            pickle.dump(self, f, protocol=pickle.HIGHEST_PROTOCOL)
+        model_name = self.__class__.__name__.lower()
+
+        if hasattr(self.clf, "save_model"):
+            # If the model has the "save_model" attribute (is an XGBoost model),
+            # To handle multiple model files as one, place them in a directory with the expected model file name.
+            # when compressing the directory, it results in a single file.
```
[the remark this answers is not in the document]
PromiseFru: I made some updates 

https://github.com/mozilla/bugbug/blob/ca616d2c3fc95b218b892b8252f677b5b2eba2e4/bugbug/model.py#L575-L577

bugbug/model.py
PromiseFru: I made some updates 

https://github.com/mozilla/bugbug/blob/ca616d2c3fc95b218b892b8252f677b5b2eba2e4/bugbug/model.py#L574

bugbug/model.py
PromiseFru: I made some updates 

https://github.com/mozilla/bugbug/blob/ca616d2c3fc95b218b892b8252f677b5b2eba2e4/bugbug/model.py#L580

scripts/trainer.py
PromiseFru: The decompression process doesn't seem to handle this yet. We could include a similar check to verify if `model_file_name` is a directory and then use the `extract_tar_zst` function for the decompression.

https://github.com/mozilla/bugbug/blob/fc6e17b2d4934c3d3c6ac9a51f91b80f325c2be9/bugbug/utils.py#L217-L233

suhaibmujahid: To have the data saved, the compressed file name should match what it is declared in `data-pipeline.yml`.

Example:
https://github.com/mozilla/bugbug/blob/d13c4b15053ec7fdbc0b6ea44acb762d59e315b6/infra/data-pipeline.yml#L983-L991

bugbug/model.py
```diff
@@ -567,8 +569,25 @@ def train(self, importance_cutoff=0.15, limit=None):
...
             self.clf.fit(X_train, self.le.transform(y_train))
 
-        with open(self.__class__.__name__.lower(), "wb") as f:
-            pickle.dump(self, f, protocol=pickle.HIGHEST_PROTOCOL)
+        model_name = self.__class__.__name__.lower()
+
+        if isinstance(self.clf, XGBClassifier):
+            # Since the workflow expects a single file
+            # we create a directory with the expected file name
+            # to store multiple model related files.
```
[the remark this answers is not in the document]
PromiseFru: I made updates 

https://github.com/mozilla/bugbug/blob/9001ce14cf75fd8cbab21ee3d84f4441b2e97da1/bugbug/model.py#L575-L576

scripts/trainer.py
PromiseFru: I attempted to compress the directory and give it a name that matches the expected file name, which is "model_file_name.zst." However, I was unsuccessful because I couldn't modify the name. So, instead, I compressed the model files in the directory and renamed the directory to match the expected file name. During the decompression process, the model files within the directory will be decompressed, and the directory will be renamed to remove the ".zst" extension.

https://github.com/mozilla/bugbug/blob/9001ce14cf75fd8cbab21ee3d84f4441b2e97da1/scripts/trainer.py#L63-L74

https://github.com/mozilla/bugbug/blob/9001ce14cf75fd8cbab21ee3d84f4441b2e97da1/bugbug/utils.py#L238-L250

PromiseFru: Hello @suhaibmujahid, thank you for the review. I've made some changes. I archived the directory and compressed it with a name that matches the expected file name. During decompression, it checks if the resulting file is a `tarfile`; if it is, it extracts the tar archive.

https://github.com/mozilla/bugbug/blob/7a7e399b66c0d5661c90474951024f525b868b45/scripts/trainer.py#L64-L77

https://github.com/mozilla/bugbug/blob/7a7e399b66c0d5661c90474951024f525b868b45/bugbug/utils.py#L237-L245

PromiseFru:
Hello @suhaibmujahid, I have made some updates to this. Please, I would appreciate your feedback. 🙏

bugbug/model.py
```diff
@@ -567,8 +569,24 @@ def train(self, importance_cutoff=0.15, limit=None):
...
+        if issubclass(type(self.clf), XGBModel):
+            # Since the workflow expects a single file, we create a directory
+            # with the expected file name to store multiple model-related files.
+            os.makedirs(model_name, exist_ok=True)
+
+            self.clf.save_model(os.path.join(model_name, "xgboost.ubj"))
+
+            # Since we save the classifier separately, we need to clear the clf
+            # attribute to prevent it from being pickled with the model object.
+            self.clf = None
```
suhaibmujahid, line 583: Using `Booster.load_model()` could cause problems when using it with Scikit-learn API.[^1] Thus, instead of replacing `self.clf` with `None`, we could replace it with any empty class. Then, when loading the mode, we could use the `self.clf.load_model()` method for that.

[^1]: https://mljar.com/blog/xgboost-save-load-python/#save-the-xgboost-booster-object

PromiseFru:
Hello @suhaibmujahid, thank you for the review. Based on your feedback, as we'll be compressing the model into a tar file, we could consistently store the models in a directory, rather than using a directory only when it's an xgboost model. This would reduce redundant if-logic and accommodate more model types.

suhaibmujahid:
> we could consistently store the models in a directory, rather than using a directory only when it's an xgboost model. This would reduce redundant if-logic and accommodate more model types.

Good idea 👍

bugbug/model.py
```diff
@@ -567,7 +569,20 @@ def train(self, importance_cutoff=0.15, limit=None):
...
+        # Since the workflow expects a single file, we create a directory
+        # with the expected file name to store multiple model-related files.
+        os.makedirs(model_name, exist_ok=True)
+
+        if issubclass(type(self.clf), XGBModel):
+            self.clf.save_model(os.path.join(model_name, "xgboost.ubj"))
+
+            # Since we save the classifier separately, we need to clear the clf
+            # attribute to prevent it from being pickled with the model object.
+            self.clf = XGBModel()
```
suhaibmujahid, line 583: In this way, before saving the model, `clf` could be `XGBClassifier` but when loading the model again, it will always be `XGBModel`.

I guess the following could work. Please test it:
```suggestion
            self.clf = self.clf.__class__(n_jobs=self.clf.n_jobs)
```

If you check the `load_model()` method, `n_jobs` is the only attribute that is used from the loading class before reading the JSON file.

bugbug/model.py
```diff
@@ -581,8 +596,13 @@ def train(self, importance_cutoff=0.15, limit=None):
 
     @staticmethod
     def load(model_file_name: str) -> "Model":
-        with open(model_file_name, "rb") as f:
-            return pickle.load(f)
+        with open(os.path.join(model_file_name, "pickled.pkl"), "rb") as f:
```
suhaibmujahid, line 599: It will be cleaner to store the file path in a variable. Less nesting:
```suggestion
        model_path = os.path.join(model_file_name, "model.pkl")
        with open(model_path, "rb") as f:
```

bugbug/model.py
```diff
@@ -581,8 +596,13 @@ def train(self, importance_cutoff=0.15, limit=None):
...
+            model = pickle.load(f)
+
+        if os.path.exists(os.path.join(model_file_name, "xgboost.ubj")):
+            model.clf.load_model(os.path.join(model_file_name, "xgboost.ubj"))
```
suhaibmujahid, line 603: Here even is more helpful to store the file path in a variable.

bugbug/model.py
```diff
@@ -4,6 +4,7 @@
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import logging
+import os
```
suhaibmujahid, line 7: Importing the path module could be cleaner:
```suggestion
from os import path
```

scripts/trainer.py
```diff
@@ -59,7 +59,8 @@ def go(self, args):
 
         model_file_name = f"{model_name}model"
```
suhaibmujahid, line 60: Since it is not a file anymore, we may need to rename it (even in other places):
```suggestion
        model_dir = f"{model_name}model"
```

bugbug/model.py
PromiseFru: This works!

bugbug/model.py
```diff
@@ -580,9 +594,17 @@ def train(self, importance_cutoff=0.15, limit=None):
         return tracking_metrics
 
     @staticmethod
-    def load(model_file_name: str) -> "Model":
-        with open(model_file_name, "rb") as f:
-            return pickle.load(f)
+    def load(model_directory: str) -> "Model":
+        model_path = path.join(model_directory, "model.pkl")
+
```
suhaibmujahid, line 599: Unnecessary white line:  
```suggestion
```

bugbug/utils.py
```diff
@@ -231,12 +231,12 @@ def download_model(model_name: str) -> str:
...
-    url = f"https://community-tc.services.mozilla.com/api/index/v1/task/project.bugbug.train_{model_name}.{version}/artifacts/public/{path}.zst"
+    url = f"https://community-tc.services.mozilla.com/api/index/v1/task/project.bugbug.train_{model_name}.{version}/artifacts/public/{path}.tar.zst"
     logger.info("Downloading %s...", url)
     updated = download_check_etag(url)
     if updated:
-        zstd_decompress(path)
-        os.remove(f"{path}.zst")
+        extract_tar_zst(f"{path}.tar.zst")
+        os.remove(f"{path}.tar.zst")
     assert os.path.exists(path), "Decompressed file exists"
```
suhaibmujahid, line 240: It is not a file anymore 😉
```suggestion
    assert os.path.exists(path), "Decompressed directory exists"
```

scripts/trainer.py
```diff
@@ -57,17 +57,18 @@ def go(self, args):
...
 
         if model_obj.store_dataset:
-            assert os.path.exists(f"{model_file_name}_data_X")
-            zstd_compress(f"{model_file_name}_data_X")
-            assert os.path.exists(f"{model_file_name}_data_y")
-            zstd_compress(f"{model_file_name}_data_y")
+            assert os.path.exists(f"{model_directory}_data_X")
+            zstd_compress(f"{model_directory}_data_X")
+            assert os.path.exists(f"{model_directory}_data_y")
+            zstd_compress(f"{model_directory}_data_y")
```
suhaibmujahid, line 71: It is a bit confusing to use `model_directory` because they have nothing to do with the model directory:
```suggestion
            assert os.path.exists(f"{model_name}model_data_X")
            zstd_compress(f"{model_name}model_data_X")
            assert os.path.exists(f"{model_name}model_data_y")
            zstd_compress(f"{model_name}model_data_y")
```

scripts/trainer.py
```diff
@@ -57,17 +57,18 @@ def go(self, args):
 
         logger.info("Training done")
 
-        model_file_name = f"{model_name}model"
-        assert os.path.exists(model_file_name)
-        zstd_compress(model_file_name)
+        model_directory = f"{model_name}model"
+        assert os.path.exists(model_directory)
+
```
suhaibmujahid, line 62: Unnecessary white line:
```suggestion
```

suhaibmujahid (commented):
@PromiseFru Looks very clean. Thank you. I have just minor comments.

scripts/trainer.py
PromiseFru: > It is a bit confusing to use `model_directory` because they have nothing to do with the model directory:

Ah right! I thought so as well 🤓

suhaibmujahid (approved):
@PromiseFru thank you! LGTM.

@marco-c we need to retrain the models before deploying this.

bugbug/model.py
```diff
@@ -567,7 +569,19 @@ def train(self, importance_cutoff=0.15, limit=None):
...
+        model_directory = self.__class__.__name__.lower()
+        makedirs(model_directory, exist_ok=True)
+
+        if issubclass(type(self.clf), XGBModel):
+            xgboost_model_path = path.join(model_directory, "xgboost.ubj")
+            self.clf.save_model(xgboost_model_path)
+
+            # Since we save the classifier separately, we need to clear the clf
+            # attribute to prevent it from being pickled with the model object.
+            self.clf = self.clf.__class__(n_jobs=self.clf.n_jobs)
```
marco-c, line 581: Why are we passing n_jobs here? What if a submodel uses other parameters than n_jobs?

marco-c:
> @marco-c we need to retrain the models before deploying this.

The models will be automatically retrained when we deploy, we won't need to do anything manually.

bugbug/model.py
suhaibmujahid: This would be valid in older versions of xgboost because it was the only parameter that would be used before copying everything from the JSON dump. However, I just noticed that this got changed in https://github.com/dmlc/xgboost/pull/8963 where none of the hyper-parameters will be saved/loaded. 

^^ This reminds us to never rely on implementation details :)

Currently, `n_jobs` is the only parameter that we use when we instantiate our xgboost models. We may file a follow-up issue to store hyper-parameters in an attribute (dictionary) so we use it when instantiating a model for `clf`.

marco-c: I wonder why it isn't just saved along with the model!

Yeah, we should file a follow-up for this.

marco-c (approved):

bugbug/model.py
suhaibmujahid: > I wonder why it isn't just saved along with the model!

It was doing that, but that got changed: "The scikit-learn interface `save_model` will save only the model and discard all hyper-parameters. This is to align with the native XGBoost interface, which distinguishes the hyper-parameter with model parameters."[^1]

[^1]: https://github.com/dmlc/xgboost/pull/8963

> Yeah, we should file a follow-up for this.

I filed an issue: https://github.com/mozilla/bugbug/issues/3794

Issue #3735, opened 2023-10-16
Remove usage of the xgboost `predictor` parameter since it is deprecated


marco-c:
E.g. https://github.com/mozilla/bugbug/blob/c025bf003fc04939b96d84b2a40c6538e8c6f82e/bugbug/models/spambug.py#L84

PR #3736, opened 2023-10-16
Removed all instances of the xgboost `predictor` parameter 

PromiseFru:
Closes: #3735

ref: https://github.com/dmlc/xgboost/blob/master/NEWS.md#new-device-parameter


suhaibmujahid (approved):
Thank you! LGTM.

Issue #793, opened 2019-07-26
Restrict the training set of the RegressionRange model only to regressions


marco-c:

Axhie:
RegressionModel
A regression model provides a function that describes the relationship between one or more independent variables and a response, dependent, or target variable. Predictive modelling techniques such as regression model may be used to determine the relationship between a dataset’s dependent (goal) and independent variables. It is widely used when the dependent and independent variables are linked in a linear or non-linear fashion, and the target variable has a set of continuous values. Thus, regression model approaches help establish causal relationships between variables, modelling time series, and forecasting. 
 Types of RegressionModel
1. Linear regression analysis is used to predict the value of a variable based on the value of another variable. The variable you want to predict is called the dependent variable. The variable you are using to predict the other variable's value is called the independent variable.
2. Logistic regression is a statistical model (also known as logit model) is often used for classification and predictive analytics. Logistic regression estimates the probability of an event occurring, such as voted or didn't vote, based on a given dataset of independent variables.
 3. Polynomial Regression is a technique of polynomial regression analysis is used to represent a      non-linear relationship between dependent and independent variables. It is a variant of the multiple linear regression model, except that the best fit line is curved rather than straight.
4. Ridge regression is a method of estimating the coefficients of multiple-regression models in scenarios where the independent variables are highly correlated. It has been used in many fields including econometrics, chemistry, and engineering.
5. Lasso regression is a regularization technique. It is used over regression methods for a more accurate prediction. This model uses shrinkage. Shrinkage is where data values are shrunk towards a central point as the mean. The lasso procedure encourages simple, sparse models (i.e. models with fewer parameters).
6. Quantile regression is a type of regression analysis used in statistics and econometrics. Whereas the method of least squares estimates the conditional mean of the response variable across values of the predictor variables, quantile regression estimates the conditional median of the response variable
7. Bayesian linear regression is a form of regression analysis technique used in machine learning that uses Bayes’ theorem to calculate the regression coefficients’ values. Rather than determining the least-squares, this technique determines the features’ posterior distribution
8. Principal component analysis (PCA) is used first to modify the training data, and then the resulting transformed samples are used to train the regression.
9. The partial least squares regression technique is a fast and efficient covariance-based regression analysis technique. It is advantageous for regression problems with many independent variables with a high probability of multicollinearity between the variables. The method decreases the number of variables to a manageable number of predictors, then is utilized in a regression.
[...]

PR #3746, opened 2023-10-23
Restrict the training set of the RegressionRange model only to regressions

PromiseFru:
Closes: #793

Hello @suhaibmujahid @marco-c,

I executed the trainer:

```bash
python scripts/trainer.py regressionrange
```

Here are the outputs before and after the changes:

[regressionrange_before.txt](https://github.com/mozilla/bugbug/files/13068242/regressionrange_before.txt)
[regressionrange_after.txt](https://github.com/mozilla/bugbug/files/13068244/regressionrange_after.txt)

However, I keep getting the error:

```bash
ValueError: The target 'y' needs to have more than 1 class. Got 1 class instead.
```

I would appreciate any assistance or insights on it. 🙏

### DIFF
--- a/bugbug/models/regressionrange.py
+++ b/bugbug/models/regressionrange.py
@@ -69,7 +69,7 @@ class RegressionRangeModel(BugModel):
         classes = {}
 
         for bug_data in bugzilla.get_bugs():
-            if not bug_data["regressions"]:
+            if "regression" not in bug_data["keywords"]:
                 continue
 
             bug_id = int(bug_data["id"])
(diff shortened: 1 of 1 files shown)


bugbug/models/regressionrange.py
```diff
@@ -69,6 +69,9 @@ def get_labels(self):
         classes = {}
 
         for bug_data in bugzilla.get_bugs():
+            if not bug_data["regressions"]:
```
suhaibmujahid, line 72: As you can see in the [bugzilla docs](https://bmo.readthedocs.io/en/latest/api/core/v1/bug.html), the `regressions` field lists the IDs of bugs that are introduced by this bug. Thus, this bug is a regressor, not a regression.

Regressions are the bugs that have the `regression` keyword.

PromiseFru: Thank you for the review. I made some updates 🙂

suhaibmujahid (approved):
LGTM. Thank you!

PR #3748, opened 2023-10-23
Include new task to train models with task cluster

### LINKED ISSUE #3688: Train models on TaskCluster whenever their corresponding files are modified
Whenever a file in the `bugbug/models` directory is modified, we could have a task to train the corresponding model.

For example, if a new Pull Request is modifying the `bugbug/models/defect.py` file, the task should train the `defect` model:
```
python -m scripts.trainer defect
```

jpangas:
Fixes #3688 

This PR includes a new task in `.taskcluster.yml` that trains models with task cluster.

Comments:
To train a model, some one has to include the keyword at the bottom of the Pull Request description. The model and arguments should then follow after the keyword.

Tested with:
Train on Taskcluster: spambug


jpangas:
The PR is now ready. 
To train a model, some one has to include the keyword `Train with task cluster:` at the bottom of the description. The model and arguments should then follow after the keyword. 
e.g. `Train with task cluster: component --classifier nn` is similar to `python3 -m scripts.trainer component --classifier nn`

jpangas:
WDYT about this approach? @suhaibmujahid and @marco-c

.taskcluster.yml
```diff
@@ -206,6 +210,35 @@ tasks:
...
 
+      - $if: 'tasks_for == "github-pull-request" && "Train with task cluster:" in pr_description'
+        then:
+          dependencies:
+            - { $eval: as_slugid("docker_build") }
+          taskId: { $eval: as_slugid("train_on_taskcluster") }
+          created: { $fromNow: "" }
+          deadline: { $fromNow: "1 day" }
+          provisionerId: proj-bugbug
+          workerType: compute-smaller
```
jpangas, line 221: I wanted to know what would be the most ideal `workerType` for training the models. Training some of the models e.g. `component` are extremely memory intensive.

.taskcluster.yml
```diff
@@ -206,6 +210,35 @@ tasks:
...
+          payload:
+            maxRunTime: 10800
+            image: python:3.10
```
suhaibmujahid, line 224: We could instead rely on the image that we already built. In that image, we have the dependencies already installed and the repository is cloned.

.taskcluster.yml
```diff
@@ -206,6 +210,35 @@ tasks:
...
+            command:
+              - "/bin/sh"
+              - "-lcx"
+              - 'apt-get -qq update &&
+                apt-get -qq install -y zstd &&
+                git clone --quiet ${repository} &&
+                cd bugbug &&
+                git -c advice.detachedHead=false checkout ${head_rev} &&
+                pip install --disable-pip-version-check --no-cache-dir --progress-bar off -r requirements.txt &&
+                train_options=$(python -c "print(''''''${pr_description}''''''.rsplit('':'', 1)[-1].strip())") &&
```
suhaibmujahid, line 234: A regex to extract the model name would be more suitable. Also, it could be better to do it as a variable instead of `pr_description`.

.taskcluster.yml
marco-c: This is an example of using an image built from an earlier task: https://github.com/mozilla/task-boot/blob/bd37ee306ba4b40f0d21818e57b59b55579a63f3/.taskcluster.yml#L143.

jpangas: Oh yes! I think we have to rely on the image from `docker_build`

.taskcluster.yml
```diff
@@ -206,6 +210,34 @@ tasks:
...
+            image:
+              type: task-image
+              path: public/bugbug/bugbug-base.tar.zst
+              taskId: { $eval: as_slugid("docker_build") }
+            env:
+              TEXT_TO_EXTRACT: "${pr_description}"
+            command:
+              - "/bin/sh"
+              - "-lcx"
+              - 'model=$(python -c ''import re, os; match = re.search(r"Train with task cluster:(.+)", os.environ.get("TEXT_TO_EXTRACT")); model = match.group(1).strip() ; print(model)'') &&
```
suhaibmujahid, line 233: Why don't we do this logic inside `scripts.trainer`?

.taskcluster.yml
```diff
@@ -206,6 +210,34 @@ tasks:
           owner: mcastelluccio@mozilla.com
           source: ${repository}/raw/${head_rev}/.taskcluster.yml
 
+      - $if: 'tasks_for == "github-pull-request" && "Train with task cluster:" in pr_description'
```
suhaibmujahid, line 213:
```suggestion
      - $if: 'tasks_for == "github-pull-request" && "Train on Taskcluster:" in pr_description'
```

.taskcluster.yml
suhaibmujahid, line 233:
```suggestion
              - 'model=$(python -c ''import re, os; match = re.search(r"Train with task cluster:\s+([a-z_1-9]+)", os.environ.get("TEXT_TO_EXTRACT")); model = match.group(1) ; print(model)'') &&
```

jpangas: I was trying to maintain the conditions of the arguments required to train the model. i.e. `model` is a required argument when running the training script. Is it okay to change this behaviour? since at the moment of running the script, with an option like `--auto-detect-model-name`, the model_name is not provided.

suhaibmujahid: I see. We could create a script in the `scripts` directory and use it to get the model name. This will be more flexible, testable, and maintainable.

jpangas: Please check the new script and let me know what you think.

suhaibmujahid: Thank you for your work. Sorry that my comment may not be clear. I did not mean to create a new trnner script. I meant to convert the code that was in the `python -c` to a standalone code and then use it to pass the options to the model trainer.

Example:
```
python -m scripts.trainer $(python -m scripts.trainer_extract_args)
```

scripts/train_on_tc.py
suhaibmujahid, line 1: It may be simplified to only extract the options that we need to pass to the trainer script. At this point, it is the model name only. Later there could be other options as well. You do not need to care about that now.

scripts/train_on_tc.py
```diff
@@ -0,0 +1,62 @@
...
+import os
+import re
+import subprocess
+import sys
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+def get_model_name(env_variable: str):
```
suhaibmujahid, line 17: We could always assume that the env var name is `PULL_REQUEST_DESCRIPTION`.

scripts/train_on_tc.py
jpangas: Sure! Done that in the latest changes. Thank you.

scripts/trainer_extract_args.py
```diff
@@ -0,0 +1,35 @@
...
+import re
+import sys
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+def get_model_name():
+    text = os.environ.get("PR_DESCRIPTION")
+
```
suhaibmujahid, line 17: Unnecessary white line:
```suggestion
```

scripts/trainer_extract_args.py
```diff
@@ -0,0 +1,35 @@
...
+    match = re.search(r"Train on Taskcluster:\s+([a-z_1-9]+)", text)
+
+    if not match:
+        logger.error("There is no match found for keyword 'Train on Taskcluster:'")
+        sys.exit(1)
+
+    model = match.group(1)
+
+    return model
```
suhaibmujahid, line 26: Nit:
```suggestion
    model_name = match.group(1)

    return model_name
```
or
```suggestion
    return match.group(1)
```

scripts/trainer_extract_args.py
```diff
@@ -0,0 +1,35 @@
...
+import logging
+import os
+import re
+import sys
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+def get_model_name():
```
suhaibmujahid, line 15: Missing the type hint: 
```suggestion
def get_model_name() -> str | None:
```

scripts/trainer_extract_args.py
```diff
@@ -0,0 +1,35 @@
...
+        sys.exit(1)
+
+    model = match.group(1)
+
+    return model
+
+
+def main():
+    model = get_model_name()
+    print(model)
```
suhaibmujahid, line 31:
```suggestion
    if model:
        print(model)
```

scripts/trainer_extract_args.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def get_model_name():
+    text = os.environ.get("PR_DESCRIPTION")
+
+    match = re.search(r"Train on Taskcluster:\s+([a-z_1-9]+)", text)
+
+    if not match:
+        logger.error("There is no match found for keyword 'Train on Taskcluster:'")
+        sys.exit(1)
```
suhaibmujahid, line 22: We should not do the flow control from this function. It is better to do it from the main:
```suggestion
        return None
```

scripts/trainer_extract_args.py
```diff
@@ -0,0 +1,35 @@
...
+logger = logging.getLogger(__name__)
+
+
+def get_model_name():
+    text = os.environ.get("PR_DESCRIPTION")
+
+    match = re.search(r"Train on Taskcluster:\s+([a-z_1-9]+)", text)
+
+    if not match:
+        logger.error("There is no match found for keyword 'Train on Taskcluster:'")
```
suhaibmujahid, line 21:
```suggestion
        logger.error("Could not identify the model name using the 'Train on Taskcluster' keyword from the Pull Request description")
```

scripts/trainer_extract_args.py
```diff
@@ -0,0 +1,35 @@
...
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+def get_model_name():
+    text = os.environ.get("PR_DESCRIPTION")
+
+    match = re.search(r"Train on Taskcluster:\s+([a-z_1-9]+)", text)
+
```
suhaibmujahid, line 19: Unnecessary white line:
```suggestion
```

scripts/trainer_extract_args.py
```diff
@@ -0,0 +1,35 @@
...
+import os
+import re
+import sys
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+def get_model_name():
+    text = os.environ.get("PR_DESCRIPTION")
```
suhaibmujahid, line 16:
```suggestion
    pr_description = os.environ.get("PR_DESCRIPTION")
    if not pr_description:
        logger.error(...)
        return None
    
```

.taskcluster.yml
```diff
@@ -206,6 +210,33 @@ tasks:
...
+          taskId: { $eval: as_slugid("train_on_taskcluster") }
+          created: { $fromNow: "" }
+          deadline: { $fromNow: "1 day" }
+          provisionerId: proj-bugbug
+          workerType: compute-large
+          payload:
+            maxRunTime: 10800
+            image:
+              type: task-image
+              path: public/bugbug/bugbug-base.tar.zst
```
suhaibmujahid, line 226: Some models require other images (e.g., `bugbug-base-nlp`). We do not need to handle that now, but it would be nice to follow up on this later.

.taskcluster.yml
```diff
@@ -206,6 +210,33 @@ tasks:
...
+              taskId: { $eval: as_slugid("docker_build") }
+            env:
+              PR_DESCRIPTION: "${pr_description}"
+            command:
+              - "/bin/sh"
+              - "-lcx"
+              - "python -m scripts.trainer $(python -m scripts.trainer_extract_args)"
+          metadata:
+            name: bugbug train on tc
```
suhaibmujahid, line 235:
```suggestion
            name: bugbug train on TC
```

.taskcluster.yml
```diff
@@ -206,6 +210,33 @@ tasks:
...
+            description: bugbug train on tc
```
suhaibmujahid, line 236:
```suggestion
            description: bugbug train on Taskcluster
```

suhaibmujahid (approved):
LGTM! Thank you, @jpangas.

suhaibmujahid (approved):

.taskcluster.yml
suhaibmujahid: I filed an issue https://github.com/mozilla/bugbug/issues/3796

PR #3752, opened 2023-10-23
Updated numpy version to 1.26.1 

gothwalritu:
because it was conflicting with pandas which was expecting a higher numpy version

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,7 +9,7 @@ matplotlib==3.8.0
 mercurial==6.5.2
 microannotate==0.0.24
 mozci==2.3.4
-numpy==1.22.4
+numpy==1.26.1 #updated numpy because of conflict with pandas==2.1.1
 orjson==3.9.9
 ortools==9.7.2996
 pandas==2.1.1


suhaibmujahid:
Thank you for the PR, @gothwalritu!

We rely on @dependabot to update the dependencies. The bot already opened a PR to update the dependency (i.e., #3731) but we did not merge it because there are some breaking changes.

BTW. I do not have a problem installing `pandas==2.1.1`; it also works on the CI.

I'm closing this PR for now, and we will merge #3731 once the breakages are fixed.

gothwalritu:
@suhaibmujahid Thanks, for the update. I was installing the requirements on my machine, when I got this error:

```
INFO: pip is still looking at multiple versions of pandas to determine which version is compatible with other requirements. This could take a while.
**ERROR**: Cannot install -r .\requirements.txt (line 14), -r .\requirements.txt (line 4), -r .\requirements.txt (line 8), numpy==1.22.4 and pandas==2.1.1 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested numpy==1.22.4
    imbalanced-learn 0.11.0 depends on numpy>=1.17.3
    matplotlib 3.8.0 depends on numpy<2 and >=1.21
    ortools 9.7.2996 depends on numpy>=1.13.3
    pandas 2.1.1 depends on numpy>=1.23.2; python_version == "3.11"

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict


```

suhaibmujahid:
@gothwalritu I see, as a workaround, using Python 10 should fix the issue.

https://github.com/pandas-dev/pandas/blob/v2.1.1/pyproject.toml#L32-L34

Issue #3759, opened 2023-10-26
Use the value of `regresed_by` in crafting the training classes for the RegressionRange model


suhaibmujahid:
Currently, we use the `regressionwindow-wanted` keyword and the value of the `cf_has_regression_range` field to determine whether there is a regression range: 

https://github.com/mozilla/bugbug/blob/798f215e4227030051f5aa2e5dcb55a24a6d15b2/bugbug/models/regressionrange.py#L76-L82

We should also use the value of the `regresed_by` field. If there is a value in the field, we should consider it as "has regression range".

PR #3764, opened 2023-10-27
[regressionrange] Use `regressed_by` instead of `cf_has_regression_range`

julietowah:
the "regressed_by" field is now being used as a factor in the categorization of regression ranges
 Fixes: #3759 

Train on Taskcluster: regressionrange

### DIFF
--- a/bugbug/models/regressionrange.py
+++ b/bugbug/models/regressionrange.py
@@ -75,12 +75,6 @@ class RegressionRangeModel(BugModel):
             bug_id = int(bug_data["id"])
             if "regressionwindow-wanted" in bug_data["keywords"]:
                 classes[bug_id] = 0
-            elif "cf_has_regression_range" in bug_data:
-                if bug_data["cf_has_regression_range"] == "yes":
-                    classes[bug_id] = 1
-                elif bug_data["cf_has_regression_range"] == "no":
-                    classes[bug_id] = 0
-            # Check the regressed_by field
             elif "regressed_by" in bug_data and bug_data["regressed_by"]:
                 classes[bug_id] = 1
             else:
--- a/bugbug/models/regressionrange.py
+++ b/bugbug/models/regressionrange.py
@@ -73,8 +73,9 @@ class RegressionRangeModel(BugModel):
                 continue
 
             bug_id = int(bug_data["id"])
-            if "regressionwindow-wanted" in bug_data["keywords"] or (
-                "regressed_by" in bug_data and bug_data["regressed_by"]
+            if (
+                bug_data.get("regressed_by")
+                or "regressionwindow-wanted" in bug_data["keywords"]
             ):
                 classes[bug_id] = 1
             else:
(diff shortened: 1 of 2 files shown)


test_suite_analysis/metadata.json
suhaibmujahid, line 1: Unrelated change.

bugbug/models/regressionrange.py
```diff
@@ -80,6 +80,11 @@ def get_labels(self):
                     classes[bug_id] = 1
                 elif bug_data["cf_has_regression_range"] == "no":
                     classes[bug_id] = 0
+            # Check the regressed_by field
```
suhaibmujahid, line 83: Comments in general should describe "why" not "what" the code is doing. I do not think the comment here is necessary.

bugbug/models/regressionrange.py
```diff
@@ -80,6 +80,11 @@ def get_labels(self):
...
+            elif "regressed_by" in bug_data and bug_data["regressed_by"]:
```
suhaibmujahid, line 84: The `cf_has_regression_range` has been removed. We can rely solely on `regressed_by` instead.

suhaibmujahid (commented):
Thank you @julietowah for the PR.

Please edit the PR title to be more descriptive. Also, the linking in the description is not pointing to an issue.

bugbug/models/regressionrange.py
julietowah: thank you so much on the feed back i will give the update soon

test_suite_analysis/metadata.json
julietowah: thank you so much 
i have updated the pr title

julietowah:
@suhaibmujahid i love the swift response i get from you and the corrections
i really appreciate
i did not join firefox on time but i i was able to find my way around to some extent because the community was so supportive 
thank you

bugbug/models/regressionrange.py
```diff
@@ -74,12 +74,11 @@ def get_labels(self):
 
             bug_id = int(bug_data["id"])
             if "regressionwindow-wanted" in bug_data["keywords"]:
+                classes[bug_id] = 1
+            elif "regressed_by" in bug_data and bug_data["regressed_by"]:
```
suhaibmujahid, line 78: This could be similified to use only one `if` statement.

test_suite_analysis/metadata.json
suhaibmujahid: Why did you add the `test_suite_analysis/metadata.json` file here? It does not seem related to #3759.

bugbug/models/regressionrange.py
julietowah: done

test_suite_analysis/metadata.json
julietowah: @suhaibmujahid please can you explain more i don't understand

julietowah: fixed it

suhaibmujahid (dismissed):
Thank you, @julietowah! LGTM!

suhaibmujahid (approved):

PR #3766, opened 2023-10-27
Update regressionrange.py

Tanya-Rawat:
@suhaibmujahid 
#3759 
If "regressed_by" is present and has a value, set the class to 1 (indicating regression range).
If none of the above conditions are met, set the class to 0 as a default.


suhaibmujahid:
Thank you @Tanya-Rawat. Unfortunately, this is a duplicate of #3764.

Tanya-Rawat:
@suhaibmujahid  Sorry, I didn't check that the other pull request was open for the issue.

PR #3778, opened 2023-10-31
Get rollback inconsistencies using multiple processes

marco-c:


bugbug/bug_snapshot.py
```diff
@@ -868,17 +869,22 @@ def parse_flag_change(change):
...
 
-def get_inconsistencies(bugs):
-    inconsistencies = []
+def _try_rollback(bug):
+    try:
+        rollback(bug, do_assert=True)
+        return None
+    except Exception:
+        logger.exception("Failed to rollback bug %s", bug["id"])
+        return bug
```
suhaibmujahid, line 878: Returning `None` does the work of distinguishing between successful and failed rollback. However, it is not intuitive why the function returns `None`  when successful and returns a bug when it fails. Thus, I suggest to be more consistent with what the `rollback()` would do:

```suggestion
def _try_rollback(bug):
    try:
        return rollback(bug, do_assert=True), True
    except Exception:
        logger.exception("Failed to rollback bug %s", bug["id"])
        return bug, False
```

bugbug/bug_snapshot.py
```diff
@@ -868,17 +869,22 @@ def parse_flag_change(change):
...
-            logger.exception("Failed to rollback bug %s", bug["id"])
-            inconsistencies.append(bug)
 
-    return inconsistencies
+def get_inconsistencies(bugs):
+    with Pool() as p:
+        return [
+            bug
+            for bug in p.imap(_try_rollback, bugs, chunksize=1024)
+            if bug is not None
```
suhaibmujahid, line 886:
```suggestion
            for bug, is_successful in p.imap(_try_rollback, bugs, chunksize=1024)
            if is_successful
```

If it were in GO, I would use `ok` instead of `is_successful` ;)

suhaibmujahid (dismissed):
LGTM! I just added some suggestions to improve the readability.

bugbug/bug_snapshot.py
marco-c: Good idea, done!

suhaibmujahid (approved):

PR #3799, opened 2023-11-09
Resolve issues concerning upgrading scikit-learn to latest version

jpangas:
**Issue:**
Closes #3798 

**Comments**:
This PR works towards investigating and resolving the issues concerning the upgrade to sci-kit learn's latest version, which is currently 1.4.1

To help with investigation, each new push will trigger the training a regression model on TC.

Train on Taskcluster: regression

PR #3810, opened 2023-11-09
Reduce `maxRunTime` to 45mins for Regression model when training on TaskCluster

### LINKED ISSUE #3809: Reduce the `maxRunTime` when training models on Taskcluster
> As the task normally takes around 30 minutes, could you reduce the timeout to a shorter time? Maybe something like 45 minutes?

_Originally posted by @marco-c in https://github.com/mozilla/bugbug/issues/3799#issuecomment-1804130532_
       

This is related to #3748.

jpangas:
**Issue:**
Closes #3809 

**Comments:**
This reduces the maximum time to 45 minutes (2700 seconds) when training the regression model in the train on TC task.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -220,7 +220,7 @@ tasks:
           provisionerId: proj-bugbug
           workerType: compute-large
           payload:
-            maxRunTime: 10800
+            maxRunTime: 2700
             image:
               type: task-image
               path: public/bugbug/bugbug-base.tar.zst
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -220,7 +220,10 @@ tasks:
           provisionerId: proj-bugbug
           workerType: compute-large
           payload:
-            maxRunTime: 2700
+            maxRunTime:
+              $switch:
+                '"Train on Taskcluster: regression" in pr_description': 2700
+                $default: 10800
             image:
               type: task-image
               path: public/bugbug/bugbug-base.tar.zst


suhaibmujahid:
@marco-c is 45 minutes enough for any of the current model types to be trained on a `compute-large` worker?

suhaibmujahid:
I see the timeout for all training tasks is `25200` (7 hours):
https://github.com/mozilla/bugbug/blob/736a76e8d75ab111f00ec185e34bdab1452ef233/infra/data-pipeline.yml#L889

marco-c:
It is not enough for all of the models. We could either use it as a safe default and change it manually to a larger value when we need, or set the default to a larger value and change it to lower when we need (e.g. for #3799).

.taskcluster.yml
```diff
@@ -220,7 +220,7 @@ tasks:
           provisionerId: proj-bugbug
           workerType: compute-large
           payload:
-            maxRunTime: 10800
+            maxRunTime: 2700
```
suhaibmujahid, line 223: Based on @marco-c's suggestion in https://github.com/mozilla/bugbug/pull/3810#issuecomment-1804800058, we could keep the current value as the default value and customize it for the models that need less time. For now, we could start with the  `regression` model. We will later add more models as required.

```suggestion
            maxRunTime:
              {
                $switch:
                  {
                    '"Train on Taskcluster: regression" in pr_description': 2700,
                    $default: 10800,
                  },
              }
```

suhaibmujahid (approved):
Thank you, @jpangas!

Issue #3794, opened 2023-11-06
Store the model hyper-parameters in a seperate attribute


suhaibmujahid:
> We may file a follow-up issue to store hyper-parameters in an attribute (dictionary) so we use it when instantiating a model for `clf`.

_Originally posted by @suhaibmujahid in https://github.com/mozilla/bugbug/pull/3707#discussion_r1382429645_

PR #3811, opened 2023-11-10
Store model hyper-parameters in a separate attribute

PromiseFru:
Resolves: #3794

### DIFF
--- a/bugbug/models/annotate_ignore.py
+++ b/bugbug/models/annotate_ignore.py
@@ -79,8 +79,8 @@ class AnnotateIgnoreModel(CommitModel):
             ]
         )
 
-        self.param = {"n_jobs": utils.get_physical_cpu_count()}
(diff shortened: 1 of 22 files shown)


bugbug/models/annotate_ignore.py
```diff
@@ -79,7 +79,8 @@ def __init__(self, lemmatization: bool = False) -> None:
             ]
         )
 
-        self.clf = xgboost.XGBClassifier(n_jobs=utils.get_physical_cpu_count())
+        self.param = {"n_jobs": utils.get_physical_cpu_count()}
```
suhaibmujahid, line 82: I would call it `hyperparameter`:
```suggestion
        self.hyperparameter = {"n_jobs": utils.get_physical_cpu_count()}
```

suhaibmujahid (approved):
LGTM. Thank you, @PromiseFru!

Issue #3838, opened 2023-11-20
Add a task to train invalid_compatibility_report model


ksy36:
Add a task to train invalid_compatibility_report model to https://github.com/mozilla/bugbug/blob/master/infra/data-pipeline.yml

PR #3840, opened 2023-11-20
Add a task to train invalid_compatibility_report model

ksy36:
Fixes https://github.com/mozilla/bugbug/issues/3838

### DIFF
--- a/infra/data-pipeline.yml
+++ b/infra/data-pipeline.yml
@@ -1238,6 +1238,44 @@ tasks:
         owner: release-mgmt-analysis@mozilla.com
         source: ${repository}/raw/master/data-pipeline.yml
 
+    - ID: train-invalidcompatibilityreport
+      created: { $fromNow: "" }
+      deadline: { $fromNow: "3 days" }
+      expires: { $fromNow: "1 year" }
+      provisionerId: proj-bugbug
+      workerType: compute-small
+      dependencies:
+        - issues-retrieval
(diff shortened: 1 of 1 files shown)


bugbug/models/invalid_compatibility_report.py
```diff
@@ -50,7 +50,8 @@ def __init__(self, lemmatization=False):
             ]
         )
 
-        self.clf = xgboost.XGBClassifier(n_jobs=utils.get_physical_cpu_count())
+        self.hyperparameter = {"n_jobs": utils.get_physical_cpu_count()}
+        self.clf = xgboost.XGBClassifier(**self.hyperparameter)
```
ksy36, line 55: (noticed this was added to all models and the training task fails without it)

suhaibmujahid: Could you please submit it in a separate PR?

ksy36: Yeah, sounds good!

marco-c (approved):
LGTM!

marco-c:
I will make a release and trigger the hook so we can see the task in action.

ksy36:
Thanks!

marco-c:
Here's the training task: https://community-tc.services.mozilla.com/tasks/GtGcPynVQB63abxYQksHjA/runs/0.

Issue #3841, opened 2023-11-20
Add hyperparameter attribute to invalid_compatibility_report model


ksy36:
```python
self.hyperparameter = {"n_jobs": utils.get_physical_cpu_count()}
self.clf = xgboost.XGBClassifier(**self.hyperparameter)
```

PR #3842, opened 2023-11-20
Add hyperparameter attribute to invalid_compatibility_report model

ksy36:
Fixes https://github.com/mozilla/bugbug/issues/3841


suhaibmujahid (approved):
Thank you, @ksy36!

Issue #3839, opened 2023-11-20
Add support for classifying invalid compatibility reports to the http service


ksy36:
The will be a scheduled ETL job with https://github.com/mozilla/docker-etl that should populate a list of reports from BQ and submit a POST request for classification for a batch of issues. The request will likely look as follows:

```json
[
   {
        "uuid": "954dbc23-91e6-4d6f-a10a-405f46663e31",
        "title": "https://www.threads.net/t/CuRtcYTNY3J",
        "body": "Loads blank page."
    },
   
]
```

PR #3846, opened 2023-11-23
Add support for classifying invalid compatibility reports to the http service

ksy36:
Fixes https://github.com/mozilla/bugbug/issues/3839

### DIFF
--- a/http_service/bugbug_http/app.py
+++ b/http_service/bugbug_http/app.py
@@ -30,6 +30,7 @@ from sentry_sdk.integrations.flask import FlaskIntegration
 from bugbug import bugzilla, get_bugbug_version, utils
 from bugbug_http.models import (
     MODELS_NAMES,
+    classify_broken_site_report,
     classify_bug,
     classify_issue,
     get_config_specific_groups,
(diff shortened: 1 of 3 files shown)


http_service/bugbug_http/app.py
```diff
@@ -720,6 +744,191 @@ def batch_prediction(model_name):
     return compress_response({"bugs": data}, status_code)
 
 
+@application.route("/<model_name>/predict/broken_site_report/batch", methods=["POST"])
+@cross_origin()
```
ksy36, line 748: I've named the endpoint `broken_site_report` to match the name of the component / telemetry ping https://github.com/mozilla/gecko-dev/blob/606cdd34860e42611fceba025314864cd293a43b/toolkit/components/reportbrokensite/pings.yaml#L8

http_service/bugbug_http/app.py
```diff
@@ -230,6 +231,29 @@ def create_bug_classification_jobs(
...
+            classify_broken_site_report, model_name, report["uuid"]
+        ).mapping_key
+        job_id_mapping[key] = job_id
+
+    redis_conn.mset(job_id_mapping)
+
+    return (
+        JobInfo(classify_broken_site_report, model_name, reports),
+        job_id,
+        BUGZILLA_JOB_TIMEOUT,
```
suhaibmujahid, line 253: nit: It looks a bit confusing to use `BUGZILLA_JOB_TIMEOUT` here since there is nothing related to Bugzilla in this job. We could drop it here, and then it will use the default timeout. Alternatively, we could use `JOB_TIMEOUT`, which is the same as the default timeout.

suhaibmujahid (dismissed):
LGTM! Thank you!

marco-c (approved):
LGTM! If we get too many, we might want to have a separate queue with a lower priority so that we don't impact the other API endpoints too much. Let's see how it looks after deployment.

ksy36:
> LGTM! If we get too many, we might want to have a separate queue with a lower priority so that we don't impact the other API endpoints too much. Let's see how it looks after deployment.

Sounds good. We've shipped the reporter in Nightly on desktop about a week ago and got about 150 reports so far (only 80 have description and would have needed classification). But I think most of these reports are from QA testing the feature :)

suhaibmujahid (approved):

ksy36:
Thanks!

marco-c:
@ksy36 when would you like to see this deployed?

ksy36:
> @ksy36 when would you like to see this deployed?

Would it be possible to deploy this some time next week? (I was thinking to start working on an ETL script that will use this endpoint then)

marco-c:
Sure! I will deploy a new version now, so we can root out any potential bugs now and retrigger a new release if necessary.

ksy36:
Perfect, thank you!

PR #3850, opened 2023-11-24
Fix the last redundant execution of string lowering in `bug_to_types()`

### LINKED ISSUE #3845: Optimize the `bug_to_types()` functions to avoid redundant the execution of string lowering
The `bug["whiteboard"].lower()` statement is performed multiple times in the function. Instead, we could store the lowercase value in a variable (e.g., `bug_whiteboard`) and reuse it where needed.

Examples:

https://github.com/mozilla/bugbug/blob/0586c6a49732f03781daf43d8726b53e2049e3db/bugbug/models/bugtype.py#L46

https://github.com/mozilla/bugbug/blob/0586c6a49732f03781daf43d8726b53e2049e3db/bugbug/models/bugtype.py#L50

sreeroop10:
Fixes #3845

### DIFF
--- a/bugbug/models/bugtype.py
+++ b/bugbug/models/bugtype.py
@@ -37,17 +37,19 @@ def bug_to_types(
 ) -> list[str]:
     types = set()
 
+    bug_whiteboard = bug["whiteboard"].lower()
+
     if any(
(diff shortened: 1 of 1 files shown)


suhaibmujahid (approved):
LGTM! Thank you, @sreeroop10!

PR #3854, opened 2023-11-25
Constrain pre-commit to use Python 10

### LINKED ISSUE #3853: The pre-commit linter picks the wrong Python version
Example: https://github.com/mozilla/bugbug/pull/3822#issuecomment-1820988443

suhaibmujahid:
Fixes #3853

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -93,4 +93,4 @@ repos:
     hooks:
       - id: check-useless-excludes
 default_language_version:
-  python: python3
+  python: python3.10


marco-c (approved):
The only downside is that we'll have to remember to update this whenever we bump the minimum version we support

suhaibmujahid:
> The only downside is that we'll have to remember to update this whenever we bump the minimum version we support

I agree. I wish that we could specify a minimum version instead of penning it on a specific version. Anyway, this will make it less tricky for new contributors

PR #3852, opened 2023-11-25
Refactor sum() calls to use boolean values directly

StarKBhaviN:
Fixes : #3821 This pull request enhances the logging statements by using the sum() function directly over boolean values. The sum() function efficiently counts the number of True values in a sequence of booleans, providing a more concise and readable approach than traditional methods.

- Also i found some logging statements that using f string formatting.

PR #3855, opened 2023-11-25
Update pre-commit repos and apply required changes

suhaibmujahid:
This should fix the lint errors in #3852

### DIFF
--- a/bugbug/repository.py
+++ b/bugbug/repository.py
@@ -1336,8 +1336,8 @@ def get_first_pushdate(repo_dir):
 
 def download_commits(
     repo_dir: str,
-    rev_start: str = None,
-    revs: list[bytes] = None,
+    rev_start: str | None = None,
+    revs: list[bytes] | None = None,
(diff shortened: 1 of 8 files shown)


bugbug/repository.py
```diff
@@ -1336,8 +1336,8 @@ def get_first_pushdate(repo_dir):
 
 def download_commits(
     repo_dir: str,
-    rev_start: str = None,
-    revs: list[bytes] = None,
+    rev_start: str | None = None,
+    revs: list[bytes] | None = None,
     branch: Optional[str] = "tip",
```
suhaibmujahid, line 1341: We should make it more consistent. We could use `... | None` instead of `Optional[...]`.

marco-c (approved):

bugbug/repository.py
suhaibmujahid: I filed an issue to follow up on this: #3856

PR #3866, opened 2023-11-27
Updated the appropriate % operator for logging statements

### LINKED ISSUE #3837: Some logging statements don't use the appropriate % operator
> This is a valid fix but unrelated to the goal of this PR. We could open a new issue and fix it in a separate PR (with similar cases if there are any)

_Originally posted by @suhaibmujahid in https://github.com/mozilla/bugbug/pull/3822#discussion_r1398670853_

wittyicon29:
Fixed #3837 

Updated the operator to %0.2f instead of %d for the logging statements.

### DIFF
--- a/bugbug/models/regressor.py
+++ b/bugbug/models/regressor.py
@@ -278,7 +278,7 @@ class RegressorModel(CommitModel):
         total_regressions = sum(is_reg for _, is_reg in results)
         average_regression_rate = total_regressions / total_landings
 
-        logger.info("Average risk is %d", average_regression_rate)
+        logger.info("Average risk is %0.2f", average_regression_rate)
 
         MIN_SAMPLE = 200
 


suhaibmujahid:
Thank you, @wittyicon29! Did you check if other cases need to be fixed?

wittyicon29:
I haven't actually, if it would be okay I'll check right away. 
For any changes made, should a new PR be created?

wittyicon29:
@suhaibmujahid I've checked for the other cases, but aren't any.

suhaibmujahid (approved):
> I haven't actually, I'll check right away. For any changes made, should a new PR be created?

Since it is within the scope of the same issue. There is no need for a new PR.


> I've checked for the other cases, but aren't any.

Thank you!

PR #3867, opened 2023-11-29
Fixed the issue #3866

ritin0204:
Fixed the issue #3866


suhaibmujahid:
Thank you for your interest!

> Fixed the issue https://github.com/mozilla/bugbug/pull/3866

What you are linking to is a pull request that fixs the issue, not the issue itself.

PR #3871, opened 2023-11-29
Delete everything related to TensorFlow

### LINKED ISSUE #3868: Drop the dependency on TensorFlow
The only usage of it is in `ComponentNNModel`, which we can drop as well since it is not used.

suhaibmujahid:
Fixes #3868


marco-c:
We can also remove https://github.com/mozilla/bugbug/blob/a59a1a74d2de21192ef0b801ec00dbbfa9445995/scripts/trainer.py#L14 and https://github.com/mozilla/bugbug/blob/a59a1a74d2de21192ef0b801ec00dbbfa9445995/scripts/trainer.py#L100-L105 and everything around that

suhaibmujahid:
Thank you! Done in 11a312620a3f68e5dbccd8eb1c079a029799bdcd.

marco-c (approved):

PR #3877, opened 2023-12-02
Refactor the pipelines to avoid fitting before splitting

### LINKED ISSUE #3818: Do not fit transformers on testing data
Currently, we fit the transformers before splitting the data, which could leak testing data into the training data, especially when using something like TF-IDF in the extraction pipeline:

Transforming: https://github.com/mozilla/bugbug/blob/02906a629108cd506d8f4d7158371a8fe94c6cb2/bugbug/model.py#L352
Splitting: https://github.com/mozilla/bugbug/blob/02906a629108cd506d8f4d7158371a8fe94c6cb2/bugbug/model.py#L368

suhaibmujahid:
Resolves #3818 

> **Note**
> The changes will be clearer if each commit is reviewed separately

- Converted `clf` to a pipeline and moved the union step to it (the step that requires fitting)
    -  Refactored the sampler to be part of the `clf` pipeline
       - The sampler requires `X` to be vectors, not dicts
       - The model train method became cleaner
- The `Files` feature is the only feature that requires `fit()`
    - Moved the part that requires fitting to the `clf` pipeline
- Replaced `fit_transform()` with `transform()`
- Adjusted the model saving process since the xgboost model now is part of a pipeline

Train on Taskcluster: annotateignore

### DIFF
--- a/bugbug/commit_features.py
+++ b/bugbug/commit_features.py
@@ -3,7 +3,6 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 import sys
-from collections import defaultdict
 from typing import Sequence
 
 import pandas as pd
@@ -808,34 +807,14 @@ class DirectoryTouchedPrev(object):
 
 
 class Files(object):
-    def __init__(self, min_freq=0.0014):
-        self.min_freq = min_freq
-
-    def fit(self, commits):
-        self.count = defaultdict(int)
-
-        self.total_commits = 0
-
-        for commit in commits:
-            self.total_commits += 1
-
-            for f in commit["files"]:
-                self.count[f] += 1
-
-        # We no longer need to store counts for files which have low frequency.
-        to_del = set(
-            f for f, c in self.count.items() if c / self.total_commits < self.min_freq
-        )
-
-        for f in to_del:
-            del self.count[f]
+    name = "files"
 
     def __call__(self, commit, **kwargs):
-        return [
-            f
-            for f in commit["files"]
-            if (self.count[f] / self.total_commits) > self.min_freq
-        ]
+        return commit["files"]
+
+
+def _pass_through_tokenizer(doc):
+    return doc
 
 
 class FileTouchedPrev(object):
@@ -1008,6 +987,7 @@ class CommitExtractor(BaseEstimator, TransformerMixin):
 
         for commit in commits():
             data = {}
+            result = {"data": data}
 
             for feature_extractor in self.feature_extractors:
                 if "bug_features" in feature_extractor.__module__:
@@ -1028,6 +1008,13 @@ class CommitExtractor(BaseEstimator, TransformerMixin):
                 else:
                     feature_extractor_name = feature_extractor.__class__.__name__
 
+                # FIXME: This is a workaround to pass the value to the
+                # union transformer independently. This will be dropped when we
+                # resolve https://github.com/mozilla/bugbug/issues/3876
+                if isinstance(feature_extractor, Files):
+                    result[sys.intern(feature_extractor_name)] = res
+                    continue
+
                 if isinstance(res, dict):
                     for key, value in res.items():
                         data[sys.intern(key)] = value
@@ -1040,7 +1027,6 @@ class CommitExtractor(BaseEstimator, TransformerMixin):
 
                 data[sys.intern(feature_extractor_name)] = res
 
-            result = {"data": data}
             if "desc" in commit:
                 for cleanup_function in self.cleanup_functions:
                     result["desc"] = cleanup_function(commit["desc"])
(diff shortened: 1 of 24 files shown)


marco-c:
> Replaced fit_transform() with transform()

Where is this done?

bugbug/commit_features.py
```diff
@@ -808,34 +809,47 @@ def __call__(self, commit, **kwargs):
...
+
+def _pass_through_tokenizer(doc):
+    return doc
+
+
+class FrequentFilesVectorizer(BaseEstimator, TransformerMixin):
     def __init__(self, min_freq=0.0014):
         self.min_freq = min_freq
+        self.vectorizer = CountVectorizer(
+            tokenizer=_pass_through_tokenizer, lowercase=False
```
marco-c, line 826: Nit: use a lambda so you don't have to define `_pass_through_tokenizer`

bugbug/commit_features.py
```diff
@@ -808,34 +809,47 @@ def __call__(self, commit, **kwargs):
...
+        )
 
-    def fit(self, commits):
-        self.count = defaultdict(int)
+    def fit(self, X, y=None):
+        min_freq = self.min_freq
```
marco-c, line 830: Nit: no need to store it in a temporary variable

bugbug/commit_features.py
```diff
@@ -808,34 +809,47 @@ def __call__(self, commit, **kwargs):
...
+            for filename in commit_files:
+                file_count[filename] += 1
 
-        for commit in commits:
-            self.total_commits += 1
+        frequent_files = {
+            filename
+            for filename, count in file_count.items()
+            if count / total_commits >= min_freq
+        }
```
marco-c, line 842: I wonder if we could use CountVectorizer's min_df to avoid doing this manually

bugbug/commit_features.py
```diff
@@ -1028,6 +1043,13 @@ def transform(self, commits):
                 else:
                     feature_extractor_name = feature_extractor.__class__.__name__
 
+                # FIXME: This is a workaround to store pass the value to the
```
marco-c, line 1046: `store pass`?

marco-c:
For confirmation, can you try running the annotateignore model before/after and see if the number of features is the same and if the metrics are the same?

suhaibmujahid:
> > Replaced fit_transform() with transform()
> 
> Where is this done?

Sorry, the commit was not pushed. It is done in 666fc465e5c3f6d90ef889e4366df8e2c2443490.



> For confirmation, can you try running the annotateignore model before/after and see if the number of features is the same and if the metrics are the same?

Here are the logs for both:
- [before - annotateignore.txt](https://github.com/mozilla/bugbug/files/13536527/before.-.annotateignore.txt)
- [after - annotateignore.txt](https://github.com/mozilla/bugbug/files/13536600/after.-.annotateignore.txt)

After the refactoring, the shape does not reflect the number of features because the vectorization happens as part of the clf pipeline. I added a log for the number of features in a1b8f450bd3b77dc28172e93830260202d976273.

The number of features has increased; it was 13572 before and is now 14465. This increase is expected since we fit on a subset of the data. For instance, during the fitting stage, the number of commits is smaller, which causes the threshold to consider a file as frequent to become lower. As a result, we ended up considering more files as features.

Regarding the rest of the metrics, they look almost the same.

bugbug/commit_features.py
suhaibmujahid: Type :( Fixed.

bugbug/commit_features.py
suhaibmujahid: I am surprised that I didn't think of it before, but it's actually possible. Done in b2a4a8e7bc3f5f9141c777edee8a3368a0317ad0. It produces the exact same number of features.

bugbug/commit_features.py
suhaibmujahid: Removed altogether in b2a4a8e7bc3f5f9141c777edee8a3368a0317ad0.

bugbug/commit_features.py
suhaibmujahid: Passing a local function (including lambda) cannot be pickled.

bugbug/models/annotate_ignore.py
```diff
@@ -71,7 +70,7 @@ def __init__(self, lemmatization: bool = False) -> None:
         )
 
         self.hyperparameter = {"n_jobs": utils.get_physical_cpu_count()}
-        self.clf = Pipeline(
+        self.clf = ImblearnPipeline(
```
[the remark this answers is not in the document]
suhaibmujahid: The `imblearn.pipeline.Pipeline` class inherits from `sklearn.pipeline.Pipeline`. The difference is that `ImblearnPipeline` supports `fit_resample`, so when fitting, if a transformer implements the `fit_resample` method, it will be used instead of `fit` or `fit_transform`.

PR #3910, opened 2023-12-09
Change `Optional[...]` to `... | None`

### LINKED ISSUE #3856: Use `... | None` instead of `Optional[...]`
> We should make it more consistent. We could use `... | None` instead of `Optional[...]`.

_Originally posted by @suhaibmujahid in https://github.com/mozilla/bugbug/pull/3855#discussion_r1405206846_

kaijif:
Fixing #3856 by changing all instances of Optional[...] to ... | None. @suhaibmujahid, let me know if this is what you had in mind.

### DIFF
--- a/bugbug/bugzilla.py
+++ b/bugbug/bugzilla.py
@@ -9,7 +9,7 @@ import math
 import re
 from datetime import datetime
 from logging import INFO, basicConfig, getLogger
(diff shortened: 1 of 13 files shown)


suhaibmujahid (approved):
Thank you, @kaijif! LGTM!

PR #3955, opened 2024-01-03
Set median fix time and fix time diff to empty when there are no fixes in a week

### LINKED ISSUE #3954: UnboundLocalError: local variable 'median_fix_time_text' referenced before assignment when generating quality report
```
  File "/usr/local/lib/python3.10/site-packages/scripts/generate_landings_risk_report.py", line 1673, in notification
    {median_fix_time_text}
UnboundLocalError: local variable 'median_fix_time_text' referenced before assignment
```

marco-c:
Fixes #3954

### DIFF
--- a/scripts/generate_landings_risk_report.py
+++ b/scripts/generate_landings_risk_report.py
@@ -1546,6 +1546,9 @@ table {
                     fix_time_diff = f"This is {verb} when compared to two weeks ago (median was {prev_median_fix_time} days)."
                 else:
                     fix_time_diff = ""
+            else:
+                median_fix_time_text = ""
+                fix_time_diff = ""
 
             top_intermittent_failures = "\n".join(
                 "{} failures ({}#{} globally{}){}".format(


scripts/generate_landings_risk_report.py
```diff
@@ -1546,6 +1546,9 @@ def get_top_crashes(team: str, channel: str) -> str | None:
                     fix_time_diff = f"This is {verb} when compared to two weeks ago (median was {prev_median_fix_time} days)."
                 else:
                     fix_time_diff = ""
+            else:
+                median_fix_time_text = ""
+                fix_time_diff = ""
```
suhaibmujahid, line 1551: nit: It could be more readable to define all the variables that have potential values to be `"` or `None` at the top; then, their values will be changed if they match any of the conditions. With that, we will have fewer need `else` statements.

suhaibmujahid (approved):

PR #3958, opened 2024-01-07
Use `LabelBinarizer()`  instead of  `LabelEncoder()` when encoding in `bugtype`  and `rcatype` models.

### LINKED ISSUE #3956: [model:bugtype] ValueError: y should be a 1d array, got an array of shape (198635, 5) instead.
> I'm wondering if the error encountered during the training of the `bugtype` model is something we should investigate.
>
> ```bash
> Traceback (most recent call last):
>   File "/home/promisefru/anaconda3/envs/bugbug/lib/python3.10/runpy.py", line 196, in _run_module_as_main
>     return _run_code(code, main_globals, None,
>   File "/home/promisefru/anaconda3/envs/bugbug/lib/python3.10/runpy.py", line 86, in _run_code
>     exec(code, run_globals)
>   File "/home/promisefru/mozilla/bugbug/scripts/trainer.py", line 145, in <module>
>     main()
>   File "/home/promisefru/mozilla/bugbug/scripts/trainer.py", line 141, in main
>     retriever.go(args)
>   File "/home/promisefru/mozilla/bugbug/scripts/trainer.py", line 41, in go
>     metrics = model_obj.train(limit=args.limit)
>   File "/home/promisefru/mozilla/bugbug/bugbug/model.py", line 377, in train
>     self.le.fit(y)
>   File "/home/promisefru/anaconda3/envs/bugbug/lib/python3.10/site-packages/sklearn/preprocessing/_label.py", line 98, in fit
>     y = column_or_1d(y, warn=True)
>   File "/home/promisefru/anaconda3/envs/bugbug/lib/python3.10/site-packages/sklearn/utils/validation.py", line 1156, in column_or_1d
>     raise ValueError(
> ValueError: y should be a 1d array, got an array of shape (198635, 5) instead.
> ```

_Originally posted by @PromiseFru in https://github.com/mozilla/bugbug/issues/3928#issuecomment-1875673580_

See also: https://github.com/mozilla/bugbug/pull/3823#pullrequestreview-1746981626

jpangas:
Issue:
Closes #3956 

Comments:
This PR introduces using `LabelBinarizer` instead of `LabelEncoder` to encode in multi-label cases. It fixes the issue which caused a Value Error to be raised when using `LabelEncoder` on non 1d arrays. 

Train with Taskcluster: bugtype

### DIFF
--- a/bugbug/model.py
+++ b/bugbug/model.py
@@ -23,7 +23,7 @@ from sklearn.feature_extraction.text import TfidfVectorizer
 from sklearn.metrics import precision_recall_fscore_support
 from sklearn.model_selection import cross_validate, train_test_split
 from sklearn.pipeline import Pipeline
-from sklearn.preprocessing import LabelBinarizer, LabelEncoder
+from sklearn.preprocessing import LabelEncoder
 from tabulate import tabulate
 from xgboost import XGBModel
 
@@ -372,14 +372,8 @@ class Model:
         # Extract features from the items.
         X = self.extraction_pipeline.transform(X_gen)
 
-        y = np.array(y)
-
-        is_multilabel = isinstance(y[0], np.ndarray)
-        is_binary = len(self.class_names) == 2
-
         # Calculate labels.
-        if is_multilabel:
-            self.le = LabelBinarizer()
+        y = np.array(y)
         self.le.fit(y)
 
         if limit:
@@ -387,6 +381,9 @@ class Model:
             y = y[:limit]
 
         logger.info(f"X: {X.shape}, y: {y.shape}")
+
+        is_multilabel = isinstance(y[0], np.ndarray)
+        is_binary = len(self.class_names) == 2
 
         # Split dataset in training and test.
         X_train, X_test, y_train, y_test = self.train_test_split(X, y)
--- a/bugbug/model.py
+++ b/bugbug/model.py
@@ -23,7 +23,7 @@ from sklearn.feature_extraction.text import TfidfVectorizer
 from sklearn.metrics import precision_recall_fscore_support
 from sklearn.model_selection import cross_validate, train_test_split
 from sklearn.pipeline import Pipeline
-from sklearn.preprocessing import LabelEncoder
+from sklearn.preprocessing import LabelBinarizer, LabelEncoder
 from tabulate import tabulate
 from xgboost import XGBModel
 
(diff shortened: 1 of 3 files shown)


jpangas:
Another bug to be looked at. 
```

File "/usr/local/lib/python3.10/site-packages/bugbug/model.py", line 546, in <genexpr>
    classified_num = sum(1 for v in y_pred_filter if v != "__NOT_CLASSIFIED__")

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

```

bugbug/model.py
```diff
@@ -372,8 +372,14 @@ def train(self, importance_cutoff=0.15, limit=None):
...
 
-        # Calculate labels.
         y = np.array(y)
+
+        is_multilabel = isinstance(y[0], np.ndarray)
+        is_binary = len(self.class_names) == 2
+
+        # Calculate labels.
+        if is_multilabel:
+            self.le = LabelBinarizer()
```
suhaibmujahid, line 382: Why do we need to change the encoder here instead of using it in the model itself in the first place?

jpangas: I thought it would be wiser to fix the error for other models too expect `bugtype` that are multi-label.

suhaibmujahid: I think, `bugtype` is the only one with multi-label. Changing the encoder in the training method can be tricky and make the codebase more complex since it is initially defined in the contractor.

jpangas: Oh, Yes. I will revert and change it for the model itself.  Thank you.

suhaibmujahid: One alternative solution would be to define `LabelEncoder` in the training method instead of in the constructor, which will make it less confusing.

bugbug/model.py
```diff
@@ -372,8 +370,15 @@ def train(self, importance_cutoff=0.15, limit=None):
...
-        # Calculate labels.
         y = np.array(y)
+        is_multilabel = isinstance(y[0], np.ndarray)
+        is_binary = len(self.class_names) == 2
+
+        # Calculate labels.
+        if is_multilabel:
+            self.le = LabelBinarizer()
+        else:
+            self.le = LabelEncoder()
```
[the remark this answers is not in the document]
jpangas: I thought so too earlier in the morning. Let us go with your earlier suggestion. Thank you.

suhaibmujahid (commented):
It seems that the `rcatype` model has the same problem since it is multi-label as well. Could you please check that?

jpangas:
Yes. I can confirm `rcatype` is also multi-label. I will work on that before EOD.

jpangas:
I have temporarily included the patch from #3963 .

jpangas:
The training task runs to completion, so I will revert the change.

suhaibmujahid (approved):
Thank you!

PR #3963, opened 2024-01-09
Calculate `classified_num`  of test set samples differently for  multi-label models

### LINKED ISSUE #3962: [model:bugtype] ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
Another bug to be looked at. 
```

File "/usr/local/lib/python3.10/site-packages/bugbug/model.py", line 546, in <genexpr>
    classified_num = sum(1 for v in y_pred_filter if v != "__NOT_CLASSIFIED__")

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

```

_Originally posted by @jpangas in https://github.com/mozilla/bugbug/issues/3958#issuecomment-1880820955_

jpangas:
**Issue**:
Closes #3962

**Comments**:
This PR fixes a bug that was checking for number of classified samples in multi-label cases the same way as binary label cases. It uses the length of the list of the classified indices to know the number of classified samples from the test set.

Train on Taskcluster: bugtype

### DIFF
--- a/bugbug/model.py
+++ b/bugbug/model.py
@@ -540,7 +540,12 @@ class Model:
                     np.array(y_pred_filter[classified_indices], dtype=int)
                 )
 
-            classified_num = sum(1 for v in y_pred_filter if v != "__NOT_CLASSIFIED__")
+            if is_multilabel:
+                classified_num = len(classified_indices)
+            else:
+                classified_num = sum(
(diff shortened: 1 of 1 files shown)


jpangas:
I would suggest merging this before #3958

suhaibmujahid (changes requested):
Could you please rebase the PR?

suhaibmujahid (approved):

PR #4030, opened 2024-02-08
Use SSL for Redis only when URL is rediss

marco-c:


http_service/bugbug_http/app.py
```diff
@@ -70,7 +70,7 @@
     host=url.hostname,
     port=url.port if url.port is not None else 6379,
     password=url.password,
-    ssl=True,
+    ssl=True if url.scheme == "rediss" else False,
```
suhaibmujahid, line 73:
```suggestion
    ssl=url.scheme == "rediss",
```

PR #4058, opened 2024-02-23
Refactor count_bugs to use libmozdata

YBCS:
Closes https://github.com/mozilla/bugbug/issues/3676
use libmozdata instead of calling api directly

### DIFF
--- a/bugbug/bugzilla.py
+++ b/bugbug/bugzilla.py
@@ -290,14 +290,14 @@ def delete_bugs(match):
 def count_bugs(bug_query_params):
     bug_query_params["count_only"] = 1
 
-    counts = [0]
+    data = {}
 
     def handler(bug):
-        counts[0] = bug["bug_count"]
+        data["bug_count"] = bug["bug_count"]
 
     Bugzilla(queries=Query(Bugzilla.API_URL, bug_query_params, handler)).wait()
 
-    return counts[0]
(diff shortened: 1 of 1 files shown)


bugbug/bugzilla.py
```diff
@@ -290,13 +290,14 @@ def delete_bugs(match):
...
 
-    r = utils.get_session("bugzilla").get(
-        "https://bugzilla.mozilla.org/rest/bug", params=bug_query_params
-    )
-    r.raise_for_status()
-    count = r.json()["bug_count"]
+    counts = [0]
+
+    def handler(bug):
+        counts[0] = bug["bug_count"]
```
suhaibmujahid, line 296: Using a list is a nice workaround. However, for the sake of consistency, I would suggest using a dictionary instead!

```suggestion
    data = {}

    def handler(bug):
        data["bug_count"] = bug["bug_count"]
```

suhaibmujahid (commented):
Thank you, @YBCS! LGTM! I just have a minor comment.

Could you please test the funcation locally?

YBCS:
image
Hi @suhaibmujahid  I have tested it locally 
PTAL

suhaibmujahid (approved):
Thank you!

suhaibmujahid (approved):

PR #4084, opened 2024-03-05
Don't show non-weighted burndown time in the quality report

### LINKED ISSUE #4079: QualityReport/MEI: Hide non-weighted BurndownTime
The non weighted burndown time is just creating confusion, as it does not match exactly to the (weighted) MEI. It might be fine to look at it in other contextes, but it's not really needed in the email.

marco-c:
Fixes #4079


suhaibmujahid (approved):
Thank you, @marco-c!

Issue #4095, opened 2024-03-12
worksforme model training task is failing (probably OOM)


marco-c:
https://community-tc.services.mozilla.com/tasks/TOyLl-ubQwehM8liNwtRdA/runs/0

PromiseFru:
> https://community-tc.services.mozilla.com/tasks/TOyLl-ubQwehM8liNwtRdA/runs/0

Oh, the training was successful on [TC](https://github.com/mozilla/bugbug/runs/22254262828). Perhaps I should try the `compute-large` worker?

marco-c:
Yeah, let's try that!

PR #4096, opened 2024-03-13
Update `worksforme` data-pipeline workerType to 'compute-large'

PromiseFru:
Resolves: #4095


suhaibmujahid (approved):
Thank you, @PromiseFru!

PR #4159, opened 2024-05-02
Dataset creation for backout commits

benjaminmah:
Script to generate dataset of bug-inducing commits, backout commits, and the subsequent fix commit. 

Intended to include:
- The hashes of the three commits.
- Metadata of each commit (`pushdate`, `desc`).
- The diff between the initial commit and the fix commit.


dataset/cleaned_backout_data.json
suhaibmujahid, line 1: Including generated data in the repo will grow its size to a point that makes it hard to maintain. Instead, we keep the data on the Taskcluster and download it when needed.

Example: https://github.com/mozilla/bugbug/blob/9aea2ba6e6a3664f144dd5388e926a4f0e89ff6f/bugbug/bugzilla.py#L27-L32

scripts/backout_data_collection.py
```diff
@@ -0,0 +1,219 @@
...
+
+    Args:
+        None
+
+    Returns:
+        None
+    """
+
+    logger.info("Downloading commits database...")
+    assert db.download(repository.COMMITS_DB, support_files_too=True)
```
suhaibmujahid, line 25: Creating a wrapping function does not add much value.

scripts/backout_data_collection.py
```diff
@@ -0,0 +1,219 @@
...
+
+    Args:
+        bug_id (int): the ID of the bug from Bugzilla
+
+    Returns:
+        bool: True if the bug is RESOLVED, False otherwise
+    """
+
+    # get bug details
+    bug_details = bugzilla.get([bug_id])
```
suhaibmujahid, line 53: The `bugzilla.get()` method is a bit expensive since it performs an API request on each call. We could utilize the bugs db for that.

scripts/backout_data_collection.py
```diff
@@ -0,0 +1,219 @@
...
+        None
+    """
+
+    with open(source_filepath, 'r') as file:
+        data = json.load(file)
+    
+    filtered_data = []
+
+    # fields to keep, rest will be discarded
+    fields_to_keep = ['node', 'bug_id', 'desc', 'pushdate', 'backsout']
```
suhaibmujahid, line 189: The goal at the first stage is to have the following information for each case:
- The defect inducing commit
- The backout commit
- The fix commit

scripts/backout_data_collection.py
```diff
@@ -0,0 +1,219 @@
...
+        json.dump(filtered_data, file, indent=4)
+
+    logger.info(f"Filtered data has been saved to {destination_filepath}")
+
+    return
+
+if __name__ == "__main__":
+    # download commits and bugs
+    download_commits()
+    download_bugs()
```
suhaibmujahid, line 205: In general, code comments should focus on explaining "why" rather than "what" the code does. Also, comments should be avoided when the code is self-explanatory. 

Leaving comments is not free. When you go through the code, you need to read both the comments and the code. If both convey the same message, then it is a waste of time. If you make any changes to the code, you may need to modify the comments as well, which would add an extra maintenance cost. If you forget to update the comment and it doesn't match with the code, then the code will become confusing and harder to maintain.

suhaibmujahid (commented):
Thank you, @benjaminmah! Please see my comments. Also, please fix the linting errors (you may want to consider installing `pre-commit`[^1]).


[^1]: https://github.com/mozilla/bugbug#auto-formatting

scripts/backout_data_collection.py
```diff
@@ -0,0 +1,169 @@
...
+        if commit["node"] == inducing_node:
+            inducing_commit_found = True
+
+    return commit
+
+
+def main():
+    download_databases()
+
+    commit_dict, bug_to_commit_dict, bug_dict = preprocess_commits_and_bugs()
```
suhaibmujahid, line 156: We may want to consider the space complexity when iterating over the whole dataset.

scripts/backout_data_collection.py
```diff
@@ -0,0 +1,169 @@
...
+
+    pbar = tqdm(total=count_limit, desc="Filtering commits")
+
+    for commit in repository.get_commits(
+        include_no_bug=False, include_backouts=True, include_ignored=False
+    ):
+        # add commit if it was backed out and the bug is fixed
+        if len(commit["backedoutby"]) > 0 and is_bug_fixed(
+            bug_id=commit["bug_id"], bug_dict=bug_dict
+        ):
```
suhaibmujahid, line 84: We could try simplifying this:
```suggestion
        bug = bug_dict[commit["bug_id"]]
        if commit["backedoutby"] and bug["resolution"] == "FIXED":
```

dataset/backout_dataset.json
suhaibmujahid, line 1: I would avoid committing data to the repository where possible.

scripts/backout_data_collection.py
benjaminmah: Removed unused keys when constructing the dictionaries and implemented a cache to use generated dictionaries from previous instances of running the code via saving them as JSON files. Let me know if this needs additional changes/fixes!

scripts/backout_data_collection.py
```diff
@@ -0,0 +1,198 @@
...
+    pbar = tqdm(total=count_limit, desc="Filtering commits")
+
+    for commit in repository.get_commits(
+        include_no_bug=False, include_backouts=True, include_ignored=False
+    ):
+        # add commit if it was backed out and the bug is fixed
+        if (
+            bug_dict.get(str(commit["bug_id"]))
+            and commit["backedoutby"]
+            and bug_dict[str(commit["bug_id"])] == "FIXED"
```
suhaibmujahid, line 112: storing the bug in a variable here will prevent retrieving it from the dict multiple times.

scripts/backout_data_collection.py
```diff
@@ -0,0 +1,198 @@
...
+                break
+
+    json_data = json.dumps(filtered_list, indent=4)
+
+    if not os.path.exists(directory_path):
+        os.makedirs(directory_path)
+        print(f"Directory {directory_path} created")
+
+    with open(destination_filepath, "w") as file:
+        file.write(json_data)
```
suhaibmujahid, line 161: For the function called `filter_commits`, thus, I would avoid storing the data to the desc within the same function, which is an unexpected side effect.

scripts/backout_data_collection.py
```diff
@@ -0,0 +1,172 @@
...
+        else:
+            bug_to_commit_dict[commit["bug_id"]].append(commit_dict[commit["node"]])
+
+    logger.info("Preprocessing bugs")
+    bug_dict = {}
+
+    num_lines = sum(1 for line in open(bugzilla.BUGS_DB, "r"))
+
+    # store bugs with their bug IDs as keys
+    with open(bugzilla.BUGS_DB, "r") as f:
```
suhaibmujahid, line 53: We do not need to read the db file directly, instead we use its class. In this case we would use `bugzilla.get_bugs()`.

scripts/backout_data_collection.py
```diff
@@ -0,0 +1,172 @@
...
+
+        if commit_dict[commit["node"]]["bug_id"] not in bug_to_commit_dict:
+            bug_to_commit_dict[commit["bug_id"]] = [commit_dict[commit["node"]]]
+        else:
+            bug_to_commit_dict[commit["bug_id"]].append(commit_dict[commit["node"]])
+
+    logger.info("Preprocessing bugs")
+    bug_dict = {}
+
+    num_lines = sum(1 for line in open(bugzilla.BUGS_DB, "r"))
```
suhaibmujahid, line 50: I don't think it is worth iterating over the whole dataset to count the bugs. We could use `tqdm` without the total parameter.

scripts/backout_data_collection.py
```diff
@@ -0,0 +1,172 @@
...
+
+def filter_commits(
+    commit_limit: int,
+    commit_dict: dict,
+    bug_to_commit_dict: dict,
+    bug_dict: dict,
+):
+    counter = 0
+    commit_limit = min(commit_limit, 709458)
+    pbar = tqdm(total=commit_limit, desc="Filtering commits")
```
suhaibmujahid, line 69: I would restrict using `tqdm` in the main method where possible.

scripts/backout_data_collection.py
```diff
@@ -0,0 +1,172 @@
...
+        counter += 1
+        pbar.update(1)
+        if commit["backedoutby"] and bug_info == "FIXED":
+            fixing_commit = find_next_commit(
+                commit["bug_id"], bug_to_commit_dict, commit["node"]
+            )
+
+            # if fixing commit could not be found or is another backing out commit, do not add it to dataset
+            if (
+                fixing_commit["node"] == commit["backedoutby"]
```
suhaibmujahid, line 86: The fixing commit should not be the backing out commit in the first place.

scripts/backout_data_collection.py
```diff
@@ -0,0 +1,172 @@
...
+        bug_info = bug_dict.get(commit["bug_id"])
+
+        counter += 1
+        pbar.update(1)
+        if commit["backedoutby"] and bug_info == "FIXED":
+            fixing_commit = find_next_commit(
+                commit["bug_id"], bug_to_commit_dict, commit["node"]
+            )
+
+            # if fixing commit could not be found or is another backing out commit, do not add it to dataset
```
suhaibmujahid, line 84: I would log cases where we cannot find the fix commit.

bugbug/repository.py
```diff
@@ -1543,6 +1543,57 @@ def trigger_pull() -> None:
...
+        elif line.startswith("+") and not line.startswith("+++"):
+            parsed_diff["added_lines"].append(
+                {"line_content": line[1:], "file": current_file}
+            )
+        elif line.startswith("-") and not line.startswith("---"):
+            parsed_diff["removed_lines"].append(
+                {"line_content": line[1:], "file": current_file}
+            )
+
+    return parsed_diff
```
suhaibmujahid, line 1563: You could try using `rs_parsepatch` instead of parsing the patch manually.

scripts/backout_data_collection.py
```diff
@@ -0,0 +1,222 @@
+import json
+import logging
+import os
+from datetime import datetime, timedelta
+from typing import Any, Dict, Generator, Tuple
```
suhaibmujahid, line 5: See https://peps.python.org/pep-0585/

benjaminmah:
Example diffs extracted:

[Backout Data Collection Validation](https://docs.google.com/document/d/1akKmKFdGZgmyf-1h7nwx9EFGyNjPqljXAsrlLC--q74/edit?usp=sharing)

bugbug/repository.py
```diff
@@ -20,7 +20,14 @@
 import threading
 from datetime import datetime
 from functools import lru_cache
-from typing import Collection, Iterable, Iterator, NewType, Set, Union
+from typing import (
+    Collection,
+    Iterable,
+    Iterator,
+    NewType,
+    Set,
```
suhaibmujahid, line 28: In Python 3.9 and higher, we could use the standard collections as generic types instead of the capitalized ones (e.g., `set` instead of `Set`).[^1]

[^1]: https://docs.python.org/3/whatsnew/3.9.html#type-hinting-generics-in-standard-collections

bugbug/repository.py
```diff
@@ -1543,6 +1550,69 @@ def trigger_pull() -> None:
...
 
+def get_diff(repo_path, original_hash, fix_hash) -> bytes:
+    client = hglib.open(repo_path)
+
+    current_rev = client.identify(id=True)
+
+    try:
+        client.rawcommand([b"shelve"])
+    except Exception:
+        pass
```
suhaibmujahid, line 1561: We should try to be more specific with the exception type. This could hide problems and make it harder to debug. I suggest specifying the exact error that we are OK with without crashing the script. Also, it may be helpful to log such cases.

bugbug/repository.py
```diff
@@ -1543,6 +1550,69 @@ def trigger_pull() -> None:
...
+    args = hglib.util.cmdbuilder(
+        str.encode("graft"), r=revs, no_commit=no_commit, f=force, tool=tool
+    )
+
+    eh = hglib.util.reterrorhandler(args)
+
+    try:
+        client.rawcommand(args, eh=eh, prompt=auto_resolve_conflict_prompt)
+    except Exception:
+        return False
```
suhaibmujahid, line 1605: Similar to https://github.com/mozilla/bugbug/pull/4159#discussion_r1624442585

scripts/backout_data_collection.py
```diff
@@ -0,0 +1,226 @@
...
+    logger.info("Generating datapoints...")
+
+    for commit in tqdm(
+        repository.get_commits(
+            include_no_bug=True, include_backouts=True, include_ignored=True
+        )
+    ):
+        counter += 1
+
+        bug_info = bug_dict.get(commit["bug_id"])
```
suhaibmujahid, line 75: The names here are a bit confusing. The `bug_inff` variable has only the resolution, so it could be more suitable to call it something like `bug_resolution`. Also, `bug_dict` is very general; I would use something more specific like `bug_resolution_map`.

bugbug/repository.py
benjaminmah: I think `Set` was in the existing code before I made any changes, however I will be sure to change any capitalized standard collections to the generic types!

bugbug/repository.py
benjaminmah: Done in ae28dcffba2c5594a1f8a184c6688c2d8d35f41b!

scripts/backout_data_collection.py
benjaminmah: Done in 37c51b640aa4b5f38ab88e1fa1e4b4bd8fc525c3!

bugbug/repository.py
suhaibmujahid: I'm not sure if logging it as `warning` would be better than `info`. It depends on the expectations. Why could it fail, and is it normal?

bugbug/repository.py
suhaibmujahid: Why is that?

benjaminmah: I'm no longer parsing the patch manually, I'm simply getting the diff between the two commits and storing the raw string so the comment is outdated

bugbug/repository.py
benjaminmah: When I tested the script, I never encountered any error grafting. However, to prevent the script from stopping due to a single grafting failure, I added the `hglib.error.CommandError` exception handler to continue to the next datapoint. Due to it not being a common occurrence, I think logging it as a `warning` is more appropriate.

suhaibmujahid: I would avoid using try-exception if I do not know what type of errors I could get and why. It is better to fail loudly, which allows you to fix the problem. On the other side, failing silently will make the problems harder to notice.

bugbug/repository.py
```diff
@@ -18,9 +18,10 @@
 import subprocess
 import sys
 import threading
+from collections.abc import Collection, Iterable, Iterator
 from datetime import datetime
 from functools import lru_cache
-from typing import Collection, Iterable, Iterator, NewType, Set, Union
+from typing import NewType, Union
```
suhaibmujahid, line 24: This is out of the scope of this PR.

bugbug/repository.py
```diff
@@ -196,7 +197,7 @@ def __init__(
         self.source_code_deleted = 0
         self.other_deleted = 0
         self.test_deleted = 0
-        self.types: Set[str] = set()
+        self.types: set[str] = set()
```
suhaibmujahid, line 200: This is out of the scope of this PR.

bugbug/repository.py
```diff
@@ -1543,6 +1544,74 @@ def trigger_pull() -> None:
...
+    current_rev = client.identify(id=True)
+
+    try:
+        client.rawcommand([b"shelve"])
+    except hglib.error.CommandError as e:
+        if b"nothing changed" in e.out:
+            logger.info(f"Nothing to shelve: {e}")
+        else:
+            logger.info(f"Error while shelving: {e}")
+            raise
```
suhaibmujahid, line 1559: I don't think logging is the right action here. You might want to consider using raise-from.

benjaminmah: Done in 66108ad83f3e2579af14ee983758cc3b5ff7fcf9

bugbug/repository.py
benjaminmah: Makes sense, I've removed the try-except block as we typically should not encounter any error while grafting. Done in 0d83fa73a8732189cab7cad7bdbaf33df0b101e7!

PR #4197, opened 2024-05-25
Add and update MEI_time_series_visualization script

### LINKED ISSUE #3717: MEI: Have a visualization of the time series of maintenance effectiveness values, aggregated by different teams/components
Once we have a time series, we could construct a nice graph visualizing the trend for some of of those values.

The exact content would need to be defined, but it should be possible to aggregate/filter using the predefined segmentation from the stored data.

A base for the graph could just be the number of open defects, over that an evolution of the years needed to burn down the defect backlog and/or the percentage of maintenance effectiveness.

gothwalritu:
Fix #3717 MEI: Have a visualization of the time series of maintenance effectiveness values, aggregated by different teams/components
@marco-c 

Done by:

Created a new script bugbug/scripts/MEI_time_series_visualization.py to generate MEI time series visualization. Script designed to be executed from bugbug folder.

### DIFF
--- a/scripts/MEI_time_series_visualization.py
+++ b/scripts/MEI_time_series_visualization.py
@@ -1,0 +1,185 @@
+import datetime
+import os
+import sys
+from datetime import timedelta
+from typing import Dict, List
+
+import matplotlib.pyplot as plt
+import pandas as pd
+
+
+def setup_bugzilla_import():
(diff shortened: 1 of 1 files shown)


marco-c:
Sorry @gothwalritu, I had completely missed this :(

@jensstutte do you want to try this?

PR #4208, opened 2024-05-30
Add missing langchain dependencies

suhaibmujahid:
We started using these dependencies in https://github.com/mozilla/bugbug/pull/4133

### DIFF
--- a/bugbug/generative_model_tool.py
+++ b/bugbug/generative_model_tool.py
@@ -6,7 +6,7 @@
 from abc import ABC, abstractmethod
 from typing import Any
 
-from langchain.llms.human import HumanInputLLM
+from langchain_community.llms import HumanInputLLM
 from langchain_openai import ChatOpenAI
 
 from bugbug.utils import get_secret
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,6 +2,9 @@ amqp==5.2.0
 beautifulsoup4==4.12.3
 boto3==1.34.108
 imbalanced-learn==0.12.2
+langchain==0.2.1
+langchain-community==0.2.1
+langchain-openai==0.1.8
 libmozdata==0.2.4
 lmdb==1.4.1
 markdown2==2.4.13


bugbug/generative_model_tool.py
```diff
@@ -6,7 +6,7 @@
 from abc import ABC, abstractmethod
 from typing import Any
 
-from langchain.llms.human import HumanInputLLM
+from langchain_community.llms import HumanInputLLM
```
benjaminmah, line 9: According to the LangChain docs, `HumanInputLLM` may need to be imported from `langchain_community.llms.human`.

See: [LangChain HumanInputLLM](https://api.python.langchain.com/en/latest/llms/langchain_community.llms.human.HumanInputLLM.html#langchain_community.llms.human.HumanInputLLM)
```suggestion
from langchain_community.llms.human import HumanInputLLM
```

benjaminmah (commented):
Thank you @suhaibmujahid! I've only suggested one change but the rest LGTM!

bugbug/generative_model_tool.py
suhaibmujahid: I notice the need for the update through the following message:
```
/lib/python3.10/site-packages/langchain/_api/module_import.py:92: LangChainDeprecationWarning: Importing HumanInputLLM from langchain.llms is deprecated. Please replace deprecated imports:

>> from langchain.llms import HumanInputLLM

with new imports of:

>> from langchain_community.llms import HumanInputLLM
You can use the langchain cli to **automatically** upgrade many imports. Please see documentation here https://python.langchain.com/v0.2/docs/versions/v0_2/ 
```

And I applied the change using `langchain-cli `:
```
langchain-cli migrate  bugbug/generative_model_tool.py
```


It seems that the class is re-exported from `langchain_community.llms`: https://github.com/langchain-ai/langchain/blob/ee32369265103ac64b412a73a758ae0d1649cafe/libs/community/langchain_community/llms/__init__.py#L281-L284 

Here is an example usage: https://github.com/langchain-ai/langchain/blob/dcec133b85381a9e40ad8afa2c6f8226b8db1e3b/libs/langchain/langchain/llms/human.py#L6

benjaminmah (approved):
LGTM

PR #4209, opened 2024-05-30
Support Vector DB

### LINKED ISSUE #4106: Add support for vector databases
As a start we want to support Qdrant.

suhaibmujahid:
Resolves #4106

Documentations will be added soon in later PRs.


bugbug/vectordb.py
```diff
@@ -0,0 +1,125 @@
...
+    def __init__(
+        self, location: str, api_key: str, collection_name: str, *args, **kwargs
+    ):
+        super().__init__(*args, **kwargs)
+        self.collection_name = collection_name
+        self.client = QdrantClient(location=location, api_key=api_key)
+
+    def setup(self):
+        config = VectorParams(
+            size=3072,
```
marco-c, line 70: What is `size`? Should it be configurable through a param?

bugbug/vectordb.py
```diff
@@ -0,0 +1,125 @@
...
+        return self.client.search(self.collection_name, query)
+
+
+class ReviewCommentsDB:
+    NAV_PATTERN = re.compile(r"\{nav, [^}]+\}")
+    WHITESPACE_PATTERN = re.compile(r"[\n\s]+")
+
+    def __init__(self, vector_db: VectorDB) -> None:
+        self.vector_db = vector_db
+        self.embeddings = OpenAIEmbeddings(model="text-embedding-3-large")
```
marco-c, line 103: We could make this configurable

bugbug/vectordb.py
```diff
@@ -0,0 +1,125 @@
...
+                )
+                for point in points
+            ],
+        )
+
+    def search(self, query: list[float]):
+        return self.client.search(self.collection_name, query)
+
+
+class ReviewCommentsDB:
```
marco-c, line 97: Maybe better to move this to a separate file?

bugbug/vectordb.py
```diff
@@ -0,0 +1,125 @@
...
+    def clean_comment(self, comment):
+        # TODO: use the nav info instead of removing it
+        comment = self.NAV_PATTERN.sub("", comment)
+        comment = self.WHITESPACE_PATTERN.sub(" ", comment)
+        comment = comment.strip()
+
+        return comment
+
+    def add_comments_by_hunk(self, items: Iterable[tuple[Hunk, InlineComment]]):
+        for batch in batched(items, 100):
```
marco-c, line 114: Why 100? Let's define it in a constant at the top of the file or in the class

scripts/review_comments_retriver.py
```diff
@@ -0,0 +1,205 @@
+# -*- coding: utf-8 -*-
```
marco-c, line 1: Typo in the name of the file :)

scripts/review_comments_retriver.py
```diff
@@ -0,0 +1,205 @@
...
+        wait=tenacity.wait_exponential(multiplier=1, min=16, max=64),
+        reraise=True,
+    )
+    def load_patch_set(self, diff_id):
+        try:
+            with open(f"patches/{diff_id}.patch", "r") as f:
+                patch = f.read()
+        except FileNotFoundError:
+            with open(f"patches/{diff_id}.patch", "w") as f:
+                patch = self.phabricator_api.load_raw_diff(diff_id)
```
marco-c, line 47: We could do like https://github.com/mozilla/bugbug/blob/7722d95d124e5bf01bb9b7eb65825773ca29b9df/scripts/code_review_tool_runner.py#L21 to make it more generic

scripts/review_comments_retriver.py
```diff
@@ -0,0 +1,205 @@
...
+                f.write(patch)
+
+        return PatchSet.from_string(patch)
+
+    def _get_revisions_count(self):
+        return sum(1 for _ in phabricator.get_revisions())
+
+    def retrieve_diff_comments(self):
```
marco-c, line 55: What if we move this in the ReviewData class, so we can make it generic here?

scripts/review_comments_retriver.py
```diff
@@ -0,0 +1,205 @@
...
+                    continue
+
+                # Ignore bot comments
+                if transaction["authorPHID"] == "PHID-USER-cje4weq32o3xyuegalpj":
+                    continue
+
+                if len(transaction["comments"]) != 1:
+                    print(transaction)
+                    # raise Exception("Unexpected, need to look into it")
+                    continue
```
marco-c, line 76: Could you file a follow-up for this?

scripts/review_comments_retriver.py
```diff
@@ -0,0 +1,205 @@
...
+
+                diff_id = transaction["fields"]["diff"]["id"]
+                comment_id = transaction["comments"][0]["id"]
+                date_created = transaction["comments"][0]["dateCreated"]
+                filename = transaction["fields"]["path"]
+                start_line = transaction["fields"]["line"]
+                end_line = (
+                    transaction["fields"]["line"] + transaction["fields"]["length"] - 1
+                )
+                # Unfortantliy, we do not have this information for a limitation
```
marco-c, line 116: `Unfortunately`

bugbug/vectordb.py
suhaibmujahid: > What is `size`? 

It is the size of the vector. So, it is 3072 here to match the size of the vector generated by the `text-embedding-3-large` model.

> Should it be configurable through a param?

At this early stage, I prefer to delay exposing configurations until they are needed. This will give us the flexibility to move things around without breaking the API.

bugbug/vectordb.py
suhaibmujahid: Same as https://github.com/mozilla/bugbug/pull/4209#discussion_r1621635828

scripts/review_comments_retriver.py
suhaibmujahid: Fixed in d227893d7d02c4b9f1d0c1728741190e99b763d8.

scripts/review_comments_retriver.py
suhaibmujahid: Fixed in d227893d7d02c4b9f1d0c1728741190e99b763d8. Spell check/auto-correct can sometimes make mistakes even worse :)

bugbug/vectordb.py
suhaibmujahid: > Why 100? 

It is an arbitrary number. I just learned that we could use `upload_points()` instead of `upsert()`, which will take care of batching, retrying, and parallelization. The default batch size in `upload_points()` is 64.

I moved to use `upload_points()` in f717028f8cae24f47e34ef49d6499d541fb4fd23.

> Let's define it in a constant at the top of the file or in the class
 
Same as https://github.com/mozilla/bugbug/pull/4209#discussion_r1621635828

bugbug/vectordb.py
suhaibmujahid: Done in 0acd3701754e9b6e7dc5f16764caa55b682bfed4.

scripts/review_comments_retriver.py
suhaibmujahid: Done in 058666bd56310b95b8d4064b685ffe5c91fb48c7.

scripts/review_comments_retriver.py
suhaibmujahid: Done as part of moving generic methods into `ReviewData`: 058666bd56310b95b8d4064b685ffe5c91fb48c7.

scripts/review_comments_retriver.py
suhaibmujahid: I filed https://github.com/mozilla/bugbug/issues/4218 to follow-up on this.

bugbug/tools/code_review.py
```diff
@@ -117,9 +138,145 @@ def get_review_request_by_id(self, review_id: int) -> ReviewRequest:
...
+
+                    for line in hunk:
+                        if (
+                            line.is_removed
+                            and line.source_line_no == comment.start_line
+                        ):
+                            return hunk
+
+            if len(matching_hunks) != 0:
+                return matching_hunks[0]
```
marco-c, line 204: Should we add a warning here, in case we find a better way to choose a matching hunk in the future?

bugbug/tools/code_review.py
```diff
@@ -117,9 +138,145 @@ def get_review_request_by_id(self, review_id: int) -> ReviewRequest:
...
+        else:
+            for hunk in patched_file:
+                if hunk.target_start <= comment.start_line < target_end(hunk):
+                    return hunk
+
+    def retrieve_comments_with_hunks(self):
+        def comment_filter(comment: InlineComment):
+            comment_content = comment.content
+
+            # Ignore very short and very log comments
```
marco-c, line 220: `long`

bugbug/tools/code_review.py
```diff
@@ -129,26 +286,87 @@ def get_review_request_by_id(self, revision_id: int) -> ReviewRequest:
...
+    def _get_revisions_count(self):
+        return sum(1 for _ in phabricator.get_revisions())
+
+    def get_all_inline_comments(
+        self, comment_filter
+    ) -> Iterable[tuple[int, list[InlineComment]]]:
+        db.download(phabricator.REVISIONS_DB)
+
+        for revision in tqdm(
+            phabricator.get_revisions(), total=self._get_revisions_count()
```
marco-c, line 308: Given it's only used here and internal, you could just inline this function or have a variable:
```Python
        revision_count = sum(1 for _ in phabricator.get_revisions())
        for revision in tqdm(phabricator.get_revisions(), total=revision_count)
```
or
```Python
        for revision in tqdm(phabricator.get_revisions(), total=sum(1 for _ in phabricator.get_revisions()))
```

Given iterating through revisions is expensive, you could also just get rid of `total` and use an indeterminate progress bar:
```Python
        for revision in tqdm(phabricator.get_revisions())
```

bugbug/tools/code_review.py
```diff
@@ -129,26 +286,87 @@ def get_review_request_by_id(self, revision_id: int) -> ReviewRequest:
...
+                if transaction["fields"]["replyToCommentPHID"] is not None:
+                    continue
+
+                # Ignore bot comments
+                if transaction["authorPHID"] == "PHID-USER-cje4weq32o3xyuegalpj":
+                    continue
+
+                if len(transaction["comments"]) != 1:
+                    print(transaction)
+                    # raise Exception("Unexpected, need to look into it")
```
marco-c, line 326: Log warning?

bugbug/tools/code_review.py
```diff
@@ -117,9 +138,145 @@ def get_review_request_by_id(self, review_id: int) -> ReviewRequest:
...
+            with open(f"patches/{diff_id}.patch", "r") as f:
+                patch = Patch(f.read())
+        except FileNotFoundError:
+            with open(f"patches/{diff_id}.patch", "w") as f:
+                patch = self.get_patch_by_id(diff_id)
+                f.write(patch.raw_diff)
+
+        return patch
+
+    def get_matching_hunk(
```
marco-c, line 169: We should add some tests for this function in a follow-up

bugbug/tools/code_review.py
```diff
@@ -117,9 +138,145 @@ def get_review_request_by_id(self, review_id: int) -> ReviewRequest:
...
+                patch_set = PatchSet.from_string(
+                    self.load_patch_by_id(diff_id).raw_diff
+                )
+            except UnidiffParseError:
+                # TODO: use log instead of print
+                print(f"Failed to parse {diff_id}")
+                continue
+
+            file_map = {
+                patched_file.path: patched_file
```
marco-c, line 261: IIRC there might be cases where the path starts with "a/" or "b/" and it should be removed (similar to https://github.com/michael-yx-wu/environment/commit/59f9fd17ea9b41c0730dd0e5ae236aef27149248.diff)

bugbug/tools/code_review.py
```diff
@@ -129,26 +286,87 @@ def get_review_request_by_id(self, revision_id: int) -> ReviewRequest:
...
+
+                diff_id = transaction["fields"]["diff"]["id"]
+                filename = transaction["fields"]["path"]
+                start_line = transaction["fields"]["line"]
+                end_line = (
+                    transaction["fields"]["line"] + transaction["fields"]["length"] - 1
+                )
+                # Unfortunately, we do not have this information for a limitation
+                # in Phabricator's API. We assume it as true as a workaround.
+                on_removed_code = True
```
marco-c, line 342: Why not using None to avoid any assumptions?

scripts/review_comments_retriever.py
```diff
@@ -0,0 +1,22 @@
...
+from bugbug.vectordb import QdrantVectorDB
+
+
+def main():
+    review_data = PhabricatorReviewData()
+    vector_db = QdrantVectorDB("diff_comments")
+    vector_db.setup()
+    comments_db = ReviewCommentsDB(vector_db)
+    # TODO: support resuming from where last run left off. We should run it from
+    # scratch only ones. Following runs should add only new comments.
```
marco-c, line 17: only "once"?

bugbug/tools/code_review.py
suhaibmujahid: That is correct when using `source_file` or `target_file`, but the getter method for the `path` property already strips that: https://github.com/matiasb/python-unidiff/blob/0f1062137b3c8d2b6b73cad3508dfb8862e66b87/unidiff/patch.py#L400-L401

bugbug/tools/code_review.py
suhaibmujahid: I already added support for None in 70ae0d0490306872b635314a0f48839cd7d44ea2, but forgot to change it here. Fixed in fe2ff948144e76db67af2567519038b02f9b410d.

scripts/review_comments_retriever.py
suhaibmujahid: Fixed in 1f3de3395c9ff9abec0cfb493d902fcc3e4b22d0

bugbug/tools/code_review.py
suhaibmujahid: Good idea. Done in 3b35f12bdda4b312c55b0e4796ad5a9cdd7c015c.

bugbug/tools/code_review.py
suhaibmujahid: Fixed in 133f66ba96570cc3a89cdbc7bf7caf1489f9f2b6

bugbug/tools/code_review.py
suhaibmujahid: I implemented the count using a separate method so we could use a cache decorator if we need to use it somewhere else later. Since we do not currently use it somewhere else, maybe we do not need it for now. 

Done in 15c2ab54f2af83528658b591a4de10418dd58e2e

bugbug/tools/code_review.py
suhaibmujahid: Done in 7aef774a887c24eb48e19346126dd72c9ea37a9b

bugbug/tools/code_review.py
suhaibmujahid: I filed https://github.com/mozilla/bugbug/issues/4223 as a follow-up

marco-c (approved):

PR #4230, opened 2024-06-06
Support Swarm as a source for code review data

olewicki:
Code to handle Swarm as a review tool

### DIFF
--- a/bugbug/swarm.py
+++ b/bugbug/swarm.py
@@ -1,0 +1,99 @@
+import json
+import subprocess
+from datetime import datetime
+from typing import Collection, NewType
+
+from P4 import P4  # from pip p4python
+
+RevisionDict = NewType("RevisionDict", dict)
+
+
+def api_revinfo(instance, rev_id):
+    u = f"https://{instance}api/v10/reviews/{rev_id}"
+    return u
+
+
+def api_filelist_v_fromto(instance, rev_id, v1=0, v2=1):
+    u = f"https://{instance}api/v10/reviews/{rev_id}/files?from={v1}&to={v2}"
+    return u
+
+
+def call(auth, g):
+    command = f"curl -u \"{auth['user']}:{auth['password']}\" \"{g}\""
+    process = subprocess.run(command, stdout=subprocess.PIPE, stderr=subprocess.DEVNULL)
+    return process.stdout
+
+
+def p4_connect(auth):
+    p4 = P4()
+    p4.port = auth["port"]
+    p4.user = auth["user"]
+    p4.password = auth["password"]
+    p4.connect()
+    p4.run_login()
+    p4.tagged = False
+    return p4
+
+
+def get_review(instance, rev_id, version_num, auth):
+    p4 = p4_connect(auth)
+    data_rev = {}
+    g = api_revinfo(instance, rev_id)
+    message = json.loads(call(auth, g))
+    data_rev = message["data"]
+
+    fl = api_filelist_v_fromto(instance, rev_id, v1=version_num[0], v2=version_num[1])
+    file_list = json.loads(call(auth, fl))
+    for what in file_list["data"]:
+        data_rev[what] = file_list["data"][what]
+
+    commit_id = data_rev["reviews"][0]["versions"][version_num[1] - 1]["change"]
+
+    diffs = {}
+    for file in data_rev["files"]:
+        filename1 = file["fromFile"] if "fromFile" in file else file["depotFile"]
+        filename2 = file["depotFile"]
+        commit_id1 = file["diffFrom"] if "diffFrom" in file else f"#{file['rev']}"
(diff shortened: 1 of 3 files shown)


bugbug/tools/code_review.py
```diff
@@ -125,10 +125,12 @@ def __init__(self, patch_id) -> None:
 
 class Patch:
     raw_diff: str
+    file_diff: str
```
suhaibmujahid, line 128: Why do we need `file_diff` here? The `raw_diff` should have the full diff in git format.

bugbug/swarm.py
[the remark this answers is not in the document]
olewicki: Done

bugbug/tools/code_review.py
olewicki: removed it

bugbug/tools/code_review.py
```diff
@@ -126,7 +126,7 @@ def __init__(self, patch_id) -> None:
 class Patch:
     raw_diff: str
 
-    def __init__(self, raw_diff) -> None:
+    def __init__(self, raw_diff, file_diff=None) -> None:
```
suhaibmujahid, line 129: I guess you forgot dropping it from the constructor:
```suggestion
    def __init__(self, raw_diff) -> None:
```

bugbug/tools/code_review.py
```diff
@@ -369,8 +369,50 @@ def get_all_inline_comments(
                 yield diff_id, comments
 
 
+class SwarmReviewData(ReviewData):
+    def __init__(self):
+        self.auth = {
+            "user": get_secret("SWARM_USER"),
+            "password": get_secret("SWARM_PASS"),
+            "port": get_secret("SWARM_PORT"),
+            "instance": get_secret("INSTANCE"),
```
suhaibmujahid, line 378: You may want to follow the same pater by adding the `SWARM` prefix to avoid mixing up vars :)
```suggestion
            "instance": get_secret("SWARM_INSTANCE"),
```

bugbug/tools/code_review.py
```diff
@@ -369,8 +369,50 @@ def get_all_inline_comments(
...
+        }
+
+    # return ReviewRequest object with patch_id = revision_id
```
suhaibmujahid, line 381: We could add this as docstrings instead of comments. Example: https://github.com/mozilla/bugbug/blob/3fd6ab0717391c3ae99bec83383fdd55ba520d7c/bugbug/bugzilla.py#L305-L314

bugbug/swarm.py
```diff
@@ -0,0 +1,104 @@
...
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import json
+import subprocess
+from datetime import datetime
+from typing import Collection, NewType
+
+from P4 import P4  # from pip p4python
+
+RevisionDict = NewType("RevisionDict", dict)
```
suhaibmujahid, line 13: This does not seem to be used.

bugbug/tools/code_review.py
suhaibmujahid: I think the full docstring is not needed here since this implements an interface. The documentation should be in the abstract class instead (I will take care of that later).

Thus, it may be more suitable to use a regular comment inside the method to describe why the patch ID is used as a revision ID.

bugbug/tools/code_review.py
```diff
@@ -369,8 +369,50 @@ def get_all_inline_comments(
...
+        revisions = swarm.get(
+            self.auth,
+            rev_ids=[int(revision_id)],
+            version_l=[version_before, version_num],
+        )
+        assert len(revisions) == 1
+
+        return Patch(
+            revisions[0]["fields"]["diff"]
+        )
```
suhaibmujahid, line 404: Why do we need this method?

suhaibmujahid (commented):
Thank you, @olewicki! I left some minor comments. 

I did not review `bugbug/swarm.py` since I'm not familiar with Swarm, but I expect that you tested it and that it is working.

bugbug/swarm.py
olewicki: fixed

bugbug/tools/code_review.py
olewicki: for the addon on swarm we need to support the different versions of the diff, I use this function to get versions later than the first instead of get_patch_by_id. I left the description of this function as it's not part of the Abstract Class and might not apply to github.

suhaibmujahid: However, the review tool will only use the methods that exist in the abstract class. I see in I see that in `get_patch_by_id()`, you use `0` to `1` as default values. What is the impact of using just that? Where will the `get_patch_by_version_fromto()` method be used?

olewicki: we had a modified run function in the api call, I added it in the code_review_tool_runner.py now

suhaibmujahid (approved):
LGTM! Thank you!

PR #4235, opened 2024-06-07
Train general component model on a single Fenix class

### LINKED ISSUE #4229: Train general component model on a single Fenix class to improve accuracy
The current [general component model](https://github.com/mozilla/bugbug/blob/master/bugbug/models/component.py) trained on all Fenix components. Due to the low number of bugs in some components and similarity with components in other products, the model's accuracy is limited. A proposal is to group all Fenix components into a single Fenix class for training the general model, which can later be refined with the [Fenix-specific component model](https://github.com/mozilla/bugbug/blob/master/bugbug/models/fenixcomponent.py).

benjaminmah:
Resolves #4229.

Groups all Fenix components into a single Fenix class (i.e. label all Fenix bugs as `Fenix`).

### DIFF
--- a/bugbug/models/component.py
+++ b/bugbug/models/component.py
@@ -132,6 +132,8 @@ class ComponentModel(BugModel):
 
     def filter_component(self, product, component):
         full_comp = f"{product}::{component}"
+        if product == "Fenix":
+            return "Fenix"
 
         if full_comp in self.CONFLATED_COMPONENTS_INVERSE_MAPPING:
             return self.CONFLATED_COMPONENTS_INVERSE_MAPPING[full_comp]
--- a/bugbug/models/component.py
+++ b/bugbug/models/component.py
@@ -131,9 +131,9 @@ class ComponentModel(BugModel):
         }
 
     def filter_component(self, product, component):
-        full_comp = f"{product}::{component}"
         if product == "Fenix":
             return "Fenix"
+        full_comp = f"{product}::{component}"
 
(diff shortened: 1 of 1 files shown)


benjaminmah:
[Training Metrics](https://github.com/user-attachments/files/15777117/output.log)

bugbug/models/component.py
```diff
@@ -132,6 +132,8 @@ def __init__(self, lemmatization=False):
 
     def filter_component(self, product, component):
         full_comp = f"{product}::{component}"
+        if product == "Fenix":
+            return "Fenix"
```
suhaibmujahid, line 136: I would move this to the beginning of the method (i.e., before `full_comp`).

suhaibmujahid (commented):
What was the aggregated percussion and recall for Fenix components  before this PR?

bugbug/models/component.py
benjaminmah: Done in 896d05dd6a91069c918c2a51f55d9292c6669344!

benjaminmah: Will generate metrics on the original model before this PR ASAP.

benjaminmah:
Aggregated results of the model before and after:
[Component Model Metrics](https://docs.google.com/document/d/11tk2N7A0GxIvA8aWorqt6eSuzKgZQpPG6_8yycH8wmM/edit?usp=sharing)

suhaibmujahid (approved):
The results looks better. Thank you!

PR #4251, opened 2024-06-13
Dataset creation for inline comments with their resolutions

benjaminmah:
Script to generate dataset of a patch, its inline comments, and the fix patch.

Intended to include:
- bug ID
- revision ID
- initial patch ID
- fix patch ID
- comments associated with initial patch
- diff between the initial patch and the  fix patch

### DIFF
--- a/bugbug/tools/code_review.py
+++ b/bugbug/tools/code_review.py
@@ -537,10 +537,6 @@ class PhabricatorReviewData(ReviewData):
                 if transaction["authorPHID"] == "PHID-USER-cje4weq32o3xyuegalpj":
                     continue
 
-                # No comments
-                if not transaction["comments"]:
-                    continue
-
                 if len(transaction["comments"]) != 1:
                     # Follow up: https://github.com/mozilla/bugbug/issues/4218
                     logger.warning(
--- a/bugbug/tools/code_review.py
+++ b/bugbug/tools/code_review.py
@@ -537,6 +537,10 @@ class PhabricatorReviewData(ReviewData):
                 if transaction["authorPHID"] == "PHID-USER-cje4weq32o3xyuegalpj":
                     continue
 
+                # No comments
+                if not transaction["comments"]:
+                    continue
+
                 if len(transaction["comments"]) != 1:
                     # Follow up: https://github.com/mozilla/bugbug/issues/4218
                     logger.warning(
--- a/bugbug/tools/code_review.py
+++ b/bugbug/tools/code_review.py
@@ -42,6 +42,8 @@ class InlineComment:
     on_removed_code: bool | None
     id: int | None = None
     date_created: int | None = None
+    date_modified: int | None = None
+    is_done: bool | None = None
 
 
 class ModelResultError(Exception):
@@ -560,6 +562,10 @@ class PhabricatorReviewData(ReviewData):
                     # in Phabricator's API.
                     on_removed_code = None
 
+                    # store the last modified date and if the comments have been marked as done
+                    date_modified = transaction_comment["dateModified"]
+                    is_done = transaction["fields"]["isDone"]
+
                     # TODO: we could create an extended dataclass for this
                     # instead of adding optional fields.
                     comment = InlineComment(
@@ -570,6 +576,8 @@ class PhabricatorReviewData(ReviewData):
                         on_removed_code,
                         comment_id,
                         date_created,
+                        date_modified,
+                        is_done,
                     )
                     continue
 
--- a/bugbug/tools/code_review.py
+++ b/bugbug/tools/code_review.py
@@ -525,70 +525,64 @@ class PhabricatorReviewData(ReviewData):
         for revision in tqdm(phabricator.get_revisions(), total=revision_count):
             diff_comments: dict[int, list[InlineComment]] = defaultdict(list)
 
-            try:
-                for transaction in revision["transactions"]:
-                    if transaction["type"] != "inline":
-                        continue
+            for transaction in revision["transactions"]:
+                if transaction["type"] != "inline":
+                    continue
 
-                    # Ignore replies
-                    if transaction["fields"]["replyToCommentPHID"] is not None:
-                        continue
+                # Ignore replies
+                if transaction["fields"]["replyToCommentPHID"] is not None:
+                    continue
 
-                    # Ignore bot comments
-                    if transaction["authorPHID"] == "PHID-USER-cje4weq32o3xyuegalpj":
-                        continue
+                # Ignore bot comments
+                if transaction["authorPHID"] == "PHID-USER-cje4weq32o3xyuegalpj":
+                    continue
 
-                    if len(transaction["comments"]) != 1:
-                        # Follow up: https://github.com/mozilla/bugbug/issues/4218
-                        logger.warning(
-                            "Unexpected number of comments in transaction %s",
-                            transaction["id"],
-                        )
-
-                    transaction_comment = transaction["comments"][0]
-                    comment_id = transaction_comment["id"]
-                    date_created = transaction_comment["dateCreated"]
-                    comment_content = transaction_comment["content"]["raw"]
-
-                    diff_id = transaction["fields"]["diff"]["id"]
-                    filename = transaction["fields"]["path"]
-                    start_line = transaction["fields"]["line"]
-                    end_line = (
-                        transaction["fields"]["line"]
-                        + transaction["fields"]["length"]
-                        - 1
-                    )
-                    # Unfortunately, we do not have this information for a limitation
-                    # in Phabricator's API.
-                    on_removed_code = None
-
-                    # store the last modified date and if the comments have been marked as done
-                    date_modified = transaction_comment["dateModified"]
-                    is_done = transaction["fields"]["isDone"]
-
-                    # TODO: we could create an extended dataclass for this
-                    # instead of adding optional fields.
-                    comment = InlineComment(
-                        filename,
-                        start_line,
-                        end_line,
-                        comment_content,
-                        on_removed_code,
-                        comment_id,
-                        date_created,
-                        date_modified,
-                        is_done,
+                if len(transaction["comments"]) != 1:
+                    # Follow up: https://github.com/mozilla/bugbug/issues/4218
+                    logger.warning(
+                        "Unexpected number of comments in transaction %s",
+                        transaction["id"],
                     )
                     continue
 
-                    if not comment_filter(comment):
-                        continue
+                transaction_comment = transaction["comments"][0]
+                comment_id = transaction_comment["id"]
+                date_created = transaction_comment["dateCreated"]
+                comment_content = transaction_comment["content"]["raw"]
 
-                    diff_comments[diff_id].append(comment)
+                diff_id = transaction["fields"]["diff"]["id"]
+                filename = transaction["fields"]["path"]
+                start_line = transaction["fields"]["line"]
+                end_line = (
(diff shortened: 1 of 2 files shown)


benjaminmah:
Sample dataset: https://drive.google.com/file/d/1QMd_zyCg9fPuavuYsQupD_q3VxWptKju/view?usp=sharing

bugbug/tools/code_review.py
```diff
@@ -314,57 +316,69 @@ def get_all_inline_comments(
...
-                # Ignore bot comments
-                if transaction["authorPHID"] == "PHID-USER-cje4weq32o3xyuegalpj":
-                    continue
-
-                if len(transaction["comments"]) != 1:
-                    # Follow up: https://github.com/mozilla/bugbug/issues/4218
-                    logger.warning(
-                        "Unexpected number of comments in transaction %s",
-                        transaction["id"],
+            try:
```
suhaibmujahid, line 319: Why are you wrapping everything in as a try-except block? This is very dangerous.

bugbug/tools/code_review.py
```diff
@@ -314,57 +316,69 @@ def get_all_inline_comments(
...
-
-                if not comment_filter(comment):
-                    continue
+                    if not comment_filter(comment):
+                        continue
 
-                diff_comments[diff_id].append(comment)
+                    diff_comments[diff_id].append(comment)
 
+            except Exception as e:
```
suhaibmujahid, line 379: We should avoid catching overly broad exception types whenever possible.

scripts/inline_comments_data_collection.py
```diff
@@ -0,0 +1,154 @@
...
+
+class PhabricatorClient:
+    def __init__(self):
+        self.api = PhabricatorAPI(get_secret("PHABRICATOR_TOKEN"))
+
+    def find_revision_from_patch(self, patch_id):
+        diffs = self.api.search_diffs(diff_id=patch_id)
+
+        if not diffs:
+            raise Exception(f"No diffs found for the given patch ID: {patch_id}")
```
suhaibmujahid, line 28: It would be better to create a custom error type instead of using `Exception` directly: https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/broad-exception-raised.html

bugbug/tools/code_review.py
benjaminmah: I originally had this as a temporary solution when retrieving all inline comments, as I encountered the following error and it would halt the entire script:
```
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/Users/bmah/Documents/GITHUB/bugbug/scripts/inline_comments_data_collection.py", line 153, in <module>
    download_inline_comments()
  File "/Users/bmah/Documents/GITHUB/bugbug/scripts/inline_comments_data_collection.py", line 59, in download_inline_comments
    for patch_id, comments in review_data.get_all_inline_comments(lambda c: True):
  File "/Users/bmah/Documents/GITHUB/bugbug/bugbug/tools/code_review.py", line 338, in get_all_inline_comments
    transaction_comment = transaction["comments"][0]
```
    
However, I've removed it in b9803344f026f9040f1ac3fb1e4ba48f5394004d as I have all the inline comments stored locally on my machine.

bugbug/tools/code_review.py
benjaminmah: I've removed it in b9803344f026f9040f1ac3fb1e4ba48f5394004d

bugbug/tools/code_review.py
benjaminmah: I've added a `continue` for transactions with no comments here 063c4750d682f09a0cf9fba905e6ceef81a6e719. This should fix the issue!

bugbug/tools/code_review.py
```diff
@@ -326,6 +328,10 @@ def get_all_inline_comments(
                 if transaction["authorPHID"] == "PHID-USER-cje4weq32o3xyuegalpj":
                     continue
 
+                # No comments
+                if not transaction["comments"]:
+                    continue
+
```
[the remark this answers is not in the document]
benjaminmah: If there are no comments present, we encounter an `IndexError` when accessing the `comments` array:
`transaction_comment = transaction["comments"][0]`

It would make sense to skip any transactions that do not contain any comments.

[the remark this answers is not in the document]
suhaibmujahid: But we already have `len(transaction["comments"]) != 1`, which should take care of this case.

bugbug/tools/code_review.py
benjaminmah: Removed in b68a1b33ce01854a6615c13bd321360b07adcc6b, as it has been resolved in #4266.

bugbug/tools/code_review.py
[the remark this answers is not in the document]
benjaminmah: Removed in b68a1b33ce01854a6615c13bd321360b07adcc6b, as it has been resolved in #4266.

scripts/inline_comments_data_collection.py
benjaminmah: Done here 5d9b655222bae7acbdc0d53fdf20f7a0cd52f56d

scripts/inline_comments_data_collection.py
```diff
@@ -0,0 +1,181 @@
...
+    def get_diff_info_from_phid(self, phid):
+        diffs = self.api.search_diffs(diff_phid=phid)
+        if not diffs:
+            raise NoDiffFoundForPHIDException(phid)
+        return diffs[0]["id"], diffs[0]["revisionPHID"]
+
+    def find_bugid_from_revision_phid(self, phid):
+        revision = self.api.load_revision(rev_phid=phid)
+        if not revision:
+            raise NoRevisionFoundException(phid)
```
suhaibmujahid, line 77: I do not think that you need to manually raise an error here. If there is no data found, the `load_revision()` method will raise `PhabricatorRevisionNotFoundException`.

scripts/inline_comments_data_collection.py
```diff
@@ -0,0 +1,172 @@
...
+        self.patch_id = patch_id
+
+
+class NoDiffFoundForPHIDException(Exception):
+    def __init__(self, phid):
+        super().__init__(f"No diff found for PHID {phid}")
+        self.phid = phid
+
+
+class PhabricatorClient:
```
suhaibmujahid, line 38: Why do we need to create a new Phabricator client class?

benjaminmah: I realized that it was unnecessary to create a new Phabricator client class, so I removed it here: df651341d45b581ef5d106652f7802feaa962c76. The script still has the same functionality!

benjaminmah:
[Updated Sample Dataset](https://github.com/user-attachments/files/16679512/inline_comment_dataset.json)

@suhaibmujahid

scripts/inline_comments_data_collection.py
```diff
@@ -0,0 +1,172 @@
...
+                }
+                save_to_dataset(data)
+
+        patch_count += 1
+        if patch_count >= patch_threshold:
+            break
+
+
+if __name__ == "__main__":
+    download_inline_comments()
```
suhaibmujahid, line 171: It is not ideal to save all comments on the desk before processing. Why do we need to do this?

scripts/inline_comments_data_collection.py
```diff
@@ -0,0 +1,172 @@
...
+
+                data = {
+                    "bug_id": to_int(bug_id),
+                    "revision_phid": revision_phid,
+                    "initial_patch_id": to_int(patch_id),
+                    "fix_patch_id": to_int(fix_patch_id),
+                    "comment": comment,
+                    "fix_patch_diff": patch_diff,
+                }
+                save_to_dataset(data)
```
suhaibmujahid, line 163: This way, we open and close the file for each data point we want to save, which adds unnecessary I/O overhead.

scripts/inline_comments_data_collection.py
```diff
@@ -0,0 +1,172 @@
...
+        return None
+    if not isinstance(value, int):
+        return int(value)
+    return value
+
+
+def process_comments(patch_threshold, diff_length_threshold):
+    comments_dir = "comments"
+    patch_count = 0
+    for file_name in os.listdir(comments_dir):
```
suhaibmujahid, line 123: Why not directly rely on `review_data.get_all_inline_comments()` instead of reading from disk?

scripts/inline_comments_data_collection.py
benjaminmah: I've changed it so we leave the dataset file open while the script is being run: 12a1e6d1f39b0ede21d2805343f8bc54899c135c

benjaminmah:
[updated_dataset.json](https://github.com/user-attachments/files/16716364/inline_comment_dataset.json)

scripts/inline_comments_data_collection.py
benjaminmah: Originally I had downloaded all comments to reduce the time to run the script multiple times, however for general use it makes sense to avoid that. Removed it here: 3fffb4900f54ffa5799a7b465e73d027b6496d6e

scripts/inline_comments_data_collection.py
benjaminmah: Similar to the other comments, I've removed any disk reading here: 3fffb4900f54ffa5799a7b465e73d027b6496d6e

scripts/inline_comments_data_collection.py
suhaibmujahid: That is a valid use case. It could be an option to have a flag to enable caching of requests when we debug locally.

scripts/inline_comments_data_collection.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+if __name__ == "__main__":
+    dataset_file_path = "dataset/inline_comment_dataset.json"
+    with open(dataset_file_path, "a") as dataset_file_handle:
+        process_comments(
+            patch_threshold=250,
+            diff_length_threshold=5000,
+            dataset_file_handle=dataset_file_handle,
+        )
```
suhaibmujahid, line 156: It is a common practice to wrap the logic in a function like `main()`.

suhaibmujahid, line 156: Instead of passing the file, I would have the function yield the results and save them to the file here.

benjaminmah: Done here: 83c7dde3c5722bf24d39c5380fc9a06396b40598

benjaminmah: Done here: 83c7dde3c5722bf24d39c5380fc9a06396b40598

benjaminmah:
Sample dataset: [dataset.json](https://github.com/user-attachments/files/16751807/dataset.json)

scripts/inline_comments_data_collection.py
```diff
@@ -0,0 +1,206 @@
...
+        for transaction in transactions
+        if transaction["type"] == "update"
+        and transaction["dateModified"] <= comment_date_modified
+    ]
+    if not updates:
+        return None
+    most_recent_update = max(
+        updates, key=lambda transaction: transaction["dateModified"]
+    )
+    return most_recent_update
```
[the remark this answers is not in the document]
benjaminmah: The transactions are ordered by descending `dateCreated`, however we look at `dateModified`, so we can't rely on the order. I've simplified the function by cutting a few lines out, however: cf9bc18c9ce3f39fd5ec99cd28b53c43eabcc806

suhaibmujahid (approved):
LGTM!

PR #4270, opened 2024-06-20
Add features based on file paths in the title and description

### LINKED ISSUE #4269: [model:component] Add features based on file paths in the title and description

benjaminmah:
Resolves #4269.

Introduces new feature that uses file paths mentioned in the title and description of a bug and splits it into sub-paths and individual directories/files.


benjaminmah:
Metrics of the newly trained model: [metrics.log](https://github.com/user-attachments/files/15919477/metrics.log)

bugbug/bug_features.py
```diff
@@ -896,3 +896,27 @@ def __call__(
...
+
+class ExtractFilePaths(SingleBugFeature):
+    """Extract file paths (partial and full) from bug data."""
+
+    name = "Extract File Paths"
+
+    def __call__(self, bug: bugzilla.BugDict, **kwargs) -> list[str]:
+        text = bug.get("summary", "") + " " + bug["comments"][0]["text"]
+
+        regex = r"\b[a-zA-Z0-9_/.\-]+(?:\.(?!com|org|net|edu|gov)[a-zA-Z]{2,4})\b"
```
suhaibmujahid, line 909: For performance optimization, we could compile the regex in the constructer and then use the compiled instance here.

bugbug/bug_features.py
```diff
@@ -896,3 +896,27 @@ def __call__(
...
+
+        all_sub_paths = []
+        for path in file_paths:
+            parts = path.split("/")
+            sub_paths = ["/".join(parts[: i + 1]) for i in range(len(parts))]
+
+            all_sub_paths.extend(sub_paths)
+            all_sub_paths.extend(parts)
+
+        return sorted(set(all_sub_paths))
```
suhaibmujahid, line 922: Why do we need them sorted? And why should they be unique?

bugbug/bug_features.py
```diff
@@ -896,3 +896,27 @@ def __call__(
...
+
+    name = "Extract File Paths"
+
+    def __call__(self, bug: bugzilla.BugDict, **kwargs) -> list[str]:
+        text = bug.get("summary", "") + " " + bug["comments"][0]["text"]
+
+        regex = r"\b[a-zA-Z0-9_/.\-]+(?:\.(?!com|org|net|edu|gov)[a-zA-Z]{2,4})\b"
+
+        file_paths = re.findall(regex, text)
+        file_paths = [path for path in file_paths if "/" in path or "." in path]
```
suhaibmujahid, line 912: Why not handle this filtering step in the regexp?

bugbug/bug_features.py
suhaibmujahid, line 909: Why the regex is looking for URLs instead of file paths?

benjaminmah: Currently, the regex excludes anything that terminates with `com`, `org`, `net`, `edu`, `gov`, so we don't accidently include URLs in the feature. This is because both file paths and URLs have the `/` character, so this helps us only focus on the former.

bugbug/bug_features.py
benjaminmah: Good idea, moved the filtering of `/` and `.` into the regex here: 62c9a0901b0004f902d4a5f5acf431888f028ebc!

bugbug/bug_features.py
benjaminmah: Moved to the constructor here 802b534acf426e35e9133e302534e7d27fafa832!

suhaibmujahid: I see that the regex is still a string and is not being compiled using `re.compile()`.
Also, the variable was moved to be a class attribute; compiling the regex string in the constructer could be a better option.

bugbug/bug_features.py
suhaibmujahid: I think that will eliminated while encoding the feature.

bugbug/bug_features.py
benjaminmah: Should be fixed here: 3ea17006acffb250fcd7adff76374993e01e28f1!

bugbug/bug_features.py
benjaminmah: Thank you for letting me know! The current script does not sort or convert the file paths into a set.

suhaibmujahid (commented):
Do you see significant improvement when adding this feature?

bugbug/bug_features.py
```diff
@@ -896,3 +896,29 @@ def __call__(
             for is_type in self.bug_type_extractors
             if is_type(bug, bug_map)
         ]
+
+
+class ExtractFilePaths(SingleBugFeature):
```
marco-c, line 901: nit: the feature names are usually without the verb, so this should be something like "FilePaths"

bugbug/bug_features.py
```diff
@@ -896,3 +896,29 @@ def __call__(
...
+        text = bug.get("summary", "") + " " + bug["comments"][0]["text"]
+
+        file_paths = self.regex.findall(text)
+
+        all_sub_paths = []
+        for path in file_paths:
+            parts = path.split("/")
+            sub_paths = ["/".join(parts[: i + 1]) for i in range(len(parts))]
+
+            all_sub_paths.extend(sub_paths)
```
marco-c, line 921: nit: you should be able to avoid creating a temporary list here by doing `all_sub_paths.extend("/".join(parts[: i + 1]) for i in range(len(parts))`. If instead you want to keep the named `sub_paths` variable for clarity reasons, you can just do `sub_paths = ("/".join(parts[: i + 1]) for i in range(len(parts)))`.

marco-c (changes requested):
Looks good in general, but could you add a few tests for the new class?

bugbug/bug_features.py
benjaminmah: Makes sense! Done here: a752033b5a083b2b8091f1177969ce4718b70f74

bugbug/bug_features.py
benjaminmah: Done here: 30265bd4e79217c51712d5574a97c409e8bd2b36!

benjaminmah:
I've converted this PR to a draft, as I realized there still needs some polishing to do with the extraction of file paths. For example, there are cases where it may mistake a URL or a step (i.e. 1.Step 1, 2.Step2) as a file path.

benjaminmah:
Current metrics: [metrics.log](https://github.com/user-attachments/files/16316806/metrics.log)

Seems to perform slightly worse than the current model and `python3 -m scripts.bug_classifier component --bug-id 1902245` classifies this bug as `Core::Widget: Gtk` (which is incorrect).

It is worth noting that the first instance of the file path feature model correctly classified the above bug as `Core::Networking`, despite it not 100% correctly retrieving the relevant file paths from the bug summary and description. Will continue to look into this.

bugbug/bug_features.py
```diff
@@ -896,3 +897,225 @@ def __call__(
...
+        """Remove URLs and strings containing specific domain extensions from the given text."""
+        for keyword in self.non_file_path_keywords:
+            if keyword in text:
+                text = re.sub(r"\S*" + re.escape(keyword) + r"\S*", "", text)
+        return text
+
+    def extract_valid_file_path(self, word: str) -> str:
+        """Extract the valid file path from the word if it contains a valid extension."""
+        for ext in self.valid_extensions:
+            ext_pattern = re.compile(rf"\.{ext}(?![a-zA-Z])")
```
suhaibmujahid, line 1095: I would compile the regexp in the constructor. 

Also, I'm not sure if it will have a better performance if we have one big regexp instead of many small ones. It may be worth testing.

bugbug/bug_features.py
```diff
@@ -896,3 +897,225 @@ def __call__(
...
+            match = ext_pattern.search(word)
+            if match:
+                ext_index = match.start()
+                prefix = word[:ext_index]
+                alphanumeric_sequence = re.findall(r"[a-zA-Z0-9/_]+", prefix)
+                if alphanumeric_sequence:
+                    return alphanumeric_sequence[-1] + f".{ext}"
```
suhaibmujahid, line 1102: Since we are using an f-string here, we could keep it consistent and use it for all concatenations:
```suggestion
                    return f"{alphanumeric_sequence[-1]}.{ext}"
```

bugbug/bug_features.py
```diff
@@ -896,3 +897,225 @@ def __call__(
...
+        return ""
+
+    def __call__(self, bug: bugzilla.BugDict, **kwargs) -> list[str]:
+        text = bug.get("summary", "") + " " + bug["comments"][0]["text"]
+        text = self.remove_urls(text)
+
+        words = text.split()
+        file_paths = [self.extract_valid_file_path(word) for word in words]
```
suhaibmujahid, line 1110: Can we create a regular expression to extract all of them at once?

bugbug/bug_features.py
benjaminmah: Done here, I've eliminated the need for the for loop by having the potential file extensions separated by `|` in the regex: 77236101c9527e079184c32f4188254905c93a25

bugbug/bug_features.py
benjaminmah: Done here: 77236101c9527e079184c32f4188254905c93a25

benjaminmah:
The current model now classifies `python3 -m scripts.bug_classifier component --bug-id 1902245` correctly as `Core::Networking`. The metrics can be found here: [metrics.log](https://github.com/user-attachments/files/16337853/metrics.log).

bugbug/bug_features.py
benjaminmah: Simplified the code somewhat here: 5c31f6cfe80db8c9d342e33171c482fa28cdf764

benjaminmah:
> Looks good in general, but could you add a few tests for the new class?

Added two tests here: a5a9c0f10abf960f37026d3df8214c2fa155ef24

benjaminmah:
Seems like the tests failed, I'll do some revisions for these ASAP.

marco-c:
What is the difference in average precision / recall? Is there any component which gets much better or much worse?

benjaminmah:
> What is the difference in average precision / recall? Is there any component which gets much better or much worse?

Here are the metrics from the model with the `FilePaths` feature included: [new_model.log](https://github.com/user-attachments/files/16473152/output.log)

Here are the metrics from the currently deployed model (which does not include the `FilePaths` feature): [old_model.log](https://github.com/user-attachments/files/16473302/output2.log)

For the 0.9 CF, the precision increased by 0.02 and recall increased by 0.01. 

Overall, there seems to be an increase in most metrics for specific product-component pairs, however feel free to consult the detailed metrics for the few cases where either the precision or recall dropped with the new model.

bugbug/bug_features.py
```diff
@@ -896,3 +897,235 @@ def __call__(
...
+            "http://",
+            "https://",
+            "www.",
+            ".com",
+            ".org",
+            ".net",
+            ".edu",
+            ".gov",
+            "@",
+        ]
```
marco-c, line 918: It'd be nice to find a more generic way to exclude these

bugbug/bug_features.py
```diff
@@ -896,3 +897,235 @@ def __call__(
...
+            "htaccess",
+            "gitignore",
+            "dockerfile",
+            "makefile",
+            "gradle",
+            "pom",
+            "bat",
+            "cmd",
+            "vbs",
+        }
```
marco-c, line 1081: same here

bugbug/bug_features.py
```diff
@@ -896,3 +897,235 @@ def __call__(
             for is_type in self.bug_type_extractors
             if is_type(bug, bug_map)
         ]
+
+
+class FilePaths(SingleBugFeature):
```
marco-c, line 902: I wonder if there's a library that we could rely upon for this, instead of manually implementing the regexes and exclusions ourselves.

benjaminmah: I have not found any libraries that can extract file paths from a string, but I will continue to do some research around this!

bugbug/bug_features.py
benjaminmah: Removed the hard-coded list, now using `pygments.lexers`: ca2a322b0cc35d8484082d9b777cad14de0ee23c

bugbug/bug_features.py
benjaminmah: I'm now using `publicsuffix2` to compile all public tlds: f7e250899ee319e30dcbd70bd9942f8a63d4b59c

tests/test_bug_features.py
```diff
@@ -187,3 +188,191 @@ def test_BugTypes(read) -> None:
         BugTypes,
         [["performance"], ["memory"], ["power"], ["security"], ["crash"]],
     )
+
+
+def test_FilePaths(read):
+    read(
+        "file_paths.json",
```
marco-c, line 195: Given the fixture is small enough, you could just write the text here directly and remove the fixture. This will make it easier to follow the test.

marco-c (commented):
Given your latest changes, was there any effect on the metrics?

bugbug/bug_features.py
```diff
@@ -896,3 +899,86 @@ def __call__(
...
+        extension_pattern_string = "|".join(
+            re.escape(ext) for ext in self.valid_extensions
+        )
+
+        self.extension_pattern = re.compile(
+            rf"\.({extension_pattern_string})(?![a-zA-Z])"
+        )
+
+        psl = PublicSuffixList()
+        tlds = set()
```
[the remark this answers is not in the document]
benjaminmah: Changed here: 75d1a239f801eb40a5ba15eafd797217cc2203a2

bugbug/bug_features.py
```diff
@@ -896,3 +899,86 @@ def __call__(
...
+            if "." not in entry:
+                tlds.add("." + entry)
+
+        filtered_tlds = [tld for tld in tlds if tld[1:] not in self.valid_extensions]
+        self.non_file_path_keywords.extend(filtered_tlds)
+
+    def remove_urls(self, text: str) -> str:
+        for keyword in self.non_file_path_keywords:
+            if keyword in text:
+                text = re.sub(r"\S*" + re.escape(keyword) + r"\S*", "", text)
```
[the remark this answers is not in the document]
benjaminmah: Changed here: 75d1a239f801eb40a5ba15eafd797217cc2203a2

tests/test_bug_features.py
benjaminmah: Done here: 301783c81fe99a70f609e25200603aeb031b41e7

benjaminmah:
> Given your latest changes, was there any effect on the metrics?

Training the model with the file path feature included and excluded, I got the following results:

| ct      | Feature Inclusion  | pre | rec | spe | f1 | geo | iba  | sup |
|---------------|--------------------|-----------|--------|-------------|----------|----------------|------|---------|
| **Training Set** | With File Path      | 0.95      | 0.95   | 1.00        | 0.95     | 0.97           | 0.95 | 73665  |
|               | Without File Path   | 0.95  | 0.95 | 1.00   | 0.95 | 0.98       | 0.95 | 73656  |
| **No CT**        | With File Path      | 0.64      | 0.63   | 0.99        | 0.62     | 0.78           | 0.60 | 8185   |
|               | Without File Path   | 0.63  | 0.62 | 0.99   | 0.61 | 0.77       | 0.59 | 8184   |
| **60% CT**       | With File Path      | 0.46      | 0.33   | 1.00        | 0.38     | 0.44           | 0.32 | 8185   |
|               | Without File Path   | 0.44  | 0.32 | 1.00   | 0.36 | 0.42       | 0.31 | 8184  |
| **70% CT**       | With File Path      | 0.47      | 0.32   | 1.00        | 0.37     | 0.42           | 0.30 | 8185   |
|               | Without File Path   | 0.45  | 0.30 | 1.00   | 0.36 | 0.41       | 0.29 | 8184  |
| **80% CT**       | With File Path      | 0.49      | 0.29   | 1.00        | 0.36     | 0.41           | 0.28 | 8185   |
|               | Without File Path   | 0.47  | 0.28 | 1.00   | 0.34 | 0.39       | 0.27 | 8184   |
| **90% CT**       | With File Path      | 0.50      | 0.26   | 1.00        | 0.33     | 0.38           | 0.25 | 8185   |
|               | Without File Path   | 0.48  | 0.25 | 1.00   | 0.32 | 0.36       | 0.24 | 8184   |


Overall, there seems to be a marginal increase in precision and recall when the file path feature is included.

bugbug/bug_features.py
```diff
@@ -896,3 +899,83 @@ def __call__(
...
+        psl = PublicSuffixList()
+        tlds = set(f".{entry}" for entry in psl.tlds if "." not in entry)
+
+        filtered_tlds = [tld for tld in tlds if tld[1:] not in self.valid_extensions]
+        self.non_file_path_keywords.extend(filtered_tlds)
+
+    def remove_urls(self, text: str) -> str:
+        for keyword in self.non_file_path_keywords:
+            if keyword in text:
+                text = re.sub(rf"\S*{re.escape(keyword)}\S*", "", text)
```
suhaibmujahid, line 944: You could gain a slight performance improvement if you pre-compile the regexp.

suhaibmujahid (commented):
@benjaminmah could you rebase and resolve the merge conflicts?

bugbug/bug_features.py
benjaminmah: Done here: f1411237124aa79009535274ee790f9abca6efe2

tests/test_bug_features.py
```diff
@@ -38,15 +39,26 @@
 
 @pytest.fixture
 def read(get_fixture_path):
-    def _read(path, feature_extractor_class, expected_results):
+    def _read(
+        path,
+        feature_extractor_class,
+        expected_results,
+        use_inline_data=False,
+        inline_data="",
```
suhaibmujahid, line 47: I would avoid having empty strings as the default value. It will be more explicit to use `None`.

benjaminmah: 0197f24e5750411eeca1d52df8f3c2f12f07270e

tests/test_bug_features.py
```diff
@@ -38,15 +39,26 @@
 
 @pytest.fixture
 def read(get_fixture_path):
-    def _read(path, feature_extractor_class, expected_results):
+    def _read(
+        path,
+        feature_extractor_class,
+        expected_results,
+        use_inline_data=False,
+        inline_data=None,
```
suhaibmujahid, line 47: Why do we have both `use_inline_data` and `inline_data`? Would it be more straightforward to rely on `inline_data` only?

benjaminmah: 63672d60d650acf49bf1bbc3863ee133b7a55512

tests/test_bug_features.py
```diff
@@ -38,15 +39,25 @@
...
+        else:
+            path = get_fixture_path(os.path.join("bug_features", path))
+            with open(path, "r") as f:
+                lines = f.readlines()
 
-        with open(path, "r") as f:
-            results = (feature_extractor(json.loads(line)) for line in f)
-            for result, expected_result in zip(results, expected_results):
-                assert result == expected_result
+            results = (feature_extractor(json.loads(line)) for line in lines)
```
suhaibmujahid, line 57: Why do we use `readlines()`? It adds the overhead of unnecessarily looping over all the lines in the file (as part of the `readlines()` method) and then looping over them again.

tests/test_bug_features.py
```diff
@@ -187,3 +198,207 @@ def test_BugTypes(read) -> None:
         BugTypes,
         [["performance"], ["memory"], ["power"], ["security"], ["crash"]],
     )
+
+
+def test_FilePaths(read):
+    inline_data = [
```
suhaibmujahid, line 204: Why do we need inline data in the first place? Why this is not stored in a file?

benjaminmah: See: https://github.com/mozilla/bugbug/pull/4270#discussion_r1806379444

tests/test_bug_features.py
```diff
@@ -187,3 +188,210 @@ def test_BugTypes(read) -> None:
...
+            "esp_lock.js",
+            "/esp_lock.js",
+            "esp_lock.js",
+        ],
+    ]
+
+    results = (FilePaths(item) for item in inline_data)
+
+    for result, expected_result in zip(results, expected_results):
+        assert result == expected_result
```
suhaibmujahid, line 397: You do not need to loop over the result:

```suggestion
    results = [FilePaths(item) for item in inline_data]
    assert results == expected_results
```

benjaminmah: 3f196d6db7fb6112b785d66525bdb8efdcb71b78

bugbug/bug_features.py
```diff
@@ -903,3 +906,85 @@ class BugType(SingleBugFeature):
...
+
+    def __init__(self):
+        self.non_file_path_keywords = [
+            "http://",
+            "https://",
+            "www.",
+            "@",
+        ]
+
+        self.valid_extensions = set(
```
suhaibmujahid, line 924: Why do we need to have `valid_extensions` as an attribute when it is only used locally?

benjaminmah: Since it's not being used anywhere except for `init`, good point! Changed to a local variable here: 228b558457e5874bd3d0d0b5e4d275492ae10ea3

bugbug/bug_features.py
```diff
@@ -903,3 +906,81 @@ class BugType(SingleBugFeature):
...
+            "@",
+        ]
+
+        valid_extensions = set(ext.lstrip(".") for ext in mimetypes.types_map.keys())
+
+        lexers = get_all_lexers()
+        lexer_extensions = set(ext[2:] for lexer in lexers for ext in lexer[2])
+
+        valid_extensions.update(lexer_extensions)
+        valid_extensions = sorted(valid_extensions, key=len, reverse=True)
```
suhaibmujahid, line 930: Is there a specific reason for sorting the extensions here? If so, it would be helpful to add it as a code comment.

bugbug/bug_features.py
```diff
@@ -903,3 +906,81 @@ class BugType(SingleBugFeature):
...
+
+        psl = PublicSuffixList()
+        tlds = set(f".{entry}" for entry in psl.tlds if "." not in entry)
+
+        filtered_tlds = [tld for tld in tlds if tld[1:] not in valid_extensions]
+        self.non_file_path_keywords.extend(filtered_tlds)
+
+        keyword_pattern_string = "|".join(
+            re.escape(keyword) for keyword in self.non_file_path_keywords
+        )
```
suhaibmujahid, line 946: Same as https://github.com/mozilla/bugbug/pull/4270#discussion_r1812586564 for `non_file_path_keywords`.

requirements.txt
```diff
@@ -22,6 +22,7 @@ orjson==3.10.7
 ortools==9.11.4210
 pandas==2.2.3
 psutil==6.0.0
+publicsuffix2==2.20191221
```
suhaibmujahid, line 25: I'm a bit skeptical about adding a non-actively maintained package.

bugbug/bug_features.py
```diff
@@ -903,3 +906,81 @@ class BugType(SingleBugFeature):
...
+        self.keyword_pattern = re.compile(rf"\S*({keyword_pattern_string})\S*")
+
+    def remove_urls(self, text: str) -> str:
+        return self.keyword_pattern.sub("", text)
```
suhaibmujahid, line 950: Why do we need to remove URLs instead of just not selecting them?

bugbug/bug_features.py
```diff
@@ -903,3 +906,81 @@ class BugType(SingleBugFeature):
...
+
+    def __init__(self):
+        self.non_file_path_keywords = [
+            "http://",
+            "https://",
+            "www.",
+            "@",
+        ]
+
+        valid_extensions = set(ext.lstrip(".") for ext in mimetypes.types_map.keys())
```
suhaibmujahid, line 924: Why don't we focus only on code-related extensions? Is this improving the results?

bugbug/bug_features.py
```diff
@@ -903,3 +906,81 @@ class BugType(SingleBugFeature):
...
+
+        lexers = get_all_lexers()
```
suhaibmujahid, line 926: The list from Lexers is not comprehensive. For example, it does not include Rust (i.e. `.rs`), a widely used language in Firefox.

bugbug/bug_features.py
benjaminmah: Can you elaborate on code-related extensions? I'm essentially extracting any instance of a file path in the title or description.

bugbug/bug_features.py
benjaminmah: When I call the generator `get_all_lexers`, one of the included extensions is: `('Rust', ('rust', 'rs'), ('*.rs', '*.rs.in'), ('text/rust', 'text/x-rust'))`, so I think it does include Rust.

bugbug/bug_features.py
benjaminmah: f8897bf56ba72a1acc299c99ead8a71bf5fd3154

bugbug/bug_features.py
benjaminmah: Yes, good point. I've added a comment here: 74affe9843a822632324e4ef807e27dd33f1aa57.

The main reason why I sort from longest to shortest extension is to avoid partial matches. For example, a file with extension ".css" could accidentally be matched with ".c" if the C extension comes before the CSS extension in the set.

requirements.txt
benjaminmah: I understand your point, but in this case I'm only using the library for a list of public suffixes. If we want to avoid using a non-actively maintained package, we could also hard-code it, but I don't think that is any better.

bugbug/bug_features.py
benjaminmah: I've changed the code to exclude URLs, rather than removing all of them from the string: b6af3c6aaf5ab49fa890a62e8197fa6a22247f6e

bugbug/bug_features.py
suhaibmujahid: How is sorting solving that? Why is this a problem?

bugbug/bug_features.py
suhaibmujahid: What value do extensions from `mime-types` add over using extensions from `lexers` only?

bugbug/bug_features.py
```diff
@@ -903,3 +906,84 @@ class BugType(SingleBugFeature):
...
+        lexer_extensions = set(ext[2:] for lexer in lexers for ext in lexer[2])
+
+        valid_extensions.update(lexer_extensions)
```
suhaibmujahid, line 929: We could simplify it as follows:
```suggestion
        valid_extensions.update(
            ext[2:] for (_, _, exts) in get_all_lexers() for ext in exts
        )
```
Improvements:
- No need to store `get_all_lexers()` in a variable since it is a simple function call. Also, it returns a generator, which means it is not reusable anyway.
- It is better for readability to unpack the table instead of pointing to the index where possible.
- There is no need to create a set to use the `set.update()` method.
- It is simpler to pass the `lexers` extensions directly to the `set.update()` method.

bugbug/bug_features.py
benjaminmah: After some testing I do realize that due to the negative lookahead in the regex, there shouldn't be any problem with incorrect matching of file extensions. b4f6139e3ccb95b34e85952bdc48b08280fa2726

bugbug/bug_features.py
benjaminmah: Good points, definitely more Pythonic! I've configured these changes here: 7eaab610851e880778ecfd0a3f82d3546f0eeccf

suhaibmujahid (commented):
@benjaminmah could you please resolve conflicts, do a self-review and check the impact on performance?

PR #4283, opened 2024-06-28
[model:component] Add sampling techniques to address imbalanced dataset

### LINKED ISSUE #4281: [model:component] Add sampling techniques to address the imbalanced training dataset
The current training set does not utilize any sampling techniques (e.g., oversampling, undersampling, SMOTE) to address the imbalance in the dataset. We could implement appropriate sampling techniques to balance the dataset and equally represent bugs of all products and components.

benjaminmah:
Resolves #4281.

Investigating and adding sampling techniques (i.e. SMOTE, SMOTEEN, RandomUndersampling) to address the imbalanced dataset of bugs.

### DIFF
--- a/bugbug/models/component.py
+++ b/bugbug/models/component.py
@@ -12,7 +12,6 @@ import xgboost
 from dateutil.relativedelta import relativedelta
 from imblearn.over_sampling import SMOTE
 from imblearn.pipeline import Pipeline as ImblearnPipeline
-from imblearn.under_sampling import RandomUnderSampler
 from sklearn.compose import ColumnTransformer
 from sklearn.feature_extraction import DictVectorizer
 from sklearn.pipeline import Pipeline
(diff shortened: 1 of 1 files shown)


benjaminmah:
Still a WIP. Metrics collected from SMOTE can be found here:  [metrics.log](https://github.com/user-attachments/files/16029517/metrics.log)

Takeaways: 
- SMOTE increases the training time to around 2 hours (as opposed to the current 30-40 minute training time)
- The precision and accuracy are extremely low

This is most likely due to the huge differences in the number of bugs in different products and components (1000+ vs 20), and SMOTE matches the number of bugs in each minority class to the majority class, making the ratio of synthetic data to real data very large.

benjaminmah:
@suhaibmujahid Since this PR is mainly experimental, and the results were proven to be much worse, I think we can close this. WDYT?

PR #4298, opened 2024-07-03
[model:component] Add initial product and component as features

### LINKED ISSUE #4297: [model:component] Experimentation with including Product and Component as a feature
Currently the component model does not include initial product and component that the bug was manually classified as. We could experiment with including this feature to see if it can increase the precision and recall of the model.

benjaminmah:
Resolves #4297.

Includes the initial manually classified product and component of a bug as features.

### DIFF
--- a/bugbug/bug_features.py
+++ b/bugbug/bug_features.py
@@ -913,7 +913,10 @@ class InitialProduct(SingleBugFeature):
         if history:
             for entry in history:
                 for change in entry["changes"]:
-                    if change["field_name"] == "product":
+                    if (
+                        change["field_name"] == "product"
+                        and change["removed"] != bug["product"]
+                    ):
                         return change["removed"]
         return None
 
@@ -924,6 +927,9 @@ class InitialComponent(SingleBugFeature):
         if history:
             for entry in history:
                 for change in entry["changes"]:
-                    if change["field_name"] == "component":
+                    if (
+                        change["field_name"] == "component"
+                        and change["removed"] != bug["component"]
+                    ):
                         return change["removed"]
         return None
--- a/bugbug/bug_features.py
+++ b/bugbug/bug_features.py
@@ -915,7 +915,7 @@ class InitialProduct(SingleBugFeature):
                 for change in entry["changes"]:
                     if change["field_name"] == "product":
                         return change["removed"]
-        return bug["product"]
+        return None
 
 
 class InitialComponent(SingleBugFeature):
@@ -926,4 +926,4 @@ class InitialComponent(SingleBugFeature):
                 for change in entry["changes"]:
                     if change["field_name"] == "component":
                         return change["removed"]
-        return bug["component"]
+        return None
--- a/bugbug/bug_features.py
+++ b/bugbug/bug_features.py
@@ -905,3 +905,24 @@ class BugType(SingleBugFeature):
 
     def __call__(self, bug, **kwargs):
         return bug["type"]
+
+class InitialProduct(SingleBugFeature):
+    def __call__(self, bug, **kwargs):
+        history = bug.get("history", [])
+        if history:
+            for entry in history:
+                for change in entry["changes"]:
+                    if change["field_name"] == "product":
+                        return change["removed"]
+        return bug["product"]
+
+
+class InitialComponent(SingleBugFeature):
+    def __call__(self, bug, **kwargs):
+        history = bug.get("history", [])
+        if history:
+            for entry in history:
+                for change in entry["changes"]:
+                    if change["field_name"] == "component":
+                        return change["removed"]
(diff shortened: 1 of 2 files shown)


benjaminmah:
Metrics of the model with the features included: [metrics.log](https://github.com/user-attachments/files/16086986/metrics_new.log)

Consistently achieves 0.59 for accuracy and recall for all confidence thresholds.

bugbug/models/component.py
```diff
@@ -84,6 +84,8 @@ def __init__(self, lemmatization=False):
             bug_features.Whiteboard(),
             bug_features.Patches(),
             bug_features.Landings(),
+            bug_features.Product(),
+            bug_features.Component(),
```
suhaibmujahid, line 88: We should get the values for when the bug was filed, not the current values. I'm not sure if the rollback guarantees that.

benjaminmah: Added two new features, `InitialProduct` and `InitialComponent` that go through the history of the bug: a9b0ed2cbbcfd38b515483fe05dc44e5dfe8946d. 

Here are the metrics: [metrics.log](https://github.com/user-attachments/files/16102039/output.log)

Overall it performs better than the current model, however it is slightly worse than the metrics I shared earlier when using only the `Product` and `Component` features.

marco-c:
Instead of having specific features for this, you could make sure the function to restore the bug to its initial version also handles changes to product and component: https://github.com/mozilla/bugbug/blob/b7b96b574408f48e23b41cfbfabe4c650850e097/bugbug/bug_snapshot.py#L584.

benjaminmah:
> Instead of having specific features for this, you could make sure the function to restore the bug to its initial version also handles changes to product and component: 

Should I create an issue for this? @marco-c

marco-c:
> > Instead of having specific features for this, you could make sure the function to restore the bug to its initial version also handles changes to product and component:
> 
> Should I create an issue for this? @marco-c

Yes thanks!

benjaminmah:
> > > Instead of having specific features for this, you could make sure the function to restore the bug to its initial version also handles changes to product and component:
> > 
> > 
> > Should I create an issue for this? @marco-c
> 
> Yes thanks!

Done here: https://github.com/mozilla/bugbug/issues/4356

suhaibmujahid (commented):
Could you please resolve the merge conflict and report the improvements caused by adding these two features?

benjaminmah:
With the feature (highest threshold):
avg / total       0.57      0.50      1.00      0.53      0.54      0.49      8986

Without the feature (highest threshold):
avg / total       0.51      0.28      1.00      0.35      0.39      0.26      8986

bugbug/bug_features.py
```diff
@@ -905,3 +905,25 @@ class BugType(SingleBugFeature):
...
+
+class InitialComponent(SingleBugFeature):
+    def __call__(self, bug, **kwargs):
+        history = bug.get("history", [])
+        if history:
+            for entry in history:
+                for change in entry["changes"]:
+                    if change["field_name"] == "component":
+                        return change["removed"]
+        return bug["component"]
```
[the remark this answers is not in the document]
benjaminmah: I've made it return `None` if the product or component has not changed: 454ab3d75562fd7e25b5666a4b7af43ff2c092ed
I've commented the results below.

bugbug/bug_features.py
```diff
@@ -905,3 +905,25 @@ class BugType(SingleBugFeature):
...
+        return None
+
+
+class InitialComponent(SingleBugFeature):
+    def __call__(self, bug, **kwargs):
+        history = bug.get("history", [])
+        if history:
+            for entry in history:
+                for change in entry["changes"]:
+                    if change["field_name"] == "component":
```
suhaibmujahid, line 927: We should avoid leaking the component name into the training set:
```suggestion
                    if change["field_name"] == "component" and change["removed"] != bug["component"]:
```

In this case, if the first component was the correct one and then moved to another component, it would return the second one instead of the initial one.

benjaminmah: Done here: b66d5a67e6aff66d1b202bdd2406febbf1070d2f

Currently training the model, will post results once finished.

suhaibmujahid: > where the model would move it to an incorrect component, when it actually belongs in the initial component

This will be kind of leaking the classification into the training data, which harms the ability of the model to learn. Also, we do not run the model on bugs that are not in general components, so this is not currently a problem.

suhaibmujahid: > The metrics actually appear to be lower

Based on this, I suggest closing the issue as not planned.

suhaibmujahid:
Closing due to https://github.com/mozilla/bugbug/pull/4298#discussion_r2045813624

PR #4316, opened 2024-07-09
[code_review] Support showing the patch with the comment examples

suhaibmujahid:
This should close #4267

### DIFF
--- a/bugbug/tools/code_review.py
+++ b/bugbug/tools/code_review.py
@@ -62,16 +62,29 @@ Please, analyze the code provided and report a summarization about the new chang
 
 {patch}"""
 
-PROMPT_TEMPLATE_REVIEW = """You will be given a task for generate a code review for the patch below. Use the following steps to solve it.
+PROMPT_TEMPLATE_REVIEW = """You will be given a task for generate a code review for the patch below. Use the following steps to solve it:
+1. Understand the changes done in the patch by reasoning about the summarization as previously reported.
+2. Identify possible code snippets that might result in possible bugs, major readability regressions, and similar concerns.
+3. Reason about each identified problem to make sure they are valid. Have in mind, your review must be consistent with the source code in Firefox.
+4. Filter out comments that focuses on documentation, comments, error handling, tests, and confirmation whether objects, methods and files exist or not.
+5. Final answer: Write down the comments and report them using the JSON format previously adopted for the valid comment examples.
+
+As valid comments, consider the examples below:
+{comment_examples}
+
+
+Here is the patch that we need you to review:
+{patch}"""
+
+
+TEMPLATE_COMMENT_EXAMPLE = """Patch example {example_number}:
 
 {patch}
 
-1. Understand the changes done in the patch by reasoning about the summarization as previously reported.
-2. Identify possible code snippets that might result in possible bugs, major readability regressions, and similar concerns.
-3. Reason about each identified problem to make sure they are valid. Have in mind, your review must be consistent with the source code in Firefox. As valid comments, consider the examples below:
-{comment_examples}
-4. Filter out comments that focuses on documentation, comments, error handling, tests, and confirmation whether objects, methods and files exist or not.
-5. Final answer: Write down the comments and report them using the JSON format previously adopted for the valid comment examples."""
+Review comments for example {example_number}:
+
+{comments}"""
+
 
 PROMPT_TEMPLATE_FILTERING_ANALYSIS = """Please, double check the code review provided for the patch below.
 Just report the comments that are:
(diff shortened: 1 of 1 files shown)


bugbug/tools/code_review.py
```diff
@@ -62,16 +62,29 @@ class HunkNotInPatchError(ModelResultError):
 
 {patch}"""
 
-PROMPT_TEMPLATE_REVIEW = """You will be given a task for generate a code review for the patch below. Use the following steps to solve it.
-
-{patch}
-
+PROMPT_TEMPLATE_REVIEW = """You will be given a task for generate a code review for the patch below. Use the following steps to solve it:
```
marco-c, line 65:
```suggestion
PROMPT_TEMPLATE_REVIEW = """You will be given a task for generating a code review for the patch below. Use the following steps to solve it:
```

bugbug/tools/code_review.py
```diff
@@ -62,16 +62,29 @@ class HunkNotInPatchError(ModelResultError):
...
 1. Understand the changes done in the patch by reasoning about the summarization as previously reported.
 2. Identify possible code snippets that might result in possible bugs, major readability regressions, and similar concerns.
-3. Reason about each identified problem to make sure they are valid. Have in mind, your review must be consistent with the source code in Firefox. As valid comments, consider the examples below:
-{comment_examples}
+3. Reason about each identified problem to make sure they are valid. Have in mind, your review must be consistent with the source code in Firefox.
```
marco-c, line 68: Note: we should make this customizable so other users can replace "Firefox" with something else. Let's open a follow-up for that

bugbug/tools/code_review.py
```diff
@@ -942,6 +1023,55 @@ def run(self, patch: Patch) -> list[InlineComment] | None:
...
 
+    def _get_comment_examples(self, patch):
+        comment_examples = []
+
+        if self.review_comments_db:
+            # TODO: use a smarter search to limit the number of comments and
+            # diversify the examples (from different hunks, files, etc.).
+            comment_examples = self.review_comments_db.find_similar_patch_comments(
+                patch, limit=10
+            )
```
marco-c, line 1034: Thought: we might want to mix comments written by developers and comments written by the bot (and accepted by developers). Perhaps it could be a number we can configure manually. This is a follow-up of course.

bugbug/tools/code_review.py
suhaibmujahid: Nice catch! The typo is out of the scope of this PR, but I fixed it in 3aea837c99613eac22e3c859499f667278409515 since we are touching this line.

bugbug/tools/code_review.py
suhaibmujahid: I filed a follow-up issue: https://github.com/mozilla/bugbug/issues/4319.

bugbug/tools/code_review.py
suhaibmujahid: I like the idea. I filed a follow-up issue: https://github.com/mozilla/bugbug/issues/4320.

bugbug/tools/code_review.py
```diff
@@ -121,24 +134,109 @@ class HunkNotInPatchError(ModelResultError):
...
++    blockList.includes(ChromeUtils.getXPCOMErrorName(status))
+   ) {
+     blockedReason = ChromeUtils.getXPCOMErrorName(status);
+   }
+
+   return { blockingExtension, blockedReason };""",
     },
 ]
 
+TEMPLATE_PATCH_FROM_HUNK = """diff --git a/{filename} b/{filename}
```
olewicki, line 234: Shouldn't the hunks be shown in the formatted patch format using the format_patch_set function? so that the examples are shown the same way the patches we ask the review for are

bugbug/tools/code_review.py
```diff
@@ -978,12 +1108,20 @@ def vector_points():
...
                 continue
 
             for hunk in patched_file:
-                yield from self.find_similar_hunk_comments(hunk)
+                for result in self.find_similar_hunk_comments(hunk):
+                    yield result
+
+                    yielded += 1
+                    if yielded >= limit:
+                        return
```
olewicki, line 1127: In this case, search for one examples per hunk in order of the patch and limit to 10 examples (according to the call), this will not scale with bigger reviews, where later hunks/files will be ignored. I would generate (at least) one for each hunk then report the top [:limit] ordered by decreasing similarity from the cosine distance that is used. By taking the top, it should also show examples that are similar to code changes that actually need reviews, as they have similar examples in the db

olewicki (commented):
Added a few comments

bugbug/tools/code_review.py
suhaibmujahid: The `format_patch_set()` function requires a parsed patch set, which we use ` PatchSet.from_string()` for. However, the hunk is a part of a patch (not a full patch) and does not show the file name. Thus, I used this template to show the hunk in a valid patch format.

bugbug/tools/code_review.py
suhaibmujahid: I agree with you. I have a todo regarding this: https://github.com/mozilla/bugbug/blob/3aea837c99613eac22e3c859499f667278409515/bugbug/tools/code_review.py#L1030-L1031

I will handle it in a separate pull request.

bugbug/tools/code_review.py
olewicki: but then the TEMPLACE_PATCH_FROM_HUNK (without the diff line) could be used as the patch format then formatted with format_patch_set(), no?

suhaibmujahid: We do pass it to `format_patch_set()`. The following is the order:
```py
format_patch_set(
    PatchSet.from_string(
        TEMPLATE_PATCH_FROM_HUNK.format(
            filename,
            raw_hunk,
        )
    )
)
```

olewicki: you are right, missed the call on line 1052 and 1053. Looks good to me!

marco-c (approved):

PR #4360, opened 2024-07-24
[searchfox_api] Consider nested blocks when determining line numbers

suhaibmujahid:
This should fix the problem when the last element is a nested code block.

### DIFF
--- a/bugbug/code_search/searchfox_api.py
+++ b/bugbug/code_search/searchfox_api.py
@@ -5,8 +5,10 @@
 
 import io
 import json
+from typing import Iterable
 
 import requests
+from lxml.html import HtmlElement
 from requests_html import HTMLSession
 
 from bugbug.code_search import searchfox_data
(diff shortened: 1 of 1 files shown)


bugbug/code_search/searchfox_api.py
```diff
@@ -18,6 +20,20 @@
 
 
 # TODO: we should use commit_hash...
+def get_line_number(elements: Iterable[HtmlElement], position):
+    assert position in ("first", "end")
```
marco-c, line 24: Shouldn't it be "start" instead of "first"? In the function call at line 51 you are using "start"

suhaibmujahid: Nice catch! Fixed in aabaf245bc291903d21bf3ac06f480fb81b3e3a5.

marco-c (approved):
I assume you tested this extensively ;)

PR #4373, opened 2024-07-31
[fenixcomponent] Include GeckoView in training set

### LINKED ISSUE #4355: Fenix component model should include GeckoView too
As GeckoView could be considered part of Fenix.

benjaminmah:
Resolves #4355.

Includes GeckoView when generating labels for the training dataset for the Fenix component model.

### DIFF
--- a/bugbug/models/fenixcomponent.py
+++ b/bugbug/models/fenixcomponent.py
@@ -89,7 +89,10 @@ class FenixComponentModel(BugModel):
                 continue
 
             # Exclude 'General' because it contains bugs that may belong to other components, thus introducing noise.
-            if bug_data["component"] == "General":
+            if (
+                bug_data["component"] == "General"
+                and bug_data["product"] != "GeckoView"
+            ):
                 continue
 
             if dateutil.parser.parse(bug_data["creation_time"]) < date_limit:
--- a/bugbug/models/fenixcomponent.py
+++ b/bugbug/models/fenixcomponent.py
@@ -85,7 +85,7 @@ class FenixComponentModel(BugModel):
         date_limit = datetime.now(timezone.utc) - relativedelta(years=2)
 
         for bug_data in bugzilla.get_bugs():
-            if bug_data["product"] != "Fenix":
+            if bug_data["product"] != "Fenix" and bug_data["product"] != "GeckoView":
                 continue
 
             # Exclude 'General' because it contains bugs that may belong to other components, thus introducing noise.
--- a/bugbug/models/fenixcomponent.py
+++ b/bugbug/models/fenixcomponent.py
@@ -85,10 +85,12 @@ class FenixComponentModel(BugModel):
         date_limit = datetime.now(timezone.utc) - relativedelta(years=2)
 
         for bug_data in bugzilla.get_bugs():
+            # Include GeckoView bugs as they are closely related to Fenix.
             if bug_data["product"] != "Fenix" and bug_data["product"] != "GeckoView":
                 continue
 
-            # Exclude 'General' because it contains bugs that may belong to other components, thus introducing noise.
+            # Exclude 'Fenix::General' because it contains bugs that may belong to other components, thus introducing unwanted noise.
+            # Include 'GeckoView::General' because the model is intended to either predict Fenix components or reclassify Fenix bugs to GeckoView.
             if (
                 bug_data["component"] == "General"
                 and bug_data["product"] != "GeckoView"
--- a/bugbug/models/fenixcomponent.py
+++ b/bugbug/models/fenixcomponent.py
@@ -85,7 +85,8 @@ class FenixComponentModel(BugModel):
         date_limit = datetime.now(timezone.utc) - relativedelta(years=2)
 
         for bug_data in bugzilla.get_bugs():
-            # Include GeckoView bugs as they are closely related to Fenix.
+            # We want the model to be aware of GeckoView bugs, as it is common
+            # for bugs filed under Fenix to end up in GeckoView.
             if bug_data["product"] != "Fenix" and bug_data["product"] != "GeckoView":
                 continue
 
--- a/bugbug/models/fenixcomponent.py
+++ b/bugbug/models/fenixcomponent.py
@@ -85,7 +85,7 @@ class FenixComponentModel(BugModel):
         date_limit = datetime.now(timezone.utc) - relativedelta(years=2)
 
         for bug_data in bugzilla.get_bugs():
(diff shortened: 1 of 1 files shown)


benjaminmah:
[Model Metrics](https://github.com/user-attachments/files/16443698/output.log)

marco-c:
What is the difference in terms of average precision / recall? Is there any component that gets significantly better or significantly worse?

marco-c:
And we should adjust bugbug/models/component.py too to ignore GeckoView in addition to Fenix

marco-c:
That is, consider GeckoView == Fenix as a single class

benjaminmah:
> That is, consider GeckoView == Fenix as a single class

I've done this in #4376, as this is related more to the general component model rather than the Fenix component model. For the Fenix model, I plan on adding all GeckoView labels to the training set, so the model can re-classify the `Fenix::General` bugs to either Fenix or GeckoView components.

benjaminmah:
> What is the difference in terms of average precision / recall? Is there any component that gets significantly better or significantly worse?

From training the model with all GeckoView and Fenix bugs vs just Fenix bugs, there is an overall decrease in precision from 0.84 to 0.81 (which is expected), but the recall remains to be 0.55.

In terms of specific components, the following drop in precision:
- Accounts and Sync (0.93 to 0.62)
- Browser Engine (0.91 to 0.58)
- Logins (0.75 to 0.33)
- Media (1.00 to 0.80)
- Performance (1.00 to 0.71)
- Top Sites (1.00 to 0.67)
- UI Tests (0.99 to 0.97)
- WebAuthn (1.00 to 0.00)
- WebExtensions (1.00 to 0.91)

However, some of these components have very few bugs, so steep drops may be due to a single bug misclassified.

Additionally, many of the components had an increase in their precision, and the full results can be found here:

Fenix and GeckoView model: [FenixAndGeckoView.log](https://github.com/user-attachments/files/16549809/output.log)

Fenix model: [Fenix.log](https://github.com/user-attachments/files/16549818/output2.log)

marco-c:
Most of those components indeed have very few bugs in the test set, so it makes sense that a small difference in absolute number can lead to a big increase in relative number. I think it's acceptable.

The only concerning one might be Browser Engine (0.91 to 0.58), as there are more bugs there. We could investigate that in a follow-up.

marco-c:
There might be components that share the same name between Fenix and GeckoView, we should add "Fenix::" and "GeckoView::" to the names.

At the moment this only applies to "Media", but it still problematic (how to differentiate between "Fenix::Media" and "GeckoView::Media"?).

benjaminmah:
> There might be components that share the same name between Fenix and GeckoView, we should add "Fenix::" and "GeckoView::" to the names.
> 
> At the moment this only applies to "Media", but it still problematic (how to differentiate between "Fenix::Media" and "GeckoView::Media"?).

Good point, done here: [results.log](https://github.com/user-attachments/files/16562365/output.log)

Other than `GeckoView::Media` which has a precision and recall of 0.00, the rest of the GeckoView product/components are above 0.67.

marco-c:
Looking at GeckoView::Media for the default confidence level, there are 4 bugs in the test set:
- 1 is assigned to Fenix::Media, which is reasonable;
- 1 is assigned to GeckoView::Media, which is correct;
- 2 are assigned to GeckoView::General, which is not ideal but better than other potential errors.

benjaminmah:
> Looking at GeckoView::Media for the default confidence level, there are 4 bugs in the test set:
> 
>     * 1 is assigned to Fenix::Media, which is reasonable;
> 
>     * 1 is assigned to GeckoView::Media, which is correct;
> 
>     * 2 are assigned to GeckoView::General, which is not ideal but better than other potential errors.

I originally had only looked at the 0.9 CF but using the default CF makes more sense! These results seem okay to me.

bugbug/models/fenixcomponent.py
```diff
@@ -85,11 +85,14 @@ def get_labels(self):
         date_limit = datetime.now(timezone.utc) - relativedelta(years=2)
 
         for bug_data in bugzilla.get_bugs():
-            if bug_data["product"] != "Fenix":
+            if bug_data["product"] != "Fenix" and bug_data["product"] != "GeckoView":
```
suhaibmujahid, line 88: nit: Adding a comment explaining why we are including `GeckoView` would be helpful for future maintainers (including us).

suhaibmujahid (approved):
LGTM! Thank you @benjaminmah!

bugbug/models/fenixcomponent.py
```diff
@@ -85,11 +85,14 @@ def get_labels(self):
...
                 continue
 
             # Exclude 'General' because it contains bugs that may belong to other components, thus introducing noise.
```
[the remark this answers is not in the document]
benjaminmah: Done here: 082ac3d0611ff5d2eac2c597fe3894fac5d78c7e

bugbug/models/fenixcomponent.py
benjaminmah: Done here: 082ac3d0611ff5d2eac2c597fe3894fac5d78c7e

bugbug/models/fenixcomponent.py
```diff
@@ -85,11 +85,16 @@ def get_labels(self):
         date_limit = datetime.now(timezone.utc) - relativedelta(years=2)
 
         for bug_data in bugzilla.get_bugs():
-            if bug_data["product"] != "Fenix":
+            # Include GeckoView bugs as they are closely related to Fenix.
```
suhaibmujahid, line 88:
```suggestion
            # We want the model to be aware of GeckoView bugs, as it is common
            # for bugs filed under Fenix to end up in GeckoView.
```

bugbug/models/fenixcomponent.py
```diff
@@ -85,11 +85,16 @@ def get_labels(self):
...
+            if bug_data["product"] != "Fenix" and bug_data["product"] != "GeckoView":
                 continue
 
-            # Exclude 'General' because it contains bugs that may belong to other components, thus introducing noise.
-            if bug_data["component"] == "General":
+            # Exclude 'Fenix::General' because it contains bugs that may belong to other components, thus introducing unwanted noise.
+            # Include 'GeckoView::General' because the model is intended to either predict Fenix components or reclassify Fenix bugs to GeckoView.
```
suhaibmujahid, line 93:
```suggestion
            # Exclude 'General' because it contains bugs that may belong to
            # other components, thus introducing noise. However, include
            # 'GeckoView::General' because the model should be able to move
            # bugs to GeckoView.
```

suhaibmujahid (approved):

PR #4376, opened 2024-08-01
[component] Combine all Fenix and GeckoView bugs in training set for component model

### LINKED ISSUE #4374: [component] Combine all Fenix and GeckoView bugs in training set for component model
Due to #4355, we would need to train the general component model on all Fenix and GeckoView bugs considered to be a single class.

benjaminmah:
Resolves #4374.

Intended to combine all Fenix and GeckoView bugs into a single class for the training set.

Linked to #4355, as these bugs will be classified generally by the component model prior to being passed into the Fenix/GeckoView-specific component model.

### DIFF
--- a/bugbug/models/component.py
+++ b/bugbug/models/component.py
@@ -131,11 +131,8 @@ class ComponentModel(BugModel):
         }
 
     def filter_component(self, product, component):
-        if product == "Fenix":
+        if product == "Fenix" or product == "GeckoView":
             return "Fenix"
-
-        if product == "GeckoView":
-            return "GeckoView::General"
 
         full_comp = f"{product}::{component}"
 
--- a/bugbug/models/component.py
+++ b/bugbug/models/component.py
@@ -133,6 +133,10 @@ class ComponentModel(BugModel):
     def filter_component(self, product, component):
         if product == "Fenix":
             return "Fenix"
+
+        if product == "GeckoView":
+            return "GeckoView::General"
+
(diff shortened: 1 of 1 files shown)


bugbug/models/component.py
```diff
@@ -133,6 +133,10 @@ def __init__(self, lemmatization=False):
     def filter_component(self, product, component):
         if product == "Fenix":
             return "Fenix"
+
+        if product == "GeckoView":
+            return "GeckoView::General"
```
marco-c, line 138: It might be better to combine all of them together as a single class, the model should perform better this way

benjaminmah: Yep, I was planning on doing so after training the model once! I'll be sure to compare the results as well. My only concern with combining them into a single class is if the bug is not classified by the Fenix/GeckoView-specific model, it will default to the component model classification (which would be our combined class). Should the combined class simply be `Fenix::General`?

benjaminmah:
[Model Metrics](https://github.com/user-attachments/files/16461667/output.log)

Performs slightly better than the current model, as all GeckoView and Fenix bugs are grouped together under the label `Fenix::General`.

bugbug/models/component.py
marco-c: I see your point.
Yes, I think it'd be ideal to fallback on `Fenix::General` when the specific model isn't sure.

marco-c (approved):

Issue #384, opened 2019-05-09
Add tests for repository._transform


marco-c:
Also change the function name, since we're there.

rajibmitra:
I would like to work on this , could you please assign this me @marco-c

marco-c:
@rajibmitra feel free to work on it! Issues are assigned automatically when there is a PR to fix them.

Iantonga:
hello I would like to work on this issue, could you give me some pointers on where to get started? Thanks!

marco-c:
@lantonga we need to add tests for the _transform function of the repository module.

Nathan-Wisner:
Can someone more intimate with the _transform function either add comments or give us pseudocode / an oversimplification so we can start understanding it?

marco-c:
The function is analyzing a commit and gathering a few metrics from it. What parts of the function are not clear? If you have specific doubts I can help clarifying them.

alihaiderrizvi:
I'm having trouble installing `rs_parsepatch`. I installed all dependencies from `requirements.txt` and `test-requirements.txt` except `rs_parsepatch`. This is the error I'm getting:

![error1](https://user-images.githubusercontent.com/45711004/111027305-52aee180-83a4-11eb-81ae-be1756a25684.png)
![error2](https://user-images.githubusercontent.com/45711004/111027323-68bca200-83a4-11eb-96e2-b8ed01be7796.png)
![image](https://user-images.githubusercontent.com/45711004/111027340-838f1680-83a4-11eb-9f44-f6b840ebea5d.png)

marco-c:
@alihaiderrizvi from the error, it looks like you need to install Visual Studio.

alihaiderrizvi:
I installed Visual Studio, and now I'm getting the following error:
![error1](https://user-images.githubusercontent.com/45711004/111079765-a579bd80-84b8-11eb-8c31-33454e823c1e.png)
![error2](https://user-images.githubusercontent.com/45711004/111079770-aca0cb80-84b8-11eb-8e86-7ee3fb182c36.png)

marco-c:
Try setting Rust nightly as your default

alihaiderrizvi:
That worked! `rs_parsepatch` is installed.
However, when I ran pytest on `tests/test_repository.py`, many tests return the following error:
`Exception: rust-code-analysis is required for code analysis`
[...]

PR #4404, opened 2024-08-06
Added initial tests for repository._transform.

flowejam:
Also added comments re. failed tests for repository._transform.
This is for #384

### DIFF
--- a/bugbug/repository.py
+++ b/bugbug/repository.py
@@ -716,6 +716,7 @@ def set_commit_metrics(
 
 
 def transform(hg: hglib.client, repo_dir: str, commit: Commit) -> Commit:
+    # TODO: rename this function and the corresponding tests
     hg_modified_files(hg, commit)
 
     if commit.ignored or len(commit.backsout) > 0 or commit.bug_id is None:
--- a/tests/test_repository.py
+++ b/tests/test_repository.py
@@ -2859,3 +2859,101 @@ void main() {
             }
         ]
     }
+
+
+def test_transform(fake_hg_repo):
+    def create_commit_obj(node, desc, bug_id, backsout, backedoutby):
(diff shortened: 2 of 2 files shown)


suhaibmujahid (commented):
Thank you for your contribution, @flowejam!

The tests are failing! Could you please fix that?

flowejam:
> Thank you for your contribution, @flowejam!
> 
> The tests are failing! Could you please fix that?

Thanks Suhaib!

Apologies if I'm missing something here, but from the log it looks like the only test that is failing is the one I added - I've included a snippet below. I think there could be an issue with the way tests are detected?

```
  # The test currently fails here with an error - might indicate an issue with
        # the is_test function in bugbug/repository.py.
        repository.transform(hg, local, commit3)
>       assert commit3.test_files_modified_num == 1
E       AssertionError: assert 0 == 1
```
E.g. for bugbug, tests are in a folder called 'tests' in the main directory of the repo, but wouldn't they be missed by this function (repository.is_test)?:

```
# This code was adapted from https://github.com/mozsearch/mozsearch/blob/2e24a308bf66b4c149683bfeb4ceeea3b250009a/router/router.py#L127
def is_test(path: str) -> bool:
    return (
        "/test/" in path
        or "/tests/" in path
        or "/mochitest/" in path
        or "/unit/" in path
        or "/gtest/" in path
        or "testing/" in path
        or "/jsapi-tests/" in path
        or "/reftests/" in path
        or "/reftest/" in path
        or "/crashtests/" in path
        or "/crashtest/" in path
        or "/gtests/" in path
        or "/googletest/" in path
    )
```

flowejam:
@suhaibmujahid @marco-c just thought I'd follow up - any chance of getting another look at this?

PR #4406, opened 2024-08-07
Rag example selection based on score

olewicki:


bugbug/tools/code_review.py
```diff
@@ -1116,15 +1116,24 @@ def find_similar_patch_comments(self, patch: Patch, limit: int):
...
+                        result.id not in max_score_per_found
+                        or result.score > max_score_per_found[result.id]
+                    ):
+                        max_score_per_found[result.id] = result
+
+        list_found_with_score = [
+            (max_score_per_found[e].score, max_score_per_found[e])
+            for e in max_score_per_found
+        ]
+        list_found_with_score.sort()  # order based on score
```
suhaibmujahid, line 1136: This could be simplified using `sorted()`:
```suggestion
        list_found_with_score = sorted(
            (max_score_per_found[e].score, max_score_per_found[e])
            for e in max_score_per_found
        )
```

bugbug/tools/code_review.py
```diff
@@ -1116,15 +1116,24 @@ def find_similar_patch_comments(self, patch: Patch, limit: int):
 
         patch_set = PatchSet.from_string(patch.raw_diff)
 
-        yielded = 0
+        max_score_per_found = {}  # to avoid repetition of same found
```
suhaibmujahid, line 1119: I expanded the comment to make it clearer for future us :)
```suggestion
        # We want to avoid returning the same comment multiple times. Thus, if
        # a comment matches multiple hunks, we will only consider it once.
        max_score_per_found = {}
```

bugbug/tools/code_review.py
```diff
@@ -1116,15 +1116,24 @@ def find_similar_patch_comments(self, patch: Patch, limit: int):
...
-                    yielded += 1
-                    if yielded >= limit:
-                        return
+                    if (
+                        result.id not in max_score_per_found
+                        or result.score > max_score_per_found[result.id]
+                    ):
+                        max_score_per_found[result.id] = result
+
+        list_found_with_score = [
```
suhaibmujahid, line 1132:
```suggestion
        results_with_score = [
```

bugbug/tools/code_review.py
suhaibmujahid, line 1119:
```suggestion
        max_score_per_comment = {}  # to avoid repetition of same found
```

suhaibmujahid (commented):
Thank you, @olewicki! It looks good to me. I left some minor comments to improve readability.

bugbug/tools/code_review.py
olewicki: done

bugbug/tools/code_review.py
olewicki: done

bugbug/tools/code_review.py
```diff
@@ -1116,15 +1116,25 @@ def find_similar_patch_comments(self, patch: Patch, limit: int):
...
+                        result.id not in max_score_per_comment
+                        or result.score > max_score_per_comment[result.id]
+                    ):
+                        max_score_per_comment[result.id] = result
+
+        results_with_score = sorted(
+            (result.score, result)
+            for result in max_score_per_comment.values()
+        )
+        list_found = [e[1] for e in results_with_score[-limit:]]
```
suhaibmujahid, line 1138: nit: this will be self-explanatory to improve the readability:
```suggestion

        return [result for score, result in results_with_score[-limit:]]
```

suhaibmujahid (commented):
Could you please fix the linting errors?

bugbug/tools/code_review.py
```diff
@@ -1116,15 +1119,23 @@ def find_similar_patch_comments(self, patch: Patch, limit: int):
...
-                    yield result
+                    if result is not None and (
+                        result.id not in max_score_per_comment
+                        or result.score > max_score_per_comment[result.id].score
+                    ):
+                        max_score_per_comment[result.id] = result
+
+        results_with_score = sorted(
+            (result.score, result) for result in max_score_per_comment.values()
+        )
```
suhaibmujahid, line 1139: We could rely on the `PayloadScore` dataclass to be sortable and then do the following:
```suggestion
        sorted_results = sorted(
            result for result in max_score_per_comment.values()
        )
```

To make the `PayloadScore` dataclass sortable, we need to pass `order=True` to `dataclass()` and re-order the fields to have `score` as the first one.

bugbug/tools/code_review.py
```diff
@@ -1116,15 +1119,23 @@ def find_similar_patch_comments(self, patch: Patch, limit: int):
...
 
-                    yielded += 1
-                    if yielded >= limit:
-                        return
+        return [e[1] for e in results_with_score[-limit:]]
```
suhaibmujahid, line 1141: I think we should return the payload only:
```suggestion
        return [result.payload for result in sorted_results[-limit:]]
```

(the suggestion depends on https://github.com/mozilla/bugbug/pull/4406#discussion_r1711690040)

bugbug/tools/code_review.py
```diff
@@ -1009,11 +1009,13 @@ def run(self, patch: Patch) -> list[InlineComment] | None:
                     "output": "Okay, I will also consider the code as additional context to the given patch."
                 },
             )
-
+        comment_examples, comment_examples_with_score = self._get_comment_examples(
+            patch
+        )
```
suhaibmujahid, line 1014: The `comment_examples_with_score` is not used; we should drop it for now, and we could add it back when needed.

suhaibmujahid: I fixed that in a62171ee2ab72acc5d724baf0fd7b48028226980

PR #4432, opened 2024-08-22
Include additional products in `get_labels()` for `spambug`

### LINKED ISSUE #2786: Train spambug model on components from all products

jpangas:
Closes #2786 

Includes the missing fix introduced in #3451 to `get_labels()` in `spambug`

### DIFF
--- a/bugbug/models/spambug.py
+++ b/bugbug/models/spambug.py
@@ -93,7 +93,10 @@ class SpamBugModel(BugModel):
     def get_labels(self):
         classes = {}
 
-        for bug_data in bugzilla.get_bugs(include_invalid=True):
+        for bug_data in bugzilla.get_bugs(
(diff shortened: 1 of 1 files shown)


marco-c (approved):
Thanks!

PR #4454, opened 2024-08-29
[inline_comments_data_collection] Run the script on Taskcluster

benjaminmah:
Makes necessary changes to register and upload dataset when run on Taskcluster, as well as adding it to the pipeline.

### DIFF
--- a/infra/data-pipeline.yml
+++ b/infra/data-pipeline.yml
@@ -234,6 +234,45 @@ tasks:
         owner: release-mgmt-analysis@mozilla.com
         source: ${repository}/raw/master/data-pipeline.yml
 
+    - ID: fixed-comments-retrieval
+      created: { $fromNow: "" }
+      deadline: { $fromNow: "2 days" }
+      expires: { $fromNow: "1 year" }
+      provisionerId: proj-bugbug
+      workerType: compute-small
+      dependencies:
+        - revisions-retrieval
+      payload:
+        env:
+          TC_SECRET_ID: project/bugbug/production
+        maxRunTime: 86400
+        image: mozilla/bugbug-base:${version}
+        command:
+          - "bugbug-data-bugzilla"
+
+        artifacts:
+          public/bugs.json.zst:
+            path: /data/fixed_comments.json.zst
+            type: file
+          public/bugs.json.version:
+            path: /data/fixed_comments.json.version
+            type: file
+
+        features:
+          taskclusterProxy: true
+      scopes:
+        - "secrets:get:project/bugbug/production"
+      routes:
+        - notify.email.release-mgmt-analysis@mozilla.com.on-failed
+        - notify.irc-channel.#bugbug.on-failed
+        - index.project.bugbug.fixed_comments.${version}
+        - index.project.bugbug.fixed_comments.latest
+      metadata:
+        name: bugbug fixed comments retrieval
+        description: bugbug fixed comments retrieval
+        owner: release-mgmt-analysis@mozilla.com
+        source: ${repository}/raw/master/data-pipeline.yml
+
     - ID: issues-retrieval
       created: { $fromNow: "" }
       deadline: { $fromNow: "2 days" }
--- a/infra/data-pipeline.yml
+++ b/infra/data-pipeline.yml
@@ -248,7 +248,7 @@ tasks:
         maxRunTime: 86400
         image: mozilla/bugbug-base:${version}
         command:
-          - "bugbug-data-bugzilla"
+          - "bugbug-fixed-comments"
 
         artifacts:
           public/bugs.json.zst:
--- a/scripts/inline_comments_data_collection.py
+++ b/scripts/inline_comments_data_collection.py
@@ -6,6 +6,7 @@ import re
 import requests
 from libmozdata.phabricator import PhabricatorAPI
 
+from bugbug.db import register, upload
 from bugbug.tools.code_review import PhabricatorReviewData
 from bugbug.utils import get_secret
 
@@ -176,6 +177,9 @@ def process_comments(patch_threshold, diff_length_threshold):
                 }
                 yield data
 
+        if patch_threshold == 0:
+            continue
+
         patch_count += 1
         if patch_count >= patch_threshold:
             break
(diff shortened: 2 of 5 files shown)


scripts/inline_comments_data_collection.py
```diff
@@ -184,11 +188,24 @@ def process_comments(patch_threshold, diff_length_threshold):
...
     with open(dataset_file_path, "a") as dataset_file_handle:
         for data in process_comments(patch_threshold=1000, diff_length_threshold=5000):
             dataset_file_handle.write(json.dumps(data) + "\n")
 
+    register(
+        path=dataset_file_path,
+        url=dataset_url,
+        version=dataset_version,
+        support_files=[],
+    )
```
suhaibmujahid, line 205: The register function should be called in the data module; in this case, we could add `FIXED_COMMENTS_DB` in `phabricator.py`, similar to the following:

https://github.com/mozilla/bugbug/blob/7b29581e03fa630243bc984e26c89017f23647cb/bugbug/phabricator.py#L23-L28

scripts/inline_comments_data_collection.py
```diff
@@ -184,11 +188,24 @@ def process_comments(patch_threshold, diff_length_threshold):
...
+
+    upload(path=dataset_file_path)
```
suhaibmujahid, line 207: I do not think that you need to use `upload()`. TC will upload the files that you specify in `data-pipeline.yml`.

scripts/inline_comments_data_collection.py
benjaminmah: Done here: 2e0d6e0dff28ab74e41f49708f2a756168630b37

scripts/inline_comments_data_collection.py
benjaminmah: Removed here: 2e0d6e0dff28ab74e41f49708f2a756168630b37

infra/data-pipeline.yml
```diff
@@ -234,6 +234,45 @@ tasks:
...
+      workerType: compute-small
+      dependencies:
+        - revisions-retrieval
+      payload:
+        env:
+          TC_SECRET_ID: project/bugbug/production
+        maxRunTime: 86400
+        image: mozilla/bugbug-base:${version}
+        command:
+          - "bugbug-data-bugzilla"
```
suhaibmujahid, line 251: This is supposed to be the comment to run the script: https://github.com/mozilla/bugbug/blob/7b29581e03fa630243bc984e26c89017f23647cb/setup.py#L47

We need to add a command for the new script and use it here.

Also, we should add it to the integration test:

https://github.com/mozilla/bugbug/blob/7b29581e03fa630243bc984e26c89017f23647cb/scripts/integration_test.sh#L16-L19

scripts/inline_comments_data_collection.py
```diff
@@ -158,7 +159,7 @@ def process_comments(patch_threshold, diff_length_threshold):
                 logger.error(f"Failed to fetch diff: {e}")
                 continue
 
-            if len(patch_diff) > diff_length_threshold:
+            if len(patch_diff) > diff_length_threshold and diff_length_threshold != 0:
```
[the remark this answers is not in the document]
benjaminmah: I'll be sure to revert these changes. Regarding `diff_length_threshold` and `patch_threshold`, should I implement these as CL arguments rather than specifying their values in `main()`? @suhaibmujahid

scripts/inline_comments_data_collection.py
```diff
@@ -176,19 +177,24 @@ def process_comments(patch_threshold, diff_length_threshold):
...
     os.makedirs("patches", exist_ok=True)
-    os.makedirs("dataset", exist_ok=True)
-    dataset_file_path = "dataset/inline_comment_dataset2.json"
-    with open(dataset_file_path, "a") as dataset_file_handle:
-        for data in process_comments(patch_threshold=1000, diff_length_threshold=5000):
+    os.makedirs("data", exist_ok=True)
+
+    with open(phabricator.FIXED_COMMENTS_DB, "a") as dataset_file_handle:
+        for data in process_comments(patch_threshold=0, diff_length_threshold=0):
             dataset_file_handle.write(json.dumps(data) + "\n")
```
suhaibmujahid, line 191:
```suggestion
            dataset_file_handle.write(orjson.dumps(elem) + b"\n")
```

Note: you need to import `orjson` :)

infra/data-pipeline.yml
benjaminmah: Added the command here: 518f04a9dce3bb5f27923b123527c45d0aa37d3d
Added an integration test here: 76b5db720f69d5950bf81374ea5165cc75f87e2f

scripts/inline_comments_data_collection.py
```diff
@@ -176,19 +177,24 @@ def process_comments(patch_threshold, diff_length_threshold):
...
 
 def main():
     os.makedirs("patches", exist_ok=True)
-    os.makedirs("dataset", exist_ok=True)
-    dataset_file_path = "dataset/inline_comment_dataset2.json"
-    with open(dataset_file_path, "a") as dataset_file_handle:
-        for data in process_comments(patch_threshold=1000, diff_length_threshold=5000):
+    os.makedirs("data", exist_ok=True)
+
+    with open(phabricator.FIXED_COMMENTS_DB, "a") as dataset_file_handle:
```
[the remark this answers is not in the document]
benjaminmah: Done here: 67f101e11c53ec6b8fcf75a1cd4eb7e341a8d4cf

scripts/inline_comments_data_collection.py
benjaminmah: Done here: 67f101e11c53ec6b8fcf75a1cd4eb7e341a8d4cf

scripts/integration_test.sh
```diff
@@ -28,6 +28,10 @@ test -d ${CACHE_DIR:-cache}/mozilla-central
 ls -lh
 ls -lh data
 
+# Then generate a sample fixed comment dataset
+bugbug-fixed-comments
+ls -lh
+ls -lh data
```
suhaibmujahid, line 34: We should not add it to the integration test without a limit, so we may want to do it in a separate PR later. Could you please file an issue for this?

benjaminmah: Done here: https://github.com/mozilla/bugbug/issues/4466

benjaminmah: Removed the integration test here: 2ac50b95598cddac70d59015c4d41e2df3936b12

suhaibmujahid (approved):
Thank you!

PR #4502, opened 2024-09-22
Fixed TC issue with `inline_comments_data_collection.py`

benjaminmah:
The original script was missing `db.register()`, so it was unable to download the `revisions.json` dataset. 

It has been added to ensure that the dataset can be downloaded and used to populate the `fixed_comments.json` dataset.

### DIFF
--- a/scripts/inline_comments_data_collection.py
+++ b/scripts/inline_comments_data_collection.py
@@ -7,7 +7,7 @@ import orjson
 import requests
 from libmozdata.phabricator import PhabricatorAPI
 
-from bugbug import db, phabricator
+from bugbug import phabricator
 from bugbug.tools.code_review import PhabricatorReviewData
 from bugbug.utils import get_secret, zstd_compress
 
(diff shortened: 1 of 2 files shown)


scripts/inline_comments_data_collection.py
```diff
@@ -18,6 +18,12 @@
 
 api = PhabricatorAPI(get_secret("PHABRICATOR_TOKEN"))
 
+db.register(
+    phabricator.REVISIONS_DB,
+    "https://community-tc.services.mozilla.com/api/index/v1/task/project.bugbug.data_revisions.latest/artifacts/public/revisions.json.zst",
+    4,
+)
+
```
suhaibmujahid, line 26: It is already registered:

https://github.com/mozilla/bugbug/blob/c98b7078911f442a99c22407bcd4ffc9beaa59e8/bugbug/phabricator.py#L30-L35

However, there is a bug in the registration; it should register `FIXED_COMMENTS_DB` instead of `REVISIONS_DB.`

suhaibmujahid (changes requested):

scripts/inline_comments_data_collection.py
benjaminmah: I see, thank you for pointing that out! I have fixed it here: 5542e679a8634432ff3d5a594a67d50b9317def9.

marco-c (approved):

PR #4507, opened 2024-09-23
Phabricator API request removal for transactions

benjaminmah:
Instead of making a request to the Phabricator API for the transactions for each revision, we load it from `revisions.json`. 

An index file has also been created to store the offset of each revision in the file to avoid searching through all revisions for each patch.

### DIFF
--- a/scripts/inline_comments_data_collection.py
+++ b/scripts/inline_comments_data_collection.py
@@ -114,8 +114,8 @@ def process_comments(limit, diff_length_limit):
                     )
                 fix_patch_id = diffs[0]["id"]
 
-            # If the most recent patch doesn't exist or is the original patch itself, skip it
-            if not fix_patch_id or fix_patch_id == patch_id:
+            # If the most recent patch is the original patch itself, skip it
+            if fix_patch_id == patch_id:
                 continue
 
             revision_phid = revision_info["phid"]
--- a/scripts/inline_comments_data_collection.py
+++ b/scripts/inline_comments_data_collection.py
@@ -47,16 +47,43 @@ def find_revision_from_patch(patch_id):
     return revision_phid
 
 
-def find_transactions_from_patch(patch_id):
-    revision_phid = find_revision_from_patch(patch_id)
-    transactions = api.request("transaction.search", objectIdentifier=revision_phid)[
-        "data"
-    ]
+def create_index(file_path, index_path):
+    index = {}
+    offset = 0
 
-    if not transactions:
+    with open(file_path, "rb") as file:
+        for line in file:
+            try:
+                entry = orjson.loads(line)
+                phid = entry.get("phid")
+                if phid:
+                    index[phid] = offset
+            except orjson.JSONDecodeError:
+                print(f"Error decoding JSON: {line}")
+            offset = file.tell()
+
+    with open(index_path, "wb") as index_file:
+        index_file.write(orjson.dumps(index))
+
+
+def find_transactions_from_patch(patch_id, revisions_index):
+    revision_phid = find_revision_from_patch(patch_id)
+    offset = revisions_index.get(revision_phid)
+
+    if offset is None:
         raise NoTransactionsFoundException(patch_id)
 
-    return transactions
+    with open(phabricator.REVISIONS_DB, "rb") as file:
+        file.seek(offset)
+        line = file.readline()
+
+        try:
+            entry = orjson.loads(line)
+            return entry.get("transactions")
+        except orjson.JSONDecodeError:
+            logger.error(f"Failed to decode JSON {line}")
+
+    raise NoTransactionsFoundException(patch_id)
 
 
 def get_diff_info_from_phid(phid):
@@ -119,11 +146,11 @@ def extract_relevant_diff(patch_diff, filename):
         return None
 
 
-def process_comments(limit, diff_length_limit):
+def process_comments(limit, diff_length_limit, revisions_index):
     patch_count = 0
 
     for patch_id, comments in review_data.get_all_inline_comments(lambda c: True):
-        transactions = find_transactions_from_patch(patch_id)
+        transactions = find_transactions_from_patch(patch_id, revisions_index)
 
         resolved_comments = [comment for comment in comments if comment.is_done]
 
@@ -206,8 +233,17 @@ def main():
     os.makedirs("patches", exist_ok=True)
     os.makedirs("data", exist_ok=True)
 
+    create_index(phabricator.REVISIONS_DB, "data/revisions_index.json")
+
+    with open("data/revisions_index.json", "rb") as index_file:
+        revisions_index = orjson.loads(index_file.read())
+
     with open(phabricator.FIXED_COMMENTS_DB, "wb") as dataset_file_handle:
-        for data in process_comments(limit=limit, diff_length_limit=diff_length_limit):
+        for data in process_comments(
+            limit=limit,
+            diff_length_limit=diff_length_limit,
+            revisions_index=revisions_index,
+        ):
             dataset_file_handle.write(orjson.dumps(data) + b"\n")
 
     zstd_compress(phabricator.FIXED_COMMENTS_DB)
--- a/scripts/inline_comments_data_collection.py
+++ b/scripts/inline_comments_data_collection.py
@@ -37,73 +37,27 @@ class NoDiffFoundForPHIDException(Exception):
         self.phid = phid
 
 
-def find_revision_from_patch(patch_id):
-    diffs = api.search_diffs(diff_id=patch_id)
+def load_revisions_maps():
+    diff_id_to_revision = {}
+    diff_phid_to_revision = {}
 
-    if not diffs:
-        raise NoDiffsFoundException(patch_id)
+    for revision in phabricator.get_revisions():
+        for transaction in revision["transactions"]:
+            if transaction.get("fields", {}).get("diff") is None:
+                continue
 
-    revision_phid = diffs[0]["revisionPHID"]
-    return revision_phid
+            diff_id_to_revision[transaction["fields"]["diff"]["id"]] = revision
+            diff_phid_to_revision[transaction["fields"]["diff"]["phid"]] = (
+                transaction["fields"]["diff"]["id"],
+                revision,
+            )
 
-
-def load_revisions_map(file_path):
-    revisions_map = {}
-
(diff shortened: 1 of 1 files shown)


scripts/inline_comments_data_collection.py
```diff
@@ -47,16 +47,43 @@ def find_revision_from_patch(patch_id):
...
+    with open(file_path, "rb") as file:
+        for line in file:
+            try:
+                entry = orjson.loads(line)
+                phid = entry.get("phid")
+                if phid:
+                    index[phid] = offset
+            except orjson.JSONDecodeError:
+                print(f"Error decoding JSON: {line}")
+            offset = file.tell()
```
marco-c, line 63: This is a bit brittle because the order of the revisions could change in the revisions.json file, rendering the index file no longer usable.

It's also an optimization that maybe is not useful, given the memory requirements are not so high. You could just read the entire revisions.json file in memory into a map.

Another option could be to change `get_all_inline_comments' to also return the revision in addition to the diff_id and comments, and then you could get the transactions from this revision object.
You'd simply need to change https://github.com/mozilla/bugbug/blob/dba47657d00fffdb185e0d69c980e5800b75e785/bugbug/tools/code_review.py#L587 to `yield revision, diff_id, comments` (and do th

scripts/inline_comments_data_collection.py
```diff
@@ -47,16 +47,43 @@ def find_revision_from_patch(patch_id):
     return revision_phid
```
marco-c, line 47: We should also be able to avoid the `api.search_diffs` calls

scripts/inline_comments_data_collection.py
benjaminmah: Implemented the first suggestion here (using a map): a5d1320ee2ba99589f56b03f060bb6d2d94fbe2b

scripts/inline_comments_data_collection.py
benjaminmah: Not 100% sure if the revisions object contains all the necessary information to replace `api.search_diffs()`, but will look into it!

marco-c: You're using it to get the `revisionPHID`, which should be in the revision object (maybe called just PHID?)

scripts/inline_comments_data_collection.py
```diff
@@ -136,18 +107,25 @@ def process_comments(limit, diff_length_limit):
             if not most_recent_update:
                 continue
 
-            fix_patch_id, revision_phid = get_diff_info_from_phid(
-                most_recent_update["fields"]["new"]
-            )
+            fix_patch_id = diff_phid_to_revisions_map.get(
+                most_recent_update["fields"].get("new"), [None]
```
marco-c, line 111: In which case could we have no diff ID?

marco-c: Given we're getting `most_recent_update` from the revisions that are in the DB, shouldn't we always find it?

scripts/inline_comments_data_collection.py
```diff
@@ -136,18 +107,25 @@ def process_comments(limit, diff_length_limit):
...
+            # If the  most recent patch doesn't exist or is the original patch itself, skip it
+            if not fix_patch_id or fix_patch_id == patch_id:
                 continue
 
-            revision_id, bug_id = find_details_from_revision_phid(phid=revision_phid)
+            revision_info = diff_phid_to_revisions_map.get(
+                most_recent_update["fields"].get("new"), [None, {}]
+            )[1]
+            revision_phid = revision_info.get("phid")
+            revision_id = revision_info.get("id")
```
marco-c, line 122: `revision_info` could just be `diff_id_to_revisions_map[patch_id]`, if you add it at line 97 then you could do:
```Python
revision_info = diff_id_to_revisions_map[patch_id]
transactions = revision_info["transactions"]
```

marco-c: And if you do this, you could change `diff_phid_to_revisions_map` to only store the ID, renaming it to `diff_phid_to_id`:
```Python
diff_phid_to_revision[transaction["fields"]["diff"]["phid"]] = transaction["fields"]["diff"]["id"]
```

scripts/inline_comments_data_collection.py
benjaminmah: That would make sense, but the mapping does not contain all diff ids/phids to revisions and I encounter `KeyError` when running it locally if I do not have the `.get()`. Will look into this.

scripts/inline_comments_data_collection.py
benjaminmah: Changed here: 255c1d6e24c63fa71cde29861300551db5b89f70. Thanks for pointing that out!

scripts/inline_comments_data_collection.py
```diff
@@ -136,18 +107,18 @@ def process_comments(limit, diff_length_limit):
...
                 continue
 
-            fix_patch_id, revision_phid = get_diff_info_from_phid(
-                most_recent_update["fields"]["new"]
-            )
+            fix_patch_id = diff_phid_to_id.get(most_recent_update["fields"].get("new"))
 
-            # If the most recent patch is the original patch itself, skip it
-            if fix_patch_id == patch_id:
+            # If the  most recent patch doesn't exist or is the original patch itself, skip it
```
marco-c, line 112:
```suggestion
            # If the most recent patch doesn't exist or is the original patch itself, skip it
```

scripts/inline_comments_data_collection.py
```diff
@@ -37,57 +34,28 @@ def __init__(self, phid):
...
-        raise NoDiffFoundForPHIDException(phid)
-    return diffs[0]["id"], diffs[0]["revisionPHID"]
+    return diff_id_to_revision, diff_phid_to_id
 
 
-def find_details_from_revision_phid(phid):
-    revision = api.load_revision(rev_phid=phid)
+def find_details_from_revision_phid(phid, revisions_map):
+    revision = revisions_map[phid]
     return revision["id"], revision["fields"]["bugzilla.bug-id"]
```
marco-c, line 56: This function is no longer used, you could remove it

scripts/inline_comments_data_collection.py
```diff
@@ -136,18 +107,18 @@ def process_comments(limit, diff_length_limit):
...
 
-            # If the most recent patch is the original patch itself, skip it
-            if fix_patch_id == patch_id:
+            # If the  most recent patch doesn't exist or is the original patch itself, skip it
+            if not fix_patch_id or fix_patch_id == patch_id:
                 continue
 
-            revision_id, bug_id = find_details_from_revision_phid(phid=revision_phid)
+            revision_phid = revision_info.get("phid")
+            revision_id = revision_info.get("id")
```
marco-c, line 117: For the keys that are always available, don't use `get`. This will make it easier to prevent errors in the future (e.g. if the key changes name, we won't start silently failing but we will see an exception).

scripts/inline_comments_data_collection.py
```diff
@@ -206,8 +177,15 @@ def main():
...
 
+    diff_id_to_revisions_map, diff_phid_to_id = load_revisions_maps()
+
     with open(phabricator.FIXED_COMMENTS_DB, "wb") as dataset_file_handle:
-        for data in process_comments(limit=limit, diff_length_limit=diff_length_limit):
+        for data in process_comments(
+            limit=limit,
+            diff_length_limit=diff_length_limit,
+            diff_id_to_revisions_map=diff_id_to_revisions_map,
+            diff_phid_to_id=diff_phid_to_id,
```
marco-c, line 187: nit: given the maps are only used in `process_comments`, you could create them in that function

scripts/inline_comments_data_collection.py
marco-c: Try finding a specific example of such a case and then it'll be easier to investigate and find why it's missing

scripts/inline_comments_data_collection.py
benjaminmah: Fixed here: 931333c920ee853a98be1e0a2f7bd9d193a31d37

scripts/inline_comments_data_collection.py
benjaminmah: Removed here: 931333c920ee853a98be1e0a2f7bd9d193a31d37

scripts/inline_comments_data_collection.py
benjaminmah: Makes sense, replaced `.get()` with direct accesses: 931333c920ee853a98be1e0a2f7bd9d193a31d37

scripts/inline_comments_data_collection.py
benjaminmah: Done here: 931333c920ee853a98be1e0a2f7bd9d193a31d37

scripts/inline_comments_data_collection.py
benjaminmah: I think it's because in the transaction of type `udpate`, it contains the `diffPHID` for the old and new patches, but not the `diffID`. So unless there is a dataset that contains all diffs with their `diffPHID` and `diffID`, then I'm not sure the map will contain all `diffID` for all `diffPHID` (i.e. a mapping/conversion between `diffPHID` and `diffID`.

I've tried populating the map with the `diffPHID` and `diffID` of the revision itself (`revision["fields"]["diffPHID"]` and `revision["fields"]["diffID"]`) but still encountered the same `KeyError`.

scripts/inline_comments_data_collection.py
```diff
@@ -136,21 +104,33 @@ def process_comments(limit, diff_length_limit):
...
+            except KeyError:
+                diffs = api.search_diffs(diff_phid=most_recent_update["fields"]["new"])
+                if not diffs:
+                    raise NoDiffFoundForPHIDException(
+                        most_recent_update["fields"]["new"]
+                    )
+                fix_patch_id = diffs[0]["id"]
+
+            # If the most recent patch doesn't exist or is the original patch itself, skip it
+            if not fix_patch_id or fix_patch_id == patch_id:
```
marco-c, line 118: Now `not fix_patch_id` is no longer needed, right?

marco-c (dismissed):

scripts/inline_comments_data_collection.py
benjaminmah: Correct, removed here: 8ac58d6720b832afff9cba22242867231a80a3eb

marco-c (approved):

PR #4537, opened 2024-10-10
Initial LLM-powered code generation script

benjaminmah:
Preliminary code generation script using models from OpenAI and Anthropic. Uses a local Qdrant vector DB for data point storage for similar commits to be injected into the single-shot and multi-shot prompts.


bugbug/tools/code_review_generation.py
```diff
@@ -0,0 +1,754 @@
+import csv
+import json
+import logging
+import re
+
+import anthropic
+import openai
```
marco-c, line 7: We shouldn't use these directly, but rely on the generic functions and classes from bugbug/generative_model_tool.py, similarly to code review tool

bugbug/tools/code_review_generation.py
```diff
@@ -0,0 +1,754 @@
...
+        metrics = calculate_metrics(reference_fix, generated_fix)
+        return metrics
+    else:
+        print(
+            f"No matching fix found in the dataset for Revision {revision_id} and Patch {initial_patch_id}."
+        )
+        return None
+
+
+def main():
```
marco-c, line 716: The CLI script should be separate and moved in the scripts folder (see for example scripts/code_review_tool_runner.py)

bugbug/tools/code_review_generation.py
```diff
@@ -0,0 +1,754 @@
...
+        url = f"https://phabricator.services.mozilla.com/D{revision_id}?id={patch_id}&download=true"
+        response = requests.get(url)
+        response.raise_for_status()
+        return response.text
+    except requests.HTTPError as e:
+        logger.error(f"HTTP error fetching diff: {e}")
+        return None
+    except Exception as e:
+        logger.error(f"Unexpected error: {e}")
+        return None
```
marco-c, line 218: For the functions that retrieve data from Phab, we should move them to the phabricator.py lib (and use already existing ones in case there are)

bugbug/tools/code_review_generation.py
```diff
@@ -0,0 +1,754 @@
...
+                                            relevant_diff,
+                                            generated_fix,
+                                            reference_fix,
+                                        ]
+                                    )
+
+                                counter += 1
+
+                    else:
+                        print(f"No relevant diff found for Comment ID {comment.id}.\n")
```
marco-c, line 659: The evaluation code should also be moved into a separate script

bugbug/tools/code_review_generation.py
benjaminmah: Good point, now relying on `generative_model_tool.py` here: 3234648f00ddc85612c8a8ce94f805814a96e724

benjaminmah: The qualitative feedback prompt is still there, but will move it to a separate script when moving all evaluation code.

bugbug/tools/code_review_generation.py
benjaminmah: Separated here: 8873dacba5332c58db1e7728784b68f76cd33a04

bugbug/tools/code_review_generation.py
benjaminmah: Moved it to `scripts/code_review_generation_runner.py`: 12b0d7cc24ade49f718b10994a91ede5025b55e0

Will move the evaluation script execution to a separate script as well!

benjaminmah: Moved the evaluation script to `scripts/code_review_generation_evaluator_runner.py`: 1ddbbdc0ef62c6a4bcecbe7bbf39947f7d82ae75

bugbug/phabricator.py
```diff
@@ -291,3 +292,17 @@ def get_pending_review_time(rev: RevisionDict) -> timedelta | None:
...
+        url = f"https://phabricator.services.mozilla.com/D{revision_id}?id={patch_id}&download=true"
+        response = requests.get(url)
+        response.raise_for_status()
+        return response.text
+    except requests.HTTPError as e:
+        logger.error(f"HTTP error fetching diff: {e}")
+        return None
+    except Exception as e:
+        logger.error(f"Unexpected error: {e}")
+        return None
```
suhaibmujahid, line 308: We already have a function to do this. We may want to consider reusing it: https://github.com/mozilla/bugbug/blob/6134ff622c1ee46f323439f77a08a18726912741/scripts/inline_comments_data_collection.py#L69-L75

bugbug/tools/code_review_generation_evaluator.py
suhaibmujahid, line 1: The evaluation script should live in the `scripts` directory.

bugbug/phabricator.py
benjaminmah: Done here: a4c1b5deb73e2ea5b1753118034bc764c88804c1

bugbug/phabricator.py
```diff
@@ -291,3 +292,18 @@ def get_pending_review_time(rev: RevisionDict) -> timedelta | None:
...
+):
+    if single_patch:
+        url = f"https://phabricator.services.mozilla.com/D{revision_id}?id={first_patch}&download=true"
+    else:
+        url = f"https://phabricator.services.mozilla.com/D{revision_id}?vs={first_patch}&id={second_patch}&download=true"
+
+    response = requests.get(url)
+    if response.status_code == 200:
+        return response.text
+    else:
```
suhaibmujahid, line 308: Using `else` is not necessary here. I would flip it and make the exception conditional. 

Note: In general, it is better to keep the regular code flow more to the lift and the irregular logic to the right.

bugbug/phabricator.py
```diff
@@ -291,3 +292,18 @@ def get_pending_review_time(rev: RevisionDict) -> timedelta | None:
...
+        raise Exception(f"Failed to download diff from URL: {url}")
```
suhaibmujahid, line 309: Generic errors like `Exception` should be avoided.

benjaminmah: c72504df8d3adc89c3b0824f6efbb075a2a92ce7

bugbug/phabricator.py
benjaminmah: c72504df8d3adc89c3b0824f6efbb075a2a92ce7

suhaibmujahid:
@benjaminmah could you please resolve the conflict?

suhaibmujahid (approved):
LGTM! Thank you!

suhaibmujahid (approved):

PR #4538, opened 2024-10-11
Misc improvements to code review tool

marco-c:
Each commit is an isolated change.

### DIFF
--- a/bugbug/tools/code_review.py
+++ b/bugbug/tools/code_review.py
@@ -1018,6 +1018,7 @@ class CodeReviewTool(GenerativeModelTool):
         function_search: Optional[FunctionSearch] = None,
         review_comments_db: Optional["ReviewCommentsDB"] = None,
         show_patch_example: bool = False,
+        verbose: bool = True,
     ) -> None:
         super().__init__()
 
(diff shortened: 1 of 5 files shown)


scripts/code_review_tool_evaluator.py
```diff
@@ -72,7 +72,7 @@ def get_file(commit_hash, path):
             (
                 "RAG",
                 code_review.CodeReviewTool(
-                    llm=llm,
+                    comment_gen_llms=[llm],
```
suhaibmujahid, line 75: Why do we pass the llm as a list here?

bugbug/tools/code_review.py
```diff
@@ -1074,11 +1082,17 @@ def run(self, patch: Patch) -> list[InlineComment] | None:
...
+        if self.verbose:
+            print(f"\u001b[33;1m\033[1;3m{output_summarization}\u001b[0m")
+
         if self.function_search is not None:
             line_code_list = self.further_context_chain.run(
                 patch=formatted_patch, summarization=output_summarization
             ).split("\n")
 
+            if self.verbose:
+                print(f"\u001b[33;1m\033[1;3m{line_code_list}\u001b[0m")
```
suhaibmujahid, line 1094: We may rap the print formatting in a function to make it more readable.

suhaibmujahid (approved):
LGTM!

scripts/code_review_tool_evaluator.py
marco-c: The idea would be to support an ensemble of LLMs (either the same model with different temperature and top_p, or different models) to generate diverse comments, and then deduplicate the comments.

suhaibmujahid: We may have such implication details in a class that uses the LLM interface. So this class doesn't need to know how the comments were generated.

bugbug/tools/code_review.py
marco-c: Done!

scripts/code_review_tool_evaluator.py
marco-c: Good idea, filed #4539 as a follow-up to do this.

PR #4569, opened 2024-10-22
Update libmozdata and use a User-Agent for all requests

marco-c:


bugbug/utils.py
```diff
@@ -495,10 +505,24 @@ def get_session(name: str) -> requests.Session:
     return session
 
 
+def get_user_agent():
+    return get_secret("USER_AGENT", "bugbug")
```
suhaibmujahid, line 509: Why do we need the user agent to be retrieved from a secret?

marco-c: Just so we can easily configure it in case we want to change it, instead of having to deploy a new version with the code change.

PR #4617, opened 2024-11-13
Build failure dataset script

benjaminmah:
Script to create the build failure dataset. Collects the Revision ID, Initial and Fix Commit Hash, Interdiff and Error Logs.

### DIFF
--- a/scripts/build_failure_data_collection.py
+++ b/scripts/build_failure_data_collection.py
@@ -219,10 +219,6 @@ def main():
         if revision_id in backed_out_revisions:
             revisions_to_commits[revision_id].append(commit["node"])
 
-    print(
-        f"NUMBER OF REVISIONS MAPPED TO BACKED OUT COMMITS: {len(revisions_to_commits)}"
-    )
-
     # 5. and 6.
 
     client_id = os.getenv("TC_CLIENT_ID")


scripts/build_failure_data_collection.py
```diff
@@ -0,0 +1,294 @@
...
+            writer.writerow(
+                [revision_id, commits[0], commits[1], commit_diff_encoded, error_lines]
+            )
+
+
+if __name__ == "__main__":
+    main()
+
+# 0. Download databases
+# 1. Identify bugs in Bugzilla that have a backout due to build failures X
```
jmaher, line 289: I normally get my regression (backed out) data from treeherder database.  The main reason why is we can get all the tasks that failed, which could help determine if this is failing ALL builds, or a certain platform or build failure.  Not required, but the additional information is good to know.

scripts/build_failure_data_collection.py
```diff
@@ -0,0 +1,294 @@
...
+    ):
+        if commit["bug_id"] not in bug_ids:
+            continue
+
+        commit_data = {
+            key: commit[key]
+            for key in ["node", "bug_id", "pushdate", "backedoutby", "backsout", "desc"]
+        }
+
+        bug_commits.setdefault(commit["bug_id"], []).append(commit_data)
```
jmaher, line 83: how will you make this work when there are multiple landings and backouts?  In general do you care about each transaction, or the end state?  Sometimes the end state is a secondary patch that hot fixes the situation, or a rebase which changes other files/code blocks not originally edited.

I think any solution should account for >1 backout.

scripts/build_failure_data_collection.py
```diff
@@ -0,0 +1,294 @@
...
+        if not commit["backsout"]:
+            continue
+
+        desc = commit["desc"]
+        if (
+            "backed out" in desc.lower()
+            and "for causing" in desc.lower()
+            and "build" in desc.lower()
+        ):
+            return commit
```
jmaher, line 153: in this case why do you need bugzilla data?

scripts/build_failure_data_collection.py
```diff
@@ -0,0 +1,294 @@
...
+    backed_out_revisions = []
+
+    for bug_id in bug_ids:
+        bug_id_commits = bug_commits.get(bug_id, None)
+        backing_out_commit = find_backing_out_commit(bug_id_commits, hg_client)
+
+        if not backing_out_commit:
+            continue
+
+        logger.info("Backing out commit found!")
```
jmaher, line 99: it would be nice to mention the commit id :)

scripts/build_failure_data_collection.py
```diff
@@ -0,0 +1,294 @@
...
+
+    # 5. get failed tasks
+    failed_tasks = set()
+
+    for task in queue_client.listTaskGroup(task_group_id)["tasks"]:
+        if task["status"]["state"] == "failed":
+            failed_tasks.add(task["status"]["taskId"])
+
+    # 6. find intersection between build tasks and failed tasks
+    failed_build_tasks = list(build_tasks & failed_tasks)
```
jmaher, line 192: if you use treeherder db (search for fixed_by_commit) instead of bugzilla to find backouts you get most of this stuff for free.

scripts/build_failure_data_collection.py
```diff
@@ -0,0 +1,294 @@
...
+                error_lines = find_error_lines(index, queue, commit)
+
+                if error_lines:
+                    break
+
+            # if not error_lines:
+            #     continue
+
+            commit_diff = repository.get_diff(
+                repo_path="hg_dir", original_hash=commits[0], fix_hash=commits[1]
```
jmaher, line 273: as many fixed revisions do not have the original code referenced, this might have more edge cases than you think.  Also for multiple cycles, it is probably best to compare commits[0] <-> commits[-1]

jmaher (commented):
I like this "in general".  Some specifics.  I think collecting this data and storing it longer term is good.  Can you store in .json artifacts- then we could ingest in the future?  Maybe fixed_by_commit revision or bugid as the filename.  Then these exist and can be ingested overtime.

The risk is that there are multiple backouts for a given bugid/original_changeset.

It would be a bonus if we could collect test regressions as well- it is ok if these are in a separate location/db/filesystem/etc.  The primary use case I can see is to determine if we are finding regressions in code we ship, or just forgetting or dealing with old test cases.  This is pretty easy by looking at the interdiff and determining which files were changed

scripts/build_failure_data_collection.py
benjaminmah: I've included the commit information when identifying the backing out commit: 460e66994e6ae0ea57b279ff8d5f6736aa2fba94

scripts/build_failure_data_collection.py
jmaher: I thought commits via hg had metadata:
https://hg.mozilla.org/mozilla-central/rev/99a8f2b2b00d85148c743f16db75d9abefb33513

if you want a short list of commits that are backed out, then bugzilla comment parsing would do.

One value add method would be to propose a change to the backout process where there is queryable metadata so everytime there is a backout, data is stored in bugzilla/treeherder/some_random_db with original commit, backed out commit, related links, list of failing tasks and related error messages.  Thinking more, developers relanding would have to go through a process to document what they were relanding, and confirm the interdiff :)

jmaher: my original comment here was assuming you were doing a massive hg log to look at commit history and parse out "backed out" commit messages to build a list.

scripts/build_failure_data_collection.py
benjaminmah: Sorry for the late response! But the current version of the script skips any revision with multiple commits. Despite skipping these cases, there are still 1000+ data points within the last 2 years.

marco-c:
This is covered by #5396.

PR #4659, opened 2024-12-04
[review_comments_retriever] Avoid populating database from scratch

### LINKED ISSUE #4644: [review_comments_retriever] Avoid starting from scratch on each run
Currently, each time we run the script, it starts retrieving comments from scratch. Ideally, the script should start from where it left off in the previous run.

https://github.com/mozilla/bugbug/blob/0d14e5fee1e79d71b9eaed23eea922f1273680b1/scripts/review_comments_retriever.py#L16-L17

benjaminmah:
Resolves #4644.

Introduces a set containing the existing comment IDs that is initialized before populating the DB to avoid unnecessarily repopulating the DB with duplicates.


bugbug/vectordb.py
```diff
@@ -46,6 +46,10 @@ def insert(self, points: Iterable[VectorPoint]):
     def search(self, query: list[float]) -> Iterable[PayloadScore]:
         ...
 
+    @abstractmethod
+    def get_existing_ids(self):
```
suhaibmujahid, line 50: As an abstract method, it is important to specify the type of the returned value.

bugbug/vectordb.py
```diff
@@ -83,3 +87,25 @@ def insert(self, points: Iterable[VectorPoint]):
...
+        while True:
+            points, next_page_offset = self.client.scroll(
+                collection_name=self.collection_name,
+                limit=100,
+                with_payload=False,
+                with_vectors=False,
+                offset=offset,
+            )
+
+            point_ids.update(point.id for point in points)
```
suhaibmujahid, line 104: Creating a generator would be a cleaner way to implement this.

suhaibmujahid (changes requested):

bugbug/vectordb.py
benjaminmah: I've changed it to be a generator: 90dd559f791b678ae12d79b5b12073cfbbc5453e. However, in `bugbug/tools/code_review.py` I still convert it to a set, as I use it to check for existing ids in O(1) time. Is there a better way we can make use of the generator and keep the same efficiency?

bugbug/vectordb.py
benjaminmah: Done here: 90dd559f791b678ae12d79b5b12073cfbbc5453e

bugbug/tools/code_review.py
```diff
@@ -1285,8 +1285,15 @@ def clean_comment(self, comment):
         return comment
 
     def add_comments_by_hunk(self, items: Iterable[tuple[Hunk, InlineComment]]):
+        point_ids = set(self.vector_db.get_existing_ids())
+
         def vector_points():
+            nonlocal point_ids
+
```
suhaibmujahid, line 1292: Since we are not assigning a value to the `point_ids`  variable here, we need to use `nonlocal`.

bugbug/vectordb.py
```diff
@@ -46,6 +46,10 @@ def insert(self, points: Iterable[VectorPoint]):
     def search(self, query: list[float]) -> Iterable[PayloadScore]:
         ...
 
+    @abstractmethod
+    def get_existing_ids(self) -> Generator[int, None, None]:
```
suhaibmujahid, line 50: Using `Iterable`, would be a better fit here.

bugbug/vectordb.py
```diff
@@ -83,3 +87,23 @@ def insert(self, points: Iterable[VectorPoint]):
     def search(self, query: list[float]) -> Iterable[PayloadScore]:
         for item in self.client.search(self.collection_name, query):
             yield PayloadScore(item.score, item.id, item.payload)
+
+    def get_existing_ids(self) -> Generator[int, None, None]:
+        offset = None
+
+        while True:
```
suhaibmujahid, line 94: It would be cleaner to use the value of `offset` to control the loop.

Example: https://github.com/mozilla/bugbug/blob/8e49ad958e9846d9c4c771ba7077924185a7800f/bugbug/vectordb.py#L127-L135

suhaibmujahid (changes requested):

suhaibmujahid:
@benjaminmah, could you please resolve the merge conflict?

bugbug/vectordb.py
benjaminmah: Changed here: 90dd559f791b678ae12d79b5b12073cfbbc5453e

bugbug/vectordb.py
benjaminmah: 90dd559f791b678ae12d79b5b12073cfbbc5453e

bugbug/tools/code_review.py
benjaminmah: 90dd559f791b678ae12d79b5b12073cfbbc5453e

suhaibmujahid: sorry, "we do not need to use nonlocal"

bugbug/vectordb.py
suhaibmujahid: The mentioned commit shows that the type hint is `Generator`, not `Iterable`.

bugbug/vectordb.py
suhaibmujahid: I'm pointing here to line 129: "while offset is not None:"

bugbug/vectordb.py
benjaminmah: I forgot to push my changes! I meant to link this commit: 7d77f7bd36d34662dc6f0daeb000f6116c1c2ea4

bugbug/vectordb.py
benjaminmah: I forgot to push my changes! I meant to link this commit: 7d77f7bd36d34662dc6f0daeb000f6116c1c2ea4

suhaibmujahid (dismissed):
LGTM! Thank you!

suhaibmujahid (dismissed):

suhaibmujahid (approved):

PR #4733, opened 2025-01-09
[code_review] Support vector db for feedback on generated comments

### LINKED ISSUE #4578: [code_review] Build “Accepted/rejected comments DB” with functionality to search by similarity
This will be needed, for example, to add examples of rejected comments in the prompt to avoid similar comments being generated.

suhaibmujahid:
Resolves #4578

### DIFF
--- a/bugbug/tools/code_review.py
+++ b/bugbug/tools/code_review.py
@@ -3,6 +3,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
+import enum
 import json
 import re
 from abc import ABC, abstractmethod
@@ -11,7 +12,7 @@ from dataclasses import asdict, dataclass
 from datetime import datetime
 from functools import cached_property
 from logging import INFO, basicConfig, getLogger
-from typing import Iterable, Optional
+from typing import Iterable, Literal, Optional
 
 import tenacity
 from langchain.chains import ConversationChain, LLMChain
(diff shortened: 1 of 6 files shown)


bugbug/tools/code_review.py
```diff
@@ -1322,3 +1323,53 @@ def find_similar_patch_comments(self, patch: Patch, limit: int):
...
+    file_path: str
+    action: Literal["APPROVE", "REJECT", "IGNORE"]
+    user: str
+
+
+class SuggestionsFeedbackDB:
+    def __init__(self, vector_db: VectorDB) -> None:
+        self.vector_db = vector_db
+        self.embeddings = OpenAIEmbeddings(
+            model="text-embedding-3-large", api_key=get_secret("OPENAI_API_KEY")
```
marco-c, line 1347: We should make this configurable, or at least define it in a constant

functions/sync-review-comments-db/database.py
```diff
@@ -0,0 +1,42 @@
...
+
+import json
+import os
+
+import pg8000
+import sqlalchemy
+from google.cloud.sql.connector import Connector, IPTypes
+
+
+def init_connection_poole_engine() -> sqlalchemy.engine.base.Engine:
```
marco-c, line 14:
```suggestion
def init_connection_pool_engine() -> sqlalchemy.engine.base.Engine:
```

functions/sync-review-comments-db/models.py
```diff
@@ -0,0 +1,189 @@
...
+    )
+
+    # TODO: Fill the columns for old suggestions, then change the column
+    # to NOT NULL.
+    llm_name: Mapped[Optional[str]]
+    llm_temperature: Mapped[Optional[float]]
+
+    # This will be filled once the comment is posted to Phabricator. It will
+    # store the inline comment ID returned by the Phabricator Conduit API.
+    inline_comment_id: Mapped[Optional[int]]
```
marco-c, line 140: Why do we need this?

functions/sync-review-comments-db/models.py
```diff
@@ -0,0 +1,189 @@
...
+    has_added_lines: Mapped[bool]
+    content: Mapped[str]
+    evaluation: Mapped[Optional["Evaluation"]] = relationship(
+        back_populates="suggestion"
+    )
+
+    # TODO: Fill the columns for old suggestions, then change the column
+    # to NOT NULL.
+    llm_name: Mapped[Optional[str]]
+    llm_temperature: Mapped[Optional[float]]
```
marco-c, line 136: There are also other parameters, like top_p, and there might be others in the future. How do we handle that?

functions/sync-review-comments-db/models.py
```diff
@@ -0,0 +1,189 @@
...
+
+    # pylint:disable=not-callable
+    created_at: Mapped[datetime] = mapped_column(server_default=func.now())
+    updated_at: Mapped[datetime] = mapped_column(
+        server_default=func.now(), server_onupdate=func.now()
+    )
+
+    @property
+    def is_recently_created(self):
+        return (datetime.utcnow() - self.created_at).total_seconds() < 240
```
marco-c, line 83: What is this going to be used for?

bugbug/tools/code_review.py
suhaibmujahid: We could use dependency injection for that. We could pass the embedding class to the contractor, allowing more flexibility in selecting the provider, not just the model.

functions/sync-review-comments-db/models.py
suhaibmujahid: Review-Helper's backend uses it.

functions/sync-review-comments-db/models.py
suhaibmujahid: We can add more columns. Alternatively, we can save the metadata as JSON, but it will be less convenient to query it later.

functions/sync-review-comments-db/models.py
suhaibmujahid: Review-Helper's backend uses it as one of the parameters to decide if it will re-process the patch.

functions/sync-review-comments-db/models.py
marco-c: The issue I see is in case we add a new parameter, we should make sure to include it here. Can you think of a way to assert that all options are captured?

marco-c (approved):

functions/sync-review-comments-db/models.py
suhaibmujahid: We have the BugBug version and the tool variant, which is a string that describes a group of parameters:

https://github.com/mozilla/bugbug/blob/cf95bc1ae77c9855a8885e93161b572dd79f7678/functions/sync-review-comments-db/models.py#L69-L70

We currently have the following values in the tool variant column: 
- RAG and CONTEXT
- CONTEXT
- RAG

suhaibmujahid: I filed an issue to follow up on this: https://github.com/mozilla/bugbug/issues/4734

Issue #4579, opened 2024-11-01
[code_review] Use rejected comments in RAG for the filtering phase


marco-c:
We are using comments as positive examples in the generation phase, but we are not making use of rejected comments.
Adding rejected comments in the filtering phase will help reduce noise.

PR #4736, opened 2025-01-10
[code_review] Use previously rejected comments as examples in the filtering phase

suhaibmujahid:
Resolves https://github.com/mozilla/bugbug/issues/4579

### DIFF
--- a/bugbug/vectordb.py
+++ b/bugbug/vectordb.py
@@ -5,7 +5,7 @@
 
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
-from typing import Iterable
+from typing import Iterable, Optional
 
 from qdrant_client import QdrantClient
 from qdrant_client.conversions import common_types as qdrant_types
@@ -34,18 +34,26 @@ class QueryFilter:
     """A filter for a vector DB query.
 
     Attributes:
-        must_match: The key is the field name and the value that must be matched.
+        must_match: The key is the field name and the value what must be matched.
+        must_not_has_id: A list of IDs to exclude from the search results.
     """
 
-    must_match: dict[str, str | int]
+    must_match: Optional[dict[str, str | int]] = None
+    must_not_has_id: Optional[list[int]] = None
 
     def to_qdrant_filter(self) -> qdrant_types.Filter:
-        return {
-            "must": [
+        qdrant_filter: qdrant_types.Filter = {}
+
+        if self.must_match:
+            qdrant_filter["must"] = [
                 {"key": key, "match": {"value": value}}
                 for key, value in self.must_match.items()
             ]
-        }
+
+        if self.must_not_has_id:
+            qdrant_filter["must_not"] = [{"has_id": self.must_not_has_id}]
+
+        return qdrant_filter or None
 
 
 class VectorDB(ABC):
--- a/bugbug/vectordb.py
+++ b/bugbug/vectordb.py
@@ -8,6 +8,7 @@ from dataclasses import dataclass
 from typing import Iterable
 
 from qdrant_client import QdrantClient
+from qdrant_client.conversions import common_types as qdrant_types
 from qdrant_client.http.exceptions import UnexpectedResponse
 from qdrant_client.models import Distance, PointStruct, VectorParams
 
@@ -28,6 +29,25 @@ class PayloadScore:
     payload: dict
 
 
+@dataclass
+class QueryFilter:
+    """A filter for a vector DB query.
+
+    Attributes:
+        must_match: The key is the field name and the value that must be matched.
+    """
+
+    must_match: dict[str, str | int]
+
+    def to_qdrant_filter(self) -> qdrant_types.Filter:
+        return {
+            "must": [
+                {"key": key, "match": {"value": value}}
+                for key, value in self.must_match.items()
+            ]
+        }
+
+
 class VectorDB(ABC):
     """Abstract class for a vector database.
 
@@ -43,7 +63,9 @@ class VectorDB(ABC):
         ...
 
     @abstractmethod
-    def search(self, query: list[float]) -> Iterable[PayloadScore]:
+    def search(
+        self, query: list[float], filter: QueryFilter | None = None, limit: int = 10
+    ) -> Iterable[PayloadScore]:
         ...
 
     @abstractmethod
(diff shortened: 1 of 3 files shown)


suhaibmujahid:
Here is a dry run:

```
---------------------------------------------------------
Review Request ID: 229230
Patch ID: 946264
---------------------------------------------------------

File                              Line  Comment
------------------------------  ------  ----------------------------------------------------------------------------------------------------
gfx/layers/ipc/PNativeLayerRem      48  The union `NativeLayerCommand` includes several command types, but the structs for these commands
ote.ipdl                                (e.g., `CreateLayer`, `AppendLayer`) are empty. Should these structs have fields or parameters to
                                        define their behavior?
gfx/layers/ipc/PNativeLayerRem      48  The `CommitNativeLayerCommands` method accepts an array of `NativeLayerCommand` objects, but there
ote.ipdl                                is no validation or description of how these commands are processed. Should there be constraints or
                                        documentation on the expected behavior for invalid or unsupported commands?
gfx/layers/NativeLayerRemote.h      21  The `NativeLayerRootRemoteGPU` class now inherits from `PNativeLayerRemoteParent`. Are there any
                                        potential conflicts or ambiguities introduced by this multiple inheritance? For example, could there
                                        be overlapping methods or responsibilities between the two base classes?
gfx/layers/NativeLayerRemote.h      46  The `SendCommitNativeLayerCommands` method is defined as a stub with an empty implementation. Should
                                        this method be implemented or removed if it is not intended to be used?






---------------------------------------------------------
Review Request ID: 230338
Patch ID: 965923
---------------------------------------------------------

File                  Line  Comment
------------------  ------  ---------------------------------------------------------------------------------------------------
gfx/gl/GLContext.h     821  The use of `make_array` here assumes that the array will always be valid for the iOS platform.
                            However, if `LOCAL_GL_COLOR_ATTACHMENT0` is not supported or behaves differently on iOS, this could
                            lead to unexpected behavior. Have we verified that this is always valid on iOS?
gfx/gl/GLContext.h     821  The conditional compilation introduces platform-specific logic, but it might be worth considering
                            whether this could be abstracted to reduce the need for preprocessor directives, which can make the
                            code harder to maintain.
gfx/gl/GLContext.h     821  The call to `fInvalidateFramebuffer` depends on the `ATTACHMENTS` array defined conditionally. If
                            the platform-specific logic is not handled correctly, it could lead to runtime issues. Have we
                            ensured that the size and data of `ATTACHMENTS` are always valid for all platforms?





---------------------------------------------------------
Review Request ID: 230304
Patch ID: 951257
---------------------------------------------------------

File                              Line  Comment
------------------------------  ------  ---------------------------------------------------------------------------------------------------
browser/modules/BrowserWindowT     323  Setting `args.data` to an empty string when `shouldRestoreLastSession` is true might lead to
racker.sys.mjs                          unexpected behavior if other parts of the code rely on `args.data` containing meaningful data.
                                        Consider verifying the implications of this change.
browser/components/sessionstor    1891  The condition `isPrivateWindow && isTaskbarTab` seems overly restrictive. Taskbar tabs and private
e/SessionStore.sys.mjs                  windows are distinct concepts, and combining them here might lead to unintended behavior. Consider
                                        separating their handling.
browser/components/sessionstor    1942  The comment and condition here imply that taskbar tabs are treated similarly to private windows.
e/SessionStore.sys.mjs                  This could lead to confusion or incorrect behavior if taskbar tabs have different requirements.
                                        Consider clarifying or separating their logic.
browser/components/sessionstor    2029  The condition for restoring the last session does not account for taskbar tabs. If taskbar tabs are
e/SessionStore.sys.mjs                  meant to be excluded, this should be explicitly stated in the condition or documented.
browser/components/sessionstor    2243  The logic for setting `_shouldRestoreLastSession` assumes that `willAutoRestore` is always accurate
e/SessionStore.sys.mjs                  at this point. If `willAutoRestore` changes dynamically, this might lead to inconsistent behavior.
browser/components/sessionstor    2379  Excluding taskbar tabs from being saved as closed windows (`!isTaskbarTab`) might lead to data loss
e/SessionStore.sys.mjs                  if taskbar tabs are expected to be restorable. Consider verifying this behavior.
browser/components/sessionstor    4492  The removal of windows from `_closedWindows` during continuous restoration (`_continousRestore`)
e/SessionStore.sys.mjs                  might lead to inconsistencies if other parts of the code rely on `_closedWindows` to track all
                                        previously closed windows.
browser/components/sessionstor    4554  The logic for removing duplicates in `_closedWindows` during continuous restoration assumes that
e/SessionStore.sys.mjs                  `closedId` is unique and consistent. If `closedId` can change or be reused, this might lead to
                                        incorrect removals.
browser/base/content/browser.j    6469  Adding the observer for `sessionstore-last-session-re-enable` without removing it later might lead
s                                       to memory leaks or unexpected behavior if the observer is triggered multiple times.
```

suhaibmujahid:
Here is a comparison with and without the rejected comments: [code_review_tool_evaluator.csv](https://github.com/user-attachments/files/18428337/code_review_tool_evaluator.csv)

suhaibmujahid:
Here is an updated dry run: [code_review_tool_evaluator.csv](https://github.com/user-attachments/files/18482810/code_review_tool_evaluator.csv)

bugbug/vectordb.py
```diff
@@ -28,6 +29,33 @@ class PayloadScore:
     payload: dict
 
 
+@dataclass
+class QueryFilter:
+    """A filter for a vector DB query.
+
+    Attributes:
+        must_match: The key is the field name and the value that must be matched.
+        must_not_has_id: A list of IDs that exclude from the search results.
```
marco-c, line 38:
```suggestion
        must_not_has_id: A list of IDs to exclude from the search results.
```

bugbug/vectordb.py
```diff
@@ -28,6 +29,33 @@ class PayloadScore:
     payload: dict
 
 
+@dataclass
+class QueryFilter:
+    """A filter for a vector DB query.
+
+    Attributes:
+        must_match: The key is the field name and the value that must be matched.
```
marco-c, line 37:
```suggestion
        must_match: The key is the field name and the value what must be matched.
```

marco-c (approved):

bugbug/vectordb.py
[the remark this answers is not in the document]
suhaibmujahid: It is a 1:1 mapping with the filtering condition used there. I used the `must_not` and `has_id`:
```py
qdrant_filter["must_not"] = [{"has_id": self.must_not_has_id}]
```

The same goes for `must_match`:
```py
qdrant_filter["must"] = [{"key": key, "match": {"value": value}}]
```

marco-c (approved):

PR #4770, opened 2025-01-23
[code_review] Improve the prompting

suhaibmujahid:
This PR, introduce some improvements to the prompts. 

Here are the main observations:
- The suggestions are shorter now.
- No questions or "ensure" suggestions.

### DIFF
--- a/bugbug/tools/code_review.py
+++ b/bugbug/tools/code_review.py
@@ -63,25 +63,18 @@ class LargeDiffError(Exception):
     """Occurs when the diff is too large to be processed."""
 
 
-TARGET_SOFTWARE = None
+TARGET_SOFTWARE: str | None = None
 
-PROMPT_TEMPLATE_SUMMARIZATION = (
-    """You are an expert reviewer for"""
-    + (f" the {TARGET_SOFTWARE}" if TARGET_SOFTWARE is not None else "")
-    + """ source code, with experience on source code reviews.
+PROMPT_TEMPLATE_SUMMARIZATION = """You are an expert reviewer for {experience_scope}, with experience on source code reviews.
 
 Please, analyze the code provided and report a summarization about the new changes; for that, focus on the coded added represented by lines that start with "+".
 
 {patch}"""
-)
 
-PROMPT_TEMPLATE_REVIEW = (
-    """You will be given a task to generate a code review for the patch below. Use the following steps to solve it:
+PROMPT_TEMPLATE_REVIEW = """You will be given a task to generate a code review for the patch below. Use the following steps to solve it:
 1. Understand the changes done in the patch by reasoning about the summarization as previously reported.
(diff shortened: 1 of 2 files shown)


suhaibmujahid:
Dry-run: [code_review_tool_evaluator.csv](https://github.com/user-attachments/files/18526739/code_review_tool_evaluator.csv)

bugbug/tools/code_review.py
```diff
@@ -63,30 +63,25 @@ class LargeDiffError(Exception):
...
-    + (f" in {TARGET_SOFTWARE}" if TARGET_SOFTWARE is not None else "")
-    + """.
-4. Filter out comments that focuses on documentation, comments, error handling, tests, and confirmation whether objects, methods and files exist or not.
-5. Filter out comments that are descriptive.
-6. Filter out comments that are praising (example: "This is a good addition to the code.").
-7. Filter out comments that are not about added lines (have '+' symbol at the start of the line).
-8. Final answer: Write down the comments and report them using the JSON format previously adopted for the valid comment examples.
+3. Reason about each identified problem to make sure they are valid. Have in mind, your review must be consistent with the {target_code_consistency} source code.
+4. Generate review comments that are short, to the point and focused on the changed code. Don't form the suggestions as questions.
+5. Avoid comments that focuses on documentation, tests, or confirmation whether something is exist.
```
marco-c, line 79:
```suggestion
5. Avoid comments that focus on documentation, tests, or confirmation whether something exists.
```

bugbug/tools/code_review.py
```diff
@@ -63,30 +63,25 @@ class LargeDiffError(Exception):
...
+6. Avoid comments that ask the author to ensure that the change correct (using words like Ensure, Verify, or Check).
```
marco-c, line 80:
```suggestion
6. Avoid comments that ask the author to ensure that the change is correct (using words like Ensure, Verify, or Check).
```

bugbug/tools/code_review.py
```diff
@@ -63,30 +63,25 @@ class LargeDiffError(Exception):
...
+7. Avoid comments that are descriptive.
+8. Avoid comments that are praising (example: "This is a good addition to the code.").
+9. Avoid comments that are not about added lines (have '+' symbol at the start of the line).
```
marco-c, line 83: In theory we could have comments on "-" lines too, maybe we should say "changed lines" instead of "added lines"?

suhaibmujahid: IIRC, we did that to prevent false positive cases, where the suggestions were about improving the removed portion of the code.

marco-c: We should re-evaluate, but maybe not now, in a follow-up. Could you file an issue?

marco-c (approved):

PR #4797, opened 2025-02-03
[code_review] Symbol prefix added to `search` function

### LINKED ISSUE #4577: [code_review] Improve Searchfox search by using "symbol" or "id" prefixes
As suggested by @asutherland:
> for something like [this symbol search](https://github.com/mozilla/bugbug/blob/c98b7078911f442a99c22407bcd4ffc9beaa59e8/bugbug/code_search/searchfox_api.py#L89) if you add the prefix "symbol:" if it's indeed a symbol or "id:" if it's an identifier search, things should go 1-to-2 orders of magnitude faster and greatly decrease the load on searchfox because you won't do a fulltext query.  There's "docs" on all the search endpoint options in https://bugzilla.mozilla.org/show_bug.cgi?id=1762817#c0.

benjaminmah:
Resolves #4577.

### DIFF
--- a/bugbug/code_search/searchfox_api.py
+++ b/bugbug/code_search/searchfox_api.py
@@ -92,7 +92,7 @@ def find_function_for_line(commit_hash, path, line):
 # TODO: we should use commit_hash...
 def search(commit_hash, symbol_name):
     r = utils.get_session("searchfox").get(
-        f"https://searchfox.org/mozilla-central/search?q={symbol_name}",
+        f"https://searchfox.org/mozilla-central/search?q=symbol:{symbol_name}",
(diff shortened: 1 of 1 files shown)


marco-c:
Unfortunately this won't work because it looks like we need to specify the full name of the symbol (e.g. for the function "once_control", we'd need `symbol:_ZN13GUIDGenerator12once_controlE', but the LLM isn't giving us that).
We need to figure out if there's a way to use a regex and do something like `symbol:*once_control*`.

marco-c:
I asked, there's no way, so I guess we can't use `symbol` yet :(

Issue #4842, opened 2025-02-26
[searchfox_api] Some patches trigger extensive number of function searches


suhaibmujahid:
Example: https://phabricator.services.mozilla.com/D215266?id=885315

It triggered the search https://searchfox.org/mozilla-central/search?q=once

Wiche caused to perform function searches in 433 files. The following is the list of paths from the `paths` variable in: https://github.com/mozilla/bugbug/blob/b766ea2c06cd4ac945131eda276a27395f463c95/bugbug/code_search/searchfox_api.py#L122
```
000: 'toolkit/content/editMenuOverlay.js'
001: 'browser/components/asrouter/modules/PageEventManager.sys.mjs'
002: 'browser/components/asrouter/tests/NimbusRolloutMessageProvider.sys.mjs'
003: 'devtools/client/shared/theme-switching.js'
004: 'browser/base/content/test/popupNotifications/head.js'
005: 'browser/components/extensions/test/browser/browser_ext_devtools_panel_zoom.js'
006: 'devtools/client/inspector/shared/tooltips-overlay.js'
007: 'toolkit/components/extensions/Extension.sys.mjs'
008: 'browser/components/sessionstore/test/browser_615394-SSWindowState_events_setTabState.js'
009: 'devtools/client/debugger/dist/parser-worker.js'
010: 'devtools/client/shared/sourceeditor/editor.js'
011: 'browser/components/AccountsGlue.sys.mjs'
012: 'browser/components/resistfingerprinting/test/browser/file_hwconcurrency_blob_iframer.html'
013: 'browser/components/asrouter/tests/browser/browser_feature_callout.js'
014: 'browser/components/places/tests/browser/head.js'
015: 'browser/base/content/test/permissions/browser_canvas_fingerprinting_resistance.js'
016: 'browser/components/extensions/test/browser/browser_ext_optionsPage_popups.js'
017: 'devtools/client/responsive/utils/e10s.js'
018: 'devtools/client/inspector/grids/grid-inspector.js'
019: 'toolkit/actors/PictureInPictureChild.sys.mjs'
020: 'accessible/tests/mochitest/actions/test_media.html'
021: 'browser/components/resistfingerprinting/test/mochitest/test_keyboard_event.html'
022: 'browser/components/genai/GenAI.sys.mjs'
023: 'browser/components/urlbar/tests/browser/browser_close_other_panels.js'
024: 'devtools/client/aboutdebugging/aboutdebugging.js'
[...]

PR #4848, opened 2025-02-27
Filter out some JS results that are most likely not functions

marco-c:
Will help, but not fix, #4842


suhaibmujahid (approved):
Thank you!

PR #4849, opened 2025-02-27
Filter out JavaScript functions better, and also Rust and C/C++

marco-c:
Will help, but not fix, #4842


suhaibmujahid (approved):
Great! Thank you!

PR #4850, opened 2025-02-27
Use identifier search when searching functions in Searchfox

marco-c:
This should be lighter on Searchfox, and will make requests faster.


suhaibmujahid (approved):
Great!

PR #4855, opened 2025-02-28
[code_review_tool_evaluator] Restrict the columns order in the CSV results

suhaibmujahid:
This is necessary because we save the results in patches.


benjaminmah (approved):
LGTM!

PR #4866, opened 2025-03-03
Release notes generator script

benjaminmah:
Includes both the tool (`bugbug/tools/release_notes.py`) and the runner (`scripts/release_notes_runner.py`).

Takes two parameters:
`--version` - the version of Firefox to generate a list of potential commits to include in release notes (e.g. `FIREFOX_BETA_136_BASE`)
`--chunk-size` - the size of the chunk (in tokens) when passing in the commit logs to the LLM


bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,236 @@
...
+        self.version1 = self.get_previous_version(version)
+        self.output_file = f"version_summary_{self.version2}.txt"
+
+        logger.info(f"Generating list of commits for version: {self.version2}")
+        url = f"https://hg.mozilla.org/releases/mozilla-release/pushloghtml?fromchange={self.version1}&tochange={self.version2}"
+        response = requests.get(url)
+        changes_output = ""
+
+        if response.status_code == 200:
+            soup = BeautifulSoup(response.text, "html.parser")
```
suhaibmujahid, line 196: Parsing HTML should be the last option. It is expensive and not reliable.

You could get JSON results instead: https://hg.mozilla.org/releases/mozilla-release/json-pushes?full=1&version=2

suhaibmujahid (changes requested):

bugbug/tools/release_notes.py
benjaminmah: Yes, originally I was going to, but I couldn't find the same commits. I just found it now, most of them were nested in the last changeset: https://hg.mozilla.org/releases/mozilla-release/json-pushes?fromchange=FIREFOX_BETA_135_BASE&tochange=FIREFOX_BETA_136_BASE&full=1. Will change now!

benjaminmah: Done here: 35ad0734172696002cfc571b2d7d58cb5c78cafe

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,238 @@
...
+    def generate_worthy_commits(self, version):
+        self.version2 = version
+        self.version1 = self.get_previous_version(version)
+        self.output_file = f"version_summary_{self.version2}.txt"
+
+        logger.info(f"Generating list of commits for version: {self.version2}")
+        url = f"https://hg.mozilla.org/releases/mozilla-release/json-pushes?fromchange={self.version1}&tochange={self.version2}&full=1"
+        response = requests.get(url)
+        changes_output = ""
+        if response.status_code == 200:
```
suhaibmujahid, line 193: I would use `response.raise_for_status()` instead.

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,238 @@
...
+            data = response.json()
+            commit_descriptions = []
+
+            for push_id in data:
```
suhaibmujahid, line 197: Why not use `data.values()` instead?

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,238 @@
...
+                push_data = data[push_id]
+                changesets = push_data.get("changesets", [])
```
suhaibmujahid, line 199: Do we have cases where `changesets` is not part of the push object?

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,238 @@
...
+
+                for changeset in changesets:
+                    desc = changeset.get("desc", "").strip()
```
suhaibmujahid, line 202: Do we have cases where `desc` does not exist?

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,238 @@
...
+                    {
+                        "role": "user",
+                        "content": prompt,
+                    }
+                ],
+                model=MODEL,
+                temperature=0.1,
+            )
+            return response.choices[0].message.content.strip()
+        except Exception as e:
```
suhaibmujahid, line 180: Wrapping a code block in a generic try-except is dangerous since it will catch all kinds of exceptions. Try to limit the code in the try block and use a more specific error type.

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,238 @@
...
+import os
+import re
+
+import requests
+import tiktoken
+from openai import OpenAI
+
+MODEL = "gpt-4o"
+
+client = OpenAI(api_key=os.environ.get("OPENAI_API_KEY"))
```
suhaibmujahid, line 11: We usually use LangCahin to swap models easily. It would be nice to do the same for consistency.

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,238 @@
...
+            for push_id in data:
+                push_data = data[push_id]
+                changesets = push_data.get("changesets", [])
+
+                for changeset in changesets:
+                    desc = changeset.get("desc", "").strip()
+                    if desc:
+                        commit_descriptions.append(desc)
+
+            changes_output = "\n".join(commit_descriptions)
```
suhaibmujahid, line 206: Why do you convert the list to a string here?

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,238 @@
...
+        ]
+        cleaned_commits = self.clean_commits(changes_output, keywords_to_remove)
+        print(f"CLEANED COMMITS: {cleaned_commits}")
+
+        logger.info("Generating summaries for cleaned commits...")
+        summaries = self.generate_summaries(cleaned_commits)
+        combined_list = "\n".join(summaries)
+
+        logger.info("Removing unworthy commits from the list...")
+        combined_list = self.remove_unworthy_commits(combined_list)
```
suhaibmujahid, line 233: I would expect to pass it as a list instead of a string.

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,238 @@
...
+
+        with open(self.output_file, "w") as file:
+            file.write(combined_list)
```
suhaibmujahid, line 236: Why do we save the results to a file?

bugbug/tools/release_notes.py
benjaminmah: Done here: 2d0030cc48cd522b6241c6beee221642e3e57794

bugbug/tools/release_notes.py
benjaminmah: Not sure. From all the objects I saw, they had `changesets`. I have removed the `.get()` and replaced it with a direct call here: 2d0030cc48cd522b6241c6beee221642e3e57794.

bugbug/tools/release_notes.py
benjaminmah: Same as `changesets`, removed here: 2d0030cc48cd522b6241c6beee221642e3e57794

bugbug/tools/release_notes.py
benjaminmah: Removed the file saving, it is now printing to console: fbb3c30c06a8c44d60fc62d68a3d534a28f38bfa

bugbug/tools/release_notes.py
benjaminmah: Changed input and output type to a list. Originally was a string because of the single LLM output but I handle that in the function now: fbb3c30c06a8c44d60fc62d68a3d534a28f38bfa

bugbug/tools/release_notes.py
benjaminmah: Originally because of the LLM input. Handled in the function now so it is now a list: fbb3c30c06a8c44d60fc62d68a3d534a28f38bfa

bugbug/tools/release_notes.py
benjaminmah: Removed the try-except block here: fbb3c30c06a8c44d60fc62d68a3d534a28f38bfa

bugbug/tools/release_notes.py
suhaibmujahid: Why do we need it printed? If we want to print the results, I think the runner is a more suitable place for that.

bugbug/tools/release_notes.py
suhaibmujahid: We create chains and use them instead of using the model directly. You can have a look at how we did that at https://github.com/mozilla/bugbug/blob/master/bugbug/tools/code_review.py

bugbug/tools/release_notes.py
benjaminmah: Done here: 1bf92b2f39807d2cbc5e86ba53bd79f034692ec2

bugbug/tools/release_notes.py
benjaminmah: Done here: 1bf92b2f39807d2cbc5e86ba53bd79f034692ec2

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,221 @@
...
+    def get_token_count(self, text):
+        if hasattr(self.llm, "model_name"):
+            model_name = self.llm.model_name
+        else:
+            raise ValueError("LLM model name not found.")
+
+        encoding = tiktoken.encoding_for_model(model_name)
+        return len(encoding.encode(text))
+
+    def split_into_chunks(self, commit_log):
```
suhaibmujahid, line 118: It would be cleaner to use `itertools.batched()`.

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,221 @@
...
+                commit_summary = block
+                cleaned_commits.append(commit_summary)
+
+        return cleaned_commits
+
+    def remove_unworthy_commits(self, summaries):
+        combined_list = "\n".join(summaries)
+        return self.cleanup_chain.run({"combined_list": combined_list}).strip()
+
+    def generate_worthy_commits(self, version):
```
suhaibmujahid, line 180: The name is a bit confusing. It does not generate commits here.

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,221 @@
...
+        for push_data in data.values():
+            changesets = push_data["changesets"]
+
+            for changeset in changesets:
+                desc = changeset["desc"].strip()
+                if desc:
+                    commit_log_list.append(desc)
+
+        if not commit_log_list:
+            logger.error("No changes found.")
```
suhaibmujahid, line 202: I do not think the log is useful here.

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,221 @@
...
+
+client = OpenAI(api_key=os.environ.get("OPENAI_API_KEY"))
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+class ReleaseNotesGenerator:
+    def __init__(self, chunk_size, llm):
+        self.chunk_size = chunk_size
```
suhaibmujahid, line 19: Could you please use type hints for all methods?

bugbug/tools/release_notes.py
benjaminmah: Fixed here: 06af4d87038764c3a72dc70630deca0799dd3e6d

bugbug/tools/release_notes.py
benjaminmah: Removed here: 06af4d87038764c3a72dc70630deca0799dd3e6d

bugbug/tools/release_notes.py
benjaminmah: a14ad873813d1965b9e8d42e891aeceb17cf39cb

bugbug/tools/release_notes.py
benjaminmah: I know we can use `itertools.batched()` for a fixed n number of lines, but would it be possible for token count? (which is what we have now)

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,221 @@
...
+import re
+from typing import List, Optional
+
+import requests
+import tiktoken
+from langchain.chains import LLMChain
+from langchain.prompts import PromptTemplate
+from openai import OpenAI
+
+client = OpenAI(api_key=os.environ.get("OPENAI_API_KEY"))
```
suhaibmujahid, line 12: Why do we need this?

benjaminmah: Removed here: 20d0b6ebf2822b76f5564f97c18ceb635fc77b63

Forgot to remove it earlier!

bugbug/tools/release_notes.py
suhaibmujahid: Why do we need to use token count? We could keep it simple with the number of comments instead.

benjaminmah: We use token count because we chunk it to fit within the context size of the model we are using. By number of comments, do you need number of commits?

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,217 @@
...
+    def shortlist_with_gpt(self, input_text: str) -> str:
+        return self.summarization_chain.run({"input_text": input_text}).strip()
+
+    def generate_commit_shortlist(self, commit_log_list: List[str]) -> List[str]:
+        commit_log_list_combined = "\n".join(commit_log_list)
+        chunks = self.split_into_chunks(commit_log_list_combined)
+        return [self.shortlist_with_gpt(chunk) for chunk in chunks]
+
+    def clean_commits(
+        self, commit_log_list: List[str], keywords: List[str]
```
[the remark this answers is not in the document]
benjaminmah: Done here: b48089fc47e80aa00ad46da20931202f0b2eb742

bugbug/tools/release_notes.py
suhaibmujahid: Yes. In many cases, the number of commits could be more related to the model capability than just the number of tokens.

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,217 @@
...
+    ) -> list[str]:
+        cleaned_commits = []
+
+        for block in commit_log_list:
+            if (
+                not any(
+                    re.search(rf"\b{keyword}\b", block, re.IGNORECASE)
+                    for keyword in keywords
+                )
+                and re.search(r"Bug \d+", block, re.IGNORECASE)
```
suhaibmujahid, line 156: I do think this is needed since we are already changing that later in `if not bug_id_match`.

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,216 @@
...
+
+    def shortlist_with_gpt(self, input_text: str) -> str:
+        return self.summarization_chain.run({"input_text": input_text}).strip()
+
+    def generate_commit_shortlist(self, commit_log_list: list[str]) -> list[str]:
+        commit_log_list_combined = "\n".join(commit_log_list)
+        chunks = self.split_into_chunks(commit_log_list_combined)
+        return [self.shortlist_with_gpt(chunk) for chunk in chunks]
+
+    def clean_commits(
```
suhaibmujahid, line 145: I do not find `clean_commits()` a descriptive name.

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,216 @@
...
+                bug_id_match = re.search(r"Bug (\d+)", block, re.IGNORECASE)
+                if not bug_id_match:
+                    continue
+
+                bug_position = re.search(r"Bug \d+.*", block, re.IGNORECASE)
+                if bug_position:
+                    block = bug_position.group(0)
+
+                commit_summary = block
+                cleaned_commits.append(commit_summary)
```
suhaibmujahid, line 170: we could use `yield` here instead of creating a list.

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,216 @@
...
+                    for keyword in keywords
+                )
+                and not re.search(
+                    r"release\+treescript@mozilla\.org", block, re.IGNORECASE
+                )
+                and not re.search(r"nightly", block, re.IGNORECASE)
+            ):
+                bug_id_match = re.search(r"Bug (\d+)", block, re.IGNORECASE)
+                if not bug_id_match:
+                    continue
```
suhaibmujahid, line 163: I just noticed that we do not use the value of `bug_id_match`, so we could just have it as part of the condition.

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,216 @@
...
+            return None
+
+        logger.info("Cleaning commit log...")
+        keywords_to_remove = [
+            "Backed out",
+            "a=testonly",
+            "DONTBUILD",
+            "add tests",
+            "disable test",
+        ]
```
suhaibmujahid, line 209: Why do we need to pass this list as an argument? Would it be better to keep it constant at the module/file level or include it in the `clean_commits()` function?

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,216 @@
...
+        return cleaned_commits
+
+    def remove_unworthy_commits(self, summaries: list[str]) -> str:
+        combined_list = "\n".join(summaries)
+        return self.cleanup_chain.run({"combined_list": combined_list}).strip()
+
+    def select_worthy_commits(self, version: str) -> Optional[str]:
+        self.version2 = version
+        self.version1 = self.get_previous_version(version)
+        self.output_file = f"version_summary_{self.version2}.txt"
```
suhaibmujahid, line 181: Why do we need `self.output_file`?

bugbug/tools/release_notes.py
benjaminmah: Ok, changed here: 2dddedb80a2a3ba8b7cbc8fdda120ddfd4345ad2

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,192 @@
...
+- The output must be only the cleaned-up list, formatted exactly the same way.
+""",
+        )
+
+        self.cleanup_chain = LLMChain(
+            llm=self.llm,
+            prompt=self.cleanup_prompt,
+        )
+
+    def get_previous_version(self, current_version: str) -> str:
```
marco-c, line 106: Could you add some tests covering this function?

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,192 @@
...
+        return f"{prefix}{previous_version_number}{suffix}"
+
+    def batch_commit_logs(self, commit_log: str) -> list[str]:
+        return [
+            "\n".join(batch)
+            for batch in batched(commit_log.strip().split("\n"), self.chunk_size)
+        ]
+
+    def shortlist_with_gpt(self, input_text: str) -> str:
+        return self.summarization_chain.run({"input_text": input_text}).strip()
```
marco-c, line 121: Given it is a single line and a single caller, I would remove this function and put the code inline in the call site

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,192 @@
...
+                    r"release\+treescript@mozilla\.org", block, re.IGNORECASE
+                )
+                and not re.search(r"nightly", block, re.IGNORECASE)
+            ):
+                bug_position = re.search(r"Bug \d+.*", block, re.IGNORECASE)
+                if bug_position:
+                    block = bug_position.group(0)
+
+                commit_summary = block
+                yield commit_summary
```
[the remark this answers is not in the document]
marco-c: Or just rename `block` to `commit_summary`

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,192 @@
...
+
+    def refine_commit_shortlist(self, summaries: list[str]) -> str:
+        combined_list = "\n".join(summaries)
+        return self.cleanup_chain.run({"combined_list": combined_list}).strip()
```
marco-c, line 152: Same here, I would just put it inline and remove the func

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,192 @@
...
+    ) -> Generator[str, None, None]:
+        for block in commit_log_list:
+            if (
+                not any(
+                    re.search(rf"\b{keyword}\b", block, re.IGNORECASE)
+                    for keyword in KEYWORDS_TO_REMOVE
+                )
+                and re.search(r"Bug \d+", block, re.IGNORECASE)
+                and not re.search(
+                    r"release\+treescript@mozilla\.org", block, re.IGNORECASE
```
marco-c, line 139: `release+treescript@mozilla.org` is in the author field, not the description. Isn't block here the `desc` of the commit?

scripts/release_notes_runner.py
```diff
@@ -0,0 +1,41 @@
...
+
+    if args.llm.startswith("openai-"):
+        model_name = args.llm.replace("openai-", "")
+        llm = ChatOpenAI(
+            model=model_name,
+            temperature=0.1,
+            openai_api_key=os.environ.get("OPENAI_API_KEY"),
+        )
+    else:
+        raise ValueError(f"Unsupported LLM provider: {args.llm}")
```
marco-c, line 33: It'd be ideal to be generic here and not specify any specific vendor.
You can use these helper functions to achieve it:
https://github.com/mozilla/bugbug/blob/c050d72a40899e16343fcbae483280e3f47a1a86/scripts/code_review_tool_runner.py#L56
and
https://github.com/mozilla/bugbug/blob/c050d72a40899e16343fcbae483280e3f47a1a86/scripts/code_review_tool_runner.py#L16

bugbug/tools/release_notes.py
benjaminmah: Done here: 030d7053530c148b429a983ed53e3eef16e1dc48

bugbug/tools/release_notes.py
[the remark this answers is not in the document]
benjaminmah: Direclty removed any of the names, just returning `bug_position.group(0)` here: 6326f668c6f0916f4008080586f79e4fcbd99dfc

bugbug/tools/release_notes.py
benjaminmah: Done here: e25aad56c7d0ebda2322ea29b71ce0c4ab12bd60

bugbug/tools/release_notes.py
benjaminmah: You're right. I added the return of the author as well from the `get_commit_logs()` function, and now filtering based on that as well: 64185519441ee9f8f7bfdf491783c89f77161562

scripts/release_notes_runner.py
benjaminmah: Done here: 4140c5256e986d84ac6b5b9b898b07c4c8b29e46

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,192 @@
...
+
+    def shortlist_with_gpt(self, input_text: str) -> str:
+        return self.summarization_chain.run({"input_text": input_text}).strip()
+
+    def generate_commit_shortlist(self, commit_log_list: list[str]) -> list[str]:
+        commit_log_list_combined = "\n".join(commit_log_list)
+        chunks = self.batch_commit_logs(commit_log_list_combined)
+        return [self.shortlist_with_gpt(chunk) for chunk in chunks]
+
+    def filter_irrelevant_commits(
```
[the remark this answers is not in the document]
benjaminmah: There are around 3000~ commits that are not excluded still (4499 to 3148).  [filtered_commits.txt](https://github.com/user-attachments/files/19253702/filtered_commits.txt)

bugbug/tools/release_notes.py
benjaminmah: I've generalized the previous version function, so it doesn't only take in `BETA` versions. c6eafb80c8388e1b451ba988e3d5c49f5a9e3158. Let me know if you still want me to add tests!

scripts/release_notes_runner.py
marco-c: It'd be better to use create_llm_to_args like I suggested above for two reasons:
1) This way you don't manually specify the `--llm` argument, which might change in create_llm_from_args
2) This way you also support the additional llm options

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,191 @@
...
+        for desc, author in commit_log_list:
+            if (
+                not any(
+                    re.search(rf"\b{keyword}\b", desc, re.IGNORECASE)
+                    for keyword in KEYWORDS_TO_REMOVE
+                )
+                and re.search(r"Bug \d+", desc, re.IGNORECASE)
+                and not re.search(
+                    r"release\+treescript@mozilla\.org", author, re.IGNORECASE
+                )
```
marco-c, line 145: No need for a regex here, we can just check for equality

bugbug/tools/release_notes.py
marco-c: Yes, it'd be best to add some tests

bugbug/tools/release_notes.py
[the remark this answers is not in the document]
marco-c:
```
Bug 1945142: back out D236033. r=tjr
Bug 1945142: Backout D234215. r=tjr
Bug 1945142: Backout D234174. r=tjr
```
could be excluded as they are backouts.

Anything that is CI related could be excluded. We could use the bug component for that (e.g. Firefox Build System::Task Configuration, Developer Infrastructure::*). An example is `Bug 1937801 - [ci] Use upstream taskgraph's logic for handling caches, r=taskgraph-reviewers,hneiva,bhearsum`.

Any commit that is only touching test files could be excluded (for example `Bug 1924052: add test for 32-bit sample size M4A. r=padenot`).

Any commit that is in the ignore list could be excluded (https://hg.mozilla.org/mozilla-central/file/tip/.hg-annotate-ignore-revs). You can also ignore commits that have `# ignore-this-changeset` in the description.

Commits fixing issues that cause tree closure can be excluded (e.g. `Bug 1942908 - Fix build bustages. CLOSED TREE`).

We could also ignore all commits that don't have a reviewer.

scripts/release_notes_runner.py
benjaminmah: 119121559e5f874262f1d59786aa8c76f8291fba

bugbug/tools/release_notes.py
benjaminmah: Done here! 51d6d9f75e3fb76289a05099918a85efbf2dba65

bugbug/tools/release_notes.py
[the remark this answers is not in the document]
benjaminmah: Added some of the suggestions here: 69af3869ff0e4071c7af133dcd4818977a9caa2f

1. Added a product/component to-ignore list
2. Extracted the commits to ignore from the ignore list
3. Excluded extra keywords such as backout, back out, test, tests, CLOSED TREE etc.

I have a couple questions and concerns. To see the product and component of the bug, I had to download the `BUGS_DB`. However, if we want to deploy this via the cloud, not sure if this is ideal. We could manually request the information of each bug, but that would be time consuming and expensive. I also need to download this DB to check if the bug has a reviewer - which leads me to my first question: Where is the bug object can I find the reviewer of the bug? I know the description has "r=..." or has a needinfo requested to a specific user, but is this sufficient to determine if a bug has a reviewer?

My second question is if there are any other product or components that should be ignored. I have a list in the code now called `PRODUCT_OR_COMPONENT_TO_IGNORE`, but only has two (the ones you mentioned).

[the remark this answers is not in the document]
marco-c: > Excluded extra keywords such as backout, back out, test, tests, CLOSED TREE etc.

I would not add "test" and "tests" to the keywords to ignore, it might be too strict.

> I have a couple questions and concerns. To see the product and component of the bug, I had to download the BUGS_DB. However, if we want to deploy this via the cloud, not sure if this is ideal.

It should be fine, as we are going to run this script only once a while.

> I also need to download this DB to check if the bug has a reviewer - which leads me to my first question: Where is the bug object can I find the reviewer of the bug? I know the description has "r=..." or has a needinfo requested to a specific user, but is this sufficient to determine if a bug has a reviewer?

The reviewer is associated to a commit, not to a bug. You can identify if a commit didn't have a reviewer by checking the absence of "r=".

> My second question is if there are any other product or components that should be ignored. I have a list in the code now called PRODUCT_OR_COMPONENT_TO_IGNORE, but only has two (the ones you mentioned).

Probably yes, but we can start with these and then add more in the future.

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,227 @@
...
+                if bug_match:
+                    bug_id = int(bug_match.group(1))
+
+                    bug = self.bug_dict.get(bug_id)
+                    if bug:
+                        if any(
+                            to_ignore in bug
+                            for to_ignore in PRODUCT_OR_COMPONENT_TO_IGNORE
+                        ):
+                            continue
```
marco-c, line 182: You can combine this into one line with an `and`

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,227 @@
...
+                and author
+                != "Mozilla Releng Treescript <release+treescript@mozilla.org>"
+                and not re.search(r"nightly", desc, re.IGNORECASE)
+                and desc not in cleaned_commits_to_ignore
+            ):
+                bug_match = re.search(r"Bug (\d+)", desc, re.IGNORECASE)
+                if bug_match:
+                    bug_id = int(bug_match.group(1))
+
+                    bug = self.bug_dict.get(bug_id)
```
marco-c, line 176: Let's rename `bug` and `bug_dict` to make it clear that it is a bug component (I was confused at first, I thought it was the bug object)

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,227 @@
...
+                    keyword.lower() in desc.lower() for keyword in KEYWORDS_TO_REMOVE
+                )
+                and re.search(r"Bug \d+", desc, re.IGNORECASE)
+                and author
+                != "Mozilla Releng Treescript <release+treescript@mozilla.org>"
+                and not re.search(r"nightly", desc, re.IGNORECASE)
+                and desc not in cleaned_commits_to_ignore
+            ):
+                bug_match = re.search(r"Bug (\d+)", desc, re.IGNORECASE)
+                if bug_match:
```
marco-c, line 173: Since we are searching `Bug \d+` above and only enter here if it is found, we don't need to check if there is a match

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,227 @@
...
+        ]
+        for desc, author in commit_log_list:
+            if (
+                not any(
+                    keyword.lower() in desc.lower() for keyword in KEYWORDS_TO_REMOVE
+                )
+                and re.search(r"Bug \d+", desc, re.IGNORECASE)
+                and author
+                != "Mozilla Releng Treescript <release+treescript@mozilla.org>"
+                and not re.search(r"nightly", desc, re.IGNORECASE)
```
marco-c, line 169: Why searching for `nightly`?

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,227 @@
...
+                and desc not in cleaned_commits_to_ignore
```
marco-c, line 170: Instead of checking for the desc, you could filter by commit hash

marco-c: It should be the same, but might be more precise

bugbug/tools/release_notes.py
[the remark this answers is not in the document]
benjaminmah: Good points. I've made the changes here: 88cf6318e05dc972a2fbf84703777c7abc26e7ab

bugbug/tools/release_notes.py
benjaminmah: Done here: 88cf6318e05dc972a2fbf84703777c7abc26e7ab

bugbug/tools/release_notes.py
benjaminmah: Sounds good! Done here: 88cf6318e05dc972a2fbf84703777c7abc26e7ab

bugbug/tools/release_notes.py
benjaminmah: Good point, reduced to a single regex here: 88cf6318e05dc972a2fbf84703777c7abc26e7ab

bugbug/tools/release_notes.py
benjaminmah: I think we discussed this in the channel with Suhaib that we wouldn't want to include `nightly`-related commits. I've moved this to the keyword list at the top of the file, but I can always remove it: 88cf6318e05dc972a2fbf84703777c7abc26e7ab

bugbug/tools/release_notes.py
benjaminmah: You're right! Done here: 88cf6318e05dc972a2fbf84703777c7abc26e7ab

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,233 @@
...
+    ) -> Generator[str, None, None]:
+        ignore_revs_url = "https://hg.mozilla.org/mozilla-central/raw-file/tip/.hg-annotate-ignore-revs"
+        response = requests.get(ignore_revs_url)
+        response.raise_for_status()
+        raw_commits_to_ignore = response.text.strip().splitlines()
+        hashes_to_ignore = [
+            line.split(" ", 1)[0]
+            for line in raw_commits_to_ignore
+            if re.search(r"Bug \d+", line, re.IGNORECASE)
+        ]
```
marco-c, line 164: If you don't need to iterate over this but only use it for checking presence, a set would be a better data format than a list

bugbug/tools/release_notes.py
marco-c: OK. I'd be suprised if there are many of these, but it doesn't hurt

bugbug/tools/release_notes.py
benjaminmah: 2c0a3ce84a16f76d2e7ea359055b3db2878c7868

bugbug/tools/release_notes.py
benjaminmah: Added a test file: 66dd826d2ac45ea132f1ca24fd3c2e0d9052a23a
Where else do I need to edit to add this new test?

functions/release_notes/main.py
```diff
@@ -0,0 +1,63 @@
...
+
+    args = build_args_from_request(request)
+
+    try:
+        llm = generative_model_tool.create_llm_from_args(args)
+        selector = ReleaseNotesCommitsSelector(chunk_size=args.chunk_size, llm=llm)
+        notes = selector.get_final_release_notes_commits(version=args.version)
+
+        if not notes:
+            return flask.Response("No user-facing commits found.", status=404)
```
suhaibmujahid, line 38: Responses might be simplified:
```suggestion
            return "No user-facing commits found.", 404
```

functions/release_notes/main.py
```diff
@@ -0,0 +1,63 @@
...
+
+        return flask.Response(notes, mimetype="text/plain")
+    except Exception as e:
```
suhaibmujahid, line 41: The exception handling should be limited to situations where an error is expected.

functions/release_notes/main.py
```diff
@@ -0,0 +1,63 @@
...
+        logger.exception("Failed to generate release notes")
+        return flask.Response(f"Internal Server Error: {str(e)}", status=500)
```
suhaibmujahid, line 43: Returning the expectation to the public is unsafe, as it may reveal unforeseen security risks.

.github/workflows/release_notes.yml
```diff
@@ -0,0 +1,41 @@
+name: Deploy Release Notes Function
+
+on:
+  push:
+    tags:
+      - "*"
```
suhaibmujahid, line 6: It would be better to trigger this manually when needed. 
```suggestion
on: workflow_dispatch
```

suhaibmujahid (changes requested):

functions/release_notes/main.py
```diff
@@ -0,0 +1,63 @@
...
+    version = request.args.get("version")
+    if not version:
+        return flask.Response("Missing 'version' query parameter", status=400)
+
+    try:
+        os.environ["OPENAI_API_KEY"] = get_secret("OPENAI_API_KEY")
+    except Exception as e:
+        return flask.Response(f"Failed to load OpenAI key: {str(e)}", status=500)
+
+    args = build_args_from_request(request)
```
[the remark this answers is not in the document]
benjaminmah: Originally it was because I wanted to reuse the functions we had in `generative_model_tool.py`. I've created a new function that doesn't require CLI args - please let me know if this looks ok: 2c5d73c1df6941b544cb751cbb6573969897514c.

functions/release_notes/main.py
benjaminmah: 2c5d73c1df6941b544cb751cbb6573969897514c

functions/release_notes/main.py
benjaminmah: Removed the exception handler here: 2c5d73c1df6941b544cb751cbb6573969897514c

functions/release_notes/main.py
benjaminmah: Removed the entire exception handler here: 2c5d73c1df6941b544cb751cbb6573969897514c

.github/workflows/release_notes.yml
benjaminmah: Changed here: 2c5d73c1df6941b544cb751cbb6573969897514c

functions/release_notes/main.py
```diff
@@ -0,0 +1,63 @@
...
+
+    try:
+        llm = generative_model_tool.create_llm_from_args(args)
+        selector = ReleaseNotesCommitsSelector(chunk_size=args.chunk_size, llm=llm)
```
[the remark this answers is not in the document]
benjaminmah: Done here: c79890e7345252a8ffcafa3928418f9a53de5b09

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,250 @@
...
+            parts = line.split(",", 3)
+            if len(parts) < 3:
+                continue
+            bug_id = parts[2].strip()
+            if bug_id not in seen:
+                seen.add(bug_id)
+                unique_lines.append(line)
+        return "\n".join(unique_lines)
+
+    def get_final_release_notes_commits(self, version: str) -> Optional[str]:
```
suhaibmujahid, line 222: Why are we returning the result as a string instead of a list of dictionaries?

functions/release_notes/main.py
```diff
@@ -0,0 +1,57 @@
...
+            chunk_size=chunk_size, llm=llm_cache["llm"]
+        )
+
+    selector = llm_cache["selector"]
+    notes = selector.get_final_release_notes_commits(version=version)
+
+    if not notes:
+        return "No user-facing commits found.", 404
+
+    return notes, 200, {"Content-Type": "text/plain"}
```
suhaibmujahid, line 57: Why is the response not JSON here?

functions/release_notes/main.py
```diff
@@ -0,0 +1,57 @@
...
+def handle_release_notes(request: flask.Request):
+    if request.method != "GET":
+        return "Only GET requests are allowed", 405
+
+    version = request.args.get("version")
+    if not version:
+        return "Missing 'version' query parameter", 400
+
+    llm_name = request.args.get("llm", "openai")
+    chunk_size = int(request.args.get("chunk_size", 100))
```
suhaibmujahid, line 34: I do not think we need to get the model name and chunk size from the request. Is there a specific reason for doing that?

functions/release_notes/main.py
```diff
@@ -0,0 +1,57 @@
...
+logger = logging.getLogger(__name__)
+
+os.environ["OPENAI_API_KEY"] = get_secret("OPENAI_API_KEY")
+
+llm_cache = {
+    "llm_name": None,
+    "llm": None,
+    "chunk_size": None,
+    "selector": None,
+}
```
suhaibmujahid, line 21: We could replace the dictionary with one value, i.e., `tool`, to store an instance of the `ReleaseNotesCommitsSelector` class.

functions/release_notes/main.py
```diff
@@ -0,0 +1,57 @@
...
+        llm_cache["chunk_size"] = chunk_size
+        llm_cache["selector"] = ReleaseNotesCommitsSelector(
+            chunk_size=chunk_size, llm=llm_cache["llm"]
+        )
+
+    selector = llm_cache["selector"]
+    notes = selector.get_final_release_notes_commits(version=version)
+
+    if not notes:
+        return "No user-facing commits found.", 404
```
suhaibmujahid, line 55: We are not retrieving the notes here; we are generating them, so it does not fit to use 404 when we were not able to generate them.

I suggest instead returning an empty list with a 200 response.

suhaibmujahid (changes requested):

functions/release_notes/main.py
benjaminmah: Would you prefer that each commit is structures as a JSON? Currently, the function returns a plaintext list of commits.

functions/release_notes/main.py
benjaminmah: I've done that here: 3e9c7f71f9e5ec3970ad22716de15151a77bfe12

functions/release_notes/main.py
benjaminmah: I've changed the output to be a Python list, subsequently making the cloud function output a JSON: 0da3e8a020c2776d86b73d8535c0bfdcf2ac76f6

functions/release_notes/main.py
```diff
@@ -0,0 +1,49 @@
...
+        return "Only GET requests are allowed", 405
+
+    version = request.args.get("version")
+    if not version:
+        return "Missing 'version' query parameter", 400
+
+    if (
+        tool is None
+        or tool.llm_name != DEFAULT_LLM_NAME
+        or tool.chunk_size != DEFAULT_CHUNK_SIZE
```
suhaibmujahid, line 36: I do not think this is necessary. It will not be changing at runtime.  
```suggestion
```

functions/release_notes/main.py
```diff
@@ -0,0 +1,49 @@
...
+        logger.info("Initializing new ReleaseNotesCommitsSelector...")
+        llm = generative_model_tool.create_llm_from_request(DEFAULT_LLM_NAME, {})
+        tool = ReleaseNotesCommitsSelector(chunk_size=DEFAULT_CHUNK_SIZE, llm=llm)
+        tool.llm_name = DEFAULT_LLM_NAME
+        tool.chunk_size = DEFAULT_CHUNK_SIZE
+
+    notes = tool.get_final_release_notes_commits(version=version)
+
+    if not notes:
+        return {"commits": []}, 200, {"Content-Type": "application/json"}
```
suhaibmujahid, line 47: I think the response type will be determined automatically, so you do not need to specify it:
```suggestion
        return {"commits": []}, 200
```

bugbug/generative_model_tool.py
```diff
@@ -147,6 +147,28 @@ def create_llm_from_args(args):
     return globals()[f"create_{args.llm}_llm"](**llm_creation_args)
 
 
+def create_llm_from_request(llm_name, request_args):
```
suhaibmujahid, line 150: Why do we need this?

benjaminmah: I created this because I wanted to keep the flexibility of choosing the LLM provider without requiring CLI args (`create_llm_from_args()`). However, I've removed it here if we're only going to be using OpenAI: 9b07b5b039f7e8ebfcb9bae36a8110f07e211a68

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,250 @@
...
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+class ReleaseNotesCommitsSelector:
+    def __init__(self, chunk_size: int, llm: LLMChain):
+        self.chunk_size = chunk_size
+        self.bug_id_to_component = {}
+        db.download(bugzilla.BUGS_DB)
```
suhaibmujahid, line 55: It is too much to download the whole bugs database to check a small set of them. We could instead deal with the ones that we need by their IDS. 

See: https://github.com/mozilla/bugbug/blob/bba1d097f968ef18ec5687f7e3064c792660c702/bugbug/bugzilla.py#L219

suhaibmujahid: Even better, you could use `libmozdata.bugzilla` to request just the bugs you need, including only the necessary fields.

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,250 @@
+import logging
+import re
+from itertools import batched
+from typing import Generator, Optional
```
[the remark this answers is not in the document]
benjaminmah: Done here: e852e9d9bb9714e55fe83189c984a17e6a2330ab

functions/release_notes/main.py
```diff
@@ -0,0 +1,45 @@
...
+import functions_framework
+
+from bugbug import generative_model_tool
+from bugbug.tools.release_notes import ReleaseNotesCommitsSelector
+from bugbug.utils import get_secret
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+os.environ["OPENAI_API_KEY"] = get_secret("OPENAI_API_KEY")
```
suhaibmujahid, line 14: When using `create_openai_llm()`, it will read the secret directly. Why do we need this?

functions/release_notes/main.py
```diff
@@ -0,0 +1,45 @@
...
+    if not version:
+        return "Missing 'version' query parameter", 400
+
+    if tool is None:
+        logger.info("Initializing new ReleaseNotesCommitsSelector...")
+
+        llm = generative_model_tool.create_openai_llm()
+        tool = ReleaseNotesCommitsSelector(chunk_size=DEFAULT_CHUNK_SIZE, llm=llm)
+        tool.llm_name = "openai"
+        tool.chunk_size = DEFAULT_CHUNK_SIZE
```
suhaibmujahid, line 38: Why do we need this here?

functions/release_notes/main.py
```diff
@@ -0,0 +1,45 @@
...
+
+    notes = tool.get_final_release_notes_commits(version=version)
+
+    if not notes:
+        return {"commits": []}, 200
```
suhaibmujahid, line 43: I do not think we need this.

functions/release_notes/requirements.txt
```diff
@@ -0,0 +1,6 @@
+bugbug
+Flask==2.2.5
+functions-framework==3.5.0
+langchain
+openai
+requests
```
suhaibmujahid, line 6: We do not use them directly in the cloud function; we could depend on bugbug to install them.

functions/release_notes/requirements.txt
```diff
@@ -0,0 +1,6 @@
+bugbug
```
suhaibmujahid, line 1: We should lock to a specific bugbug version.

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,266 @@
...
+                continue
+            bug_id = parts[2].strip()
+            if bug_id not in seen:
+                seen.add(bug_id)
+                unique_lines.append(line)
+        return "\n".join(unique_lines)
+
+    def get_final_release_notes_commits(self, version: str) -> Optional[list[str]]:
+        self.version2 = version
+        self.version1 = get_previous_version(version)
```
suhaibmujahid, line 233: Two issues here:
1. The names `version1` and `version2` are a bit confusing. We could use something like `target_version` and `preceding_version`.
2. We should not store them as attributes, as this will be a side effect in the class since we could reuse the class across different target versions. If multiple requests are sent at the same time, we will not guarantee that the results belong to the provided version.

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,266 @@
...
+        commit_shortlist = self.generate_commit_shortlist(filtered_commits)
+
+        if not commit_shortlist:
+            return None
+
+        logger.info("Refining commit shortlist...")
+        combined_list = "\n".join(commit_shortlist)
+        cleaned = self.cleanup_chain.run({"combined_list": combined_list}).strip()
+
+        logger.info("Removing duplicates...")
```
suhaibmujahid, line 264: I'm not a fan of too much logging. It will be useful to log in to places where it could help with tracing and debugging.

benjaminmah: Makes sense. I've removed some logging here: 11a6444800631fbf8c9322c21d9b9c54be607190

bugbug/tools/release_notes.py
benjaminmah: This makes sense! I've renamed the variables and have not stored them as attributes: 11a6444800631fbf8c9322c21d9b9c54be607190

functions/release_notes/requirements.txt
benjaminmah: I've locked it to the most recent bugbug version, is this okay? 11a6444800631fbf8c9322c21d9b9c54be607190

functions/release_notes/requirements.txt
benjaminmah: Removed here: 11a6444800631fbf8c9322c21d9b9c54be607190

functions/release_notes/main.py
benjaminmah: Removed here: 11a6444800631fbf8c9322c21d9b9c54be607190

functions/release_notes/main.py
benjaminmah: I think I forgot to remove this when I refactored the tool to use `create_openai_llm()`. Removed here: 11a6444800631fbf8c9322c21d9b9c54be607190

functions/release_notes/main.py
```diff
@@ -0,0 +1,29 @@
...
+
+    version = request.args.get("version")
+    if not version:
+        return "Missing 'version' query parameter", 400
+
+    if tool is None:
+        llm = generative_model_tool.create_openai_llm()
+        tool = ReleaseNotesCommitsSelector(chunk_size=DEFAULT_CHUNK_SIZE, llm=llm)
+
+    notes = tool.get_final_release_notes_commits(target_version=version)
```
suhaibmujahid, line 27: Are they comment messages or notes?

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,262 @@
...
+            if len(parts) < 3:
+                continue
+            bug_id = parts[2].strip()
+            if bug_id not in seen:
+                seen.add(bug_id)
+                unique_lines.append(line)
+        return "\n".join(unique_lines)
+
+    def get_final_release_notes_commits(
+        self, target_version: str
```
suhaibmujahid, line 232: Would taking the release and the channel separately, instead of one string, be cleaner?

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,262 @@
...
+        preceding_version = get_previous_version(target_version)
+        logger.info(f"Generating commit shortlist for: {target_version}")
+        commit_log_list = self.get_commit_logs(preceding_version, target_version)
+
+        if not commit_log_list:
+            return None
+
+        bug_ids = []
+        for desc, _, _ in commit_log_list:
+            match = re.search(r"Bug (\d+)", desc, re.IGNORECASE)
```
suhaibmujahid, line 243: We parse the commit message twice, once here and again in `filter_irrelevant_commits()`. It would be nice to do it once here; we could have a list of structured data about each commit, e.g., a tuple or dict with the bug number and message.

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,262 @@
...
+
+3. Be Aggressive in Filtering:
+    - If you're unsure whether a commit impacts end users, EXCLUDE it.
+    - Do not list developer-focused changes.
+
+4. Select Only the Top 10 Commits:
+    - If there are more than 10 relevant commits, choose the most impactful ones.
+
+5. Input:
+   Here is the chunk of commit logs you need to focus on:
```
suhaibmujahid, line 102:
```suggestion
   Here is the list of commits you need to focus on:
```

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,262 @@
...
+   {input_text}
```
suhaibmujahid, line 103: Could you please use a descriptive variable name in the template?

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,262 @@
...
+Strict Filtering Criteria - REMOVE the following:
+- Overly technical web platform changes (e.g., spec compliance tweaks, behind-the-scenes API adjustments).
+- Developer-facing features that have no direct user impact.
+- Minor UI refinements (e.g., button width adjustments, small animation tweaks).
+- Bug fixes that don’t impact most users.
+- Obscure web compatibility changes that apply only to edge-case websites.
+- Duplicate entries or similar changes that were already listed.
+
+Here is the list to filter:
+{combined_list}
```
suhaibmujahid, line 135: Why is this not at the end?

benjaminmah: I've moved it to the end. I've also done the same for the other prompts: aebee0ae86ea27d28ae80769d788bc71fd3771be

bugbug/tools/release_notes.py
benjaminmah: Changed the variable name to `commit_list`: aebee0ae86ea27d28ae80769d788bc71fd3771be

bugbug/tools/release_notes.py
benjaminmah: Changed here: aebee0ae86ea27d28ae80769d788bc71fd3771be

bugbug/tools/release_notes.py
benjaminmah: Done here! aebee0ae86ea27d28ae80769d788bc71fd3771be

functions/release_notes/main.py
benjaminmah: They're just a list of commits. I've change the variable name accordingly here: aebee0ae86ea27d28ae80769d788bc71fd3771be

bugbug/tools/release_notes.py
benjaminmah: I think the way we're getting the commit logs from the URL, it should be fine. However, if you think it would be better, I can change it to have two inputs rather than a single string for the release and channel.

suhaibmujahid: I think it would be clear from an API point of view.

bugbug/tools/release_notes.py
suhaibmujahid: Did you verify whether this affected accuracy?

bugbug/tools/release_notes.py
benjaminmah: Done here: 38499c365d2734ffc43dcb5566262f1e8cda0e0d

bugbug/tools/release_notes.py
benjaminmah: From visual inspection, it seemed to me more or less the same due to the slight variability of selected commits. However, I believe that this overall will improve accuracy due to the raw commits list being at the end of the prompt.

bugbug/tools/release_notes.py
```diff
@@ -0,0 +1,273 @@
...
+    if not match:
+        raise ValueError("No number found in the version string")
+
+    number = match.group(0)
+    decremented_number = str(int(number) - 1)
+    return (
+        target_version[: match.start()]
+        + decremented_number
+        + target_version[match.end() :]
+    )
```
suhaibmujahid, line 43: Is this still needed?

tests/test_release_notes.py
```diff
@@ -0,0 +1,7 @@
+from bugbug.tools.release_notes import get_previous_version
+
+
+def test_get_previous_version():
+    assert get_previous_version("FIREFOX_BETA_135_BASE") == "FIREFOX_BETA_134_BASE"
+    assert get_previous_version("FIREFOX_NIGHTLY_132") == "FIREFOX_NIGHTLY_131"
+    assert get_previous_version("FIREFOX_RELEASE_130_2") == "FIREFOX_RELEASE_129_2"
```
suhaibmujahid, line 7: How do we now handle cases like `FIREFOX_RELEASE_130_2`?

bugbug/tools/release_notes.py
benjaminmah: No, thank you for pointing this out. I've removed it!

PR #4868, opened 2025-03-06
[code_review] Attach comments to the line returned by the LLM

### LINKED ISSUE #4643: [code_review] Attache the comment to the line returned by the LLM instead of its chunk
Suggested by @marco-c

suhaibmujahid:
Resolves #4643

### DIFF
--- a/bugbug/tools/code_review.py
+++ b/bugbug/tools/code_review.py
@@ -45,6 +45,8 @@ class InlineComment:
     date_created: int | None = None
     date_modified: int | None = None
     is_done: bool | None = None
+    hunk_start_line: int | None = None
+    hunk_end_line: int | None = None
 
 
 class ModelResultError(Exception):
@@ -1042,11 +1044,10 @@ def generate_processed_output(output: str, patch: PatchSet) -> Iterable[InlineCo
                 if scope["has_added_lines"]
                 else patched_file.source_file[2:]
             ),
-            # Note(suhaib): We were previously using the hunk's scope instead of
-            # the line number returned by the model. This was changed to reduce
-            # confusion: https://github.com/mozilla/bugbug/issues/4643
             start_line=line_number,
             end_line=line_number,
+            hunk_start_line=scope["line_start"],
+            hunk_end_line=scope["line_end"],
             content=comment["comment"],
             on_removed_code=not scope["has_added_lines"],
         )
--- a/bugbug/tools/code_review.py
+++ b/bugbug/tools/code_review.py
@@ -1042,8 +1042,11 @@ def generate_processed_output(output: str, patch: PatchSet) -> Iterable[InlineCo
                 if scope["has_added_lines"]
                 else patched_file.source_file[2:]
             ),
-            start_line=scope["line_start"],
-            end_line=scope["line_end"],
+            # Note(suhaib): We were previously using the hunk's scope instead of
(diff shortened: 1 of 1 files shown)


marco-c:
Does this affect the evaluation in any way?

suhaibmujahid:
> Does this affect the evaluation in any way?

Currently, we do not test if the matched comments belong to the same hunk, which could reduce the FP in the matching process. To match based on hunks, the hunk scope (which we just dropped) will be needed. However, we could generate it on the fly (it would be a bit slower since we need to retrieve and parse the diff again).

Ideally, we could return and store the hunk scope alongside the actual line numbers.

marco-c (dismissed):

marco-c (approved):

kitsiosk:
Hi! 🤖 The test below is automatically generated and serves as a regression test for this PR because it:
- passes, and
- fails in the codebase before the PR.

```python
def test_generate_processed_output_attaches_comment_to_correct_line():
    from bugbug.tools.code_review import generate_processed_output
    from unidiff import PatchSet
    from bugbug.tools.code_review import InlineComment

    # Mock output from the model
    output = json.dumps([
        {
            "file": "example.py",
            "code_line": 10,
            "comment": "This is a test comment."
        }
    ])

    # Create a mock patch set
    patch_content = """\
diff --git a/example.py b/example.py
--- a/example.py
+++ b/example.py
@@ -8,0 +9,2 @@
+def example_function():
+    pass
"""
    patch = PatchSet(patch_content)

    # Generate processed output
    comments = list(generate_processed_output(output, patch))

    # Check that the comment is attached to the correct line
    assert len(comments) == 1
    assert isinstance(comments[0], InlineComment)
    assert comments[0].filename == "example.py"
    assert comments[0].start_line == 10
    assert comments[0].end_line == 10
    assert comments[0].content == "This is a test comment."
```

If you find this regression test useful, feel free to insert it to your test suite.
Our automated pipeline inserted the test at the end of the `tests/test_repository.py` file before running it.

This is part of our research at the [ZEST](https://www.ifi.uzh.ch/en/zest.html) group of University of Zurich in collaboration with [Mozilla](https://www.mozilla.org).
If you have any suggestions, questions, or simply want to learn more, feel free to contact us at konstantinos.kitsios@uzh.ch and mcastelluccio@mozilla.com.

<details>
<summary> Click to see which lines were covered.</summary>

```diff
diff --git a/bugbug/tools/code_review.py b/bugbug/tools/code_review.py
--- a/bugbug/tools/code_review.py
+++ b/bugbug/tools/code_review.py
@@ -1042,8 +1042,11 @@
                 if scope["has_added_lines"]
                 else patched_file.source_file[2:]
             ),
-            start_line=scope["line_start"],
-            end_line=scope["line_end"],
+            # Note(suhaib): We were previously using the hunk's scope instead of# ✅ Covered by above test
+            # the line number returned by the model. This was changed to reduce# ✅ Covered by above test
+            # confusion: https://github.com/mozilla/bugbug/issues/4643# ✅ Covered by above test
+            start_line=line_number,# ✅ Covered by above test
+            end_line=line_number,# ✅ Covered by above test
             content=comment["comment"],
             on_removed_code=not scope["has_added_lines"],
         )

```

Line coverage\* achieved: 100.0%

\* Line coverage is calculated over the lines added in this PR.

<details>

PR #4923, opened 2025-04-02
Increase the bugzilla dataset version to trigger data fetching from scratch

### LINKED ISSUE #4922: The current Bugzilla dataset uses the product name "Fenix" instead of "Firefox for Android"
The scheduled dataset updates will add new bugs to the dataset and will update modified bugs, but they will not reflect the product rename on bugs that are already in the dataset.

suhaibmujahid:
Resolves #4922 

We need to release a new version after merging this.


marco-c:
This should not be necessary, as this case is covered by https://github.com/mozilla/bugbug/blob/c050d72a40899e16343fcbae483280e3f47a1a86/scripts/bug_retriever.py#L50-L61

suhaibmujahid:
@marco-c Thank you!

PR #4932, opened 2025-04-03
[bug_retriever] Re-download bugs that are missing history

### LINKED ISSUE #4931: Re-download bugs that are missing history
> let's do the same for history, as currently we're not retrying downloading them

_Originally posted by @marco-c in https://github.com/mozilla/bugbug/pull/4930#discussion_r2027351812_

suhaibmujahid:
Resolves #4931

### DIFF
--- a/scripts/bug_retriever.py
+++ b/scripts/bug_retriever.py
@@ -193,67 +193,59 @@ class Retriever(object):
             bugzilla.download_bugs(inconsistent_bug_ids)
 
         # TODO: Figure out why.
-        missing_history_bug_ids = {
-            bug["id"]
-            for bug in bugzilla.get_bugs(
-                include_invalid=True,
-                include_additional_products=bugzilla.ADDITIONAL_PRODUCTS,
-            )
-            if "history" not in bug
-        }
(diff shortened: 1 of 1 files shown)


scripts/bug_retriever.py
```diff
@@ -193,67 +193,59 @@ def retrieve_bugs(self, limit: int | None = None) -> None:
...
 
 
-def handle_missing_comments(trial_number: int = 1, max_tries: int = 2) -> int:
-    """Handle bugs that are missing comments.
-
-    This function will try to re-download the bugs that are missing comments
-    and delete them if they are still missing comments after the maximum number
-    of tries.
+def handle_missing_field(
+    field_name: str, trial_number: int = 1, max_tries: int = 2
```
marco-c, line 205: nit: it would be faster if we supported a list of fields (so we iterate through bugs fewer times in total)

suhaibmujahid: Yes, I thought of that, but we would lose the detailed logging of the number of bugs for each case.

suhaibmujahid: Done in 6a637176a56066060adf0146ba81bbe76a0fb155

marco-c (approved):

PR #5005, opened 2025-05-08
[code_review] Add explanation to the generated comments

### LINKED ISSUE #4785: [code_review] Add explanations to generated comments
This could help developers to understand if a comment is correct or not.

Also, it might elicit more reasoning.

The explanation would be shown but not published. Only the comment would be published.

suhaibmujahid:
Resolves #4785

### DIFF
--- a/bugbug/tools/code_review.py
+++ b/bugbug/tools/code_review.py
@@ -51,6 +51,7 @@ class InlineComment:
     hunk_start_line: int | None = None
     hunk_end_line: int | None = None
     is_generated: bool | None = None
+    explanation: str | None = None
 
 
 class ModelResultError(Exception):
@@ -118,7 +119,7 @@ Generate code review comments for the patch provided below.
 **Output Format**:
 
 - Write down the comments in a JSON list as shown in the valid comment examples.
-- Do **not** include any explanations about your choices.
+- Include your justification about your choices as part of each JSON object under the key `explanation`.
 - Only return the JSON list.
 
 **Valid Comment Examples**:
@@ -152,7 +153,7 @@ Exclude comments that:
 - include praising;
 - ask if changes are intentional or ask to ensure things exist.
 
-Do not report any explanation about your choice. Only return a valid JSON list.
+Only return a valid JSON list. Do not drop any key from the JSON objects.
 
 Comments:
 {comments}
@@ -181,6 +182,7 @@ Adopt the template below as the report format:
         "file": "com/br/main/Pressure.java",
         "code_line": 458,
         "comment" : "In the third code block, you are using `nsAutoStringN<256>` instead of `nsString`. This is a good change as `nsAutoStringN<256>` is more efficient for small strings. However, you should ensure that the size of `tempString` does not exceed 256 characters, as `nsAutoStringN<256>` has a fixed size."
+        "explanation": "THE JUSTIFICATION GOES HERE"
     }}
 ]
 Do not report any explanation about your choice. Only return a valid JSON list.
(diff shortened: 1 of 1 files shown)


suhaibmujahid:
The evaluation results:

```
--------------------
Variant Name: with-explanation
--------------------
New Comments: 331
New Valid Comments: 46
New Invalid Comments: 97
New Unevaluated Comments: 188
--------------------
Old Comments: 273
Old Valid Comments: 84
Old Invalid Comments: 189
--------------------
Recalled comments: 55.67765567765568
Recalled valid comments: 54.761904761904766
Recalled invalid comments: 56.08465608465608
--------------------
Missed valid comments: 45.23809523809524
Missed invalid comments: 43.91534391534391
```

This can be compared to https://github.com/mozilla/bugbug/pull/5002#issuecomment-2859802412. The results look similar, with a minor improvement.

bugbug/tools/code_review.py
```diff
@@ -1401,6 +1407,9 @@ def _get_comment_examples(self, patch, created_before: datetime | None = None):
 
         if not comment_examples:
             comment_examples = STATIC_COMMENT_EXAMPLES
+        else:
+            for example in comment_examples:
+                example["comment"]["explanation"] = "THE JUSTIFICATION GOES HERE"
```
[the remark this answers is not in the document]
suhaibmujahid: Invalid suggestion. The explanation needs to be part of the comment dict.

bugbug/tools/code_review.py
```diff
@@ -51,6 +51,7 @@ class InlineComment:
     hunk_start_line: int | None = None
     hunk_end_line: int | None = None
     is_generated: bool | None = None
+    explanation: str | None = None
```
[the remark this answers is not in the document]
suhaibmujahid: That will not be consistent with the rest of the attributes.

bugbug/tools/code_review.py
```diff
@@ -152,7 +153,7 @@ class LargeDiffError(Exception):
 - include praising;
 - ask if changes are intentional or ask to ensure things exist.
 
-Do not report any explanation about your choice. Only return a valid JSON list.
+Only return a valid JSON list. Do not drop any key from the JSON objects.
```
marco-c, line 156: Is the second part needed?

bugbug/tools/code_review.py
```diff
@@ -212,6 +214,7 @@ class LargeDiffError(Exception):
             "filename": "netwerk/streamconv/converters/mozTXTToHTMLConv.cpp",
             "start_line": 1211,
             "content": "You are using `nsAutoStringN<256>` instead of `nsString`. This is a good change as `nsAutoStringN<256>` is more efficient for small strings. However, you should ensure that the size of `tempString` does not exceed 256 characters, as `nsAutoStringN<256>` has a fixed size.",
+            "explanation": "THE JUSTIFICATION GOES HERE",
```
marco-c, line 217: We should maybe add one

bugbug/tools/code_review.py
marco-c, line 1412: Here we should reuse the old generated explanations when there are some

suhaibmujahid: I'm not sure that would add value, especially since it will not be consistent when some comments have justification value and others do not. But it will definitely consume more tokens.

I added the placeholder here to avoid situations where the explanation is omitted.

bugbug/tools/code_review.py
suhaibmujahid: The generated explanation seems reasonable, even without it.

bugbug/tools/code_review.py
suhaibmujahid: At the beginning, it was without the sound part. However, the filtering step was omitting the `explanation` key from he final result.

bugbug/tools/code_review.py
marco-c: Let's file a follow-up issue to add it, and then we can compare the results

bugbug/tools/code_review.py
marco-c: Let's file a follow-up issue to consider adding it (and we could auto-generate it when it is missing)

marco-c (approved):

bugbug/tools/code_review.py
suhaibmujahid: I filed https://github.com/mozilla/bugbug/issues/5017

bugbug/tools/code_review.py
suhaibmujahid: I filed https://github.com/mozilla/bugbug/issues/5017

PR #5019, opened 2025-05-14
[code_review] Rank comments based on importance

### LINKED ISSUE #5018: [code_review] Experiment with ranking comments based on importance
Ranking comments based on importance could help limit their number while retaining the most significant ones.

suhaibmujahid:
Resolves #5018 

I will report the experiment results here: https://docs.google.com/spreadsheets/d/1CIskQot4TVxHFXxOpXpYMReERuOobqeYQh25bmPHjxQ

### DIFF
--- a/bugbug/tools/code_review.py
+++ b/bugbug/tools/code_review.py
@@ -52,6 +52,7 @@ class InlineComment:
     hunk_end_line: int | None = None
     is_generated: bool | None = None
     explanation: str | None = None
+    order: int | None = None
 
 
 class ModelResultError(Exception):
@@ -120,6 +121,7 @@ Generate code review comments for the patch provided below.
 
 - Write down the comments in a JSON list as shown in the valid comment examples.
 - Include your justification about your choices as part of each JSON object under the key `explanation`.
+- Include the key `order` for each comment, which starts from 1 and goes up to the number of comments.
 - Only return the JSON list.
 
 **Valid Comment Examples**:
@@ -1109,6 +1111,7 @@ def generate_processed_output(output: str, patch: PatchSet) -> Iterable[InlineCo
             content=comment["comment"],
             on_removed_code=not scope["has_added_lines"],
             explanation=comment["explanation"],
+            order=comment["order"],
         )
 
 
--- a/bugbug/tools/code_review.py
+++ b/bugbug/tools/code_review.py
@@ -81,58 +81,64 @@ Please, analyze the code provided and report a summarization about the new chang
 
 PROMPT_TEMPLATE_REVIEW = """**Task**:
 
-Generate code review comments for the patch provided below.
+Generate high-quality code review comments for the patch provided below.
 
 **Instructions**:
 
-1. **Understand the Changes**:
+1. **Analyze the Changes**:
 
-   - Analyze the changes made in the patch.
-   - Use the previously reported summarization as context.
+   * Understand the intent and structure of the changes in the patch.
+   * Use the provided summarization for context, but prioritize what’s visible in the diff.
 
-2. **Identify Potential Issues**:
+2. **Identify Issues**:
 
-   - Look for code that might result in possible bugs.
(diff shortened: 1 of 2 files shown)


bugbug/tools/code_review.py
```diff
@@ -80,53 +81,65 @@ class LargeDiffError(Exception):
...
 
 **Instructions**:
 
-1. **Understand the Changes**:
+1. **Analyze the Changes**:
 
-   - Analyze the changes made in the patch.
-   - Use the previously reported summarization as context.
+   * Understand the intent and structure of the changes in the patch.
+   * Use the provided summarization for context, but prioritize what’s visible in the diff.
```
marco-c, line 91:
```suggestion
   * Use the provided summarization for context, but prioritize what's visible in the diff.
```

scripts/code_review_tool_evaluator.py
```diff
@@ -33,7 +33,7 @@
 from bugbug.vectordb import QdrantVectorDB
 
 code_review.TARGET_SOFTWARE = "Mozilla Firefox"
-VERBOSE_CODE_REVIEW = False
+VERBOSE_CODE_REVIEW = True
```
marco-c, line 36: Was this change for debugging or do you intend to set it to True by default?

bugbug/tools/code_review.py
```diff
@@ -80,53 +81,65 @@ class LargeDiffError(Exception):
...
+   * Focus strictly on code-related concerns.
+   * Avoid hedging language (e.g., don’t use “maybe”, “might want to”, or form questions).
+   * Avoid repeating what the code is doing unless it supports your critique.
 
-- **Avoid Comments That**:
+**Avoid Comments That**:
 
-  - Focus on documentation or tests.
-  - Ask for confirmation of existence (e.g., "Does this function exist?").
-  - Instruct the author to ensure correctness (avoid "Ensure", "Verify", "Check").
```
marco-c, line 114: Shouldn't we keep this to avoid comments asking if something exists?

scripts/code_review_tool_evaluator.py
suhaibmujahid: It was for debugging. I did undo it in 5e94ddceca596aac5d3dcb78e14313b11d62688b.

bugbug/tools/code_review.py
suhaibmujahid: It is kind of covered with:
> Ask for verification or confirmation (e.g., “Check if…”)

bugbug/tools/code_review.py
```diff
@@ -80,53 +81,65 @@ class LargeDiffError(Exception):
...
-  - Merely describe the code without providing constructive feedback.
-  - Offer praise (e.g., "This is a good addition to the code.").
-  - Refer to code not added in this patch (lines without a '+' at the start).
+* Refer to unmodified code (lines without a `+` prefix).
+* Ask for verification or confirmation (e.g., “Check if…”).
+* Provide praise or restate obvious facts.
+* Focus on documentation or testing.
```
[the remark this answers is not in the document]
suhaibmujahid: IIRC, we added it because we had cases where many comments were about adding documentation comments.

I will remove it, and we can add it back if we see something we do not want.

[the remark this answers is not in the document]
suhaibmujahid: Done in 359af8d15c0d8ff04a539dc515326b559997259b

bugbug/tools/code_review.py
marco-c: Could you check if the number of comments asking to verify something exists are more or less the same before/after the change (by using a LLM, even if not perfect)?
Perhaps this could be an additional metric in the evaluation? Maybe file a follow-up for this

suhaibmujahid: Filed https://github.com/mozilla/bugbug/issues/5046

PR #5038, opened 2025-05-23
Opst 767

steveprokopienko:


.github/workflows/docker.yml
```diff
@@ -0,0 +1,59 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install docker-compose
+        run: sudo apt update && sudo apt install -y docker-compose
+      - name: Generate version.json
+        shell: bash
+        run: |
+          printf '{"commit":"%s","version":"%s","source":"https://github.com/%s","build":"%s/%s/actions/runs/%s"}\n' \
+          "$GITHUB_SHA" \
+          "$GITHUB_REF_NAME" \
```
suhaibmujahid, line 22: Regarding the use of `GITHUB_REF_NAME` as the version, I want to point out that we currently use `latest` as a tag when things are based on the master branch. See: https://github.com/mozilla/bugbug/blob/359af8d15c0d8ff04a539dc515326b559997259b/.taskcluster.yml#L175-L178

.github/workflows/docker.yml
suhaibmujahid, line 1: @marco-c, are we ok with moving forward with GitHub Actions here instead of Taskcluster? This might add a bit of complexity when we want to coordinate this with other tasks on Taskcluster as dependencies.

steveprokopienko: This GHA workflow will not be replacing any existing setup with Taskcluster.
This workflow just ensures we have the images in GAR

marco-c: The only issue I see is that the http service Docker image can't be built in CI, it needs to be built by the pipeline after training the models.
In CI we can only build the docker images used in the pipeline itself, but the docker image for the http service needs to be built within the pipeline after training the models because it needs to download the models: https://github.com/mozilla/bugbug/blob/e6d439baa58e252341ba63a6138782c0f158852c/http_service/Dockerfile.bg_worker#L24.

marco-c: Note we don't necessarily need this behavior, it was mostly done for two reasons:
- performance optimization to reduce startup time of the image (particularly important on Heroku not having to redownload models all the time, given they restart workers daily);
- full reproducibility (an old image will always be reproducible because it contains all it needs to run).

marco-c (approved):
LGTM

PR #5052, opened 2025-05-28
Don't regenerate bug_map dict on every call to get_blocking_of

marco-c:


scripts/generate_landings_risk_report.py
```diff
@@ -895,7 +899,9 @@ def go(self, days: int) -> None:
             ):
                 bug_map[bug["id"]] = bug
 
-        self.generate_landings_by_date(bug_map, bugs, self.get_blocking_of(meta_bugs))
+        self.generate_landings_by_date(
+            bug_map, bugs, self.get_blocking_of(bug_map, meta_bugs)
```
suhaibmujahid, line 903: As a follow-up, what do you think about adding bug_map as a cached property (i.e., `@cached_property`)? In that way, we do not need to pass it with the function calls.

scripts/generate_landings_risk_report.py
```diff
@@ -357,9 +357,11 @@ def get_landed_and_filed_since(self, days: int) -> list[int]:
         return list(set(commit["bug_id"] for commit in commits) | set(timespan_ids))
 
     def get_blocking_of(
-        self, bug_ids: list[int], meta_only: bool = False
+        self,
+        bug_map: dict[int, bugzilla.BugDict],
+        bug_ids: list[int],
+        meta_only: bool = False,
     ) -> dict[int, list[int]]:
-        bug_map = {bug["id"]: bug for bug in bugzilla.get_bugs()}
```
suhaibmujahid, line 362: The map that we are passing now is not the same one. Is that ok?

suhaibmujahid (approved):

PR #5077, opened 2025-06-11
[code_review] Integrate ML filter for the generated comments

ArezouAmini:


bugbug/ml_filter_finetune_tool.py
```diff
@@ -0,0 +1,97 @@
+from abc import ABC, abstractmethod
+from pathlib import Path
+
+import torch
+from datasets import Dataset
+from torch.nn.functional import softmax
+from transformers import (
```
suhaibmujahid, line 7: I think some of the dependencies are not in the `requirements.json`.

bugbug/ml_filter_trainer_tool.py
```diff
@@ -0,0 +1,51 @@
...
+        min_recall: float = 0.9,
+        thr_metric: str = "acceptance_rate",
+        tmpdir: Path = Path(""),
+    ):
+        self.min_recall = min_recall
+        self.thr_metric = thr_metric
+        self.tmpdir = tmpdir
+
+    @abstractmethod
+    def train_test_split(self, data, test_size=0.5, random_split=True): ...
```
suhaibmujahid, line 20: What should the data look like? I do not see the method used anywhere!

suhaibmujahid (commented):
@ArezouAmini Thank you for submitting the PR! 

Could you please clarify the relation between the three classes that you added?

PR #5228, opened 2025-09-01
Add MCP code review server module

suhaibmujahid:
Introduces a new MCP server for code review functionality, including Dockerfile, project metadata, and server implementation. The server provides patch review prompts and fetches patch data from Phabricator for automated code review workflows.


mcp/Dockerfile
```diff
@@ -0,0 +1,26 @@
+# Build stage
```
marco-c, line 1: need to add this folder to dependabot if it isn't already

suhaibmujahid: Done here 2f7344a41f0309873194c170883f0a56cfe4dcaa

mcp/src/bugbug_mcp/server.py
```diff
@@ -0,0 +1,240 @@
...
+
+from bugbug import phabricator, utils
+from bugbug.code_search.searchfox_api import FunctionSearchSearchfoxAPI
+from bugbug.tools.code_review import PhabricatorPatch
+from bugbug.utils import get_secret
+
+mcp = FastMCP("BugBug Code Review MCP Server")
+
+
+PATCH_REVIEW_PROMPT = """**Task**:
```
marco-c, line 20: We'll need to find a way to sync the prompts between mcp and the tool

suhaibmujahid: I will add a method from the `code_review` module to return the base prompt with context, making the MCP a thinner layer.

marco-c: Sounds good, let's file a follow-up issue for that so we don't forget

suhaibmujahid: I was doing that ;)

Here it is: https://github.com/mozilla/bugbug/issues/5285

Issue #5239, opened 2025-09-02
get_first_pushdate is failing with `IndexError: tuple index out of range`


marco-c:
```
Traceback (most recent call last):
  File "/usr/local/bin/bugbug-data-commits", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/usr/local/lib/python3.12/site-packages/scripts/commit_retriever.py", line 67, in main
    retriever.retrieve_commits(args.limit)
  File "/usr/local/lib/python3.12/site-packages/scripts/commit_retriever.py", line 40, in retrieve_commits
    repository.download_commits(self.repo_dir, revs=revs[i : (i + chunk_size)])
  File "/usr/local/lib/python3.12/site-packages/bugbug/repository.py", line 1370, in download_commits
    first_pushdate = get_first_pushdate(repo_dir)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/bugbug/repository.py", line 1338, in get_first_pushdate
    return hg_log(hg, [b"0"])[0].pushdate
           ~~~~~~~~~~~~~~~~~~^^^
IndexError: tuple index out of range
```

marco-c:
Maybe https://github.com/mozilla/bugbug/pull/5185 caused this?

marco-c:
It looks like it is working again now, must have been an intermittent issue.

PR #5255, opened 2025-09-09
Use "default" instead of "tip" for Mercurial

marco-c:
"tip" is an ambiguous reference, it can be the current tip of any branch in the repository. In many cases on hg.mozilla.org, this ends up being a tags branch.

"default" is safer as it will always be the tip of the default branch.

This will maybe help with #5239


suhaibmujahid (approved):
LGTM!

PR #5268, opened 2025-09-17
Integrate agent tools for code review context expansion

### LINKED ISSUE #4513: [code_review] Switch to tool/function calling for RAG and Context
Should be better than "asking" the LLM

suhaibmujahid:
Resolves #4513

The changes are split into separate commits to make it easier to review. The first four commits are prerequisites. The last commit does the following:

- Added tools for expanding code context and finding function definitions 
- Refactored code_review.py to use the tools via a LangGraph agent
- Updated patch classes to support file retrieval from Phabricator
- Removed legacy context/further info chains


bugbug/tools/code_review.py
```diff
@@ -1259,26 +1283,17 @@ def __init__(
...
             verbose=verbose,
         )
 
-        self.function_search = function_search
+        tools = [expand_context, create_find_function_definition_tool(function_search)]
+        self.agent = create_react_agent(
+            llm,
+            tools,
+            prompt=f"You are an expert reviewer for {experience_scope}, with experience on source code reviews.",
+            debug=True,
```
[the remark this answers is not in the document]
suhaibmujahid: Removed.

bugbug/tools/agent_tools.py
```diff
@@ -0,0 +1,79 @@
...
+        Returns:
+            The function definition.
+        """
+        functions = function_search.get_function_by_name(
+            # TODO: We may want to use the patch base commit hash here instead of "tip".
+            "tip",
+            file_path,
+            function_name,
+        )
+
```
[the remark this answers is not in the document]
suhaibmujahid: Done!

bugbug/tools/code_review.py
```diff
@@ -1206,9 +1234,8 @@ class CodeReviewTool(GenerativeModelTool):
 
     def __init__(
         self,
-        comment_gen_llms,
-        llm=None,
-        function_search: Optional[FunctionSearch] = None,
+        llm: BaseLanguageModel,
+        function_search: FunctionSearch,
```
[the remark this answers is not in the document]
suhaibmujahid: This is intentional.

scripts/code_review_tool_evaluator.py
```diff
@@ -219,127 +219,64 @@ def print_evaluation_matches(matching_results: list[dict]):
...
     """
-
-    def is_variant_selected(*target_variants):
-        return variants is None or any(
-            target_variant in variants for target_variant in target_variants
-        )
-
     # Step 1: we start with instantiating the dependencies based on the selected
     # variants.
+    repo_dir = "../mozilla-unified"
```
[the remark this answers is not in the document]
suhaibmujahid: We are not using it at the moment.

scripts/code_review_tool_runner.py
```diff
@@ -30,7 +30,10 @@ def run(args) -> None:
         show_patch_example=False,
     )
 
-    review_data = code_review.review_data_classes[args.review_platform]()
+    if args.review_platform == "phabricator":
+        review_data = code_review.PhabricatorReviewData()
```
marco-c: What was the problem before this change?

bugbug/tools/code_review.py
```diff
@@ -1206,8 +1188,7 @@ class CodeReviewTool(GenerativeModelTool):
 
     def __init__(
         self,
-        comment_gen_llms,
```
marco-c, line 1209: This was introduced to support generation of comments through multiple LLMs (e.g. Claude and OpenAI). Why do we need to remove it?

scripts/code_review_tool_runner.py
```diff
@@ -15,11 +15,9 @@
...
     llm = generative_model_tool.create_llm_from_args(args)
 
-    function_search = (
-        function_search_classes[args.function_search_type]()
-        if args.function_search_type is not None
-        else None
-    )
+    assert args.function_search_type is not None, "Function search type is required"
+
+    function_search = function_search_classes[args.function_search_type]()
```
marco-c: Can't we keep it optional?

suhaibmujahid: We can, which will require making the function search tool optional as well.

Done here: https://github.com/mozilla/bugbug/compare/51c0373c552c3103fb9105386b77674981d1e3f8..e654b3b979e813fb6745374cf09bea596a2146b5

bugbug/tools/code_review.py
suhaibmujahid: Keeping it as is conflicts with this PR. I thought it wasn't worth refactoring because we don't use it, and supporting multiple models simultaneously could be done differently after this PR. For example, we could use LangGraph with the model section for that.

scripts/code_review_tool_runner.py
suhaibmujahid: mypy complained that we cannot instantiate the `ReviewData` class since it is abstract. 

It could be a bug with mypy, since we are not instantiating `ReviewData`, but it is the parent for the classes that we might instantiate.

bugbug/tools/code_review.py
marco-c: Let's file a follow-up to bring it back.
I would avoid using langgraph to avoid vendor lock-in

scripts/code_review_tool_runner.py
marco-c: Weird, let's add a comment to skip mypy here then, we shouldn't reduce functionality to satisfy mypy

bugbug/tools/code_review.py
```diff
@@ -153,6 +165,12 @@ class LargeDiffError(Exception):
 
 ---
 
+**Patch Summary**:
+
+{patch_summarization}
```
marco-c, line 170: We were not including the summarization? :S

bugbug/tools/agent_tools.py
```diff
@@ -0,0 +1,82 @@
...
+    )
+
+    if not functions:
+        # Fallback: return a few lines around the line number.
+        lines = file_content.splitlines()
+        start = max(0, line_number - 20)
+        end = min(len(lines), line_number + 20)
+        return "\n".join(lines[start:end])
+
+    return functions[0].source
```
marco-c: Can't we just use the file contents from Phabricator here always, without hitting Searchfox?

bugbug/tools/code_review.py
```diff
@@ -1238,22 +1282,19 @@ def __init__(
...
-            prompt=PromptTemplate.from_template(PROMPT_TEMPLATE_FURTHER_INFO),
-            llm=self.llm,
-            verbose=verbose,
-        )
+        tools = [expand_context]
+        if function_search:
+            tools.append(create_find_function_definition_tool(function_search))
 
-        self.function_search = function_search
+        self.agent = create_react_agent(
```
marco-c, line 1298: Does this use the native LLM tool calling capabilities? If not, maybe that's why you don't see OpenAI models using tools?

bugbug/tools/code_review.py
suhaibmujahid: It was included, but it was provided as part of the conversation history: https://github.com/mozilla/bugbug/blob/9b6ac62ab114477a70b1f8192a23c6c6360e734f/bugbug/tools/code_review.py#L1219-L1224

bugbug/tools/agent_tools.py
suhaibmujahid: Yes, we can. 

I tested the Searchfox method and found it not super accurate, but improving it is out of the scope of this PR. I was planning to do that in a follow-up using tree-sitter (mentioned in the TODO comment). 

Alternatively, we could return the expanding lines scope as we do for fallback, without considering method/function blocks.

marco-c: Yep, that's what I was thinking, just returning the expanding lines scope without method/function blocks

suhaibmujahid: Done here: https://github.com/mozilla/bugbug/compare/e654b3b979e813fb6745374cf09bea596a2146b5..2d31294119c92ccbf9faf7ab69e54bd69ad21910

bugbug/tools/code_review.py
suhaibmujahid: Filed: https://github.com/mozilla/bugbug/issues/5269

bugbug/tools/code_review.py
suhaibmujahid: It appears to be broken for GPT-5. I tested it with GPT-4.1, and it seems to work fine.

See https://community.openai.com/t/gpt-5-breaks-the-agents-sdk-and-tool-calling/1341727/3:
> OpenAI’s own engineers on GitHub said to keep using 4.1 until they fix 5.

marco-c: 🫢

PR #5293, opened 2025-09-26
Adds a new MCP resource endpoint to read bugs from Bugzilla

### LINKED ISSUE #5292: Allow accessing an LLM friendly view of Bugzilla bugs through the MCP server
For now, this should not include private bugs.

suhaibmujahid:
Resolves #5292

Introduces a Bug class and markdown export utilities to bugbug.tools.code_review.py for representing Bugzilla bugs, including their history and comments. Adds a new MCP resource endpoint in server.py to retrieve and render bug details in markdown format for a given bug ID.

### DIFF
--- a/bugbug/tools/code_review.py
+++ b/bugbug/tools/code_review.py
@@ -517,6 +517,204 @@ class PhabricatorPatch(Patch):
         return self._revision_metadata["fields"]["title"]
 
 
+def create_bug_timeline(comments: list[dict], history: list[dict]) -> list[str]:
+    """Create a unified timeline from bug history and comments."""
+    events = []
+
+    ignored_fields = {"cc", "flagtypes.name"}
+
+    # Add history events
+    for event in history:
+        changes = [
+            change
+            for change in event["changes"]
+            if change["field_name"] not in ignored_fields
+        ]
+        if not changes:
+            continue
+
+        events.append(
+            {
+                "time": event["when"],
+                "type": "change",
+                "who": event["who"],
+                "details": changes,
+            }
+        )
+
+    # Add comments
+    for comment in comments:
+        print(comment)
+        events.append(
+            {
+                "time": comment["time"],
+                "type": "comment",
+                "who": comment["author"],
+                "id": comment["id"],
+                "count": comment["count"],
+                "text": comment["text"],
+            }
+        )
+
+    # Sort by timestamp
+    events.sort(key=lambda x: (x["time"], x["type"] == "change"))
+
+    # Format timeline
+    timeline = []
+
+    last_event = None
+    for event in events:
+        date = event["time"][:10]
+        time = event["time"][11:19]
+
+        if last_event and last_event["time"] != event["time"]:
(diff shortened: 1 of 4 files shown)


bugbug/tools/code_review.py
```diff
@@ -517,6 +517,203 @@ def patch_title(self) -> str:
...
+        bug_data = bugs[0]
+        assert bug_data["id"] == bug_id
+
+        return Bug(bug_data)
+
+    def to_md(self) -> str:
+        """Return a markdown representation of the bug."""
+        return bug_dict_to_markdown(self.metadata)
+
+
```
marco-c, line 715: I would move this code in the Bugzilla module (or a generic new "bug" module), it makes more sense to have it there

mcp/src/bugbug_mcp/server.py
```diff
@@ -103,10 +103,15 @@ def get_code_review_tool():
     from bugbug.tools.code_review import CodeReviewTool, ReviewCommentsDB
     from bugbug.vectordb import QdrantVectorDB
 
+    # FIXME: This is a workaround, we should refactor CodeReviewTool to not avoid this.
```
marco-c:
```suggestion
    # FIXME: This is a workaround, we should refactor CodeReviewTool to avoid this.
```

bugbug/tools/code_review.py
```diff
@@ -517,6 +517,203 @@ def patch_title(self) -> str:
...
+    md_lines.append(
+        f"# Bug {bug.get('id', 'Unknown')} - {bug.get('summary', 'No summary')}"
+    )
+    md_lines.append("")
+
+    # Basic Information
+    md_lines.append("## Basic Information")
+    md_lines.append(f"- **Status**: {bug.get('status', 'Unknown')}")
+    md_lines.append(f"- **Severity**: {bug.get('severity', 'Unknown')}")
+    md_lines.append(f"- **Priority**: {bug.get('priority', 'Unknown')}")
```
marco-c: Probably not worth having this, since it's not consistent

marco-c (dismissed):

bugbug/tools/code_review.py
suhaibmujahid: This is LLM-related logic, similar to PhabricatorPatch, which lives in the same file.

I agree with you that it needs better organization. I want to do a general refactoring to reduce unused dependencies between ML/models and LLM/agents modules, which will also organize this.

mcp/src/bugbug_mcp/server.py
suhaibmujahid: Funny typo. Nice catch. Fixed here: 0751c80c969402a890c3c0a766e329668eea1696.

bugbug/tools/code_review.py
suhaibmujahid: Done here: 94d093c17927e4741a668df04c123731ffa5b82a

marco-c (approved):

PR #5366, opened 2025-10-21
Add patch-based test selection API endpoint

### LINKED ISSUE #1718: Define an API to select tests POSTing the contents of a patch
This way we can:
1) Start the test selection jobs even earlier than now (currently we start them after a Mercurial push);
2) Run the selector locally without a push to try (to select tests to run locally).

suhaibmujahid:
Resolves #1718

Introduces a new /patch/<base_rev>/<patch_hash>/schedules endpoint to submit and retrieve test selection results for patches.

### DIFF
--- a/bugbug/repository.py
+++ b/bugbug/repository.py
@@ -1554,17 +1554,14 @@ def pull(repo_dir: str, branch: str, revision: str) -> None:
 
 
 def generate_commit_from_raw_patch(
-    repo_dir: str, base_rev: str, patch: bytes, commit_msg: str
+    repo_dir: str, base_rev: str, patch: bytes
 ) -> Commit:
     with hglib.open(repo_dir) as hg:
         logger.info("Updating to base revision %s...", base_rev)
         hg.update(base_rev.encode("utf-8"), clean=True)
 
         logger.info("Applying the patch ...")
-        hg.import_(
-            patches=io.BytesIO(patch),
-            message=commit_msg.encode("utf-8"),
-        )
+        hg.import_(patches=io.BytesIO(patch))
 
         logger.info("Retrieving commit information...")
         commits = hg_log(hg, [b"."])
--- a/http_service/bugbug_http/models.py
+++ b/http_service/bugbug_http/models.py
@@ -318,7 +318,6 @@ def schedule_tests_from_patch(base_rev: str, patch_hash: str) -> str:
         REPO_DIR,
         hg_base_rev,
         patch=patch_data_raw,
-        commit_msg="Applied patch for test selection",
     )
 
     data = _analyze_patch([commit.node.encode("ascii")], "default")


http_service/bugbug_http/models.py
```diff
@@ -332,3 +312,100 @@ def get_config_specific_groups(config: str) -> str:
...
+        return "NOK"
+
+    LOGGER.info(f"Found {len(files)} files in patch")
+
+    # Create a minimal Commit object
+    commit = repository.Commit(
+        node=patch_hash,
+        author="patch-user",
+        desc="Patch-based test selection",
+        pushdate=datetime.utcnow(),
```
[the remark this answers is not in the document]
suhaibmujahid: This method is used in multiple places across the project. Filed an issue to refactor that as a follow-up: #5367

http_service/bugbug_http/app.py
```diff
@@ -997,6 +998,116 @@ def push_schedules(branch, rev):
...
+@application.route("/patch/<base_rev>/<patch_hash>/schedules", methods=["GET", "POST"])
+@cross_origin()
+def patch_schedules(base_rev, patch_hash):
+    """
+    ---
+    get:
+      description: Get results of patch-based test selection.
+      summary: Get test selection results for a previously submitted patch.
+      parameters:
+      - name: base_rev
```
suhaibmujahid, line 1010: We are currently not pulling the repo, so `base_rev` is not used at the moment. If we end up with that, we should remove it before merging.

http_service/bugbug_http/models.py
```diff
@@ -332,3 +299,96 @@ def get_config_specific_groups(config: str) -> str:
...
+    )
+
+    data = _analyze_patch(commits)
+
+    setkey(job.result_key, orjson.dumps(data), compress=True)
+
+    return "OK"
+
+
+def _analyze_patch(commits: Iterable[repository.CommitDict]):
```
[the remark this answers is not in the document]
suhaibmujahid: Fixed here: 034b369ee1ff487a1c07a92ed6c10b3eb71417e6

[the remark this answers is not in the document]
suhaibmujahid: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/mozilla/bugbug/pull/5366#discussion_r2495779588)

http_service/bugbug_http/models.py
```diff
@@ -332,3 +299,99 @@ def get_config_specific_groups(config: str) -> str:
     )
 
     return "OK"
+
+
+def schedule_tests_from_patch(base_rev: str, patch_hash: str) -> str:
+    from libmozdata import vcs_map
```
marco-c: This doesn't work anymore, we need to use Lando

marco-c: The API I shared with you the other day

marco-c: Or we need to update libmozdata to use Lando

bugbug/repository.py
```diff
@@ -1552,6 +1553,37 @@ def trigger_pull() -> None:
...
+    ]
+    with hglib.open(repo_dir, configs=configs) as hg:
+        logger.info("Updating to base revision %s...", base_rev)
+        hg.update(base_rev.encode("utf-8"), clean=True)
+
+        logger.info("Applying the patch ...")
+        hg.import_(
+            patches=io.BytesIO(patch),
+            message=commit_msg.encode("utf-8"),
+            user="bugbot <bugbug@mozilla.tld>".encode("utf-8"),
```
marco-c: We might need the actual user, I think there are some features based on author experience

bugbug/repository.py
```diff
@@ -1552,6 +1553,37 @@ def trigger_pull() -> None:
...
 
 
+def generate_commit_from_raw_patch(
+    repo_dir: str, base_rev: str, patch: bytes, commit_msg: str
+):
+    # Open the hg repository with minimal extensions to avoid compatibility issues
+    configs = [
+        b"extensions.evolve=!",  # Disable evolve extension
+        b"extensions.mozext=!",  # Disable mozext extension
+    ]
```
marco-c: We are not doing this anywhere else, it's only a problem when running locally with these enabled.
I would remove this and just assume they are not enabled

suhaibmujahid: Done here: 95d0879f32a2f3f7a683be4ec9605b86cea6e654

http_service/bugbug_http/models.py
suhaibmujahid: Done here 489515d042fbca4ccea5aa34ee7800e9da29a5e2, and I filed an issue to do it in libmozdata (https://github.com/mozilla/libmozdata/issues/266).

http_service/bugbug_http/models.py
```diff
@@ -332,3 +299,94 @@ def get_config_specific_groups(config: str) -> str:
...
+    if not patch_data_raw:
+        LOGGER.error(f"Patch not found in Redis for hash {patch_hash}")
+        return "NOK"
+
+    hg_base_rev = utils.git2hg(base_rev)
+    LOGGER.info(f"Mapped git base rev {base_rev} to hg rev {hg_base_rev}")
+
+    # Pull the base revision to the local repository
+    LOGGER.info("Pulling base revision from the remote repository...")
+    repository.pull(REPO_DIR, "mozilla-central", hg_base_rev)
```
marco-c:
```suggestion
    repository.pull(REPO_DIR, "autoland", hg_base_rev)
```

many people are basing their work on autoland, so some revisions might not be available yet on mozilla-central

http_service/bugbug_http/models.py
```diff
@@ -332,3 +299,94 @@ def get_config_specific_groups(config: str) -> str:
...
+
+    LOGGER.info("Generating commit from patch...")
+    commit = repository.generate_commit_from_raw_patch(
+        REPO_DIR,
+        hg_base_rev,
+        patch=patch_data_raw,
+        commit_msg="Applied patch for test selection",
```
marco-c: The commit message also might be used for feature selection, like the user

http_service/bugbug_http/models.py
```diff
@@ -332,3 +299,94 @@ def get_config_specific_groups(config: str) -> str:
...
+
+    # Analyze the patch.
+    commits = repository.download_commits(
+        REPO_DIR,
+        revs=[commit.node.encode("ascii")],
+        branch="default",
+        save=False,
+        use_single_process=True,
+        include_no_bug=True,
+    )
```
marco-c: Nit: you might also move this to `_analyze_patch`, most of it is shared between the two functions. I guess you just need to add a `branch` parameter to `_analyze_patch`.

marco-c (dismissed):
There are two things missing:
- getting the right author and commit message. They could be parsed from the patch posted to the API.
- support multiple patches (many developers use patch stacks): right now we have a base and a patch, but we might actually have a base and multiple patches one on top of the other

We could do these as fast follow-ups to iterate faster.

http_service/bugbug_http/models.py
suhaibmujahid: The use case I am targeting is to have test selection before committing, so we do not have a commit message at that stage.

bugbug/repository.py
suhaibmujahid: Fixed in 72741582fec0af4203089a3d740d2e7b2fad4d25, so the import command will automatically pick the author from the patch.

http_service/bugbug_http/models.py
suhaibmujahid: Nice catch! Done in fd6a179c90b498b3526f50526ff6fb3a35e89750.

suhaibmujahid:
>     * getting the right author and commit message. They could be parsed from the patch posted to the API.

Yes, that was a simple fix: https://github.com/mozilla/bugbug/pull/5366#discussion_r2496781448


>     * support multiple patches (many developers use patch stacks): right now we have a base and a patch, but we might actually have a base and multiple patches one on top of the other

Related to https://github.com/mozilla/bugbug/pull/5366#discussion_r2496630198. In this approach, even if there are multiple commits, we can send them as a single patch that includes the uncommitted changes (with an option to control what to include).

http_service/bugbug_http/models.py
marco-c: We could make it optional, but I can see people running the command after commit too.

marco-c: Especially common when you have a stack: you might have a few commits, and then some uncommitted changes on top.

marco-c: Anyway, it currently isn't used as a feature in the testselect model, so we can ignore it for now (file a follow-up so we do it for future-proofing).

suhaibmujahid: I removed the comment message in da39a35341a436936f15d175b805a4fae1b5d4e0. So it will read it from the patch.  Will use a default commit message on the client side if changes haven't been committed yet.

bugbug/repository.py
```diff
@@ -1552,6 +1553,37 @@ def trigger_pull() -> None:
     trigger_pull()
 
 
+def generate_commit_from_raw_patch(
+    repo_dir, base_rev: str, patch: bytes, commit_msg: str
+):
```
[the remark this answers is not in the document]
suhaibmujahid: Fixed in e700d7e072b60912a03c6982cb281016f441b6e5

http_service/bugbug_http/models.py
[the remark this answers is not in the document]
suhaibmujahid: Fixed in 82a36b8187e537ff960f1a25402cb17cf342de30

PR #5392, opened 2025-10-30
A few improvements to the DB implementation

marco-c:


bugbug/db.py
```diff
@@ -186,21 +206,21 @@ def _db_open(path, mode):
...
         else:
             dctx = zstandard.ZstdDecompressor()
             with open(path, mode) as f:
                 with dctx.stream_reader(f) as reader:
-                    yield store_constructor(reader)
+                    yield store_constructor(reader, use_mmap=False)
     else:
         with open(path, mode) as f:
-            yield store_constructor(f)
+            yield store_constructor(f, use_mmap=True)
```
suhaibmujahid, line 223: Why do we need to disable mmap here?

marco-c: It can only work with plain files, but not with compressed ones. For compressed ones, we would have to mmap the compressed file and not the one we're decompressing

suhaibmujahid: I see. With the limited performance improvement, I'm unsure if the added complexity is worth it.

suhaibmujahid (dismissed):

bugbug/db.py
marco-c: In Calixte's case the difference was much bigger (34.5 s vs 54 s for the repository DB)

suhaibmujahid: Make sense.

bugbug/tools/code_review.py
```diff
@@ -0,0 +1,2123 @@
...
+        return PhabricatorPatch(patch_id)
+
+    def get_all_inline_comments(
+        self, comment_filter
+    ) -> Iterable[tuple[int, list[InlineComment]]]:
+        db.download(phabricator.REVISIONS_DB)
+
+        for revision in tqdm(
+            phabricator.get_revisions(), total=phabricator.count_revisions()
+        ):
```
suhaibmujahid: This lives now in https://github.com/mozilla/bugbug/blob/ae21431167b1f5383516b39f0e96eeed3660dce0/bugbug/tools/core/platforms/phabricator.py#L488-L489

marco-c: Conflict fixed!

bugbug/tools/code_review.py
suhaibmujahid, line 1: I believe this ended up here by mistake.

marco-c: Oops, removed.

PR #5396, opened 2025-10-31
Add a task to retrieve CI failures, their logs and their fixes

marco-c:


scripts/retrieve_ci_failures.py
```diff
@@ -0,0 +1,307 @@
...
+        fixed_by_commit_pushes[bug_id]["failures"].append(
+            {
+                "task_name": element["job_type_name"],
+                "task_id": element["task_id"],
+                "retry_id": element["retry_id"],
+                "failure_lines": element["failure_lines"],
+            }
+        )
+
+    print(f"Analyzing {len(fixed_by_commit_pushes)} 'fixed by commit' pushes.")
```
suhaibmujahid: It would be nice to use `logger` instead of `print`.

scripts/retrieve_ci_failures.py
```diff
@@ -0,0 +1,307 @@
...
+    # Skip cases where there is no backout (and so the fix was a bustage fix).
+    no_backouts = set()
+    for bug_id in fixed_by_commit_pushes.keys():
+        if bug_id not in backouts_by_bug_id:
+            print(f"No backouts on {bug_id}")
+            no_backouts.add(bug_id)
+
+    print(
+        f"{len(no_backouts)} cases removed because there were no backouts in the bug."
+    )
```
suhaibmujahid: We have not removed them yet; we could move them after removing them or tweak the wording to reflect that.

scripts/retrieve_ci_failures.py
```diff
@@ -0,0 +1,307 @@
...
+    # Skip cases where there are multiple backouts associated to the same bug ID.
+    multiple_backouts = set()
+    for bug_id, backouts in backouts_by_bug_id.items():
+        if backouts > 1:
+            if bug_id in fixed_by_commit_pushes:
+                multiple_backouts.add(bug_id)
+
+    print(
+        f"{len(multiple_backouts)} cases removed because there were multiple backouts in the same bug."
+    )
```
suhaibmujahid: Similar to https://github.com/mozilla/bugbug/pull/5396/files#r2481691238

scripts/retrieve_ci_failures.py
```diff
@@ -0,0 +1,307 @@
...
+            except requests.exceptions.HTTPError:
+                pass
+
+
+def diff_failure_vs_fix(repo, failure_commits, fix_commits):
+    try:
+        fd, idx = tempfile.mkstemp(prefix="gitidx_")
+        os.close(fd)
+
+        try:
```
suhaibmujahid, line 182: Nested try-catch blocks make code less readable. If unavoidable, I would move that part to a separate function and call it here. This way, the logic with the nested try-catch would be encapsulated within the new function.

scripts/retrieve_ci_failures.py
```diff
@@ -0,0 +1,307 @@
...
+                os.remove(log_path)
+
+                if upload:
+                    utils.upload_s3([log_path])
+            except requests.exceptions.HTTPError:
+                pass
+
+
+def diff_failure_vs_fix(repo, failure_commits, fix_commits):
+    try:
```
suhaibmujahid, line 178: I would avoid big try-catch blocks.

scripts/retrieve_ci_failures.py
```diff
@@ -0,0 +1,307 @@
...
+                check=True,
+            )
+            for commit in fix_commits:
+                patch = subprocess.check_output(
+                    ["git", "-C", repo, "show", "--pretty=format:", "-p", commit]
+                )
+                # Apply the commit's patch into the index (3-way to tolerate drift)
+                subprocess.run(
+                    ["git", "-C", repo, "apply", "--cached", "--3way"],
+                    env=env,
```
suhaibmujahid, line 196: @cgsheeh, do you think something like the following would be better to apply the fix commits?
 
```sh
git rebase --onto <failure_commits>^ <fix_commit>^ <fix_commit>
```

suhaibmujahid (dismissed):

scripts/retrieve_ci_failures.py
marco-c: I meant to do that but then forgot, fixed in 6239a30d4874f429a88709101e8fdb67e414ac91.

scripts/retrieve_ci_failures.py
marco-c: Done in 67e3fadb22fd90614695eb5113f3134040d05043.

scripts/retrieve_ci_failures.py
marco-c: Done in 67e3fadb22fd90614695eb5113f3134040d05043.

scripts/retrieve_ci_failures.py
marco-c: I thought about that, but then splitting this into multiple functions would make it even less readable

scripts/retrieve_ci_failures.py
marco-c: We need that because any of the subprocess calls might fail with the same exception

suhaibmujahid: We could have more than one localized try-catch.

scripts/retrieve_ci_failures.py
suhaibmujahid: A descriptive function name would help with that.

scripts/retrieve_ci_failures.py
cgsheeh: I agree that using Git internals to re-create the commit would be better than collecting the patch and applying it to the working copy. IIUC it looks like we're taking a commit that has landed in a remote branch on the Firefox repo (`main`, `beta` etc) and applying it to a new local branch so we can compare the result against the same patch applied to a different parent. That's technically the same thing as `git cherry-pick <commit>`, so I'd recommend using that instead for simplicity and readability.

suhaibmujahid: Both commits are on the same branch. So, one of them is the original landed patch, and the other is a new version of that patch after being backed out. We are trying to have both based on the same parent to compare them and determine the diff (which will be the fix).

marco-c: I will land this as it is for now, as the cherry-pick alternative takes way longer (I tested with 5 patches, and it took around 2:30 minutes vs 20 seconds).
We can add the cherry-pick as a fallback in case the faster fails. I will file a follow-up.

marco-c: https://github.com/mozilla/bugbug/issues/5398

PR #5429, opened 2025-11-09
Update notification email to bugbug-team in infra configs

suhaibmujahid:
Replaces release-mgmt-analysis@mozilla.com with bugbug-team@mozilla.com and updates owner fields accordingly across all pipeline YAML and Taskcluster hook JSON files. This change ensures notifications and ownership are directed to the bugbug team.

### DIFF
--- a/infra/check-pipeline.yml
+++ b/infra/check-pipeline.yml
@@ -14,12 +14,12 @@ tasks:
         - component
 
     routes:
-      - notify.email.release-mgmt-analysis@mozilla.com.on-failed
+      - notify.email.bugbug-team@mozilla.com.on-failed
       - notify.irc-channel.#bugbug.on-failed
     metadata:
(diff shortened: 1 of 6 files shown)


infra/check-pipeline.yml
```diff
@@ -14,12 +14,12 @@ tasks:
         - component
 
     routes:
-      - notify.email.release-mgmt-analysis@mozilla.com.on-failed
+      - notify.email.bugbug-team@mozilla.com.on-failed
       - notify.irc-channel.#bugbug.on-failed
```
marco-c, line 18: We should remove this here and everywhere else, as we don't have an IRC channel anymore :)
(not in this PR)

suhaibmujahid: Filed #5430 as follow-up

Issue #5402, opened 2025-10-31
[inline_comments_data_collection] Don't restart from scratch on every run


marco-c:
We are currently re-retrieving everything.
We should instead store the results and, on the next run, download the previous ones and start from those.

Similarly to what we're doing in other places, e.g. https://github.com/mozilla/bugbug/blob/e45a2421dcd3d9cbc878068e7e22a36de119e36e/scripts/retrieve_ci_failures.py#L32 is downloading the previous DB first and then adding on top of it.

PR #5445, opened 2025-11-10
Avoid re-retrieving fixed comments from patches that were already analyzed and skipped

marco-c:
We are doing this for patches that we analyzed and got some comments from, but we are re-analyzing all skipped patches

Related to #5402

### DIFF
--- a/scripts/inline_comments_data_collection.py
+++ b/scripts/inline_comments_data_collection.py
@@ -87,7 +87,7 @@ def process_comments(limit, diff_length_limit):
     except FileNotFoundError:
         already_done_patches = set()
 
-    logger.info(f"{len(already_done_patches)} were already analyzed.")
+    logger.info("%d were already analyzed.", len(already_done_patches))
 
     for patch_id, comments in review_data.get_all_inline_comments(lambda c: True):
         revision_info = diff_id_to_revisions_map[patch_id]


scripts/inline_comments_data_collection.py
```diff
@@ -78,15 +79,22 @@ def process_comments(limit, diff_length_limit):
...
-    already_done_patches = {
-        data["initial_patch_id"] for data in db.read(phabricator.FIXED_COMMENTS_DB)
-    }
+    try:
+        with open(
+            os.path.join("data", phabricator.FIXED_COMMENTS_ALREADY_ANALYZED_DB), "rb"
+        ) as f:
+            already_done_patches = pickle.load(f)
+    except FileNotFoundError:
+        already_done_patches = set()
```
suhaibmujahid, line 88: I would add a logging statement here.

scripts/inline_comments_data_collection.py
```diff
@@ -78,15 +79,22 @@ def process_comments(limit, diff_length_limit):
...
 
     for patch_id, comments in review_data.get_all_inline_comments(lambda c: True):
+        revision_info = diff_id_to_revisions_map[patch_id]
+
         if patch_id in already_done_patches:
             continue
+        elif revision_info["fields"]["status"]["closed"]:
```
suhaibmujahid:
```suggestion

        if revision_info["fields"]["status"]["closed"]:
```

suhaibmujahid (dismissed):

scripts/inline_comments_data_collection.py
marco-c: Done in efc55ec3e2cfb1f6e0f16d1d96470eb86450eebf.

scripts/inline_comments_data_collection.py
marco-c: Done in d511878fd3456a7d3bf9099a15a44f9c437da2f2.

scripts/inline_comments_data_collection.py
```diff
@@ -78,15 +79,25 @@ def process_comments(limit, diff_length_limit):
...
-    }
+    try:
+        with open(
+            os.path.join("data", phabricator.FIXED_COMMENTS_ALREADY_ANALYZED_DB), "rb"
+        ) as f:
+            already_done_patches = pickle.load(f)
+    except FileNotFoundError:
+        already_done_patches = set()
+
+    logger.info(f"{len(already_done_patches)} were already analyzed.")
```
suhaibmujahid: nit:
```suggestion
    logger.info("%d were already analyzed.", len(already_done_patches))
```

PR #5456, opened 2025-11-13
Refactor code review prompt generation and reuse it in the MCP server

### LINKED ISSUE #5285: Reuse the prompt from `code_review` in the MCP
> We'll need to find a way to sync the prompts between mcp and the tool

_Originally posted by @marco-c in https://github.com/mozilla/bugbug/pull/5228#discussion_r2379528007_

suhaibmujahid:
Resolves #5285

Moved prompt formatting logic into CodeReviewTool with support for JSON and text output formats. Updated the MCP server to use the new generate_initial_prompt method, removing redundant prompt construction and simplifying patch review endpoint.

### DIFF
--- a/mcp/src/bugbug_mcp/server.py
+++ b/mcp/src/bugbug_mcp/server.py
@@ -34,7 +34,7 @@ def get_code_review_tool():
     from bugbug.vectordb import QdrantVectorDB
 
     return CodeReviewTool(
-        llm=init_chat_model("gpt-4.1"),
+        llm=init_chat_model("gpt-5.1"),
         review_comments_db=ReviewCommentsDB(QdrantVectorDB("diff_comments")),
     )
 


bugbug/tools/code_review/agent.py
```diff
@@ -130,10 +133,10 @@ def __init__(
     def count_tokens(self, text):
         return len(self._tokenizer.encode(text))
 
-    def _generate_suggestions(self, patch: Patch):
+    def generate_initial_prompt(
+        self, patch: Patch, output_format: Literal["JSON", "TEXT"] = "JSON"
+    ) -> str:
         formatted_patch = format_patch_set(patch.patch_set)
```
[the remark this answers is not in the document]
suhaibmujahid: I do not see a case where we would have an empty patch.

mcp/src/bugbug_mcp/server.py
```diff
@@ -21,171 +21,44 @@
...
-        def bind_tools(self, *args, **kwargs):
-            return self
-
-    review_comments_db = ReviewCommentsDB(QdrantVectorDB("diff_comments"))
-
-    tool = CodeReviewTool(
-        MockLLM(),
-        review_comments_db=review_comments_db,
+    return CodeReviewTool(
+        llm=init_chat_model("gpt-4.1"),
```
marco-c: Why `gpt-4.1`?

marco-c (dismissed):

mcp/src/bugbug_mcp/server.py
suhaibmujahid: Ooh! I meant to use GPT-5.1 since it is faster than GPT-5 and GPT-4.1. Also, it is cheaper than GPT-4.1.

suhaibmujahid: Fixed in f634205a8d0c9cfbe774cf828a5b89fc592a1a05

PR #5457, opened 2025-11-14
Replace lambda tenacity.retry patterns with Retrying code blocks

### LINKED ISSUE #5423: Use tenacity.Retrying to retry code blocks instead of decorating lambda functions
In some cases, e.g. https://github.com/mozilla/bugbug/blob/59f1d1a69b0b1ce19941403123992075fe7d796d/scripts/microannotate_generator.py#L81, we have defined lambda functions just so we could use tenacity decorators.
We should instead use https://tenacity.readthedocs.io/en/latest/#retrying-code-block.

WooJJam:
This PR implements the change requested in issue #5423 .

## Summary 
Replaced the pattern of using `tenacity.retry(...)(lambda: ...)` with the `tenacity.Retrying` class for retrying code blocks, applying the approach suggested in the issue.

## Related Issue
Closed #5423

### DIFF
--- a/bugbug/phabricator.py
+++ b/bugbug/phabricator.py
@@ -69,14 +69,18 @@ def get_transactions(rev_phid: str) -> Collection[TransactionDict]:
     data = []
 
     while after is not None:
-        out = tenacity.retry(
+        for attempt in tenacity.Retrying(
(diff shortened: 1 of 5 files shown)


suhaibmujahid (approved):
LGTM! Thank you!

PR #5459, opened 2025-11-14
Improvements to the code review prompt

### LINKED ISSUE #5257: Use XML tags within prompts to clearly delineate different parts and sections
This structured approach will help the LLM better parse and interpret the prompt, which could lead to more accurate responses.

Related docs:
- https://cookbook.openai.com/examples/gpt4-1_prompting_guide#delimiters
- https://docs.anthropic.com/en/docs/build-with-claude/prompt-engineering/use-xml-tags

suhaibmujahid:
Resolves #5257

### DIFF
--- a/bugbug/tools/code_review/prompts.py
+++ b/bugbug/tools/code_review/prompts.py
@@ -118,6 +118,7 @@ Respond only with a **JSON list**. Each object must contain the following fields
 * `"code_line"`: The number of the specific changed line of code that the comment refers to.
 * `"comment"`: A concise review comment.
 * `"explanation"`: A brief rationale for the comment, including how confident you are and why.
+* `"order"`: An integer representing the priority of the comment, with 1 being the highest confidence/importance.
 """
 
 OUTPUT_FORMAT_TEXT = """


bugbug/tools/code_review/prompts.py
```diff
@@ -26,82 +26,89 @@
...
+* **Banned phrases**: "maybe", "might want to", "consider", "possibly", "could be", "you may want to".
+* **Use directive language**: "Fix", "Remove", "Change", "Add", "Validate", "Check" (not "Consider checking").
+* Avoid repeating what the code is doing unless it supports your critique.
+</write_comments>
+
+<use_tools>
+**Use available tools to verify concerns**:
+* Use tools to gather context when you suspect an issue but need verification.
+* Use `find_function_definition` to check if error handling or validation exists elsewhere.
+* Use `expand_context` to see if edge cases are handled in surrounding code.
```
marco-c, line 72: These are just examples, but the tools could be used for other things (e.g. expand_context is not necessarily only to see if edge cases are handled in surrounding code, it might be also to check how a variable is defined earlier or how else it is used).

suhaibmujahid: Yes, these are just examples. The tools themselves have descriptions, and the first point directs to that.

marco-c (approved):

PR #5465, opened 2025-11-18
Add heuristics to choose manifests that we must definitely run on test selection

### LINKED ISSUE #5450: Use fixed heuristics to select relevant tests to complement the model suggestions
_From @marco-c:_

Adding some static rules
- If a test is modified, run the manifest that includes it
- If a support file is modified, run the manifest that includes it
- If a file in a directory is modified, run the manifest that is in the same directory (e.g. if I modify https://searchfox.org/firefox-main/source/dom/battery/BatteryManager.cpp, we should run https://searchfox.org/firefox-main/source/dom/battery/test/chrome.toml and https://searchfox.org/firefox-main/source/dom/battery/test/mochitest.toml)

marco-c:
Fixes #5450

### DIFF
--- a/bugbug/test_scheduling.py
+++ b/bugbug/test_scheduling.py
@@ -948,36 +948,43 @@ def find_manifests_for_paths(repo_dir_str: str, paths: list[str]) -> set[str]:
                 ].add(str(toml_rel))
 
             # Collect support files.
-            support_files = data.get("DEFAULT", {}).get("support-files", [])
-            if isinstance(support_files, str):
-                support_files = [support_files]
+            def collect_support_files(value):
+                support_files = value.get("support-files", [])
+                if isinstance(support_files, str):
+                    support_files = [support_files]
 
-            for support_file in support_files:
-                if not support_file.strip():
-                    continue
+                for support_file in support_files:
+                    if not support_file.strip():
+                        continue
 
-                if support_file.startswith("!"):
-                    support_file = support_file[1:]
+                    if support_file.startswith("!"):
+                        support_file = support_file[1:]
 
-                if support_file.startswith("/"):
-                    support_file_path = (repo_dir / support_file[1:]).resolve()
-                else:
-                    support_file_path = (toml_dir / support_file).resolve()
+                    if support_file.startswith("/"):
+                        support_file_path = (repo_dir / support_file[1:]).resolve()
+                    else:
+                        support_file_path = (toml_dir / support_file).resolve()
 
-                if "*" in support_file:
-                    files = [
-                        Path(f)
-                        for f in glob.glob(str(support_file_path), recursive=True)
-                    ]
-                else:
-                    files = [support_file_path]
+                    if "*" in support_file:
+                        files = [
+                            Path(f)
+                            for f in glob.glob(str(support_file_path), recursive=True)
+                        ]
+                    else:
+                        files = [support_file_path]
 
-                for f in files:
-                    manifest_by_path[str(f.relative_to(repo_dir))].add(str(toml_rel))
+                    for f in files:
+                        manifest_by_path[str(f.relative_to(repo_dir))].add(
+                            str(toml_rel)
+                        )
+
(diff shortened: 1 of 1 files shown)


bugbug/test_scheduling.py
```diff
@@ -895,3 +898,95 @@ def get_test_info(date: datetime) -> dict[str, Any]:
...
+
+    if manifest_by_path is None:
+        manifest_by_path = collections.defaultdict(set)
+
+        for toml_path in repo_dir.rglob("*.toml"):
+            if (
+                toml_path.name == "Cargo.toml"
+                or toml_path.parent.name == "test-manifest-toml"
+                or "third_party" in toml_path.parts
+            ):
```
jmaher, line 922: what is the risk if there is a future .toml file that isn't a test manifest but not in this list?  we try to schedule it but it does nothing?  should we make a uniform comment for code that makes hardcoded assumptions and add it throughout bugbug so we can find all hacks easier?

bugbug/test_scheduling.py
```diff
@@ -895,3 +898,95 @@ def get_test_info(date: datetime) -> dict[str, Any]:
...
+            for head_file in head_files:
+                if not head_file.strip():
+                    continue
+
+                manifest_by_path[
+                    str((toml_dir / head_file).resolve().relative_to(repo_dir))
+                ].add(str(toml_rel))
+
+            # Collect support files.
+            support_files = data.get("DEFAULT", {}).get("support-files", [])
```
jmaher: support-files are also defined for individual tests, not just `DEFAULT`, should we add that in?  probably 90% of the cases are defined in DEFAULT.

jmaher (dismissed):
logic and tests seem good.  I have some potential future work in my comments.

bugbug/test_scheduling.py
marco-c: I added some code to try to protect against toml files that aren't really manifests, and added a "HACK:" comment so we know that this is something that might need to be updated (however, we don't have a way to tell when it might need to be updated).

bugbug/test_scheduling.py
marco-c: Fixed in ff41b9f40914842caadd39a817527435cc2f92b4.

bugbug/test_scheduling.py
```diff
@@ -895,3 +898,109 @@ def get_test_info(date: datetime) -> dict[str, Any]:
...
+    global manifest_by_path
+
+    repo_dir = Path(repo_dir_str)
+
+    manifests = set()
+
+    if manifest_by_path is None:
+        manifest_by_path = collections.defaultdict(set)
+
+        for toml_path in repo_dir.rglob("*.toml"):
```
suhaibmujahid, line 916: What if we get the list of manifests from the models' result (including low-confidence ones)?

http_service/bugbug_http/models.py
```diff
@@ -363,6 +363,10 @@ def _analyze_patch(revs: list[bytes], branch: str | None) -> dict:
     )
 
     groups = testgroupselect_model.select_tests(commits, test_selection_threshold)
+    for group in test_scheduling.find_manifests_for_paths(
+        REPO_DIR, list(set(path for commit in commits for path in commit["files"]))
+    ):
+        groups[group] = 1.0
```
suhaibmujahid, line 369: What if we include them separately and let the client decide how to use them?

bugbug/test_scheduling.py
marco-c: It would be a bit faster, but slightly less precise (in case there's a patch that adds a new manifest landing after the model was trained). However, this is already pretty fast and we need to do it only once for the first request, so it shouldn't matter much in the end.

http_service/bugbug_http/models.py
marco-c: I did it this way so we don't need to update all clients. They should still be able to differentiate if they really want to (as these groups will be the only ones with 1.0).

PR #5485, opened 2025-11-19
Fix logic to ignore bugs with no relandings

marco-c:
The old logic was hard to follow and wrong.

### DIFF
--- a/scripts/retrieve_ci_failures.py
+++ b/scripts/retrieve_ci_failures.py
@@ -134,7 +134,7 @@ def get_fixed_by_commit_pushes():
     # Skip cases where there is no relanding.
     no_relanding_bugs = set()
     for bug_id, obj in fixed_by_commit_pushes.items():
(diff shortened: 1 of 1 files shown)


scripts/retrieve_ci_failures.py
```diff
@@ -134,7 +134,7 @@ def get_fixed_by_commit_pushes():
     # Skip cases where there is no relanding.
     no_relanding_bugs = set()
     for bug_id, obj in fixed_by_commit_pushes.items():
-        if not any(not commit["backedoutby"] for commit in obj["commits"]):
+        if sum(1 for commit in obj["commits"] if commit["backedoutby"]) == 0:
```
suhaibmujahid, line 137:
```suggestion
        if not all(commit["backedoutby"] for commit in obj["commits"]):
```

marco-c:
My bad, this was not the issue.

PR #5488, opened 2025-11-19
[retrieve_ci_failures] Avoid double adding the same bug to no_backouts

suhaibmujahid:
Follow up on #5487

### DIFF
--- a/scripts/retrieve_ci_failures.py
+++ b/scripts/retrieve_ci_failures.py
@@ -173,7 +173,7 @@ def get_fixed_by_commit_pushes():
             no_backouts.add(bug_id)
 
         # This is needed because sometimes v-c-t fails to identify backouts.
-        if not any(commit["backedoutby"] for commit in obj["commits"]):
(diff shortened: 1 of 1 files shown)


scripts/retrieve_ci_failures.py
```diff
@@ -169,11 +169,10 @@ def get_fixed_by_commit_pushes():
     # Skip cases where there is no backout (and so the fix was a bustage fix).
     no_backouts = set()
     for bug_id, obj in fixed_by_commit_pushes.items():
-        if bug_id not in backouts_by_bug_id:
-            no_backouts.add(bug_id)
-
-        # This is needed because sometimes v-c-t fails to identify backouts.
```
marco-c, line 175: I find this option to be clearer because the comment is clearly separated

suhaibmujahid: So let us go with `elif` to avoid double adding the bug id (it is just a silly optimization): e91c232be5a3f08d1ed48b94bcd50609449d70d8

PR #5549, opened 2025-12-16
[code_review] Misc improvements

suhaibmujahid:
These improvements could be reviewed commit by commit.

### DIFF
--- a/bugbug/tools/core/platforms/phabricator.py
+++ b/bugbug/tools/core/platforms/phabricator.py
@@ -14,9 +14,10 @@ from typing import Iterable, Optional
 import tenacity
 from tqdm import tqdm
 
-from bugbug import bugzilla, db, phabricator, utils
+from bugbug import db, phabricator, utils
 from bugbug.tools.core.data_types import InlineComment, ReviewRequest
 from bugbug.tools.core.platforms.base import Patch, ReviewData
+from bugbug.tools.core.platforms.bugzilla import Bug
 from bugbug.utils import get_secret
 
 logger = getLogger(__name__)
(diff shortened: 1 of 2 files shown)


scripts/code_review_tool_evaluator.py
```diff
@@ -421,6 +448,18 @@ def main(args):
         )
 
     for review_request_id, review_request in selected_review_requests:
+        if review_request_id in [227266, 233414]:
+            print(
+                f"Skipping Review Request ID {review_request_id} because it is known to cause issues."
+            )
+            continue
```
marco-c, line 460: What issues?

scripts/code_review_tool_evaluator_report.py
```diff
@@ -2,10 +2,11 @@
 
 import pandas as pd
 
-from scripts.code_review_tool_evaluator import get_latest_evaluation_results_file
+import scripts.code_review_tool_evaluator as evaluator_script
 
 evaluation_results = pd.read_csv(
-    get_latest_evaluation_results_file("../evaluation_results")
+    # evaluator_script.get_latest_evaluation_results_file("../evaluation_results")
```
marco-c, line 8: I guess we can remove this line

bugbug/tools/core/platforms/phabricator.py
```diff
@@ -150,8 +150,32 @@ def _get_file(self, file_path: str, is_before_patch: bool) -> str:
...
+        return r.text
+
     def get_old_file(self, file_path: str) -> str:
-        return self._get_file(file_path, is_before_patch=True)
+        if file_path.startswith("b/") or file_path.startswith("a/"):
+            file_path = file_path[2:]
+
+        try:
+            return self._get_file_from_patch(file_path, is_before_patch=True)
+        except FileNotFoundError:
```
marco-c, line 176: In which cases does this happen?

scripts/code_review_tool_evaluator.py
suhaibmujahid: They are not accessible without a login.

scripts/code_review_tool_evaluator_report.py
suhaibmujahid: I switch sometimes to use it, so I kept it for convenience.

bugbug/tools/core/platforms/phabricator.py
suhaibmujahid: When requesting a file that has not been modified in the patch.

PR #5567, opened 2025-12-20
Trigger the data pipeline hook on new versions

suhaibmujahid:
I already added the route binding to the hook via the web UI. Let us merge this PR once it is confirmed to work as expected.


suhaibmujahid:
It worked as expected:

image

PR #5579, opened 2025-12-24
Update `uv.lock` to fix `uv sync` with Python 3.14

loganrosen:
`uv sync` currently fails in the `mcp` module with the following:

```
> uv sync --all-extras
  × No solution found when resolving dependencies for split (markers: python_full_version >= '3.14'):
  ╰─▶ Because there is no version of llvmlite{python_full_version >= '3.14'}==0.46.0b1 and shap==0.50.0 depends on llvmlite{python_full_version >= '3.14'}==0.46.0b1, we can conclude that shap==0.50.0 cannot
      be used.
      And because bugbug==0.0.616 depends on shap[plots]==0.50.0 and only bugbug==0.0.616 is available, we can conclude that all versions of bugbug cannot be used.
      And because your project depends on bugbug and your project requires bugbug-mcp[llms-txt], we can conclude that your project's requirements are unsatisfiable.

      hint: `llvmlite` was requested with a pre-release marker (e.g., llvmlite==0.46.0b1), but pre-releases weren't enabled (try: `--prerelease=allow`)

      hint: While the active Python version is 3.12, the resolution failed for other Python versions supported by your project. Consider limiting your project's supported Python versions using
      `requires-python`.
```

There are two options: (1) restrict the `mcp` module to Python versions less than 3.14, or (2) allow pre-release versions of dependencies. We need to do one of these until [`shap` releases a new version containing this change](https://github.com/shap/shap/pull/4235). I chose (2) to allow use of the module with Python 3.14, but please let me know if you prefer (1).

I ran `uv sync --prerelease=allow` to generate this change. That triggered it to ignore the current lockfile and update the dependencies to the latest compatible versions.

### DIFF
--- a/mcp/uv.lock
+++ b/mcp/uv.lock
@@ -2,9 +2,13 @@ version = 1
 revision = 3
 requires-python = ">=3.12"
 resolution-markers = [
-    "python_full_version >= '3.13'",
+    "python_full_version >= '3.14'",
+    "python_full_version == '3.13.*'",
     "python_full_version < '3.13'",
 ]
+
+[options]
+prerelease-mode = "allow"
 
 [[package]]
 name = "absl-py"
@@ -120,7 +124,7 @@ wheels = [
 
 [[package]]
 name = "anthropic"
-version = "0.72.0"
+version = "0.75.0"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "anyio" },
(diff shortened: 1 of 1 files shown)


suhaibmujahid:
Thanks, @loganrosen, for taking the time to fix this! 

Since this was generated by `uv sync-- prerelease=allow`, running `uv sync` would change it back, and running `uv sync-- locked` would fail.

For now, you could run `uv venv --python 3.12`. The issue should resolve itself once the dependencies support Python 3.14. 

NOTE: We should even consider whether we still need shape, since it causes issues from time to time.

PR #5580, opened 2025-12-24
Add Bugzilla quicksearch to MCP server

loganrosen:
Adding support for Bugzilla quicksearch to the MCP server so that people working on Firefox code can easily find relevant bugs while using AI assistance in their editor. Example output:

image

I've also made the API keys optional because I don't think they should be required for someone developing on this MCP server. Please let me know if you think otherwise.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -139,6 +139,28 @@ tasks:
 
         - $mergeDeep:
             - { $eval: default_task_definition }
+            - taskId: { $eval: as_slugid("mcp_tests_task") }
+              workerType: compute-smaller
+              payload:
+                env:
+                  CODECOV_TOKEN: 66162f89-a4d9-420c-84bd-d10f12a428d9
+                command:
+                  - "/bin/bash"
+                  - "-lcx"
+                  - "git clone --quiet ${repository} &&
+                    cd bugbug &&
+                    git -c advice.detachedHead=false checkout ${head_rev} &&
+                    pip install --disable-pip-version-check --no-cache-dir --progress-bar off . &&
+                    pip install --disable-pip-version-check --no-cache-dir --progress-bar off -r test-requirements.txt &&
+                    pip install --disable-pip-version-check --no-cache-dir --progress-bar off ./services/mcp[dev] &&
+                    pytest --cov=bugbug_mcp services/mcp/tests/ -vvv &&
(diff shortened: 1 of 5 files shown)


mcp/src/bugbug_mcp/server.py
```diff
@@ -109,6 +119,75 @@ def get_bugzilla_bug(bug_id: int) -> str:
...
+    limit: Annotated[int, "Maximum number of bugs to return (default: 20)"] = 20,
+) -> str:
+    """Search for bugs in Bugzilla using quick search syntax.
+
+    Quick search allows natural language searches and supports various shortcuts:
+    - Bug numbers: "12345" or "bug 12345"
+    - Keywords: "crash", "regression"
+    - Status: "NEW", "FIXED", "ASSIGNED"
+    - Products/Components: "Firefox", "Core::DOM"
+    - Combinations: "firefox crash NEW"
```
suhaibmujahid: The Quick Search does not really support natural-language search. The Quick Search enables shortcuts, e.g, `Status:RESOLVED` (without space after `:`), which get converted to an advanced search.

Here is how to use it: https://bugzilla.mozilla.org/page.cgi?id=quicksearch.html

I think something like a [Skill](https://code.claude.com/docs/en/skills) would be more suitable for this, since explaining it in the tool description would be overkill.


I'm curious to learn about your use case for this tool.

loganrosen: Makes sense. My use case is wanting to find any existing bugs that might be related to something I discover in the code while working on the Firefox codebase vs. having to manually search Bugzilla in my browser. I can clarify the guidance in a skill and keep the tool lightweight for performing the queries. What do you think?

suhaibmujahid: @loganrosen,  sorry for the late reply. It seems that I missed the notification. 

Creating a lightweight tool to perform the queries sounds good to me. We could try it without a skill at first, just by pointing to the [doc's URL](https://bugzilla.mozilla.org/page.cgi?id=quicksearch.html). The URL could trigger relevant context, and the model might know how to use it even without fetching the URL.

loganrosen: Sounds good — I've updated the docstring to drop the natural-language claim and just link to the quicksearch docs. Let me know if the updated description looks alright.

PR #5654, opened 2026-02-06
MCP server filtering improvements

padenot:
This implements the policy as discussed on the 2026-02-05.

### DIFF
--- a/bugbug/tools/core/platforms/phabricator.py
+++ b/bugbug/tools/core/platforms/phabricator.py
@@ -598,9 +598,7 @@ class PhabricatorPatch(Patch):
             current_phid = self._revision_metadata["phid"]
             patch_map = {
                 phid: (
-                    self
-                    if phid == current_phid
-                    else PhabricatorPatch(revision_phid=phid)
+                    self if phid == current_phid else self.__class__(revision_phid=phid)
                 )
                 for phid in stack_graph.keys()
             }


bugbug/tools/core/platforms/phabricator.py
```diff
@@ -651,21 +696,83 @@ def to_md(self) -> str:
...
+
 class PhabricatorReviewData(ReviewData):
     @tenacity.retry(
         stop=tenacity.stop_after_attempt(7),
         wait=tenacity.wait_exponential(multiplier=2, min=2),
         reraise=True,
     )
     def get_patch_by_id(self, patch_id: str | int) -> Patch:
-        return PhabricatorPatch(patch_id)
+        return SanitizedPhabricatorPatch(patch_id)
```
suhaibmujahid: This is not used by the MCP, so we do not need to change it.

We should instead change it in the mcp server: https://github.com/mozilla/bugbug/blob/23ae9820d94679e9da14c6a1463b9e712a71fa4d/mcp/src/bugbug_mcp/server.py#L123-L126

bugbug/tools/core/platforms/bugzilla.py
```diff
@@ -7,12 +7,45 @@
...
+from functools import cached_property
 
 from libmozdata.bugzilla import Bugzilla, BugzillaBase
 
 logger = logging.getLogger(__name__)
 
 MOZILLA_CORP_GROUP_ID = 42
+EDITBUGS_GROUP_ID = 9
+EDITBUGS_CUTOFF_DAYS = 365
+TRUST_BEFORE_DATE = datetime(2022, 1, 1, tzinfo=timezone.utc)
```
suhaibmujahid, line 18: I recall that we agreed not to use the activity data since it is not useful.

bugbug/tools/core/platforms/bugzilla.py
```diff
@@ -51,8 +86,36 @@ def user_handler(user, data):
             return
 
         groups = user.get("groups", [])
-        # Check for mozilla-corporation group (ID 42) only
-        is_trusted = any(g.get("id") == MOZILLA_CORP_GROUP_ID for g in groups)
+        group_ids = {g.get("id") for g in groups}
+
+        # Check if user is in mozilla-corporation group
+        is_moco = MOZILLA_CORP_GROUP_ID in group_ids
```
suhaibmujahid: IIRC, all ppl in MOZILLA_CORP_GROUP are inherently in EDITBUGS_GROUP, so we do not need to check both; we only need EDITBUGS_GROUP. Especially since we will not use the activity as a condition for ppl in EDITBUGS_GROUP.

bugbug/tools/core/platforms/bugzilla.py
```diff
@@ -368,12 +448,124 @@ def get(bug_id: int) -> "Bug":
         bug_data = bugs[0]
         assert bug_data["id"] == bug_id
 
-        return Bug(bug_data)
+        return SanitizedBug(bug_data)
```
suhaibmujahid: In this way, only `SanitizedBug` will be used, even in places where we do not want to perform sanitization. 

To target the mcp, we should modify it's method to use `SanitizedBug` instead of. `Bug`, here: https://github.com/mozilla/bugbug/blob/23ae9820d94679e9da14c6a1463b9e712a71fa4d/mcp/src/bugbug_mcp/server.py#L107-L110


With that, this method should become a class method (i.e., `@classmethod`) instead of a static method (i.e., `@staticmethod`). Then we use `cls`:


```suggestion
        return cls(bug_data)
```

bugbug/tools/core/platforms/phabricator.py
```diff
@@ -453,6 +463,72 @@ def patch_description(self) -> str:
...
+        if self.author_phid == self.diff_author_phid:
+            return None
+        return self._format_user_display(self.diff_author_phid)
+
+    @property
+    def summary_display(self) -> str | None:
+        return self._revision_metadata["fields"].get("summary") or None
+
+    @property
+    def test_plan_display(self) -> str | None:
```
suhaibmujahid: Why are we using the `_display` postfix for some properties?

bugbug/tools/core/platforms/phabricator.py
```diff
@@ -53,6 +57,12 @@ def get_phabricator_client(
     # This is awkward since PhabricatorAPI does not accept user agent directly
     set_default_value("User-Agent", "name", user_agent)
 
+    if not user_agent:
+        user_agent = get_user_agent()
+
+    # This is awkward since PhabricatorAPI does not accept user agent directly
+    set_default_value("User-Agent", "name", user_agent)
+
```
suhaibmujahid: It seems a rebase error
```suggestion
```

bugbug/tools/core/platforms/phabricator.py
```diff
@@ -478,85 +554,82 @@ def to_md(self) -> str:
...
+                    md_lines.append(f"    {from_id}[{stack_title} - CURRENT]")
                     md_lines.append(f"    style {from_id} fill:#105823")
                 else:
-                    md_lines.append(f"    {from_id}[{from_patch.patch_title}]")
+                    md_lines.append(f"    {from_id}[{stack_title}]")
 
                 for dep_phid in dependencies:
-                    dep_id = f"D{patch_map[dep_phid].revision_id}"
+                    dep_id = f"D{revision_ids[dep_phid]}"
                     md_lines.append(f"    {from_id} --> {dep_id}")
```
suhaibmujahid, line 560: I do not see the goal of the change here. The title should be sanitized automatically by using the patch_title property without any changes to this part.

bugbug/tools/core/platforms/phabricator.py
```diff
@@ -478,85 +554,82 @@ def to_md(self) -> str:
...
-        if revision["fields"]["authorPHID"] != diff["authorPHID"]:
-            md_lines.append(f"- **Diff Author**: {diff['authorPHID']}")
+        if self.diff_author_display:
+            md_lines.append(f"- **Diff Author**: {self.diff_author_display}")
         md_lines.append("")
         md_lines.append("")
 
-        stack_graph = revision["fields"].get("stackGraph")
-        if len(stack_graph) > 1:
+        if len(self.stack_graph) > 1:
```
suhaibmujahid: It would be great to minimize the changes where possible:
```suggestion
        stack_graph = revision["fields"].get("stackGraph")
        if len(stack_graph) > 1:
```

This will also minimize changes in later lines.

bugbug/tools/core/platforms/bugzilla.py
```diff
@@ -352,6 +414,24 @@ class Bug:
...
+            "keywords",
+            "cc",
+            "blocks",
+            "depends_on",
+            "regressed_by",
+            "duplicates",
+            "see_also",
+        ):
+            return []
+        return value
```
suhaibmujahid, line 412: I'm not fond of Python magic, but it's ok for now. We can look for a better method later.

bugbug/tools/core/platforms/bugzilla.py
suhaibmujahid, line 82: We might trust all the `@mozilla.com` emails:

```py
if email.endswith("@mozilla.com"):
    data[email] = True
```

bugbug/tools/core/platforms/bugzilla.py
padenot: yes, sorry, lefover.

bugbug/tools/core/platforms/bugzilla.py
padenot: We decided against this because checking emails is very hard (because unicode is hard, etc.), and checking group is fast (it can be batched -- same duration to check one or 30, I measured this against live BMO). I had this in the first version of the first patch iirc but took it out after a quick chat with Luke.

If we decide we want this, we'll want to use the industry standard package to do this.

bugbug/tools/core/platforms/phabricator.py
suhaibmujahid: I think the change here can be reversed. The information is already sanitized, unless I'm missing something.

bugbug/tools/core/platforms/bugzilla.py
suhaibmujahid: I remember Luke's comment on the first PR, but didn't grasp the issue. Why is email validation complex here? What edge cases does `email.endswith("@mozilla.com")` miss that we need to handle?"

bugbug/tools/core/platforms/bugzilla.py
```diff
@@ -368,12 +419,124 @@ def get(bug_id: int) -> "Bug":
...
+    @property
+    def creator_detail(self) -> dict:
+        return self._metadata.get("creator_detail", {})
+
+    @property
+    def assignee_detail(self) -> dict:
+        return self._metadata.get("assigned_to_detail", {})
+
+    @property
+    def reporter(self) -> str | None:
```
suhaibmujahid: The name `reporter` is unclear here; is it an email or a name? We could have both in two separate properties.

bugbug/tools/core/platforms/bugzilla.py
```diff
@@ -368,12 +419,124 @@ def get(bug_id: int) -> "Bug":
...
+        if not detail:
+            return None
+        name = detail.get(
+            "real_name", detail.get("nick", detail.get("email", "Unknown"))
+        )
+        email = detail.get("email", "No email")
+        return f"- **Reporter**: {name} ({email})"
+
+    @property
+    def assignee(self) -> str | None:
```
suhaibmujahid: Same as https://github.com/mozilla/bugbug/pull/5654/changes#r2785583692

bugbug/tools/core/platforms/bugzilla.py
```diff
@@ -368,12 +419,124 @@ def get(bug_id: int) -> "Bug":
...
+        if not detail:
+            return None
+        name = detail.get(
+            "real_name", detail.get("nick", detail.get("email", "Unknown"))
+        )
+        email = detail.get("email", "No email")
+        return f"- **Assignee**: {name} ({email})"
+
+    @property
+    def timeline_comments(self) -> list[dict]:
```
suhaibmujahid:
```suggestion
    def comments(self) -> list[dict]:
```

bugbug/tools/core/platforms/bugzilla.py
```diff
@@ -368,12 +419,124 @@ def get(bug_id: int) -> "Bug":
...
+        return self._metadata.get("comments", [])
+
+    @property
+    def timeline_history(self) -> list[dict]:
```
suhaibmujahid:
```suggestion
    def history(self) -> list[dict]:
```

bugbug/tools/core/platforms/bugzilla.py
```diff
@@ -368,12 +419,124 @@ def get(bug_id: int) -> "Bug":
...
+    @property
+    def reporter(self) -> str | None:
+        detail = self.creator_detail
+        if not detail:
+            return None
+        name = detail.get(
+            "real_name", detail.get("nick", detail.get("email", "Unknown"))
+        )
+        email = detail.get("email", "No email")
+        return f"- **Reporter**: {name} ({email})"
```
suhaibmujahid: The markdown formatting does not belong here. It should be moved `bug_to_markdown()`.

bugbug/tools/core/platforms/bugzilla.py
```diff
@@ -368,12 +419,124 @@ def get(bug_id: int) -> "Bug":
...
+    @property
+    def assignee(self) -> str | None:
+        detail = self.assignee_detail
+        if not detail:
+            return None
+        name = detail.get(
+            "real_name", detail.get("nick", detail.get("email", "Unknown"))
+        )
+        email = detail.get("email", "No email")
+        return f"- **Assignee**: {name} ({email})"
```
suhaibmujahid: Same as https://github.com/mozilla/bugbug/pull/5654/changes#r2785592511

suhaibmujahid (commented):
Thank you. It looks good. I just left a few minor comments.

Splitting the changes into scoped commits made the review much easier. I really appreciated it.

bugbug/tools/core/platforms/bugzilla.py
padenot: Discussed on chat.m.o. We can do this is an email parser if we want.

PR #5692, opened 2026-02-17
Fit the extraction pipeline before applying it

marco-c:
This is a requirement before bumping scikit-learn to 1.8.0 (#5545).

### DIFF
--- a/bugbug/model.py
+++ b/bugbug/model.py
@@ -370,7 +370,7 @@ class Model:
         X_gen, y = split_tuple_generator(lambda: self.items_gen(classes))
 
         # Extract features from the items.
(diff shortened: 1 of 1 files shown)


bugbug/model.py
```diff
@@ -370,7 +370,7 @@ def train(self, importance_cutoff=0.15, limit=None):
         X_gen, y = split_tuple_generator(lambda: self.items_gen(classes))
 
         # Extract features from the items.
-        X = self.extraction_pipeline.transform(X_gen)
```
suhaibmujahid, line 373: We did this intentionally back then. See https://github.com/mozilla/bugbug/issues/3818. We might need to address that issue differently.

PR #5742, opened 2026-02-23
Remove Searchfox function search tool

### LINKED ISSUE #5741: Remove find_function_definition from the MCP server
It is redundant with searchfox_cli, so we should remove it to avoid increasing the context.

suhaibmujahid:
Resolves #5741

### DIFF
--- a/bugbug/tools/code_review/prompts.py
+++ b/bugbug/tools/code_review/prompts.py
@@ -28,7 +28,7 @@ Follow this systematic approach to review the patch:
 **Step 3: Verify and Assess Confidence**
 - Use available tools when you need to verify concerns or gather additional context
 - Only include comments where you are at least 80% confident the issue is valid
-- When uncertain about an issue, use tools like `find_function_definition` or `expand_context` to verify before commenting
+- When uncertain about an issue, use available tools to verify before commenting
 - Do not suggest issues you cannot verify with available context
 
 **Step 4: Sort and Order Comments**


services/mcp/src/bugbug_mcp/server.py
```diff
@@ -12,8 +12,6 @@
 from fastmcp.resources import FileResource
 from pydantic import Field
 
-from bugbug import utils
-from bugbug.code_search.searchfox_api import FunctionSearchSearchfoxAPI
 from bugbug.tools.code_review.prompts import SYSTEM_PROMPT_TEMPLATE
```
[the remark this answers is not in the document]
suhaibmujahid: Fixed in c600ea07f5a72219c080ec495b744ee5964ce0ee

bugbug/tools/code_review/prompts.py
```diff
@@ -28,7 +28,7 @@
 **Step 3: Verify and Assess Confidence**
 - Use available tools when you need to verify concerns or gather additional context
 - Only include comments where you are at least 80% confident the issue is valid
-- When uncertain about an issue, use tools like `find_function_definition` or `expand_context` to verify before commenting
+- When uncertain about an issue, use available tools to verify before commenting
```
marco-c, line 31: Does this mean review helper won't be able to search for now, until it'll be able to use searchfox-cli itself?

suhaibmujahid: No, Review helper does not use the mcp, we provide the search functionality as a LangChan tool.

PR #5762, opened 2026-03-03
Build Repair Agent: initial evaluation harness

evgenyrp:
Here's an [example](https://wandb.ai/moz-bugbug/bugbug-build-repair-eval/weave/traces?view=traces_default&peekPath=%2Fmoz-bugbug%2Fbugbug-build-repair-eval%2Fcalls%2F019cb006-3afa-7ed5-b990-11f13560b5d1%3FhideTraceTree%3D1) of running it on a dataset of 85 examples. An outage at Anthropic contributed to some fails but it works overall.

The baseline agent is very simple; most of the complexity is about integration with Weave and building Firefox.

What works so far:
- Creating a dataset of one commit build failures (with one commit fixes)
- Running evaluation from a local machine under docker-compose
- Creating and cleaning worktrees for each example
- Basic sandboxing
- Tracing agent's steps in Weave
- Running multiple trials per example
- Separate stages for analysis and fixing (allows deploying analysis only)
- Building Firefox locally inside Docker to verify the fix
- Basic scorers
- Cost tracking
- Verifying data contamination (meaning the LLM is trained before the example date)

Out of scope for this PR:
- Reusing a pool of worktrees not to run `./mach bootstrap` on every example (I have the code, but it didn't work well, so I kept the simple approach)
- Pushing to TRY to verify a specific build: the main issue is propagating credentials to Docker (an alternative of using `./mach load-task` also didn't work for me)
- LLM-as-a-judge scorers
- Better integration with Weave (hacky implementation of tracing, verify metrics when all scorers work, verify error handling)
- Production deployment-related work
- Improving the agent itself (reading through the Claude Agents SDK docs and applying best practices, making sure the Firefox MCP and skills work, maybe doing another iteration if the build is still failing etc.)
- Improving sandboxing


bugbug/tools/build_repair/__init__.py
```diff
@@ -0,0 +1,8 @@
+# -*- coding: utf-8 -*-
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from bugbug.tools.build_repair.agent import AgentResponse, BuildFailure, BuildRepairTool
+
+__all__ = ["AgentResponse", "BuildFailure", "BuildRepairTool"]
```
suhaibmujahid: This is not needed since we already import from the submodules directly.

docker/build_repair/Dockerfile
suhaibmujahid, line 1: I would move the Dockerfile to `services/build_repair/Dockerfile` to be part of a service.

`bugbug/tools/` contains agents and tools; `services/` contains deployment logic — Dockerfiles, API applications, etc.

bugbug/tools/build_repair/worktree.py
```diff
@@ -0,0 +1,75 @@
...
+        logger.info(f"Cleaning up worktree {name}")
+        subprocess.run(
+            [
+                "git",
+                "worktree",
+                "remove",
+                "--force",
+                "--force",
+                str(self.base_dir / name),
+            ],
```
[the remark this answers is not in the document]
evgenyrp: See docs https://git-scm.com/docs/git-worktree#_options

> To remove a locked worktree, specify --force twice.

.gitignore
```diff
@@ -49,3 +49,6 @@ node_modules/
 *.log
 # Desktop Service Store
 *.DS_Store
+
+# JetBrains IDEs
+.idea
```
suhaibmujahid: This one falls a bit outside our scope, but I think the cleanest solution here is a **user-scoped global gitignore file** — that way it applies across all your repos without any per-project configuration.

I personally keep a `~/.gitignore_global` with entries like:

```
*~
.DS_Store
.idea/*
.vscode/*
cmake-build-debug/*
.claude/settings.local.json
```

Just don't forget to register it with Git:

```sh
git config --global core.excludesfile ~/.gitignore_global
```

docker-compose.dev.yml
suhaibmujahid, line 1: Similer to https://github.com/mozilla/bugbug/pull/5762/changes#r2907229399

docker/build_repair/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+
+WORKDIR /app
+
+RUN apt-get install -y python3-pip python3-venv
+
+RUN python3 -m venv /opt/venv
+ENV PATH="/opt/venv/bin:$PATH"
+
+RUN apt-get install -y git nodejs npm && rm -rf /var/lib/apt/lists/*
+RUN pip install weave>=0.52.29 pydantic claude-agent-sdk requests
```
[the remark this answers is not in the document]
evgenyrp: I'd want to have separate requirements + lock files but wasn't sure where to put it.

bugbug/tools/build_repair/worktree.py
```diff
@@ -0,0 +1,75 @@
...
+        subprocess.run(
+            [
+                "git",
+                "worktree",
+                "add",
+                "--force",
+                "--force",
+                str(worktree_path),
+                commit_hash,
+            ],
```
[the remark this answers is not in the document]
evgenyrp: This is correct, see docs https://git-scm.com/docs/git-worktree#_options

docker/build_repair/Dockerfile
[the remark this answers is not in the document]
suhaibmujahid: These dependencies are already installed by bugbug — the only exception is `claude-agent-sdk`, which you've already added.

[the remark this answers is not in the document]
evgenyrp: I noticed it was very slow to rebuild the Docker image with all the deps together. But yeah, probably it's better to keep everything in one file, since the build repair tool can access some other modules from bugbug.

bugbug/tools/build_repair/worktree.py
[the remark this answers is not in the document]
evgenyrp: I can probably add a comment about it

docker-compose.dev.yml
evgenyrp: It's not clear where this link leads

evgenyrp: Ok, it seems it's the comment about Dockerfile

scripts/build_repair_eval.py
```diff
@@ -0,0 +1,355 @@
...
+        gh_failure_commits: list[str],
+        failures: list[dict],
+        fix_commit_date: str,
+        **kwargs,
+    ) -> dict:
+        wt_name = f"bug-{bug_id}-{uuid.uuid4().hex[:8]}"
+        logger.info(
+            f"Invoking bug {bug_id} "
+            f"(commit={gh_failure_commits[0][:12]}, {len(failures)} failures)"
+        )
```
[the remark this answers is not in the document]
evgenyrp: It's very much expected it's not empty, so IndexError here would be ok

bugbug/tools/build_repair/scorer.py
```diff
@@ -0,0 +1,163 @@
...
+                r.get("cache_read_input_tokens", 0) for r in score_rows
+            ),
+            "total_cache_creation_tokens": sum(
+                r.get("cache_creation_input_tokens", 0) for r in score_rows
+            ),
+            "num_examples": n,
+        }
+        if self.num_trials > 1:
+            summary.update(_pass_at_k(score_rows, self.num_trials, "successful"))
+        logger.info(f"BasicMetrics summary: {summary}")
```
suhaibmujahid, line 101: Avoid f-string interpolation in logging calls — the string is eagerly evaluated even when the log level is disabled. Use %s lazy formatting instead. See: https://pylint.readthedocs.io/en/stable/user_guide/messages/warning/logging-fstring-interpolation.html

scripts/build_repair_eval.py
```diff
@@ -0,0 +1,355 @@
...
+            cutoff = max(
+                MODEL_CUTOFF_DATES[self.tool.analysis_model],
+                MODEL_CUTOFF_DATES[self.tool.fix_model],
+            )
+            if datetime.fromisoformat(fix_commit_date).date() < cutoff:
+                logger.warning(
+                    f"Skipping bug {bug_id}: fix date {fix_commit_date} "
+                    f"is before model cutoff {cutoff}"
+                )
+                raise ValueError("skipped_data_contamination")
```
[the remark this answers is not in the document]
evgenyrp: I don't think this is valid. I tested it and it works fine:

```python
>>> datetime.fromisoformat('2025-09-10T09:34:13Z').date()
datetime.date(2025, 9, 10)
```

docker-compose.dev.yml
evgenyrp: done

.gitignore
evgenyrp: done

docker/build_repair/Dockerfile
evgenyrp: moved

bugbug/tools/build_repair/__init__.py
evgenyrp: removed

docker/build_repair/Dockerfile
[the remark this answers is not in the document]
evgenyrp: Ok, I managed to build it with full bugbug dependencies and my image, but I had to use `requirements.txt` directly, not to rebuild the image every time after the bugbug code changes (if we do just uv sync in the end). Let's do another iteration on this when we understand better how it will be deployed. It's likely all this will change.

bugbug/tools/build_repair/scorer.py
evgenyrp: I would push back on this one. I always prefer readability to optimizations that are not strictly necessary. This makes a lot of sense when we call a function 1000 times per second (which we don't do here). In other cases, f-string is better. So this pylint rule is too strict.

services/buildrepair/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+RUN wget https://www.python.org/ftp/python/3.12.8/Python-3.12.8.tgz && \
+    tar -xf Python-3.12.8.tgz && \
+    cd Python-3.12.8 && \
+    ./configure --enable-optimizations --prefix=/usr/local && \
+    make -j$(nproc) && \
+    make install && \
+    cd .. && rm -rf Python-3.12.8 Python-3.12.8.tgz
+
+RUN python3.12 -m venv /opt/venv
+ENV PATH="/opt/venv/bin:$PATH"
```
suhaibmujahid, line 23: Why not use the official `python:3.12-bookworm` image?

evgenyrp: It's a very good question. I think it's possible, but the sheer number of things required to install to build Firefox is harder to add right than just installing Python. So I went the easiest path here. I suggest we investigate this again when doing the deployment work. It will really depend on where the agent lives. If it's Taskcluster, likely reusing the appropriate Firefox build image is optimal and keeps it always in sync with whatever is required for the latest Firefox. If elsewhere, we likely would need to come up with a self-contained image, as I think Taskcluster images are not published anywhere.

PR #5768, opened 2026-03-07
Treat Phabricator bot accounts as trusted in comment filtering

### LINKED ISSUE #5767: Phabricator MCP: reviewhelper-bot comments redacted as untrusted
The Phabricator trust filtering in the MCP tool redacts comments from `reviewhelper-bot` when they appear after the last trusted (MOCO member) comment. This is because the bot is not a member of the Mozilla Corporation Phabricator group, so it fails the `moco_members` check in `bugbug/tools/core/platforms/phabricator.py`.

The filtering walks comments backwards to find the last trusted comment, and redacts all untrusted comments after it. This means the bot's comments are visible when a MOCO reviewer comments after them, but get redacted when the bot comments last (e.g. on a new revision update).

Since `reviewhelper-bot` is a Mozilla-operated bot, its comments should not be redacted. One possible fix would be to whitelist known bot accounts in the trust check.

Relevant code: `bugbug/tools/core/platforms/phabricator.py` lines 88-112 (trust determination) and lines 142-180 (comment filtering).

loganrosen:
The trust filtering logic redacts comments from users not in the bmo-mozilla-employee-confidential group. This causes comments from Mozilla-operated bots like reviewbot (Review Helper Bot) to be redacted when they appear after the last trusted user's comment.

This adds a `TRUSTED_BOT_PHIDS` set and a separate `is_trusted_bot` field so bot content is shown without propagating trust to surrounding untrusted comments. A bot commenting does not shift `last_trusted_index`, so earlier untrusted comments remain redacted.

The reviewbot account isn't marked with the Phabricator `bot` role (only 5 accounts are: email-bot, updatebot, wptsync, lando-bot, and a disabled maintenance bot), so a role-based check isn't sufficient here.

Also extracts the reviewbot PHID into a `REVIEWBOT_PHID` constant, replacing the hardcoded string in the existing inline comment filter.

Fixes #5767

### DIFF
--- a/bugbug/tools/core/platforms/phabricator.py
+++ b/bugbug/tools/core/platforms/phabricator.py
@@ -105,11 +105,13 @@ def _get_users_info_batch_impl(user_phids: set[str]) -> dict[str, dict]:
     for user_data in users_response.get("data", []):
         user_phid = user_data["phid"]
         fields = user_data.get("fields", {})
-        is_trusted = user_phid in moco_members or user_phid in TRUSTED_BOT_PHIDS
+        is_trusted = user_phid in moco_members
+        is_trusted_bot = user_phid in TRUSTED_BOT_PHIDS
 
         result[user_phid] = {
             "email": fields.get("username", "unknown"),  # Username is typically email
             "is_trusted": is_trusted,
+            "is_trusted_bot": is_trusted_bot,
             "real_name": fields.get("realName", ""),
         }
 
@@ -173,8 +175,9 @@ def _sanitize_comments(comments: list, users_info: dict[str, dict]) -> tuple[lis
     sanitized_comments = []
 
     for i, comment in enumerate(comments):
-        comment_is_trusted = users_info.get(comment.author_phid, {}).get(
-            "is_trusted", False
+        author = users_info.get(comment.author_phid)
+        comment_is_trusted = author and (
+            author["is_trusted"] or author["is_trusted_bot"]
         )
 
         # Create a shallow copy to avoid modifying the original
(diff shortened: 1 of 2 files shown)


bugbug/tools/core/platforms/phabricator.py
```diff
@@ -101,7 +105,7 @@ def _get_users_info_batch_impl(user_phids: set[str]) -> dict[str, dict]:
     for user_data in users_response.get("data", []):
         user_phid = user_data["phid"]
         fields = user_data.get("fields", {})
-        is_trusted = user_phid in moco_members
+        is_trusted = user_phid in moco_members or user_phid in TRUSTED_BOT_PHIDS
```
suhaibmujahid: The tricky part is that while content posted by the bot is trusted, that trust cannot be extended retroactively to earlier comments. To handle this, we need two distinct trust types: one that marks content as trusted in isolation, and another that propagates trust to preceding comments. Otherwise, any content would be incorrectly marked as trusted simply because the bot commented on the same revision.

loganrosen: Good catch — addressed in 00179c7. Bot accounts now have a separate `is_trusted_bot` field. The backward walk for `last_trusted_index` only considers `is_trusted` (MOCO members), so bot comments are shown without propagating trust to surrounding untrusted comments. Updated the test to verify this: an untrusted comment between a MOCO comment and a bot comment is still redacted.

bugbug/tools/core/platforms/phabricator.py
suhaibmujahid: I would do the change here, then I do not need to change anything else below:

```suggestion
author = users_info.get(comment.author_phid)
comment_is_trusted = author and (author["is_trusted"] or author["is_trusted_bot"])
```

suhaibmujahid (commented):
Thank you! Just a minor comment — everything else looks good to me.

bugbug/tools/core/platforms/phabricator.py
loganrosen: Done in dce91d2. Also added `is_trusted_bot` to the existing test fixtures so they match the updated data model.

PR #5813, opened 2026-03-17
Add TodoListMiddleware for code review planning

### LINKED ISSUE #5812: Give the agent a TODO list tool that it uses to plan its review and then follow that plan step by step
Give the agent a structured way to organize its work — plan what to check, then execute against that plan.

suhaibmujahid:
Resolves #5812

Integrate TodoListMiddleware into the code review agent to enable tracking of investigation tasks during reviews.

### DIFF
--- a/bugbug/tools/code_review/prompts.py
+++ b/bugbug/tools/code_review/prompts.py
@@ -187,6 +187,27 @@ STATIC_COMMENT_EXAMPLES = [
     },
 ]
 
+CODE_REVIEW_TODO_PROMPT = """
+## Review Planning with `write_todos`
+
+Use the `write_todos` tool to track investigation tasks as you review.
+
+- After your initial scan, create todos for any concerns that need deeper investigation
+  (e.g., "Verify that removed error handler is covered elsewhere", "Check callers of
+  renamed function for breakage")
+- As the review progresses, add new todos when you discover additional concerns
+- Remove or complete todos that turn out to be non-issues after verification
+- For small or straightforward patches, skip todos entirely — just review directly
+"""
+
+CODE_REVIEW_TODO_TOOL_DESCRIPTION = (
+    "Track investigation tasks during code review. Add items for concerns that need "
+    "tool-based verification (expand_context, find_function_definition). Evolve the "
+    "list as you go — add new items when you discover concerns, remove irrelevant ones. "
+    "Do not use this as a file checklist."
+)
+
+
 TEMPLATE_PATCH_FROM_HUNK = """diff --git a/{filename} b/{filename}
 --- a/{filename}
 +++ b/{filename}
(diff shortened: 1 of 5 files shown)


JiwaniZakir (commented):
The import `from langchain.agents.middleware import TodoListMiddleware` is worth scrutinizing — `langchain.agents.middleware` does not appear to be a documented or stable LangChain module, and there's no pinned version or changelog reference to confirm this API exists. If this is a custom internal class or a very recent addition, it should either be vendored locally or accompanied by a minimum version constraint in the project's dependencies to prevent silent import failures.

The `CODE_REVIEW_TODO_TOOL_DESCRIPTION` in `prompts.py` mentions `expand_context` and `find_function_definition` by name as examples of "tool-based verification," which is a useful constraint — but the `CODE_REVIEW_TODO_PROMPT` doesn't echo this same framing. Since the prompt is what the model actually sees at runtime, it would be more effective to include those concrete tool examples there too, rather than only in the tool description that the model may see in a different context window position.

There are no tests covering the new middleware integration — at minimum, a unit test asserting that the agent is constructed with the `TodoListMiddleware` instance (checking `middleware` kwarg contents) would catch regressions if the `create_agent` signature changes. Given that `middleware` appears to be a new parameter path, this seems like a reasonable place to guard.

bugbug/tools/code_review/prompts.py
```diff
@@ -187,6 +187,27 @@
...
+- As the review progresses, add new todos when you discover additional concerns
+- Remove or complete todos that turn out to be non-issues after verification
+- For small or straightforward patches, skip todos entirely — just review directly
+"""
+
+CODE_REVIEW_TODO_TOOL_DESCRIPTION = (
+    "Track investigation tasks during code review. Add items for concerns that need "
+    "tool-based verification (expand_context, find_function_definition). Evolve the "
+    "list as you go — add new items when you discover concerns, remove irrelevant ones. "
+    "Do not use this as a file checklist."
```
marco-c, line 207: ?

marco-c: What is a "file checklist"?

marco-c (approved):
Did you run validation? What is the impact?

suhaibmujahid:
> Did you run validation? What is the impact?

Yes, I see improved numbers. Though the experiment itself in not perfect. 

Where able to complete with fewer errors. Generated 3 time more comments, which yielded higher recall. With the capabilities that we are planning to add, the to-do will be valuable to improve things further.

image

bugbug/tools/code_review/prompts.py
suhaibmujahid: This to avoid listing the files from the patch as to-do items.

PR #5832, opened 2026-03-18
pre-commit optimizations

marco-c:


.github/dependabot.yml
```diff
@@ -1,19 +1,41 @@
...
-    directory: "/http_service"
-    schedule:
-      interval: weekly
-    open-pull-requests-limit: 99
-  - package-ecosystem: docker
-    directory: "/infra"
+    directories:
+      - "/http_service"
+      - "/infra"
+      - "/mcp"
```
suhaibmujahid:
```suggestion
      - "/services/mcp"
```

.github/dependabot.yml
```diff
@@ -1,19 +1,41 @@
 version: 2
 updates:
   - package-ecosystem: docker
-    directory: "/http_service"
-    schedule:
-      interval: weekly
-    open-pull-requests-limit: 99
-  - package-ecosystem: docker
-    directory: "/infra"
+    directories:
```
suhaibmujahid, line 4: I would add `/services/reviewhelper-api` as well.

Or even better, we could do `/services/*`, and then we do not need even to add `/services/mcp`.

.github/dependabot.yml
```diff
@@ -1,19 +1,41 @@
...
+        applies-to: version-updates
+        patterns:
+          - "*"
+        update-types:
+          - minor
+          - patch
+    cooldown:
+      semver-major-days: 14
+      semver-minor-days: 7
+      semver-patch-days: 3
```
suhaibmujahid, line 23: Is there a specific reason for this pattern?

suhaibmujahid: I mean what is the motivation.

marco-c: Allowing some time for the dependencies to have other updates. E.g. often a major version will be quickly followed by a patch version to fix some unexpected issue

.github/dependabot.yml
marco-c: I don't know if `/services/*` means "Dockerfiles within /services" or "Dockerfiles within subdirectories of /services".

Issue #4407, opened 2024-08-09
Support including data in the training dataset but not in the testing dataset


suhaibmujahid:
This would be helpful in cases where we need to include more data, but these data do not represent what the model will be exposed to once it is deployed in production.

PR #5892, opened 2026-03-31
Add a component_specific model to identify 'general' bugs that should be moved into more specific components

marco-c:
This is a v1, a better approach would be to have a model that is trained on the output of the normal component model, where the options to learn would be something like:
- "keep in general"
- "specific component 1" and related confidence
- "specific component 2" and related confidence
- "specific component 3" and related confidence

I'll file a follow-up issue to implement that.

Train on Taskcluster: componentspecific

### DIFF
--- a/bugbug/models/component_specific.py
+++ b/bugbug/models/component_specific.py
@@ -1,0 +1,136 @@
+# -*- coding: utf-8 -*-
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import logging
+from datetime import datetime, timezone
+
+import dateutil.parser
+import xgboost
+from dateutil.relativedelta import relativedelta
+from sklearn.compose import ColumnTransformer
+from sklearn.feature_extraction import DictVectorizer
+from sklearn.pipeline import Pipeline
+
+from bugbug import bug_features, bugzilla, feature_cleanup, utils
+from bugbug.model import BugModel
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+class ComponentSpecificModel(BugModel):
+    def __init__(self, lemmatization=False, product="Firefox", component="General"):
+        BugModel.__init__(self, lemmatization)
+
+        self.product = product
+        self.component = component
+
+        feature_extractors = [
+            bug_features.HasSTR(),
+            bug_features.Severity(),
+            bug_features.Keywords(),
+            bug_features.HasCrashSignature(),
+            bug_features.HasURL(),
+            bug_features.HasW3CURL(),
+            bug_features.HasGithubURL(),
+            bug_features.Whiteboard(),
+            bug_features.Patches(),
+            bug_features.Landings(),
+        ]
+
+        cleanup_functions = [
+            feature_cleanup.fileref(),
+            feature_cleanup.url(),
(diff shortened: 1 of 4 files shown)


bugbug/models/component_specific.py
```diff
@@ -0,0 +1,136 @@
...
+        classes = {}
+
+        for bug_data in bugzilla.get_bugs():
+            if dateutil.parser.parse(bug_data["creation_time"]) < datetime.now(
+                timezone.utc
+            ) - relativedelta(years=3):
+                continue
+
+            # Only bugs that were moved out of General and into a specific component
+            # Or the opposite
```
marco-c, line 93: I couldn't think of other ways to find labels.
We can't rely on bugs that are currently in Firefox::General because some of them might be wrong. Even the fixed ones could have been wrong.

suhaibmujahid, line 93: What about bugs that got resolved without being moved out of the general component?

suhaibmujahid: Ooh, you already answered my question in https://github.com/mozilla/bugbug/pull/5892#discussion_r3016574897

marco-c: Yeah... Unfortunately we are not great when it comes to moving bugs in the right components out of Firefox::General...
Maybe we could pick a subset that we think should be really good, e.g. uplifted bugs. But I would do this as a follow-up because it needs to be evaluated (and maybe we want to use them for training but not for testing, which means we must wait for #4407).

suhaibmujahid: We could later use the changed files in the linked patch could as heuristic (e.g., utilizing [mots](https://github.com/mozilla-conduit/mots)).

bugbug/models/component_specific.py
```diff
@@ -0,0 +1,136 @@
...
+                    ),
+                ),
+                (
+                    "estimator",
+                    xgboost.XGBClassifier(n_jobs=utils.get_physical_cpu_count()),
+                ),
+            ]
+        )
+
+    def get_labels(self):
```
suhaibmujahid, line 83: The dataset is small and not balanced (https://community-tc.services.mozilla.com/tasks/Woy2F_GaTXeJ29eCHoh1bg/runs/0/logs/public/logs/live.log):

```
                 1       0.90      0.80      0.33      0.85      0.52      0.28        71
                 0       0.50      0.11      0.99      0.18      0.33      0.10         9
__NOT_CLASSIFIED__       0.00      0.00      0.81      0.00      0.00      0.00         0
       avg / total       0.86      0.72      0.41      0.78      0.50      0.26        80
╒════════════╤═════════════════╤═════════════════╤══════════════════╕
│            │   1 (Predicted) │   0 (Predicted) │   Not classified │
╞════════════╪═════════════════╪═════════════════╪══════════════════╡
│ 1 (Actual) │              57 │               1 │               13 │
├────────────┼─────────────────┼─────────────────┼──────────────────┤
│ 0 (Actual) │               6 │               1 │                2 │
╘════════════╧═════════════════╧═════════════════╧══════════════════╛
```

marco-c: I'll file a follow-up to investigate using imbalanced-learn.

marco-c: https://github.com/mozilla/bugbug/issues/5903

bugbug/models/component_specific.py
marco-c: https://github.com/mozilla/bugbug/issues/5904

PR #5961, opened 2026-04-23
Remove bugbug-nlp image because it is unused

marco-c:
We pay compute time and storage for this image, but we never use it.


marco-c:
The test failure is unrelated.

suhaibmujahid (approved):

PR #5970, opened 2026-04-27
Add hackbot-api service

suhaibmujahid:


scripts/run_bug_fix.py
```diff
@@ -1,16 +1,23 @@
-"""Run the bug_fix tool locally."""
+"""Run the bug_fix tool locally.
+
+Uses an in-process Bugzilla MCP server for simplicity — no separate
+broker needed. The orchestrated production path is different: the agent
+runs with no Bugzilla credentials and talks to a broker sidecar.
```
marco-c: Not sure it's worth mentioning here

bugbug/tools/bug_fix/firefox_mcp.py
```diff
@@ -209,8 +210,41 @@ async def evaluate_js_shell_tool(args):
         )
         return _jtext(crash_info)
 
+    @tool(
```
marco-c, line 213: Do we need this as a tool or can we always pre-bootstrap it?

suhaibmujahid: With pre-bootstrap, the overhead cost will always be paid even if no build needed. We need to optimize this anyway, so this is a temporary workaround.

scripts/run_bug_fix.py
suhaibmujahid: I agree, removed in c2efaee0e798c06b18517d33f9a560aad0eb0832.

marco-c (approved):

PR #5971, opened 2026-04-27
hg: update version-control-tools revision for pushlog wireproto updates (Bug 2024739)

cgsheeh:
This allows bugbug to use the streaming pushlog wire protocol command
which should fix the persistent 503s from hg.mozilla.org while cloning
against try.


marco-c (approved):
Thanks!

PR #5976, opened 2026-04-28
Don't use login shell to avoid overwriting environment variables set by Docker images

### LINKED ISSUE #5975: Integration tests failing (probably due to switch to uv)
```
+ bugbug-data-bugzilla --limit 500
./scripts/integration_test.sh: line 17: bugbug-data-bugzilla: command not found
```

marco-c:
Fixes #5975


marco-c:
I'll merge so we can test this.

PR #5982, opened 2026-04-28
Use python instead of /usr/local/bin/python3 so we use the uv version

### LINKED ISSUE #5981: Pipeline spawning failing after switch to uv
```
[taskcluster 2026-04-28T10:40:01.375Z] Executing command 0: docker run -t --name taskcontainer_YOSXGwp_Tv-TisSf5n6dcA --memory-swap -1 --pids-limit -1 --pull=never --log-driver=none '--add-host=localhost.localdomain:127.0.0.1' --add-host=taskcluster:host-gateway --env-file 'env.list' 'mozilla/bugbug-spawn-pipeline:v0.0.626' /usr/local/bin/python3 '/code/spawn_pipeline.py' '/code/data-pipeline.yml'
  File "/code/spawn_pipeline.py", line 24, in <module>
    import jsone
ModuleNotFoundError: No module named 'jsone'
```

marco-c:
Fixes #5981

### DIFF
--- a/infra/taskcluster-hook-data-pipeline.json
+++ b/infra/taskcluster-hook-data-pipeline.json
@@ -37,7 +37,7 @@
         "taskclusterProxy": true
       },
       "command": [
-        "python3",
+        "python",
         "/code/spawn_pipeline.py",
         "/code/data-pipeline.yml"
       ],
--- a/infra/taskcluster-hook-data-pipeline.json
+++ b/infra/taskcluster-hook-data-pipeline.json
@@ -37,7 +37,7 @@
         "taskclusterProxy": true
       },
       "command": [
-        "/usr/local/bin/python3",
+        "python3",
         "/code/spawn_pipeline.py",
         "/code/data-pipeline.yml"
       ],
--- a/infra/taskcluster-hook-check-models-start.json
+++ b/infra/taskcluster-hook-check-models-start.json
@@ -31,7 +31,7 @@
         "taskclusterProxy": true
       },
       "command": [
-        "python3",
+        "python",
         "/code/spawn_pipeline.py",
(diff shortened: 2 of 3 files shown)


suhaibmujahid (dismissed):
We could try. But these command use the `mozilla/bugbug-spawn-pipeline` image, and in that image we do install `uv`, but we do not install a different python version or use a `uv` image with preinstalled python.

infra/taskcluster-hook-data-pipeline.json
```diff
@@ -37,7 +37,7 @@
         "taskclusterProxy": true
       },
       "command": [
-        "/usr/local/bin/python3",
+        "python3",
```
suhaibmujahid: The issue is likely that the venv isn't activated. In the image itself, we use `python`; I would use the same here for constancy:
https://github.com/mozilla/bugbug/blob/48e0a12fa6d74a9d3ad5d7f8cf020953739cd87b/infra/dockerfile.spawn_pipeline#L18

marco-c:
I made the change manually temporarily in the hook and it worked: https://community-tc.services.mozilla.com/tasks/JaBZu3LUSgCiH3_4ki-UBA.

suhaibmujahid:
> I made the change manually temporarily in the hook and it worked: https://community-tc.services.mozilla.com/tasks/JaBZu3LUSgCiH3_4ki-UBA.

It makes sense; we activate the venv by adding it to `PATH`, so using `python`/`python3` would point to the one from the `venv`.

infra/taskcluster-hook-data-pipeline.json
marco-c: Done!

PR #5989, opened 2026-04-29
hg: set `mozext.disable_local_database = true` (Bug 2024739)

cgsheeh:
This disables the local `changetracker.db` database, which still uses
the old `pushlog` wire protocol command instead of `pushlog-stream`.

### DIFF
--- a/infra/hgrc
+++ b/infra/hgrc
@@ -7,7 +7,7 @@ pushlog = /version-control-tools/hgext/pushlog
 mozext = /version-control-tools/hgext/mozext
 
 [mozext]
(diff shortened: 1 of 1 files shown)


cgsheeh:
After looking closer, I think we might need a slightly more involved fix here. Converting to draft while I investigate further.

marco-c:
@cgsheeh do you want to convert out of draft and test this on staging?

cgsheeh:
> @cgsheeh do you want to convert out of draft and test this on staging?

Moved out of draft, not sure how to deploy to stage.

suhaibmujahid (approved):
I will merge it and release a new version to deploy.

PR #5997, opened 2026-05-06
Use multiple processes for mining commits in the HTTP service

### LINKED ISSUE #5387: Use multiple processes for mining commits in the HTTP service
Since we will no longer host bugbug on Heroku, we are not as concerned with memory usage as before.

Basically reverting b74bf96400d0e845705ae41ec308119125fa31ab and 2ac0bc4c12bf0bd37abcbd631ef4f503a46aeee5 will do the trick.

marco-c:
Fixes #5387


suhaibmujahid (approved):
LGTM! Thank you!

PR #6019, opened 2026-05-11
Add Test Plan Generation Tool

jpangas:


bugbug/tools/test_plans_generator/data_types.py
suhaibmujahid, line 1: I would call it something like `test_plans_generator`, the key here is "paln".

bugbug/tools/test_generation/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from bugbug.tools.test_generation.agent import TestGenerationTool
+from bugbug.tools.test_generation.data_types import TestGenerationResult
+
+__all__ = [
+    "TestGenerationResult",
+    "TestGenerationTool",
+]
```
suhaibmujahid: I do not think we need this, at least for now.

suhaibmujahid (commented):
Thank you! Looks clean! Did you do a dry-run for this version? How does the results looks like?

jpangas:
I have performed a dry-run, and it is working correctly.

PR #6035, opened 2026-05-14
Bug 1830760 - For manifests selected with very high confidence (e.g. manifests with modified tests), run on all configs

marco-c:
Ideally we'd be smarted and e.g. run a modified tests only on all configs where the test actually runs, but I think this is still better than the current state.
The current algorithm uses history about previous failures, so if a test failed in the same way e.g. on Linux and Windows, it knows it can schedule on each independently. When a test or manifest is modified though, this assumption breaks because the failure mode could change.

### DIFF
--- a/bugbug/models/testselect.py
+++ b/bugbug/models/testselect.py
@@ -280,7 +280,7 @@ def reduce_configs(
 
 
 def select_configs(
-    groups: Collection[str],
+    group_confidences: dict[str, float],
     min_redundancy_confidence: float,
     max_configurations: int = 3,
 ) -> dict[str, list[str]]:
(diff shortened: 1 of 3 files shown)


bugbug/models/testselect.py
```diff
@@ -290,6 +290,12 @@ def select_configs(
     all_configs_by_group = pickle.loads(failing_together[b"$CONFIGS_BY_GROUP$"])
     config_costs = {config: _get_cost(config) for config in all_configs}
 
+    all_groups = group_confidences.keys()
+    high_confidence_groups = {
+        group for group in all_groups if group_confidences.get(group, 0.0) >= 0.99
+    }
+    groups = [group for group in all_groups if group not in high_confidence_groups]
```
suhaibmujahid, line 297: nit: I guess the name `groups` used here to minimize the change, but since it is not including all groups, it would be a bit confusing. I would name it something `other_groups`.

suhaibmujahid (approved):

PR #6051, opened 2026-05-25
Add more tools (based on searchfox-py) to the LLM review agent, and make it patch stack aware to a certain extent

padenot:
The two patches can be reviewed independently but are close in location so I'm putting one branch for both.


padenot:
https://github.com/padenot/searchfox-cli/compare/v0.10.12...v0.13.0 are the changes in `searchfox-py` for this (API tightening, adding async, etc.).

bugbug/tools/code_review/agent.py
```diff
@@ -94,7 +103,18 @@ def __init__(
...
+            search_text,
+            get_blame,
+            get_field_layout,
+            find_definition,
+            search_identifier,
+            calls_from,
+            calls_to,
+            calls_between,
+            check_can_gc,
+        ]
```
marco-c: It would be preferable to have this as a list somewhere and then use it here and import it above, so we don't have to remember when we add a new searchfox tool that we also need to add it here

bugbug/tools/code_review/langchain_tools.py
```diff
@@ -7,29 +7,52 @@
...
+    reraise=True,
+)
 
 
 def _tool_error(message: str, *, fatal: bool = False) -> str:
     prefix = "Fatal" if fatal else "Warning"
     return f"{prefix}: {message}"
 
 
+logger = getLogger(__name__)
```
marco-c: Nit: move this back above so it doesn't show as diff (unless it's necessary for linting)

bugbug/tools/code_review/langchain_tools.py
```diff
@@ -7,29 +7,52 @@
...
+
+LangStr = Literal[
+    "cpp", "c", "js", "webidl", "java", "kotlin", "rust", "python", "html", "css"
+]
```
marco-c, line 39: Nit: we could get a list of langs from bugbug/repository.py (SOURCE_CODE_TYPES_TO_EXT.keys())

bugbug/tools/code_review/langchain_tools.py
```diff
@@ -7,29 +7,52 @@
...
 @dataclass
 class CodeReviewContext:
     patch: Patch
 
 
+def _get_client() -> AsyncSearchfoxClient:
+    global _client
+    if _client is None:
+        _client = AsyncSearchfoxClient()
+    return _client
```
marco-c: Nit: you could use `cache` from functools to avoid managing it yourself

bugbug/tools/code_review/searchfox_tools.py
```diff
@@ -60,8 +83,20 @@ async def expand_context(
...
+    client = _get_client()
+
+    async def fetch(path: str) -> str:
+        if revision:
+            try:
+                return await _retry(client.get_file_at_revision)(path, revision)
+            except Exception:
+                # searchfox raises plain Exception; fall back to tip
+                pass
+        return await _retry(client.get_file)(path)
```
marco-c: We could use the old cose as fall back and get the `before` code from Phab

bugbug/tools/code_review/agent.py
```diff
@@ -94,7 +103,18 @@ def __init__(
...
         if function_search:
             tools.append(create_find_function_definition_tool(function_search))
```
marco-c: We don't need this anymore, do we?

bugbug/tools/code_review/langchain_tools.py
```diff
@@ -7,29 +7,52 @@
...
-from typing import Optional
+from typing import Literal, Optional
 
+import tenacity
 from langchain.tools import tool
 from langgraph.runtime import get_runtime
 from requests import HTTPError
+from searchfox import AsyncSearchfoxClient
 
 from bugbug.code_search.function_search import FunctionSearch
```
marco-c: We had different types of function search tools implementing this FunctionSearch interface, your approach is to instead just use searchfox directly.
Should we implement yours in the FunctionSearch interface too?

Currently we don't really need the abstraction, but we might want it if we add support for GitHub where we don't have Searchfox. WDYT?

bugbug/tools/code_review/langchain_tools.py
padenot: They're too dissimilar. This comes form searchfox-py API.

bugbug/tools/code_review/agent.py
padenot: Removed, and implemented the one thing that was missing (but was a good idea: find the function from a line).

padenot: Removed, and implemented the one thing that was missing (but was a good idea: find the function from a line).

bugbug/tools/code_review/langchain_tools.py
marco-c: We discussed this on Matrix, and it's better to avoid the abstraction and just give the agent different tools depending on the service (given that different services have different features).

tests/test_code_review.py
```diff
@@ -1,12 +1,228 @@
...
+        def patch_set(self):
+            return PatchSet.from_string("")
+
+    fake = FakePatch()
+    stack = fake.patch_stack
+    assert len(stack) == 3  # A, C, E
+
+
+# ---------------------------------------------------------------------------
+# find_comment_scope (pre-existing test)
```
marco-c:
```suggestion
# find_comment_scope
```

or just remove the whole comment here

bugbug/tools/code_review/langchain_tools.py
```diff
@@ -6,92 +6,400 @@
...
+    get_blame,
+    get_field_layout,
+    find_definition,
+    get_function_at_line,
+    search_identifier,
+    calls_from,
+    calls_to,
+    calls_between,
+    check_can_gc,
+]
```
marco-c, line 451: Maybe move in a separate file? I can imagine these could be used by other agents

padenot:
@marco-c last one is a fix for a new failure on main, ptal

marco-c (dismissed):

bugbug/tools/code_review/langchain_tools.py
suhaibmujahid, line 1: I'd keep the filename `langchain_tools.py`, consistent with the naming pattern used by other tools, so it can accommodate additional tools in the future. If it grows too large and separation becomes necessary, we can convert `langchain_tools` from a file to a directory and re-export everything from its `__init__.py`.

bugbug/tools/code_review/searchfox_tools.py
```diff
@@ -0,0 +1,406 @@
...
+        if revision:
+            try:
+                return await _retry(client.get_file_at_revision)(path, revision)
+            except Exception:
+                pass
+        try:
+            return await patch.get_old_file(path)
+        except Exception:
+            pass
+        return await _retry(client.get_file)(path)
```
suhaibmujahid: I would a void deeply nested try-except.

bugbug/tools/code_review/searchfox_tools.py
```diff
@@ -0,0 +1,406 @@
...
+
+    revision = await patch.get_base_revision()
+    client = _get_client()
+
+    async def fetch(path: str) -> str:
+        if revision:
+            try:
+                return await _retry(client.get_file_at_revision)(path, revision)
+            except Exception:
+                pass
```
suhaibmujahid: Using `Exception` is too generic, I would avoid it where possible.

suhaibmujahid (dismissed):

bugbug/tools/core/platforms/patch_apply.py
```diff
@@ -0,0 +1,87 @@
...
+    for patched_file in patch_set:
+        if file_path in {
+            patched_file.path,
+            strip_diff_prefix(patched_file.target_file),
+        }:
+            return patched_file
+    return None
+
+
+def apply_patched_file(base_content: str, patched_file: PatchedFile) -> str:
```
suhaibmujahid, line 30: Can't we get it from Phab directly?

suhaibmujahid (approved):
LGTM! Thank you!

PR #6115, opened 2026-06-03
Add a demo interface for Hackbot

suhaibmujahid:
Add a new Next.js based Hackbot web interface under services/hackbot-ui to be used for demonstration.


marco-c (approved):
Mostly a rubber stamp since it's for testing purposes

PR #6128, opened 2026-06-04
Introduce a way to load external content when an agent review code

padenot:
See the doc and example in the last commit first.

### DIFF
--- a/bugbug/tools/code_review/agent.py
+++ b/bugbug/tools/code_review/agent.py
@@ -70,6 +70,10 @@ class CodeReviewTool(GenerativeModelTool):
         target_software: str = "Mozilla Firefox",
         todo_enabled: bool = True,
         skills: Optional[list[Skill]] = None,
+        review_context_repo: Optional[str] = None,
+        review_context_branch: str = "main",
+        extra_context_toml: Optional[str] = None,
+        content_overrides: Optional[dict[str, str]] = None,
     ) -> None:
         super().__init__()
 
@@ -125,6 +129,11 @@ class CodeReviewTool(GenerativeModelTool):
         self.show_patch_example = show_patch_example
 
         self.verbose = verbose
+
+        self._review_context_repo = review_context_repo
+        self._review_context_branch = review_context_branch
+        self._extra_context_toml = extra_context_toml
+        self._content_overrides = content_overrides
 
     @property
     def _agent_model_name(self) -> str:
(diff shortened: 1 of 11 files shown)


bugbug/tools/code_review/agent.py
```diff
@@ -214,14 +219,47 @@ async def generate_review_comments(
...
 
-    async def run(self, patch: Patch) -> CodeReviewToolResponse:
+    async def run(
+        self,
+        patch: Patch,
+        review_context_repo: Optional[str] = None,
+        review_context_branch: str = "main",
+        extra_context_toml: Optional[str] = None,
+        content_overrides: Optional[dict[str, str]] = None,
+    ) -> CodeReviewToolResponse:
```
suhaibmujahid: The `run` method should stay as it is (i.e., `run(self, patch: Patch)`) to not break the interface, which we use in the extermination pipeline as well.

bugbug/tools/code_review/review_context.py
```diff
@@ -0,0 +1,591 @@
...
+        bugs: list[dict] = []
+        Bugzilla(
+            bug_id,
+            include_fields=["product", "component"],
+            bughandler=lambda bug, data: data.append(bug),
+            bugdata=bugs,
+        ).get_data().wait()
+        if not bugs:
+            return None
+        return f"{bugs[0]['product']}::{bugs[0]['component']}"
```
suhaibmujahid: We could get that from the patch class instead of fetching the bug here.

bugbug/tools/code_review/prompts.py
```diff
@@ -63,6 +63,7 @@
 <patch_summary>
 {patch_summarization}
 </patch_summary>
+{external_context}
```
suhaibmujahid, line 68: Let us wrap it with an XML tag to make a clear boundary.
```suggestion

<external-resources>
{external_context}
</external-resources>
```

bugbug/tools/code_review/agent.py
```diff
@@ -214,14 +219,47 @@ async def generate_review_comments(
...
+                review_context_branch=review_context_branch,
+                bug_id=bug_id,
+                extra_context_toml=extra_context_toml,
+                content_overrides=content_overrides,
+            )
+            if content_items:
+                external_context = format_external_content(content_items)
+                external_content_manifest = build_external_content_manifest(
+                    content_items
+                )
```
suhaibmujahid: I suggest moving this into `bugbug.tools.code_review.review_context`

bugbug/tools/code_review/agent.py
```diff
@@ -214,14 +219,47 @@ async def generate_review_comments(
...
+
         unfiltered_suggestions = await self.generate_review_comments(
-            patch, patch_summary
+            patch, patch_summary, external_context
         )
```
suhaibmujahid: I suggest retrieving `external_context` inside `generate_review_comments()` so we don't need to refactor the experimentation pipeline to accommodate this.

suhaibmujahid (dismissed):
Thank you! Just minor comments to address and then we should land it ASAP.

bugbug/tools/code_review/agent.py
padenot: I don't know what that is, it's not tested?

padenot: Fixed in any case.

bugbug/tools/code_review/review_context.py
padenot: Moved.

suhaibmujahid (approved):
LGTM! Thank you!

PR #6138, opened 2026-06-08
Clarify error message for private Phabricator revisions

### LINKED ISSUE #5992: Clarify the MCP error message when requesting access to private revision
Displaying an error message stating that a private revision cannot be accessed would be less confusing for both LLMs and humans.

padenot:
Previously, attempting to access a private revision would surface a confusing internal error. Now it raises a `ToolError` with an explicit message explaining the revision is private or not found.

I'm not entirely certain that it's that useful to distinguish, because our token isn't capable of seeing all revisions, but maybe that will change in the future, and feels right anyway.

Fixes #5992

### DIFF
--- a/services/mcp/src/bugbug_mcp/server.py
+++ b/services/mcp/src/bugbug_mcp/server.py
@@ -18,6 +18,7 @@ from bugbug.tools.core.platforms.phabricator import (
     PhabricatorPatch,
     SanitizedPhabricatorPatch,
 )
+from libmozdata.phabricator import PhabricatorRevisionNotFoundException
 
 mcp = FastMCP("Firefox Development MCP Server")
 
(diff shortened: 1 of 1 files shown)


services/mcp/src/bugbug_mcp/server.py
```diff
@@ -139,20 +140,34 @@ def bughandler(bug):
     return result
 
 
+def _get_revision_md(revision_id: int) -> str:
+    try:
+        patch = SanitizedPhabricatorPatch(revision_id=revision_id)
+        if not patch.is_public():
+            raise ToolError(
+                f"Revision D{revision_id} is private and cannot be accessed."
```
suhaibmujahid: This won’t work since we will always get `PhabricatorRevisionNotFoundException` on secure revisions.

suhaibmujahid (changes requested):

services/mcp/src/bugbug_mcp/server.py
padenot: Ok, but this is what I said above.

suhaibmujahid: Since the Web UI itself can differentiate between 404 and secure revisions, we could fall back to scraping the web URL when a revision is not found, if you think the distinction would be useful.

What you have for blocking secure revisions could be helpful anyway as a defensive mechanism; so if the account accidentally gains access to a secure revision, it won't be publicly available.

Nit: we do the check differently in Review Helper, we could align it for consistency:
https://github.com/mozilla/bugbug/blob/15ceb27b989081a717337d3c93b24204238a9072/services/reviewhelper-api/app/review_processor.py#L63

suhaibmujahid (dismissed):
Could you please fix the linting error?

suhaibmujahid (approved):

PR #6139, opened 2026-06-08
Include diff ID in review comment

### LINKED ISSUE #6029: Add Diff ID for reviewhelper-bot's Summary of Changes
Because the summary is provided as a global comment, it's a bit harder to see at what point the review happened - It'd be nice if it had something like (Reviewing Diff N, ID <...>).

padenot:
Adds a linkified diff reference to the top of the main review comment, so reviewers can tell which diff was reviewed at a glance.

Fixes #6029

### DIFF
--- a/services/reviewhelper-api/app/review_processor.py
+++ b/services/reviewhelper-api/app/review_processor.py
@@ -167,7 +167,8 @@ def create_main_review_comment(
     review_request: ReviewRequest, generated_comments: Collection[GeneratedComment]
 ) -> str:
     """Create the main review comment that summarizes the review results."""
-    parts = []
+    diff_url = f"{settings.phabricator_url}/D{review_request.revision_id}?id={review_request.diff_id}"
+    parts = [f"(Reviewing [Diff {review_request.diff_id}]({diff_url}))"]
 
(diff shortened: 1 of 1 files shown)


services/reviewhelper-api/app/review_processor.py
```diff
@@ -167,7 +167,8 @@ def create_main_review_comment(
     review_request: ReviewRequest, generated_comments: Collection[GeneratedComment]
 ) -> str:
     """Create the main review comment that summarizes the review results."""
-    parts = []
+    diff_url = f"{settings.phabricator_url}/D{review_request.revision_id}?id={review_request.diff_id}"
+    parts = [f"(Reviewing [[{diff_url}|Diff {review_request.diff_id}]])"]
```
suhaibmujahid: I suggest using Markdown instead of Remarkup, since it's more intuitive to read in logs. Phabricator supports both!


```suggestion
    parts = [f"(Reviewing [Diff {review_request.diff_id}]({diff_url}))"]
```

suhaibmujahid (dismissed):

suhaibmujahid (approved):

PR #6145, opened 2026-06-09
Simplify hackbot agent structure into self-contained agents/<name>/folders

suhaibmujahid:
Co-locate each agent's logic, entrypoint, and deployment in one folder; move shared building blocks (Reporter, actions_server_for, Bugzilla MCP) into hackbot-runtime; derive Job env from the input schema (model_to_env) instead of per-agent build_env. Migrate bug-fix as the reference.


libs/agent-tools/tests/test_firefox.py
```diff
@@ -0,0 +1,26 @@
...
+
+async def _list(server):
+    return (
+        await server.request_handlers[ListToolsRequest](
+            ListToolsRequest(method="tools/list")
+        )
+    ).root.tools
+
+
+async def test_exposes_firefox_tools(tmp_path):
```
marco-c, line 16: Is it worth it?

suhaibmujahid: We have generic tools to be shared across all SDKs, and SDK-specific wrappers around them. This test verifies that everything is wired up correctly.

libs/agent-tools/agent_tools/registry.py
```diff
@@ -0,0 +1,107 @@
+"""Framework-neutral declaration of agent tools.
```
marco-c, line 1: Hopefully it won't be too painful to maintain this framework-neutral adapter

suhaibmujahid: For now, this is just a thin wrapper. If it grows complex with edge cases down the road, it can be dropped in favor of an independent implementation at the cost of some duplication. Conceptually, though, this is scoped to tools only; not a generic, SDK-wide abstraction, so it should not grow to be complex.

agents/README.md
```diff
@@ -0,0 +1,113 @@
...
+  pyproject.toml          # dist "hackbot-agent-<name>"; packages = ["hackbot_agents"]; deps: hackbot-runtime[claude-sdk] + agent-specific
+  hackbot.toml            # declares platform capabilities: [source], [firefox]
+  Dockerfile              # multi-stage: builder / agent [/ broker]
+  compose.yml             # local run; sets static env (e.g. the broker URL)
+  hackbot_agents/         # shared PEP 420 namespace — NO __init__.py here
+    <name_snake>/         # the agent package (e.g. bug_fix)
+      __init__.py         # empty package marker
+      agent.py            # run_<name>() logic + helpers (the reusable agent library)
+      __main__.py         # entrypoint: AgentInputs(BaseSettings) + async def main(ctx) -> dict + run_async(main)
+      prompts/ rules/     # assets read via Path(__file__).parent
```
marco-c: Could you write a better description for these folders?

agents/README.md
```diff
@@ -0,0 +1,113 @@
...
+  results/artifacts/actions plumbing (`ctx.actions`, `ctx.publish_file`,
+  `ctx.publish_json`).
+- `from hackbot_runtime import ensure_source_repo` — the lower-level shallow-clone/refresh
+  primitive (you normally don't call this directly; `ctx.source_repo` does it for you).
+- `from hackbot_runtime.claude import Reporter` — renders streamed claude-agent-sdk
+  messages to stdout/log. Call `reporter.header("...")` per work item, `reporter.message(msg)` per message.
+- `from hackbot_runtime.actions.claude_sdk import actions_server_for` — returns
+  `(recorder, mcp_server)`; write actions land in `summary.json` instead of mutating anything.
+
+Reusable MCP **tool servers** live in the separate `agent-tools` package, each behind its
```
marco-c: Do they have to be MCP?
Also, how about CLI tools?

agents/README.md
```diff
@@ -0,0 +1,113 @@
...
+   (omit either otherwise).
+2. `agents/<name>/hackbot_agents/<name_snake>/__main__.py` — define `AgentInputs(BaseSettings)`
+   (domain inputs only), an `async def main(ctx: HackbotContext) -> dict` that returns
+   findings on success and raises `AgentError` to fail, and end with `run_async(main)` (it
+   discovers `hackbot.toml` — cwd, then up to the agent root — and exits the process itself).
+3. `agents/<name>/hackbot_agents/<name_snake>/agent.py` — your prompts/logic, exposing the
+   `run_<name>()` entrypoint `main` calls (leave `<name_snake>/__init__.py` empty). Do **not**
+   create `agents/<name>/hackbot_agents/__init__.py`.
+4. Copy `pyproject.toml`, `Dockerfile`, `compose.yml` from `bug-fix/` and rename (the
+   Dockerfile CMDs become `python -m hackbot_agents.<name>` / `… .broker`, and it copy
```
marco-c:
```suggestion
   Dockerfile CMDs become `python -m hackbot_agents.<name>` / `… .broker`, and it copies
```

agents/bug-fix/hackbot_agents/bug_fix/__main__.py
```diff
@@ -0,0 +1,58 @@
...
+    max_turns: int | None = None
+    effort: str | None = None
+
+    model_config = SettingsConfigDict(extra="ignore")
+
+
+async def main(ctx: HackbotContext) -> dict:
+    inputs = AgentInputs()
+    # Fail fast if the platform did not provide Anthropic credentials.
+    ctx.anthropic.api_key
```
marco-c: should we do assert?

marco-c (dismissed):

agents/README.md
suhaibmujahid: This only applies to tools passed directly to the agent. Claude Code, for example, accepts MCP tools exclusively; so these generic tools need to be wrapped in an MCP server. Other agentic harnesses have their own integration methods, which is why we use wrappers rather than exposing raw MCP tools from the start.

For CLI tools, they can be used through the `Bash` tool. They do not need to live here.

marco-c: Do we want to plan some sort of common structure to also add CLI tools? Or will each agent do its own thing?

agents/README.md
suhaibmujahid: Fixed in 5a08a060f243d72d25bff4cf161dd62aedf99734

agents/bug-fix/hackbot_agents/bug_fix/__main__.py
suhaibmujahid: Removed in 2cff913c76999fe108b6de93221fbf76097dc3e8; a more comprehensive solution will be implemented in #6150

agents/README.md
suhaibmujahid: Improved in 5a08a060f243d72d25bff4cf161dd62aedf99734

agents/README.md
suhaibmujahid: Yes, and I've filed a follow-up issue (#6151). In the meantime, each agent can install the package as part of its build in the Dockerfile, so this is not blocked.

PR #6148, opened 2026-06-10
Add git hashes in addition to Mercurial ones in the commits DB

### LINKED ISSUE #6094: Include git commit hash in commits db
The commits DB currently only records the hg changeset hash for each commit. Converting these to git commits seems to require per-hash round-trips to the Lando hg2git API or a local git cinnabar checkout.

How hard would it be to add a git_node (or similar) field to each commit dict in the commits DB containing the corresponding git commit hash?

marco-c:
Fixes #6094

### DIFF
--- a/bugbug/repository.py
+++ b/bugbug/repository.py
@@ -4,6 +4,7 @@
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import argparse
+import collections
 import concurrent.futures
 import copy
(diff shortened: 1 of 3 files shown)


suhaibmujahid (changes requested):
LGTM! Just the tests need to reflect the changes.

suhaibmujahid (approved):

PR #6149, opened 2026-06-11
Improve exception handing in langchain_tools.py

padenot:
The searchfox-lib python bindings weren't using typed exceptions, but this got added in 0.16.

This changes the following:
- If the tool call fails because of a logic error (file not found, using wrong arguments, etc.), we log with error level, this goes to sentry. We want to track those and improve API/documentation/etc. -- this is generally actionnable
- If the tool call fails because of a searchfox infra error (network, load, timeout etc.), we allow ourselve to retry, but if that still fails it is logged as a warning -- we can't do much.


suhaibmujahid (approved):
LGTM! Thank you!

PR #6170, opened 2026-06-15
Add webcompat-triage agent

ksy36:
This is a first part of a webcompat triage agent. I'm planning to add a second part (chrome mask extension install) in a  follow-up PR.

### DIFF
--- a/agents/autowebcompat-repro/Dockerfile
+++ b/agents/autowebcompat-repro/Dockerfile
@@ -1,0 +1,64 @@
+FROM python:3.12 AS builder
+
+COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/
+
+ENV UV_PROJECT_ENVIRONMENT=/opt/venv
+
+WORKDIR /app
+
+# Install external deps without building workspace members.
+RUN --mount=type=cache,target=/root/.cache/uv \
+    --mount=type=bind,source=pyproject.toml,target=pyproject.toml \
+    --mount=type=bind,source=uv.lock,target=uv.lock \
+    --mount=type=bind,source=VERSION,target=VERSION \
+    uv sync --frozen --no-dev --no-install-workspace --package hackbot-agent-autowebcompat-repro
+
+RUN --mount=type=cache,target=/root/.cache/uv \
+    --mount=type=bind,target=/app,rw \
+    uv sync --locked --no-dev --no-editable --package hackbot-agent-autowebcompat-repro
+
+FROM python:3.12 AS base
+
+COPY --from=builder /opt/venv /opt/venv
+WORKDIR /app
+
+ENV PYTHONUNBUFFERED=1
+ENV PYTHONDONTWRITEBYTECODE=1
+ENV PATH="/opt/venv/bin:$PATH"
+
+FROM base AS agent
+
+# The Firefox DevTools MCP server is an npm package launched via `npx`, so the
+# agent image needs Node.js + npm (the python base ships neither). It also
+# needs the shared libraries Firefox requires to run headless; the Firefox
+# binary itself is downloaded at agent startup (a fresh Nightly per run) via
+# mozdownload/mozinstall, not baked in here.
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+        nodejs npm \
+        ca-certificates \
+        libgtk-3-0 libdbus-glib-1-2 libx11-xcb1 libxtst6 libxt6 \
+        libasound2 libpci3 \
+    && rm -rf /var/lib/apt/lists/*
+
+# hackbot.toml lives at the agent root (not inside the package), so copy it into
+# the working dir; the runtime discovers it there (cwd) at startup.
+COPY agents/autowebcompat-repro/hackbot.toml /app/hackbot.toml
+
+RUN useradd --create-home --shell /bin/bash agent \
+    && mkdir -p /workspace \
+    && chown agent:agent /workspace
+
+USER agent
+
+CMD ["python", "-m", "hackbot_agents.autowebcompat_repro"]
+
+FROM base AS broker
+
+RUN useradd --create-home --shell /bin/bash broker
+
+USER broker
+
+EXPOSE 8765
+
+CMD ["python", "-m", "hackbot_agents.autowebcompat_repro.broker"]
--- a/agents/autowebcompat-repro/compose.yml
+++ b/agents/autowebcompat-repro/compose.yml
@@ -1,0 +1,31 @@
+services:
+  autowebcompat-repro-broker:
+    build:
+      context: ../..
+      dockerfile: agents/autowebcompat-repro/Dockerfile
+      target: broker
+    environment:
+      BUGZILLA_API_URL: ${BUGZILLA_API_URL}
+      BUGZILLA_API_KEY: ${BUGZILLA_API_KEY}
+    expose:
+      - "8765"
+
+  autowebcompat-repro-agent:
+    build:
+      context: ../..
+      dockerfile: agents/autowebcompat-repro/Dockerfile
+      target: agent
+    environment:
+      - RUN_ID
+      - BUG_DATA
+      - BUG_ID
+      - BUGZILLA_MCP_URL=http://autowebcompat-repro-broker:8765/mcp
+      - ANTHROPIC_API_KEY=${ANTHROPIC_API_KEY:?error}
+      # No uploader locally: summary/logs/attachments are written under
+      # /artifacts/<run_id>, bind-mounted to the host's ~/hackbot/artifacts.
+      - ARTIFACTS_DIR=/artifacts
+    volumes:
+      - ${HOME}/hackbot/artifacts:/artifacts
+    depends_on:
+      autowebcompat-repro-broker:
+        condition: service_started
--- a/agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/__main__.py
+++ b/agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/__main__.py
@@ -1,0 +1,43 @@
+from hackbot_runtime import HackbotContext, run_async
+from pydantic_settings import BaseSettings, SettingsConfigDict
+
+from .agent import AutowebcompatReproResult, run_autowebcompat_repro
+from .firefox_install import install_firefox_nightly
+
+
+class AgentInputs(BaseSettings):
+    bugzilla_mcp_url: str
+    bug_data: str | None = None
+    bug_id: int | None = None
+    model: str | None = None
+    max_turns: int | None = None
+    effort: str | None = None
+
+    model_config = SettingsConfigDict(extra="ignore")
+
+
+async def main(ctx: HackbotContext) -> AutowebcompatReproResult:
+    inputs = AgentInputs()
+
+    # Provision a fresh Nightly at startup so each run reproduces against a
+    # current build; drive the binary the install reports back.
+    firefox_path = str(install_firefox_nightly())
+
+    return await run_autowebcompat_repro(
+        bugzilla_mcp_server={
+            "type": "http",
+            "url": inputs.bugzilla_mcp_url,
+        },
+        bug_data=inputs.bug_data,
+        bug_id=inputs.bug_id,
+        model=inputs.model,
+        max_turns=inputs.max_turns,
+        effort=inputs.effort,
+        firefox_path=firefox_path,
+        log=ctx.log_path,
+        verbose=True,
+    )
+
+
+if __name__ == "__main__":
+    run_async(main)
--- a/agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/agent.py
+++ b/agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/agent.py
@@ -1,0 +1,155 @@
+"""Firefox web-compatibility reproduction agent.
+
+Drives an agent that reproduces a broken-site report in Firefox
+using the Firefox DevTools MCP. The bug is passed either inline as ``bug_data``
+text or a Bugzilla ``bug_id`` (read via Bugzilla broker).
+"""
+
+from __future__ import annotations
+
+import logging
+from pathlib import Path
+
+from claude_agent_sdk import (
+    ClaudeAgentOptions,
+    ClaudeSDKClient,
+    McpServerConfig,
+    ResultMessage,
+)
+from hackbot_runtime import AgentError, HackbotAgentResult
+from hackbot_runtime.claude import Reporter
+
+from .config import BUGZILLA_READ_TOOLS, DEVTOOLS_TOOLS
+from .devtools_mcp import build_devtools_server
+from .result import (
+    RESULT_SERVER_NAME,
+    SUBMIT_RESULT_TOOL,
+    ReproductionResult,
+    ResultCollector,
+    build_result_server,
+)
+
+HERE = Path(__file__).resolve().parent
+
+logger = logging.getLogger("autowebcompat-repro")
+
+
+class AutowebcompatReproResult(HackbotAgentResult):
+    result: ReproductionResult | None = None
+
+
+def load_system_prompt() -> str:
+    return (HERE / "prompts" / "system.md").read_text()
+
+
+def build_user_prompt(bug_data: str | None, bug_id: int | None) -> str:
+    if bug_data:
+        return (
+            "Here is the web-compatibility report to work on:\n\n"
+            f"{bug_data}\n\n"
+            "Follow your task procedure."
+        )
+    if bug_id is not None:
+        return (
+            f"The web-compatibility report to work on is Bugzilla bug {bug_id}. "
+            "Fetch it using the Bugzilla MCP tools, then follow your task procedure."
+        )
+    raise AgentError("neither bug_data nor bug_id was provided")
+
+
+async def run_autowebcompat_repro(
+    *,
+    bugzilla_mcp_server: McpServerConfig,
+    bug_data: str | None = None,
+    bug_id: int | None = None,
+    model: str | None = None,
+    max_turns: int | None = None,
+    effort: str | None = None,
+    firefox_path: str | None = None,
+    verbose: bool = False,
+    log: Path | None = None,
+) -> AutowebcompatReproResult:
+    """Reproduce a web-compat issue and return the agent's findings.
+
+    Returns a :class:`AutowebcompatReproResult` on success; raises
+    :class:`AgentError` if the agent ends in an error.
+    """
+    subject = bug_data if bug_data else f"bug {bug_id}"
+    preview = subject if len(subject) <= 200 else f"{subject[:200]}..."
+    logger.info("reproducing %s", preview)
+
+    devtools_server = build_devtools_server(
+        firefox_path=Path(firefox_path) if firefox_path else None,
+        headless=True,
+        enable_script=True,
+    )
+
+    # Structured-result MCP server (in-process): the agent calls submit_result
+    # once at the end, giving a predictable JSON result instead of free text.
+    result_collector = ResultCollector()
+    result_server = build_result_server(result_collector)
+
+    # Only wire up Bugzilla when there's a bug to fetch. With inline bug_data
+    # there's nothing to read, so the bugzilla MCP is not available
+    mcp_servers: dict[str, McpServerConfig] = {
+        "firefox-devtools": devtools_server,
+        RESULT_SERVER_NAME: result_server,
+    }
+    bugzilla_tools: list[str] = []
+    if bug_id is not None:
+        mcp_servers["bugzilla"] = bugzilla_mcp_server
+        bugzilla_tools = BUGZILLA_READ_TOOLS
+
+    system_prompt = load_system_prompt()
+
+    options = ClaudeAgentOptions(
+        system_prompt=system_prompt,
+        mcp_servers=mcp_servers,
+        permission_mode="bypassPermissions",
+        allowed_tools=[
+            "Read",
+            "Grep",
+            "Glob",
+            "Bash",
+            *bugzilla_tools,
+            *DEVTOOLS_TOOLS,
+            SUBMIT_RESULT_TOOL,
+        ],
+        model=model,
+        max_turns=max_turns,
+        **({"effort": effort} if effort else {}),
+        setting_sources=[],
+        # DevTools snapshots/screenshots of complex pages serialize to JSON that
+        # can exceed the SDK's default 1 MiB message buffer (the reader dies
+        # fatally if it does). Raise it well above that ceiling.
+        max_buffer_size=10 * 1024 * 1024,
+    )
+
+    user_prompt = build_user_prompt(bug_data, bug_id)
+
+    result_msg: ResultMessage | None = None
+    with Reporter(verbose=verbose, log_path=log) as reporter:
+        reporter.header(subject)
+        async with ClaudeSDKClient(options=options) as client:
+            await client.query(user_prompt)
+            async for msg in client.receive_response():
+                reporter.message(msg)
+                if isinstance(msg, ResultMessage):
+                    result_msg = msg
+
+    if result_msg is None:
+        raise AgentError(f"{subject}: agent produced no result message")
+    if result_msg.is_error:
+        raise AgentError(
+            f"{subject} investigation failed: {result_msg.result or result_msg.subtype}"
+        )
+    if result_collector.result is None:
+        raise AgentError(
+            f"{subject}: agent finished without submitting a result via submit_result"
+        )
+
+    return AutowebcompatReproResult(
+        result=result_collector.result,
+        num_turns=result_msg.num_turns,
+        total_cost_usd=result_msg.total_cost_usd,
+    )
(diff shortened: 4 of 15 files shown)


services/hackbot-api/app/schemas.py
```diff
@@ -67,3 +67,18 @@ class BugFixInputs(BaseModel):
     model: str | None = None
     max_turns: int | None = None
     effort: str | None = None
+
+
+class AutoWebcompatInputs(BaseModel):
+    bug_data: str | None = None
+    bug_id: int | None = None
+    mode: Literal["triage", "chrome-mask"] = "triage"
```
ksy36: If we decide to have a single agent for all webcompat work (triage/diagnosis/ intervention), we could pass a mode here and have a prompt specific to the task, while keeping generic rules in system.md. I've added `triage` prompt as an example.

services/hackbot-api/app/schemas.py
```diff
@@ -67,3 +67,18 @@ class BugFixInputs(BaseModel):
...
+    model: str | None = None
+    max_turns: int | None = None
+    effort: str | None = None
+
+    @model_validator(mode="after")
+    def _require_subject(self) -> "AutoWebcompatInputs":
+        if self.bug_data is None and self.bug_id is None:
+            raise ValueError("provide at least one of bug_data or bug_id")
+        return self
```
ksy36, line 83: It accepts either `bug_data` or a `bug_id`. If `bug_id` is provided, it'll fetch the bug data using bugzilla MCP. In case if `bug_data` is provided, it will work with that. 

It can be a report text, something like:
`https://aifaceswap.io/#face-swap-playground — after generating a face-swapped image, clicking the Download button under the generated image does nothing in Firefox (works in Chrome)`. 

It is currently a `str`, but we can agree on format that would come from bugbot.

agents/autowebcompat/hackbot_agents/autowebcompat/prompts/chrome_mask.md
```diff
@@ -0,0 +1,4 @@
+## Task: Chrome Mask test
+
+Run the **Chrome Mask test** to determine whether the broken behavior stems from
+User-Agent sniffing.
```
ksy36: This is just a placeholder for now, I'll add the actual prompt and a creation of a profile with an xpi in a follow up PR.

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/agent.py
```diff
@@ -0,0 +1,156 @@
...
+        raise AgentError(
+            f"{subject} investigation failed: {result_msg.result or result_msg.subtype}"
+        )
+
+    return AutoWebcompatResult(
+        mode=mode,
+        result=result_msg.result,
+        num_turns=result_msg.num_turns,
+        total_cost_usd=result_msg.total_cost_usd,
+    )
```
ksy36, line 155: Looks like result of the run is stored as an artifact in a json file. We should decide what the result should look like. So for a sample triage run it produced this in summary.json:

```
{
  "status": "ok",
  "error": null,
  "findings": {
    "num_turns": 78,
    "total_cost_usd": 3.2635555,
    "mode": "triage",
    "result": "I've confirmed the reproduction. Here's my report:\n\n## Reproduction Report\n\n**URL:** https://aifaceswap.io/#face-swap-playground\n**Browser:** Firefox (headless via Firefox DevTools MCP)\n**Date tested:** 2026-06-15\n\n### Steps taken\n1. Navigated to the Face Swap Playground.\n2. Selected the Mona Lisa template via the \"use it\" button, then replaced the Original image by uploading `face.jpg` (a woman portrait fetched from the site's own template set).\n3. Uploaded `face2.jpg` (a man portrait, also from the site's templates) into the **Face image** slot.\n4. Clicked **Start face swapping**. The queue progressed from ~25% \u2192 95% \u2192 complete (~35 seconds).\n5. A swapped result image appeared in the right pane, with three action buttons below it: a small circular **Download** button (download icon), **Upscale**, and **Enhance**.\n6. Clicked the Download button (`<button class=\"btn btn-icon btn-sm btn-primary me-5\"><i class=\"bi bi-download\"></i></button>`).\n\n### Result \u2014 bug reproduced\nClicking Download **does nothing** in Firefox:\n- No file save dialog appears.\n- No new tab opens.\n- No network request is issued after the click.\n- No console messages are logged.\n- The page state is unchanged.\n\nI tried both a UID-targeted click via DevTools and a programmatic `.click()` from JavaScript \u2014 both produced no visible effect and no observable side effects.\n\nThe reported web-compatibility issue is reproducible as described."
  },
  "actions": []
```
The sample triage prompt didn't ask to leave a comment or attach testcases/screenshots, but if we add that and there is a bug id passed as well, these could be in the `actions` list:
```
ENABLED_ACTION_TYPES = [
    "bugzilla.update_bug",
    "bugzilla.add_comment",
    "bugzilla.add_attachment",
    "bugzilla.create_bug",
]
```
For example,
```
  "actions": [
    {
      "type": "bugzilla.add_comment",
      "params": {
        "bug_id": 2046445,
        "text": "Reproduced on Firefox Nightly using firefox-devtools MCP tooling on Linux.\n\n**Steps performed:**\n1. Navigated to https://aifaceswap.io/#face-swap-playground\n2. Uploaded two images via the \"Original image\" and \"Face image\" upload widgets (programmatically populated the two `input[type=\"file\"]` elements with image/webp files and dispatched a `change` event)\n3. Clicked \"Start face swapping\" \u2014 the face-swap pipeline ran, the progress bar advanced to completion, and the swapped image appeared in the result panel along with the action toolbar (download icon button, Upscale, Enhance)\n4. Clicked the download icon button (`button.btn.btn-icon.btn-sm.btn-primary.me-5` containing `<i class=\"bi bi-download\">`)\n\n**Result:** Nothing happens.\n- No file download is initiated (no \"Save As\" dialog, no entry added to downloads).\n- No new tab/page is opened.\n- No navigation occurs.\n- No visible UI change on the page.\n- No anchor element with a `download` attribute is programmatically clicked (verified by instrumenting `HTMLAnchorElement.prototype.click`).\n- No console messages are emitted.\n\nI also tried dispatching a full `mousedown` \u2192 `mouseup` \u2192 `click` MouseEvent sequence at the button's center coordinates in addition to a plain `.click()` \u2014 the behavior was identical (no download, no UI change).\n\nThis matches the actual behavior described in comment 0. Issue reproduces on the current site build.\n\n*This is an automated analysis result. If this result is incorrect please add a needinfo and feel free to correct the error.* ",
        "is_private": false
      },
      "reasoning": "Confirming reproduction of the broken download button for the webcompat triage. Reporting only what was observed, no root-cause analysis."
    }
```

agents/autowebcompat/hackbot_agents/autowebcompat/__main__.py
```diff
@@ -0,0 +1,44 @@
...
+from pydantic_settings import BaseSettings, SettingsConfigDict
+
+from .agent import AutoWebcompatResult, run_autowebcompat
+
+
+class AgentInputs(BaseSettings):
+    bugzilla_mcp_url: str
+    bug_data: str | None = None
+    bug_id: int | None = None
+    mode: str = "triage"
```
jgraham: Define this as an Enum.

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/__main__.py
```diff
@@ -0,0 +1,44 @@
+from hackbot_runtime import HackbotContext, run_async
+from pydantic_settings import BaseSettings, SettingsConfigDict
+
+from .agent import AutoWebcompatResult, run_autowebcompat
+
+
+class AgentInputs(BaseSettings):
+    bugzilla_mcp_url: str
+    bug_data: str | None = None
```
jgraham, line 10: I think we should consider making this explicitly a JSON object rather than just a string (although maybe in the end it doesn't make much difference?).

agents/autowebcompat/hackbot_agents/autowebcompat/__main__.py
jgraham: Let's make this an enum here and throughout.

agents/webcompat-triage/hackbot_agents/webcompat_triage/agent.py
```diff
@@ -0,0 +1,159 @@
...
+        actions_recorder, types=ENABLED_ACTION_TYPES
+    )
+    enabled_action_tools = actions_to_tool_names(ENABLED_ACTION_TYPES)
+
+    system_prompt = load_system_prompt(mode)
+
+    options = ClaudeAgentOptions(
+        system_prompt=system_prompt,
+        mcp_servers={
+            "bugzilla": bugzilla_mcp_server,
```
jgraham: I wonder if we should configure things so that if we don't pass in a bug id we don't give it any access to bugzilla?

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/agent.py
jgraham: My preference, which I don't think is universal, is that we try to get it to produce those artifacts without modelling it as an update to the bug (even if it doesn't actually perform the updates). The bug is intrinsically shared global state, and the more we depend on the model interacting with that directly, the harder it is likely to be to run experiments or use in non-bugzilla contexts (e.g. with the dashboard).

agents/autowebcompat/scripts/install-firefox-nightly.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
```
jgraham: I don't love this being a shell script. It's not really a blocker, but I think pretty soon we're going to want something like mozdownload / mozinstall that knows how to work with multiple platforms.

agents/autowebcompat-repro/compose.yml
```diff
@@ -0,0 +1,32 @@
...
+    expose:
+      - "8765"
+
+  autowebcompat-agent:
+    build:
+      context: ../..
+      dockerfile: agents/autowebcompat/Dockerfile
+      target: agent
+    environment:
+      - RUN_ID
```
jgraham, line 19: Should we prefix these either like `AUTOWEBCOMPAT_` or `BUGBUG_`? We can change the settings class to strip the prefix.

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/broker.py
```diff
@@ -0,0 +1,71 @@
...
+Sidecar container that holds the Bugzilla API key and serves the
+bugzilla MCP tools over HTTP. The agent process (in a sibling container
+in the same Cloud Run Job task) reaches us at `127.0.0.1:<port>/mcp`.
+The agent container itself binds no Bugzilla credentials.
+"""
+
+import logging
+from contextlib import asynccontextmanager
+
+import bugsy
```
jgraham, line 12: I appreciate this is not your fault, but I'm rather sad that bugbot is using bugsy, which is not maintained.

services/hackbot-api/app/schemas.py
suhaibmujahid: I prefer having an agent for each workflow, so each agent will be less complex and easier to improve and expand without complex side effects on other workflows. For redundant components, we could always import from other agents, or even better, moving to a common package to import from.

services/hackbot-api/app/schemas.py
suhaibmujahid: I suggest having a separate agent for each workflow.  One agent could handle bugs, while another could handle text reports.  Each agent would need different tools, and the prompts might require different tweaks.

agents/webcompat-triage/hackbot_agents/webcompat_triage/agent.py
suhaibmujahid: As mentioned in https://github.com/mozilla/bugbug/pull/6170#discussion_r3421486634, having a separate agent for each workflow would be simpler.  Therefore, a workflow that doesn’t use Bugzilla wouldn’t need the Bugzilla MCP in the first place.

jgraham: But in this case the actual workflows would be more or less identical, with very similar prompts and the same outputs, just the source of input data would be slightly different. It feels very heavyweight to say that we need to have a different workflow for every combination of configuration options (even if it makes some sense to separate things out when the agent is fundamentally trying to do different things).

suhaibmujahid: I totally agree with you. I see it as the agent more or less should adapt to your workflow where you want to inject its results, not serving multiple workflows that are fundamentally different in their nature, we do not need all the team's use cases in one agent. We defiantly do not want to create a new agent for every combination of configuration options, it is more about what the type of work that the agent will be doing, not the inputs and configurations. 

Contexts: While we are building this, my goal is to reduce the blueprint when creating new agents as much as possible (we still have a considerable amount of blueprints), but also I do not want to have things to be magically, because that would make things harder, not easier. For that, I wanted to onboard more agents with identical use cases before starting to optimize on that.

agents/autowebcompat-repro/compose.yml
suhaibmujahid: It will not be necessary since this will not be shared with other agents or so. Keeping it shorter would make it cleaner when running it locally.

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/broker.py
suhaibmujahid: @jgraham This was temporary :) 

We have a draft PR to migrate to libmozdata: #6180


That said, even though libmozdata is maintained and widely used across our tools, I think we need a more modern library that offers:

- **Type safety** with proper type hints/annotations (perhaps Pydantic)
- **Async support**
- **A clean query builder API** for constructing advanced queries
- **Sensible retry defaults** for when Bugzilla is having bad time


It might be worth starting a conversation with you, me, and @dklawren to figure out the best path forward.

agents/autowebcompat/scripts/install-firefox-nightly.sh
ksy36: I've dded a `firefox_install.py` script with mozdownload / mozinstall that runs before the agent starts

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/agent.py
ksy36: Sounds good, I've changed it to this format for now (defined in [result.py](https://github.com/mozilla/bugbug/pull/6170/changes#diff-64663b9eb1cd7eb30d24c438360320bb23b74a4ff84640306a6e97dd33a140faR12)):

```
{
  "status": "ok",
  "error": null,
  "findings": {
    "num_turns": 19,
    "total_cost_usd": 0.47287900000000005,
    "result": {
      "reproduced": true,
      "summary": "The reported issue reproduces in Firefox Nightly. After loading https://www.mbusa.com/en/owners/manuals, the page renders the site header at the top and jumps straight to the \"Stay connected to all things Mercedes-Benz\" newsletter section and footer. The main content region \u2014 where the vehicle model catalog (the model picker grid users select to view their manuals) should appear \u2014 is empty. Inspection of the DOM shows the `<main>` element contains a `<css-css-oom-generic>` custom element host whose laid-out height is 0px and which contributes no visible text, so the catalog never becomes visible. A cookie/consent dialog (Usercentrics) overlays the page on first load but dismissing it does not bring the catalog back; the area between header and footer remains empty. This matches the report's expected vs. actual behavior.",
      "steps": "1. Launch Firefox with a clean profile (Firefox 154 Nightly or Release 151 \u2014 both reproduce per the report).\n2. In a new tab, navigate to https://www.mbusa.com/en/owners/manuals .\n3. Wait for the page to finish loading. Observe that the Mercedes-Benz site header (logo and Vehicles / Electric / Shopping / Owners / My Account / Find a Dealer / Search) appears at the top.\n4. If a Usercentrics \"Analytics and Advertising\" consent dialog appears at the bottom of the viewport, click \"Confirm\" to dismiss it (this step is not required to see the bug, but removes the overlay so the empty area is clearly visible).\n5. Scroll down from the top of the page and observe that immediately below the site header the page shows the \"Stay connected to all things Mercedes-Benz\" newsletter band followed by the footer (Vehicles / Shopping Tools / Electric / Owners Info / Discover Mercedes-Benz columns), with no vehicle model catalog rendered in between.\n6. (Optional verification) Open DevTools Console and run `document.querySelector('main').getBoundingClientRect()` \u2014 note the `<main>` element's height is ~80px and `document.querySelector('main').innerText.length` returns 0, confirming the catalog content area is empty.\n7. (Optional comparison) Load the same URL in Chrome with a clean profile and observe that the vehicle model catalog renders correctly in the main area."
    }
  },
  "actions": []
}
```

agents/webcompat-triage/hackbot_agents/webcompat_triage/agent.py
ksy36: After discussion with James, we decided we want to keep both `bug_id` and `bug_data`. I made it so bugzilla MCP is not available if `bug_id` is not passed. I've also removed `mode` and made this a "triage" agent only

agents/webcompat-triage/hackbot_agents/webcompat_triage/prompts/system.md
```diff
@@ -0,0 +1,56 @@
...
+
+If using Bugzilla MCP tools:
+
+- **Always request `whiteboard` and `keywords` explicitly** in `include_fields`. This Bugzilla proxy drops them from `_all` / `_default`.
+- **The history endpoint is not exposed** on this proxy. Do not try to fetch change history — infer it from comments if you need it.
+- **Bulk fetch whenever possible.** `get_bugs` takes a list of IDs and makes one request. Do not call `get_bugs` in a loop with single IDs.
+- **Inaccessible bugs are silently dropped.** `get_bugs` reports them under `inaccessible` — log and skip those.
+- **Search parameters are ANDed.** `search_bugs({{"blocks": 123, "keywords": "sec-low"}})` returns bugs that block 123 _and_ have keyword sec-low.
+
+Use **only** these tools for accessing Bugzilla, nothing else.
```
suhaibmujahid: The "Bugzilla MCP tools" section seems not relevant to the current flow anymore.

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/prompts/system.md
```diff
@@ -0,0 +1,56 @@
...
+- `summary`: a concise account of what you observed.
+- `steps`: the ordered steps you took, as a single numbered list (`1.`, `2.`,
+  `3.`, ... one step per line), written so another agent could reproduce them
+  with no extra context. Each step must be self-contained: whenever you introduce
+  an input or artifact the report did not provide (a file, image, account, or any
+  other test data), state its exact origin — the URL you fetched it from, the
+  command you ran, or how you generated it — not just that you "used" or "saved"
+  it. A reader must be able to obtain the same inputs.
+
+Do not call `submit_result` until the investigation is complete.
```
suhaibmujahid, line 32: nit: These details could be part of the tool description instead of the system prompt.

agents/webcompat-triage/hackbot_agents/webcompat_triage/result.py
```diff
@@ -0,0 +1,66 @@
...
+    **TriageResult.model_json_schema(),
+    "additionalProperties": False,
+}
+
+
+class ResultCollector:
+    """Holds the result submitted by the agent, if any."""
+
+    def __init__(self) -> None:
+        self.result: TriageResult | None = None
```
suhaibmujahid: We should add support to support something like that as part of hackbot-runtime. I filed a follow-up issue.

services/hackbot-api/app/agents.py
```diff
@@ -48,4 +48,13 @@ def model_to_env(inputs: BaseModel) -> dict[str, str]:
...
     ),
+    "webcompat-triage": AgentSpec(
+        name="webcompat-triage",
+        description=(
+            "Reproduce a Firefox web-compatibility issue in headless Firefox "
+            "(from inline report text or a Bugzilla bug id) and return findings."
+        ),
+        job_name="hackbot-agent-webcompat-triage",
+        input_schema=WebcompatTriageInputs,
+    ),
```
suhaibmujahid: Adding a new agent should not require doing pluming to the API service. I filed a follow-up issue.

pyproject.toml
```diff
@@ -128,7 +128,7 @@ include = ["/bugbug", "/scripts", "/VERSION"]
 packages = ["bugbug", "scripts"]
 
 [tool.uv.workspace]
-members = ["http_service", "services/hackbot-api", "agents/bug-fix", "libs/hackbot-runtime", "libs/agent-tools"]
+members = ["http_service", "services/hackbot-api", "agents/bug-fix", "agents/webcompat-triage", "libs/hackbot-runtime", "libs/agent-tools"]
```
suhaibmujahid: Adding a new agent should not require editing the root `pyproject.toml` file. I filed a follow-up issue.

suhaibmujahid (dismissed):
LGTM! Thank you!

I left minor comments. For the comments that I filed a follow-up issue for, I will address them later; they are out of the scope of this PR.

agents/webcompat-triage/hackbot_agents/webcompat_triage/prompts/system.md
```diff
@@ -0,0 +1,56 @@
+# Firefox Web-Compatibility Triage Agent
```
jgraham: Let's rename this agent to something like `autowebcompat-repro` and change the prompt to remove the word "triage" (we don't at this moment want it to try to assess severity or similar) and instead call it a "Web-Compatibility Report Reproduction" agent or something like that.

agents/webcompat-triage/hackbot_agents/webcompat_triage/prompts/system.md
```diff
@@ -0,0 +1,56 @@
...
+## Reporting your result
+
+When you finish the investigation, call the `submit_result` tool exactly once to
+record your result. This is how your result is captured — a prose message is not
+enough. Provide:
+
+- `reproduced`: `true` if the reported issue reproduced in Firefox, otherwise `false`.
+- `summary`: a concise account of what you observed.
+- `steps`: the ordered steps you took, as a single numbered list (`1.`, `2.`,
+  `3.`, ... one step per line), written so another agent could reproduce them
```
jgraham: The next step here will be to have the agent produce a script, then have the harness rerun the problem using only the script as input and verify that it could reproduce the issue. But absolutely not a blocker for landing this.

agents/webcompat-triage/hackbot_agents/webcompat_triage/agent.py
```diff
@@ -0,0 +1,152 @@
...
+    verbose: bool = False,
+    log: Path | None = None,
+) -> WebcompatTriageResult:
+    """Reproduce a web-compat issue and return the agent's findings.
+
+    Returns a :class:`WebcompatTriageResult` on success; raises
+    :class:`AgentError` if the agent ends in an error.
+    """
+    subject = bug_data if bug_data else f"bug {bug_id}"
+    print(f"[webcompat-triage] triaging {subject}", file=sys.stderr)
```
jgraham: Why are we using `print` rather than configuring a logger?

Also in some cases the bug data could be rather large; it's unclear that logging it all makes a lot of sense.

agents/autowebcompat-repro/Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+
+ENV PYTHONUNBUFFERED=1
+ENV PYTHONDONTWRITEBYTECODE=1
+ENV PATH="/opt/venv/bin:$PATH"
+
+FROM base AS agent
+
+# The Firefox DevTools MCP server is an npm package launched via `npx`, so the
+# agent image needs Node.js + npm (the python base ships neither). It also
+# needs the shared libraries Firefox requires to run headless; the Firefox
```
jgraham, line 33: I wonder if we're going to run into bugs that don't reproduce in headless mode. Might be worthwhile in the future to use a normal Firefox window.

jgraham (commented):
I think we should change the name and remove the references to "triage" from the prompt, but otherwise I think we should land and iterate.

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/prompts/bugzilla.md
suhaibmujahid, line 1: I do not think the Bugzilla prompt is needed in our use case here.

jgraham (approved):

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/prompts/bugzilla.md
ksy36: Sounds good, I've removed it :)

suhaibmujahid (approved):

agents/webcompat-triage/hackbot_agents/webcompat_triage/agent.py
ksy36: Added a logger instead of print and logging only first 200 chars of the `bug_data`

ksy36:
@suhaibmujahid I was wondering how can we get the agent deployed? James added the caller functionality and I think we're ready to test the full flow :)

suhaibmujahid:
> I was wondering how can we get the agent deployed? James added the caller functionality and I think we're ready to test the full flow :)

Sure, I will deploy it today.

PR #6196, opened 2026-06-17
Periodically write a /tmp/heartbeat file in the bugbug http service worker

marco-c:
@emaydeck-mozilla does this work? Another alternative would be to write the last heartbeat from the rq worker.


suhaibmujahid (approved):
Why do we need this?

marco-c:
@suhaibmujahid this would be used as a liveness probe to identify if a worker crashed.

PR #6201, opened 2026-06-20
Broaden workspace membership patterns in the root pyproject

### LINKED ISSUE #5653: Use uv workspaces to manage the diffrent services/libs
Related to #5022

This repository contains multiple services and libraries, with more expected as the project grows. We've opted for a monorepo structure to reduce maintenance overhead and minimize interdependency issues. However, managing multiple Python packages in a single repository introduces challenges around dependency conflicts, complex dependency resolution, and keeping package dependencies synchronized. Adopting [uv workspaces](https://docs.astral.sh/uv/concepts/projects/workspaces) will address these issues by providing unified dependency resolution across all packages, simplified handling of internal dependencies, and better tooling for managing multiple related Python projects. This will reduce configuration duplication, ensure consistent dependency versions across services, and simplify the development workflow

suhaibmujahid:
Resolves #5653


.github/dependabot.yml
```diff
@@ -34,44 +34,6 @@ updates:
...
-        update-types:
-          - patch
-    cooldown:
-      semver-major-days: 14
-      semver-minor-days: 7
-      semver-patch-days: 3
-    open-pull-requests-limit: 99
-    allow:
-      - dependency-type: direct
   - package-ecosystem: uv
```
marco-c, line 37: The split was because Dependabot was running out of time when looking for dependency updates. If we do this, we might run into that problem again.

marco-c (approved):

.github/dependabot.yml
suhaibmujahid: Now it is different since they share the same uv.lock file, so it is one update. If we keep it as it is, it won’t work.

PR #6226, opened 2026-06-22
Add Chrome mask test to autowebcompat-repro agent

ksy36:
This PR adds a script that creates a profile and installs Chrome mask extension from AMO and passes the profile to devtools MCP.


agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/prompts/system.md
```diff
@@ -1,20 +1,41 @@
...
-Reproduce the reported issue. Do not attempt to debug or perform root cause analysis.
+Reproduce the reported issue, then test whether Chrome Mask fixes it. Do not
+attempt to debug or perform root cause analysis.
 
 ### Procedure
 
 1. Identify the affected URL and the described broken behavior.
-2. Navigate to the URL using the Firefox DevTools MCP and try to reproduce the issue.
-3. Submit your findings via `submit_result` (see "Reporting your result").
+2. Baseline: Navigate to the URL with the Firefox DevTools MCP and
```
jgraham, line 23: I think this is OK for now, but a lot of this is a set of deterministic steps. We don't need an LLM to execute things that we can represent directly in code; it's more token-heavy and less deterministic. I'd rather put this control flow in the harness and just use the LLM for the parts that aren't just running through an algorithm.

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/setup_profile.py
```diff
@@ -0,0 +1,166 @@
...
+# extensions folder (see firefox-devtools-mcp src/firefox/profile.ts,
+# resolveProfilePath). To bypass that we create the subdir ourselves; when it
+# already exists the MCP just uses it as-is, extensions included.
+MCP_PROFILE_DIR_NAME = "firefox_devtools_mcp_profile"
+
+
+def resolve_xpi_url(slug: str) -> tuple[str, str]:
+    """Return (download_url, version) for the latest signed xpi of an AMO addon."""
+    url = AMO_API_TMPL.format(slug=slug)
+    req = urllib.request.Request(url, headers={"User-Agent": "webcompat-setup"})
```
jgraham: Let's not use `urllib` directly or copy the same setup code across multiple functions; that's not going to help with maintainance later on. I'm pretty sure that requests is already a transitive dependency, so let's add that to pyproject.toml and use it (with a wrapper function that configures the default headers &c.).

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/setup_profile.py
```diff
@@ -0,0 +1,166 @@
...
+    """
+    parent = Path(tempfile.mkdtemp(prefix="ff-profile-"))
+    profile_dir = parent / MCP_PROFILE_DIR_NAME
+    profile_dir.mkdir(parents=True, exist_ok=True)
+
+    installed = [
+        install_amo_extension(profile_dir, parent, slug) for slug in extensions
+    ]
+
+    if not installed:
```
jgraham: This can only happen if  `extensions` is empty, right? In which case we should return even earlier to make it clear that none of the steps other than creating the parent directory are meaningful.

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/setup_profile.py
```diff
@@ -0,0 +1,166 @@
...
+
+    ``extensions`` is a list of AMO addon slugs (e.g. ``["chrome-mask"]``); each
+    is downloaded and installed. With no extensions a plain profile is built and
+    no warm launch happens. The returned path is meant to be passed as the
+    devtools MCP's ``--profile-path`` (``build_devtools_server(profile_path=...)``).
+
+    Raises ``RuntimeError`` if an extension does not end up registered and
+    enabled in the profile.
+    """
+    parent = Path(tempfile.mkdtemp(prefix="ff-profile-"))
```
jgraham, line 201: It looks like we end up leaking this temporary directory. Is that expected (e.g. because we end up destroying the entire container) or should there be some teardown code that runs at the end of each agent run?

jgraham (changes requested):

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/__main__.py
```diff
@@ -23,6 +24,9 @@ async def main(ctx: HackbotContext) -> AutowebcompatReproResult:
     # current build; drive the binary the install reports back.
     firefox_path = str(install_firefox_nightly())
 
+    # Build a profile with Chrome Mask preinstalled
+    chrome_mask_profile = setup_profile(firefox_path, extensions=["chrome-mask"])
+
```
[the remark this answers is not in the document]
ksy36: I think we want to stop the run and return an error if it's unable to install the extension as this pilot is mainly about reproduction with and without Chrome mask.

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/setup_profile.py
```diff
@@ -0,0 +1,166 @@
...
+
+    print("warm-launching Firefox to register the extensions", file=sys.stderr)
+    warm_launch(firefox_path, profile_dir)
+    time.sleep(1)
+
+    for ext_id in installed:
+        if not verify_registered(profile_dir, ext_id):
+            raise RuntimeError(
+                f"{ext_id} did not register and enable in {profile_dir}/extensions.json"
+            )
```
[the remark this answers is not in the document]
ksy36: added

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/setup_profile.py
ksy36: Added that, thanks!

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/setup_profile.py
ksy36: I've added a cleanup on error and after a successful run. But yes I assumed a container is destroyed after each run

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/prompts/system.md
ksy36: Yeah, another option could be to have one profile without Chrome mask and another with Chrome mask but enabled globally. Or maybe it's possible to enable it for a given domain with a script that the agent will run. I can try a few options after we launch this first version.

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/__main__.py
```diff
@@ -23,20 +26,28 @@ async def main(ctx: HackbotContext) -> AutowebcompatReproResult:
...
+            max_turns=inputs.max_turns,
+            effort=inputs.effort,
+            firefox_path=firefox_path,
+            chrome_mask_profile=chrome_mask_profile,
+            log=ctx.log_path,
+            verbose=True,
+        )
+    finally:
+        if chrome_mask_profile is not None:
+            shutil.rmtree(chrome_mask_profile, ignore_errors=True)
```
suhaibmujahid: You do not need to clean the profile. We always use a fresh container for each task.

suhaibmujahid (dismissed):

jgraham (approved):

ksy36:
@suhaibmujahid could you merge / enable automerge for this as well, please?

suhaibmujahid:
The linter errors seem unrelated.

PR #6228, opened 2026-06-23
Fix a few errors reported on sentry

padenot:
This patch series is composed of independent commits, only related by the fact that each of those fixes one or more issues surfaced by sentry. They should be reviewed independently.

The goal is to fix some sentry errors, and prequalify some of them (usually not actionable, e.g. a transient failure of another service) to something that doesn't create a sentry entry.

### DIFF
--- a/bugbug/tools/code_review/database.py
+++ b/bugbug/tools/code_review/database.py
@@ -135,7 +135,7 @@ class EvaluationAction(enum.Enum):
 
 @dataclass
 class SuggestionFeedback:
-    id: int
+    id: int | str
     comment: str
     file_path: str
     action: Literal["APPROVE", "REJECT", "IGNORE"]
--- a/bugbug/tools/code_review/langchain_tools.py
+++ b/bugbug/tools/code_review/langchain_tools.py
@@ -203,7 +203,7 @@ async def search_text(
     except SearchfoxNetworkError as e:
         return _tool_error(f"search failed: {e}")
     except SearchfoxRequestError as e:
-        logger.error(
+        logger.warning(
             "Bad request searching for %r (path=%r, langs=%r): %s",
             query,
             path_filter,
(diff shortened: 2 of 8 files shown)


marco-c:
What about keeping transient errors in Sentry, but ignoring them there? So if we see a huge uptick we could investigate

padenot:
Update w/ the 50 -> 100 turn change dropped, and marco's comment addressed

marco-c (dismissed):
LGTM, but there's still a linting error

marco-c (approved):

services/reviewhelper-api/app/routers/request.py
```diff
@@ -80,7 +81,19 @@ async def create_or_get_review_request(
...
+        await db.rollback()
+        existing_request = await db.scalar(stmt)
+        if existing_request is None:
+            raise
+        return JSONResponse(
+            {
+                "status": existing_request.status.value,
+                "message": _build_response_message(existing_request),
+            }
+        )
```
suhaibmujahid, line 96: Why are we doing this? We already have a mechanism to handle this and do the rollback on error in `get_db()`.

bugbug/tools/code_review/langchain_tools.py
```diff
@@ -233,7 +233,7 @@ async def get_field_layout(
     except SearchfoxNetworkError as e:
         return _tool_error(f"field layout fetch failed: {e}")
     except SearchfoxRequestError as e:
-        logger.error("Bad request fetching field layout for %r: %s", class_name, e)
+        logger.warning("Bad request fetching field layout for %r: %s", class_name, e)
```
suhaibmujahid, line 236: Why are we downgrading the log level? We can ignore it in Sentry until it spikes. Sentry supports that.

bugbug/tools/code_review/database.py
```diff
@@ -135,7 +135,7 @@ class EvaluationAction(enum.Enum):
 
 @dataclass
 class SuggestionFeedback:
-    id: int
+    id: int | str
```
suhaibmujahid, line 138: Do we have an error regarding this?

PR #6229, opened 2026-06-23
Increase recursion limit to 100 and stop retrying on GraphRecursionError

suhaibmujahid:
Adds RecursionLimitError to core exceptions, raised by the agent when GraphRecursionError is hit, and caught in the API as ReviewProcessingError to prevent retries.


marco-c:
This was done in https://github.com/mozilla/bugbug/pull/6228 too, do you want to land independently or we just land that one instead and close this?

padenot:
I'll drop mine from the stack, this should land separately asap.

marco-c (approved):

PR #6230, opened 2026-06-23
code_review: prompt suggests splitting large/multi-concern patches

sylvestre:


marco-c:
To avoid affecting the results of the review from the perspective of finding issues, etc., I would add this as a separate prompt that is run independently.

marco-c:
Like, before the main review prompt, or after the main review prompt.

sylvestre:
@suhaibmujahid @marco-c done!

suhaibmujahid (approved):
LGTM! Than thank you!

PR #6129, opened 2026-06-04
Make two tools non-strict to sidestep Anthropic's schema size restriction

padenot:
Fixes #6109
 
https://docs.langchain.com/oss/python/integrations/chat/anthropic#strict-tool-use
https://platform.claude.com/docs/en/agents-and-tools/tool-use/strict-tool-use

PR #6234, opened 2026-06-25
code_review: fix schema compilation error (ToolStrategy + remove _make_non_strict)

padenot:
## Summary

- Switches `response_format` from `ProviderStrategy` to `ToolStrategy` in `CodeReviewTool`
- Removes the `_make_non_strict` workaround from `SEARCHFOX_TOOLS`

## Background

`ProviderStrategy(AgentResponse)` causes `langchain_anthropic` to send the `AgentResponse` JSON schema as `output_config.format`. When combined with thinking enabled and 12 tool schemas (11 SEARCHFOX + `write_todos` from `TodoListMiddleware`), this exceeds Anthropic's schema compilation budget:

```
BadRequestError: Schema is too complex for compilation.
Try reducing the number of tools or simplifying tool schemas.
```

Sentry issue: REVIEWHELPER-API-Z — 400+ occurrences since the tools were expanded in PR #6051.

`ToolStrategy` sends `AgentResponse` as a regular tool call instead, removing `output_config` from the request. Thinking stays enabled. The compilation limit is no longer hit regardless of tool count.

## Why _make_non_strict is also removed

PR #6129 made `expand_context` and `find_definition` non-strict to stay under a separate 16-anyOf-in-strict-tools limit. That limit only applies to strict tools; `ToolStrategy` doesn't change it. But the non-strict flag only affects validation behaviour — Anthropic still compiles non-strict tool schemas — so it never actually fixed the compilation error we were seeing. Now that the root cause is addressed there is no reason to keep the workaround.

## Tradeoff

`ToolStrategy` relies on the model following the `AgentResponse` schema via tool-call conventions rather than having it enforced server-side by `output_config`. In practice Claude follows tool schemas reliably; a malformed response surfaces as a Pydantic parsing error in the app rather than being caught at the API boundary. Application-level validation is already present.

## Test plan

- `_schema_test.py` (repo root) demonstrates the four cases against the live API
- `pytest tests/test_code_review.py` — 61 tests, all passing

### DIFF
--- a/_schema_test.py
+++ b/_schema_test.py
@@ -1,0 +1,164 @@
+"""
+Demonstrates the schema compilation limit hit by the code review agent.
+
+Sentry issue: REVIEWHELPER-API-Z
+GitHub issue: https://github.com/mozilla/bugbug/issues/6109
+
+Three factors combine to exceed Anthropic's compilation limit:
+  1. thinking (adaptive or enabled)
+  2. output_config — sent by langchain_anthropic when ProviderStrategy is used;
+     carries the full AgentResponse/GeneratedReviewComment JSON schema
+  3. 12 tool schemas — 11 SEARCHFOX tools + write_todos injected by TodoListMiddleware
+
+Any two of these three alone is fine. All three together triggers:
+  "Schema is too complex for compilation. Try reducing the number of tools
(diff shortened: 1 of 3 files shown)


padenot:
Wrong branch

PR #6255, opened 2026-06-29
Pre-fill agent run form from URL query parameters

msujaws:
Deep-link support for the agent run page: TriggerForm seeds each field from snake_case query params (agent, bug_id, bug_data, model, max_turns, effort) via useSearchParams, matching the /api/runs inputs contract so a link mirrors what the run receives. Invalid/missing params fall back to the existing empty defaults; an invalid agent falls back to bug-fix.

Wrap TriggerForm in a Suspense boundary as required by useSearchParams in Next 15.

### DIFF
--- a/services/hackbot-ui/app/page.tsx
+++ b/services/hackbot-ui/app/page.tsx
@@ -1,3 +1,5 @@
+import { Suspense } from "react";
+
 import { RecentRuns } from "@/components/RecentRuns";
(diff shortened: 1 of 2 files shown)


suhaibmujahid (dismissed):
LGTM! We have some new fields that were added. Could you please resolve the merge conflict ad support them as well?

suhaibmujahid (dismissed):

suhaibmujahid (approved):

PR #6268, opened 2026-06-30
Add bug_map as a cached property for LandingsRiskReportGenerator

### LINKED ISSUE #5053: Add bug_map as a cached property for LandingsRiskReportGenerator
> As a follow-up, what do you think about adding bug_map as a cached property (i.e., `@cached_property`)? In that way, we do not need to pass it with the function calls.

_Originally posted by @suhaibmujahid in https://github.com/mozilla/bugbug/pull/5052#discussion_r2111405600_

Karthikvenugopal:
## Summary

Implements the follow-up from #5053: `bug_map` is now a `@cached_property` on
`LandingsRiskReportGenerator` instead of a local dict in `go()` passed as an
explicit argument to every method that needs it.

## Changes

- Add `from functools import cached_property`.
- Add a `bug_map` cached property reproducing the existing filtering (keep bugs
  we are interested in, bugs blocked by other bugs, and bugs that caused
  regressions).
- Promote the locals the property depends on (`bugs`, `test_info_bugs`,
  `meta_bugs`) to instance attributes set in `go()` before `bug_map` is first
  accessed.
- Drop the `bug_map` parameter from `get_blocking_of`,
  `generate_landings_by_date`, `generate_component_connections`, and
  `generate_component_test_stats`, and update their bodies and the call sites in
  `go()` to use `self.bug_map`.

## Why the dependencies had to move too

`bug_map` is filtered against `bugs_set`, which is derived from the `bugs`,
`test_info_bugs`, and `meta_bugs` locals computed earlier in `go()`. A
`cached_property` only receives `self`, so those inputs had to live on the
instance for the property to rebuild the same set. They are assigned in `go()`
before any consumer touches `self.bug_map`.

## Notes

- No method mutates `bug_map`, so sharing a single cached dict across consumers
  is safe.
- The generator is instantiated once and `go()` runs once per process, so
  keeping these values on the instance does not change the effective memory
  lifetime.

## Testing

- `pre-commit run --all-files` passes (ruff, ruff-format, mypy, codespell, etc.).
- No existing test exercises this script.

Fixes #5053

### DIFF
--- a/scripts/generate_landings_risk_report.py
+++ b/scripts/generate_landings_risk_report.py
@@ -18,6 +18,7 @@ import textwrap
 import traceback
 import urllib.parse
 from datetime import datetime, timedelta, timezone
+from functools import cached_property
 from typing import Any, Set, cast
 
 import bs4
@@ -359,14 +360,13 @@ class LandingsRiskReportGenerator(object):
 
     def get_blocking_of(
         self,
-        bug_map: dict[int, bugzilla.BugDict],
         bug_ids: list[int],
         meta_only: bool = False,
     ) -> dict[int, list[int]]:
         return {
-            bug_id: bugzilla.find_blocking(bug_map, bug_map[bug_id])
+            bug_id: bugzilla.find_blocking(self.bug_map, self.bug_map[bug_id])
             for bug_id in bug_ids
-            if not meta_only or "meta" in bug_map[bug_id]["keywords"]
+            if not meta_only or "meta" in self.bug_map[bug_id]["keywords"]
         }
 
     def get_meta_bugs(self, days: int) -> list[int]:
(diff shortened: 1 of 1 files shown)


Karthikvenugopal:
Hi @suhaibmujahid, this PR implements the `@cached_property` follow-up you suggested in #5053. Since `bug_map` is filtered against `bugs_set` (derived from the `bugs`, `test_info_bugs`, and `meta_bugs` locals in `go()`), I promoted those three to instance attributes so the property can build it from `self` alone; the four methods that previously took `bug_map` as a parameter now read `self.bug_map`. `pre-commit run --all-files` passes. Would appreciate your review when you get a chance. Thanks!

Issue #6233, opened 2026-06-25
Failed to start execution: length cannot exceed 32768


suhaibmujahid:
This happened when passing long input to the `autowebcompat-repro` agent, the input included encoded attachments. 

Run ID: `6f2278a7-e260-4e1c-9c3b-51525acab1cb`

```
Failed to start execution: 400 Violation in RunJobRequest.overrides.container_overrides[0].env[5].value: length cannot exceed 32768. [field_violations {
  field: "Violation in RunJobRequest.overrides.container_overrides[0].env[5].value"
  description: "length cannot exceed 32768."
}
]
```

jgraham:
This is because environment variables on GCP are limited to 32kB. We should support some other mechanism to pass data into agents other than just environment variables.

PR #6275, opened 2026-07-02
Pass agent inputs through a file rather than environment variables

jgraham:
To pass inputs ot agents in production, upload a file to a GCS bucket, and pass a URL to that file into the agent environment.

This enables having inputs which don't fit inside the 32kB limits of enironment variables for GCS.

For local runs using docker, data is still pased via environment variables.


jgraham:
This is a possible approach for fixing https://github.com/mozilla/bugbug/issues/6233

suhaibmujahid (commented):
Thanks for the PR! I can see the need for supporting a larger input payload.

That said, passing the environment variables does simplify the wiring in a lot of use cases, so I'd love to find a way to support both. One option: when we have a large input, we include an `attachments` field that lists the items and uploads them to a separate bucket, with a link the agent can use to read them from there. We'd also give the agent read-only permission for that bucket.

Let's discuss and decide how we want to move forward on this.

PR #6282, opened 2026-07-03
Support multiple tasks in the autowebcompat agent

jgraham:


jgraham:
CC @ksy36

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/agent.py
```diff
@@ -8,151 +8,475 @@
...
+                    default_config,
+                    tracker,
+                    firefox_browser.esr,
+                    setup_profile(firefox_browser.esr),
+                    input_data,
+                    bugzilla_mcp_server,
+                )
+                reproduced_channel = "esr"
+        if repro_task is not None:
+            repro_result = await repro_task.run()
```
ksy36: I think this might rerun it in Nightly again, even if the first Nightly run has just failed?

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/agent.py
```diff
@@ -8,151 +8,475 @@
...
     RESULT_SERVER_NAME,
     SUBMIT_RESULT_TOOL,
+    ChromeMaskResult,
     ReproductionResult,
     ResultCollector,
+    ResultT,
+    TestPlanResult,
     build_result_server,
 )
+from .setup_profile import setup_profile
```
ksy36, line 42: looks like the file is deleted, but function is still being used

PR #6287, opened 2026-07-07
Refactor autowebcompat agent into a series of tasks

jgraham:
Currently the agent does two things: checks in nightly and checks with Chrome Mask enabled. These are done in a single Firefox session and a single invocation of the LLM. However this approach won't scale; if we want to check on multiple browsers, multiple Firefox versions, etc. putting all the logic into the LLM prompt will be fragile and hard to debug.

Instead create a base Task class that knows how to invoke the LLM and get results out. Then create a specific subclass for each kind of operation we're interested in, and write some business logic to deterministically wire them together according to our needs.

### DIFF
--- a/agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/agent.py
+++ b/agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/agent.py
@@ -10,6 +10,7 @@ from __future__ import annotations
 import logging
 from abc import ABC, abstractmethod
 from dataclasses import dataclass, field
+from datetime import datetime
 from pathlib import Path
 from typing import Any, Generic, Literal
 
@@ -82,19 +83,58 @@ class TaskConfig:
 
 
 @dataclass
-class RunTracker:
-    num_turns: int = 0
-    total_cost_usd: float = 0.0
-    tasks: list[str] = field(default_factory=lambda: [])
+class TaskRun:
+    name: str
+    start_time: datetime
+    end_time: datetime
+    num_turns: int
+    total_cost_usd: float | None
 
-    def update(self, result_msg: ResultMessage) -> None:
-        self.num_turns += result_msg.num_turns
-        if result_msg.total_cost_usd is not None:
-            self.total_cost_usd += result_msg.total_cost_usd
+
+class RunTracker:
+    def __init__(self):
+        self.task_runs: list[TaskRun] = field(default_factory=lambda: [])
+        self.current_task: tuple[str, datetime] | None = None
+
+    @property
+    def num_turns(self) -> int:
+        return sum(item.num_turns for item in self.task_runs)
+
+    @property
+    def total_cost_usd(self) -> float:
+        return sum(
+            item.total_cost_usd
+            for item in self.task_runs
+            if item.total_cost_usd is not None
+        )
+
+    def start_task(self, name: str) -> None:
+        self.current_task = name, datetime.now()
+
+    def end_task(self, name: str, result_msg: ResultMessage) -> None:
+        if self.current_task is None:
+            logger.warning("Got end_task without start_task")
+            return
+        current_name, start_time = self.current_task
+        if current_name != name:
+            logger.warning(
+                "Got end_task with name %s but current_task was %s", name, current_name
+            )
+            self.current_task = None
+            return
+        self.task_runs.append(
+            TaskRun(
+                name=name,
+                start_time=start_time,
+                end_time=datetime.now(),
+                num_turns=result_msg.num_turns,
+                total_cost_usd=result_msg.total_cost_usd,
+            )
+        )
 
 
 class Task(ABC, Generic[ResultT]):
-    name: str
+    name: str = "unnamed-task"
     result_server_name: str = RESULT_SERVER_NAME
     submit_result_tool: str = SUBMIT_RESULT_TOOL
     result_cls: type[ResultT]
@@ -144,7 +184,7 @@ class Task(ABC, Generic[ResultT]):
         )
 
     async def run(self) -> ResultT:
-        self.run_tracker.tasks.append(self.name)
+        self.run_tracker.start_task(self.name)
         subject = self.subject()
         preview = str(subject)
         if len(preview) > 200:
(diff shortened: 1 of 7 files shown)


jgraham:
CC @ksy36

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/agent.py
```diff
@@ -8,151 +8,376 @@
...
+            logger.warning("Got end_task without start_task")
+            return
+        current_name, start_time = self.current_task
+        if current_name != name:
+            logger.warning(
+                "Got end_task with name %s but current_task was %s", name, current_name
+            )
+            self.current_task = None
+            return
+        self.task_runs.append(
```
ksy36, line 131: This is failing for me with:
```
autowebcompat-repro-agent-1  |     self.task_runs.append(
autowebcompat-repro-agent-1  |     ^^^^^^^^^^^^^^^^^^^^^
autowebcompat-repro-agent-1  | AttributeError: 'Field' object has no attribute 'append'
```
Changing `self.task_runs` to `self.task_runs: list[TaskRun] = []` makes it work

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/agent.py
```diff
@@ -8,151 +8,376 @@
...
+        run_tracker: RunTracker,
+        firefox_path: Path,
+        profile_path: Path,
+        input_data: AutoWebcompatInput,
+        bugzilla_mcp_server: McpServerConfig,
+    ):
+        super().__init__(task_config, run_tracker)
+        self.input_data = input_data
+        self.add_mcp_server(
+            "firefox_devtools",
```
ksy36: It used to be `firefox-devtools`, so all `DEVTOOLS_TOOLS` in the config file have hyphens in the name, i.e. `mcp__firefox-devtools__list_pages`. They should probably match for consistency if we're changing to underscore. It doesn't seem to break anything though because we're using `bypassPermissions` permission_mode and all tools are allowed

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/result.py
```diff
@@ -32,6 +65,17 @@ class ReproductionResult(BaseModel):
...
         ),
     )
+    screenshot: str | None = Field(
+        description=(
+            """A base64 encoded screenshot showing the issue. This must only be
+            set for issues where the breakage is visual in nature
+            i.e. incorrect site layout rather than broken interaction.
+            Otherwise it must be null"""
+        ),
+    )
```
ksy36, line 79: I've added the artifact upload in this draft PR https://github.com/mozilla/bugbug/pull/6289

libs/hackbot-runtime/hackbot_runtime/__init__.py
```diff
@@ -10,6 +10,7 @@
 __all__ = [
     "ActionsRecorder",
     "AgentError",
+    "BaseAgentInputs",
```
suhaibmujahid: This seems out of the scope of this PR.

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/agent.py
```diff
@@ -8,151 +8,411 @@
...
+
+    def system_prompt(self) -> str:
         return (
-            "Here is the web-compatibility report to work on:\n\n"
-            f"{bug_data}\n\n"
-            "Follow your task procedure."
+            super()
+            .system_prompt()
+            .format(
+                task_details=f"""
```
suhaibmujahid, line 274: For consistency and easy discovery, I suggest adding all prompts in the `/prompts` directory.

suhaibmujahid (commented):
Have we evaluated the results on a single task and a multi-task?  Did we see improvements when we tackled them separately?

libs/hackbot-runtime/hackbot_runtime/__init__.py
jgraham: Apologies, I pushed the wrong commits to the PR. This has been removed.

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/agent.py
jgraham: I'm not very convinced that this will be better. In this case they're not really prompts, but fragments of prompts that often need additional details from the code (e.g. input data). We could insist that all prompt fragments are stored as jinja templates or something, but I don't see what the concrete benefits would be,

jgraham:
> Have we evaluated the results on a single task and a multi-task? Did we see improvements when we tackled them separately?

No, but for later cases we want to be able to deterministically control what the agent does, rather than just writing text and hoping that it follows the instructions; see e.g. https://github.com/mozilla/bugbug/commit/a4451a8cf0435acd33ec0a0a0d3c4da1c5596b4e That also allows us to only provide the agent with the permissions it needs and the resources it needs at each stage (e.g. we only enable access to the browser chrome when strictly needed to enable chrome mask, we only download the specific version of Firefox when we actually want to test that one). We could try to put some of that logic into MCPs and put the model in control of more, but it makes more sense to me to just make the deterministic steps normal code and call out to the model only when required. This will also allow us to use different efforts or different models for different steps in the future (e.g. Chrome Mask probably doesn't need to be high effort)

ksy36 (approved):

libs/hackbot-runtime/hackbot_runtime/__init__.py
suhaibmujahid: It seems to be still in the PR.

jgraham: Oh interesting, I guess `__all__` containing a symbol that doesn't exist doesn't break anything. Makes sense I suppose. Anyway it's removed now.

suhaibmujahid (approved):

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/agent.py
suhaibmujahid: It is not a blocker, but when exploring the prompts for an agent, you could discover them easily without even going through the code.

suhaibmujahid:
> but it makes more sense to me to just make the deterministic steps normal code and call out to the model only when required

What makes more sense is not necessarily what performs better for the agent. Providing the instruction in the initial prompt is not the only way; there are many ways to go with that. We could land this for now and iterate later.

PR #6288, opened 2026-07-07
Agent tracing in Weave

### LINKED ISSUE #6266: Observability for agents
There should be a way to track agentic steps in production for easier inspection and debugging. 

We use Weights and Biases Weave for evaluation (see example of [evaluation](https://wandb.ai/moz-bugbug/bugbug-build-repair-eval/weave/evaluations?view=evaluations_default&peekPath=%2Fmoz-bugbug%2Fbugbug-build-repair-eval%2Fcalls%2F019effc4-aa0b-7974-aa70-619642ddf56c%3FhideTraceTree%3D1%26traceRootStartedAt%3D2026-06-25T17%253A12%253A19.211125Z) and [traces](https://wandb.ai/moz-bugbug/bugbug-build-repair-eval/weave/agents)). We should explore how suitable it is for production workflows.

Also, we need dashboards that show API request rates, processed queue messages, error counts, AI costs, and other useful metrics for the Hackbot services. Ideally, those should be customizable (think Grafana way). Maybe GCP dashboards can be sufficient.

evgenyrp:
Enable Weave tracing for all agents. The agent name will be the same as its folder name.

See test example traces: https://wandb.ai/moz-bugbug/hackbot-test/weave/agents?filters[agent_name]=build-repair

Auth through WIF similar to Anthropic (need to test that on deployment).

fixes #6266


agents/README.md
```diff
@@ -100,6 +100,40 @@ the repo's root `docker-compose.yml`, so running it is three steps:
...
+`compose.yml`:
+
+```yaml
+environment:
+  - WANDB_API_KEY=${WANDB_API_KEY:-}
+```
+
+**In deployment**, the agent container holds no long-lived key: it authenticates
+via W&B [Identity Federation](https://docs.wandb.ai/platform/hosting/iam/identity_federation).
+
```
suhaibmujahid: Let us drop this out to keep it short. We could have advanced docs for Hackbot in `/docs`, we could do this later since it is out of the scope of this PR.

libs/hackbot-runtime/hackbot_runtime/wandb_wif.py
suhaibmujahid, line 1: Let us file a follow-up issue to reduce duplication between `wandb_wif.py` and `anthropic_wif.py`.

libs/hackbot-runtime/pyproject.toml
```diff
@@ -11,7 +11,7 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
-claude-sdk = ["claude-agent-sdk>=0.1.30", "agent-tools[claude-sdk]"]
+claude-sdk = ["claude-agent-sdk>=0.1.30", "agent-tools[claude-sdk]", "weave>=0.53.1"]
```
suhaibmujahid: I do not think this needs to be specific to the Claude SDK.

agents/README.md
evgenyrp: I'd rather add docs together with functionality, especially since we already have them here. I can move it to another place.

libs/hackbot-runtime/hackbot_runtime/wandb_wif.py
evgenyrp: I'll add the issue when we land it

libs/hackbot-runtime/pyproject.toml
evgenyrp: good catch

agents/README.md
evgenyrp: moved to /docs, we can refactor later if we don't like it

libs/hackbot-runtime/hackbot_runtime/runtime.py
```diff
@@ -9,7 +9,7 @@
 
 from pydantic import ValidationError
 
-from hackbot_runtime import anthropic_wif
+from hackbot_runtime import anthropic_wif, tracing, wandb_wif
```
suhaibmujahid:
```suggestion
from hackbot_runtime import anthropic_wif, wandb_wif
from hackbot_runtime.tracing import trace_agent
```

suhaibmujahid (dismissed):
LGTM! Thank you!

agents/README.md
suhaibmujahid: Looks good!

libs/hackbot-runtime/hackbot_runtime/wandb_wif.py
evgenyrp: https://github.com/mozilla/bugbug/issues/6317

evgenyrp:
@suhaibmujahid maybe let's try to deploy it first before landing, as discussed. Let me know if something else is missing.

suhaibmujahid (approved):

PR #6289, opened 2026-07-08
Autowebcompat screenshot artifact

ksy36:
This depends on https://github.com/mozilla/bugbug/pull/6287

It saves a screenshot to a specified location using Firefox devtools mcp and uploads it as an artifact using exisitng `publish_file` method.


ksy36:
This was merged to the base branch, so no need for a separate PR

PR #6292, opened 2026-07-08
Hackbot pulse listener with email notifications

### LINKED ISSUE #6261: Listener for Taskcluster build failures
The listener should subscribe to the Taskcluster build failure [exchange](https://firefox-ci-tc.services.mozilla.com/pulse-messages?bindings%5B0%5D%5Bexchange%5D=exchange%2Ftaskcluster-queue%2Fv1%2Ftask-failed&bindings%5B0%5D%5Bpattern%5D=%23), filter the desired messages, and trigger the Hackbot API.

It should live outside the Hackbot API as a separate service, as it has a different lifecycle and scalability.

Initially, it will trigger the build repair agent, but later on it can also handle test failures, intermittent failures, etc.

Notification about the results will not be a part of this service. It will live either in the API itself or in a separate service, as discussed with @suhaibmujahid. If we wanted to include the notifications here, the service would need to poll the API to wait for the job completion (as it was implemented in a draft prototype https://github.com/mozilla/bugbug/pull/6292).

evgenyrp:
Here's what I deployed in my personal GCP project to start testing it and getting feedback. It's temporary until we land a proper notification system and deduplication.

fixes #6261


ui/changes/snowpack.config.js
suhaibmujahid, line 1: The changes to the is file are out of scope; rebasing would allow dropping the changes without failing the CI.

evgenyrp: Rebased and removed the change, let's see if CI will be green

suhaibmujahid (approved):

PR #6297, opened 2026-07-09
Add CODEOWNERS with team mappings

suhaibmujahid:
Introduce a new .github/CODEOWNERS file to assign default and path-specific code owners. Sets @mozilla/bugbug-leads as the default owner and adds specific owners for Hackbot agents.


jgraham:
Thanks. lgtm but since I wouldn't be able to approve this if it was merged I won't ;)

marco-c (approved):

PR #6299, opened 2026-07-09
Apply agent-recorded actions to Bugzilla/Phabricator after a Hackbot run

suhaibmujahid:


libs/hackbot-runtime/hackbot_runtime/actions/handlers/phabricator_handler.py
```diff
@@ -0,0 +1,249 @@
...
+def _api_key() -> str:
+    token = os.environ.get("PHABRICATOR_API_KEY", "")
+    if not token:
+        raise RuntimeError("PHABRICATOR_API_KEY is not configured")
+    return token
+
+
+def _conduit_request(method: str, **payload: Any) -> dict:
+    payload["__conduit__"] = {"token": _api_key()}
+    response = requests.post(
```
marco-c, line 48: might want an user agent to make sure phab can identify the originator easily

libs/hackbot-runtime/hackbot_runtime/actions/phabricator.py
```diff
@@ -0,0 +1,107 @@
...
+            default=None,
+            description=(
+                "An existing Phabricator revision to attach a new diff to. "
+                "Omit to create a brand-new revision instead — never inferred "
+                "automatically, so pass this explicitly whenever you intend "
+                "to update rather than create."
+            ),
+        ),
+    ] = None,
+    reviewers: Annotated[
```
marco-c, line 42: will the agent fill this? we might want to just leave it empty and let the harald rules decide the reviewer

marco-c (approved):

libs/hackbot-runtime/hackbot_runtime/actions/handlers/phabricator_handler.py
suhaibmujahid: Should be address as part of #6306

libs/hackbot-runtime/hackbot_runtime/actions/phabricator.py
suhaibmujahid: Should be address as part of #6308

PR #6301, opened 2026-07-10
Run build repair only on regressions

### LINKED ISSUE #6296: Most build failures are unrelated to the push revision
Marco:
> It might also be because we'd only need to analyze the failure on the actual push where the failure started
so you'd need to change the pulse listener to not only deduplicate, but also look at when the failure first started
load the revision, check the parent revision, check if the build was already failing in the parent revision
and loop until you find the first revision where the build was not failing

We should investigate how to allow the build repair agent to do that. For now, I'm disabling notifications if there's no patch.

evgenyrp:
- walk the tree of pushes to determine if the build failure is new
- if a build for a parent is still in progress, poll for up to an hour (blocking the consumer thread is fine, I do not want to ack the message without processing)
- do not trigger the agent if the failure is old
- add queue name to test locally without breaking prod

fixes #6296


evgenyrp:
I redeployed it again with the latest changes, but overall, it seems to be working as we get fewer build repair notifications now. 

It's a bit hard to debug, and I feel like I lack some knowledge about Firefox CI. @Archaeopteryx does it make sense to you?

services/hackbot-pulse-listener/app/regression.py
```diff
@@ -0,0 +1,164 @@
...
+    real regression. Retriggers are collapsed: any green run means 'passed'; only
+    a genuine build failure counts as 'failed'.
+    """
+    label_tasks = [t for t in push.tasks if t.label == label]
+    if label_tasks:
+        results = [t.result for t in label_tasks]
+        if any(r in _PASSED_RESULTS for r in results):
+            return "passed"
+        if any(r in _FAILED_RESULTS for r in results):
+            return "failed"
```
marco-c: IIRC there's a `failed` attribute that you can use instead of reimplementing this

evgenyrp: I reused that, but our definition is a bit different. We don't treat "Exception" as failed because it's likely a Taskcluster infra issue

services/hackbot-pulse-listener/app/consumer.py
```diff
@@ -97,24 +120,38 @@ def process(body: dict, executor: Executor) -> str | None:
...
             message.ack()
 
     return on_message
 
 
 def _build_queues(user: str) -> list[Queue]:
+    # Both local and prod authenticate as the same pulse user, so the queue name
+    # must also vary by environment; otherwise both consumers bind to the same
+    # durable queue and steal each other's messages. Production keeps the plain
+    # name for continuity.
```
marco-c:
```suggestion
    # durable queue and steal each other's messages.
```

services/hackbot-pulse-listener/app/regression.py
```diff
@@ -0,0 +1,165 @@
...
+import time
+
+from mozci.errors import ParentPushNotFound
+from mozci.push import Push
+
+logger = logging.getLogger(__name__)
+
+# Maximum number of ancestor pushes to walk back over coalescing gaps before
+# giving up. Mirrors mozci's own MAX_DEPTH.
+MAX_DEPTH = 20
```
marco-c: Can we use MAX_DEPTH from mozci directly?

services/hackbot-pulse-listener/app/regression.py
```diff
@@ -0,0 +1,165 @@
...
+# infra exception and be auto-retried, so we poll until it settles or the
+# deadline elapses (well above a normal build + one retry). Fails open after the
+# deadline so a real regression is never silently dropped.
+POLL_INTERVAL_SECONDS = 120
+MAX_WAIT_SECONDS = 60 * 60
+
+# A green build; mozci reports it as "passed" (Taskcluster) or "success"
+# (Treeherder). Failures are read from mozci's Task.failed attribute instead of
+# reimplementing the vocabulary.
+_PASSED_RESULTS = ("passed", "success")
```
marco-c, line 22: Add a comment explaining why we can't just do "not task.failed"

services/hackbot-pulse-listener/app/regression.py
```diff
@@ -0,0 +1,165 @@
...
+
+# A build in one of these states, or with one of these (infra) results, has not
+# settled: it is still running or was retried after an exception, so its outcome
+# is not knowable yet and we wait for it. Anything else that is not a decisive
+# pass/fail (unscheduled, canceled, superseded, ...) or a build that never ran
+# at all (coalesced) is treated as non-decisive and skipped.
+_UNSETTLED_STATES = ("pending", "running", "exception")
+_UNSETTLED_RESULTS = ("exception", "retry")
```
marco-c, line 30: Might be worth adding an attribute in mozci Task class for this

services/hackbot-pulse-listener/app/regression.py
```diff
@@ -0,0 +1,165 @@
...
+        # mozci's Task.failed also counts `exception` as a failure, but we treat
+        # an exceptioned build as unsettled (infra, likely auto-retried) and wait
+        # for the retry rather than declaring the parent failed.
+        if any(t.failed and t.result not in _UNSETTLED_RESULTS for t in label_tasks):
+            return "failed"
+        if any(
+            t.state in _UNSETTLED_STATES or t.result in _UNSETTLED_RESULTS
+            for t in label_tasks
+        ):
+            return _PENDING
```
marco-c, line 61: If there's one successful, we return "passed". Should we move this before the "failed" case then, so when a pending one in the end passes we'd return "passed"?

marco-c (dismissed):

services/hackbot-pulse-listener/app/regression.py
evgenyrp: yes, it makes sense to return pending first

services/hackbot-pulse-listener/app/regression.py
evgenyrp: added

suhaibmujahid (approved):

PR #6302, opened 2026-07-10
Add Chrome support to autowebcompat repro agent

ksy36:


agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/browser.py
```diff
@@ -1,14 +1,24 @@
...
+import requests
 
 logger = logging.getLogger("autowebcompat-repro")
 
+CHROME_VERSIONS_URL = (
+    "https://googlechromelabs.github.io/chrome-for-testing/"
+    "last-known-good-versions-with-downloads.json"
+)
+CHROME_DOWNLOAD_TIMEOUT = 120
+CHROME_CHUNK_SIZE = 1 << 20
```
jgraham: I don't really agree with Claude's preference for these things being module-level variables. If there are situations where we might want to adjust them we should make them arguments to the actual download function. If there aren't we can just define them inline in the function itself.

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/browser.py
```diff
@@ -72,3 +82,82 @@ def esr(self) -> Path:
...
+        with archive.open("wb") as out:
+            for chunk in response.iter_content(chunk_size=CHROME_CHUNK_SIZE):
+                if chunk:
+                    out.write(chunk)
+
+    with zipfile.ZipFile(archive) as zf:
+        zf.extractall(install_dir)
+    archive.unlink()
+
+    binary = install_dir / f"chrome-{cft_platform}" / "chrome"
```
jgraham: It feels like this is probably wrong on Windows (missing `.exe`), which we are nominally supporting. (oh that's also what Copiolt said below)

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/browser.py
```diff
@@ -72,3 +82,82 @@ def esr(self) -> Path:
...
+    if not binary.exists():
+        raise RuntimeError(f"Chrome binary not found at {binary} after unpacking")
+
+    # zipfile does not preserve the executable bit; restore it.
```
jgraham: Did you actually test if this is true? AFAICT the implementation in wpt doesn't do this: https://searchfox.org/firefox-main/source/testing/web-platform/tests/tools/wpt/browser.py#1545-1553

agents/autowebcompat-repro/Dockerfile
```diff
@@ -28,17 +28,22 @@ ENV PATH="/opt/venv/bin:$PATH"
 
 FROM base AS agent
 
-# The Firefox DevTools MCP server is an npm package launched via `npx`, so the
-# agent image needs Node.js + npm (the python base ships neither). It also
-# needs the shared libraries Firefox requires to run headless; the Firefox
-# binary itself is downloaded at agent startup (a fresh Nightly per run) via
-# mozdownload/mozinstall, not baked in here.
+# The Firefox and Chrome DevTools MCP servers are npm packages launched via
```
jgraham: This comment doesn't really help me understand which dependencies are required for each thing.

I think it would be easier if the comment was inline like
```
RUN apt-get update \
    && apt-get install -y --no-install-recommends \
        # Node, required for devtools mcp
        nodejs npm \
        # Required for Firefox
        ca-certificates \
        libgtk-3-0 libdbus-glib-1-2 libx11-xcb1 libxtst6 libxt6 libpci3 \
        # Required for Chrome
        […]
```
(note that I actually don't know which are required for each)

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/chrome_devtools_mcp.py
```diff
@@ -0,0 +1,42 @@
+from __future__ import annotations
+
+from pathlib import Path
+
+from claude_agent_sdk.types import McpStdioServerConfig
+
+PACKAGE = "chrome-devtools-mcp@latest"
```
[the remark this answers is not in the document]
jgraham: Rather than using npx with a hardcoded package name, maybe we should setup a npm project at the root and install the dependencies into the container? That would allow dependabot to update them (although we probably need to figure out some e2e tests to use to ensure the update works).

I'm also not sure we shouldn't put this in the same file as the firefox mcp configuration and just call it `mcp_servers.py` or something.

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/agent.py
```diff
@@ -644,8 +699,8 @@ async def next_repro_task(
 
     screenshots_dir = Path(tempfile.mkdtemp(prefix="autowebcompat-screenshots-"))
 
-    # Always try in nightly first
-    await next_repro_task(FirefoxChannel.nightly)
+    # Always try in nightly first, cross-checking in Chrome.
+    await next_repro_task(FirefoxChannel.nightly, cross_check_chrome=True)
```
jgraham: I don't think this extra parameter is needed. We should _always_ use Chrome as a baseline when doing the `BugReproduction` steps. However there's a case when we should bail early: if the behaviour matched the description of the bug report but it happened in both Firefox and Chrome. In that case we should set the failure reason to non_webcompat and after the first run we should stop if we think it's not a webcompat issue rather than going on to try multiple browsers.

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/agent.py
```diff
@@ -317,9 +319,13 @@ def __init__(
         input_data: AutoWebcompatInput,
         bugzilla_mcp_server: McpServerConfig,
         screenshot_dir: Path,
+        chrome_path: Path | None = None,
     ):
         super().__init__(task_config, run_tracker)
         self.input_data = input_data
+        # Chrome cross-check is only added when a Chrome binary is supplied
+        # (the initial nightly attempt). Later channel attempts run Firefox-only.
+        self.cross_check_chrome = chrome_path is not None
```
jgraham: I don't think we need this extra variable.

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/agent.py
```diff
@@ -334,29 +340,64 @@ def __init__(
...
+        if self.cross_check_chrome:
+            chrome_step = """
+3. Cross-check in Chrome: run the same steps in Chrome using the Chrome DevTools
+   MCP. A genuine web-compat issue reproduces in Firefox but not in in Chrome.
+   - If the behavior is identical in both browsers, your steps may be wrong or
+     this may not be a compat issue. Iterate: refine the steps and re-check both
+     browsers before concluding. Use the difference (or lack of it) between the
+     two browsers to decide whether you have really reproduced the reported
+     issue.
+   - Set `reproduced` for the Firefox outcome and `chrome_reproduced` for the
```
jgraham: The key thing here should be that the failure reason is `non_webcompat`.

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/result.py
jgraham: Update this to also consider the case where the issue reproduces in both Firefox and Chrome (or add a different status to cover that).

jgraham (changes requested):

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/browser.py
ksy36: Yeah it doesn't work without this, I get:

```
autowebcompat-repro-agent-1  | --- turn 28 ---
autowebcompat-repro-agent-1  | [agent→tool] mcp__chrome-devtools__new_page({"url": "https://drmeth.com/"})
autowebcompat-repro-agent-1  |   [tool←ERROR] MCP error -32000: Connection closed
```
Looks like wpt's `install` function uses `unzip` helper that does set permissions here https://searchfox.org/firefox-main/source/testing/web-platform/tests/tools/wpt/utils.py#98. It also has additional code to support unzipping on Window ans MacOS, so maybe I should do the same, though I can't try it to verify them with the container we have

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/agent.py
ksy36: sounds good, I changed it to run on every BugReproduction run and added a check for `non_compat`

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/chrome_devtools_mcp.py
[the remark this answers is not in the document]
ksy36: Ah yes good idea, I've added the npm project and combined the two mcp servers together

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/browser.py
ksy36: True, I've added that

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/agent.py
```diff
@@ -507,6 +530,22 @@ def __init__(self, publish_file: PublishFile, plan_result: TestPlanResult):
     def reproduced(self) -> bool:
         return self.initial_repro is not None
 
+    @property
+    def chrome_reproduced(self) -> bool | None:
+        # Prefer the Chrome verdict from the channel that reproduced in Firefox,
```
jgraham: This comment feel unnecessary.

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/agent.py
```diff
@@ -507,6 +530,22 @@ def __init__(self, publish_file: PublishFile, plan_result: TestPlanResult):
...
+        # since that is the meaningful cross-check. If nothing reproduced in
+        # Firefox, still surface the first Chrome verdict we got (since the cross-check
+        # ran on every BugReproduction attempt).
+        if self.initial_repro is not None:
+            return self.initial_repro.chrome_reproduced
+        for result in self.results.values():
```
jgraham: It's not an invariant enforced by the types, but in practice we shouldn't get into a state where we take this branch; either `initial_repro` is not None or we don't have any reproduction at all. So I think we can just delete the loop here.

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/agent.py
```diff
@@ -647,6 +693,13 @@ async def next_repro_task(
     # Always try in nightly first
     await next_repro_task(FirefoxChannel.nightly)
 
+    # If the reported behavior reproduced in both Firefox and Chrome it is not a
+    # Firefox web-compat issue: stop early.
+    if repro_results.reproduced and repro_results.chrome_reproduced:
+        result = repro_results.into_result()
```
jgraham, line 696: Might be interesting to log a warning if `result.failure_reason != "non_compat"`

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/mcp_servers.py
```diff
@@ -0,0 +1,109 @@
...
+the image with ``npm ci`` (see the Dockerfile).
+"""
+
+from __future__ import annotations
+
+from pathlib import Path
+
+from claude_agent_sdk.types import McpStdioServerConfig
+
+NODE_PROJECT_DIR = Path("/app/node")
```
jgraham: These constants should also be parameters to functions (if we might ever change them) or just hardcoded (if we won't).

jgraham (dismissed):
Some minor fixups, but I don't think this requires more review to land.

jgraham (approved):

PR #6315, opened 2026-07-14
Add severity assessment to frontend-triage agent

msujaws:
Extend the frontend-triage agent with a new triage judgment driven by rules and prompt (no ML models or new tools).

## What it does
- **Severity assessment:** assesses an appropriate Mozilla severity (S1–S4) from user impact.

At **high confidence** the agent records a `bugzilla_update_bug` proposal for the severity for human review; at **medium/low confidence** it reports the assessment only.

## Changes
- Adds a `severity-assessment.md` rule file.
- Adds the corresponding job step and structured-output key in the system prompt.
- Adds a `SeverityAssessment` submodel parsed into `FrontendTriageResult`.

### DIFF
--- a/agents/frontend-triage/hackbot_agents/frontend_triage/rules/component-verification.md
+++ b/agents/frontend-triage/hackbot_agents/frontend_triage/rules/component-verification.md
@@ -1,0 +1,50 @@
+# Product / component verification
+
+Bugs are frequently filed against the wrong `Product :: Component`. Part of triage is
+confirming the bug is filed where it belongs and, when it isn't, proposing the correct
+`Product :: Component` so it reaches the right team.
+
+Do this **after** you have localized the affected code (you need concrete file paths to
+verify ownership), and record the result in the `component_assessment` structured-output
+object.
+
+## How to verify
+
+1. **Find the ownership map.** The Firefox source tree ships a module-ownership file,
+   `mots.yaml` (the maintainer/module metadata; commonly at the repo root or under
+   `source/`). Glob for it (`**/mots.yaml`) and Read it. Each module lists path globs it
+   owns (e.g. an `includes:` list) and its Bugzilla product/component (e.g. under a
+   `meta`/`components` mapping — field names may vary slightly, read the file to see the
+   shape).
+2. **Match your localized paths to a module.** Take the file path(s) you identified as
+   the root cause and find the module whose path globs own them. That module's Bugzilla
+   product/component is the _expected_ component for a bug in that code. Use Searchfox to
+   confirm ownership when a path spans directories or the mapping is ambiguous.
+3. **Compare with the bug.** Compare the expected `Product :: Component` against the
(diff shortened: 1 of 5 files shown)


agents/frontend-triage/hackbot_agents/frontend_triage/rules/component-verification.md
```diff
@@ -0,0 +1,50 @@
+# Product / component verification
```
marco-c: I think we could do this in bugbot instead, without an LLM, so it'll be cheaper and more deterministic.

msujaws: I do think the LLM can cover bases that today might be in a grey area, though agree that we can separate these two out and get severity in sooner which will be helpful for flagging high severity issues

crythms (approved):
looks good. Thanks

agents/frontend-triage/hackbot_agents/frontend_triage/rules/component-verification.md
marco-c: There's a WIP patch for this here: https://github.com/mozilla/bugbug/pull/6505. It requires a bit more work, but we could deploy that first and then figure out if that's enough or not.

agents/frontend-triage/hackbot_agents/frontend_triage/rules/severity-assessment.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Severity definitions
+
+- **S1 — catastrophic.** Crash, hang, data loss, security issue, or a bug that blocks
+  major functionality with **no workaround**. Affects a large number of users.
+- **S2 — serious.** Major functionality is broken or a severe UX problem, and the
+  workaround (if any) is painful or non-obvious. Affects many users.
+- **S3 — normal.** Blocks non-critical functionality, or a reasonable workaround exists.
+  **This is the default for most frontend papercuts.**
+- **S4 — minor / trivial.** Cosmetic issues, small polish, or edge cases with negligible
```
marco-c, line 16: I do wonder if we shouldn't update these to make them more clear. They are often confusing for users, so they will also be for LLMs :)

marco-c (approved):

agents/frontend-triage/hackbot_agents/frontend_triage/rules/severity-assessment.md
msujaws: Yeah, agreed. We can do that in a follow-up though, as there will probably be some discussion with others as to what we want them to be.

marco-c: Yes, it's not the right time/place to change them! It does make me think that maybe we should have some sort of CI check or something that alerts us if definitions change and the definitions here are not updated.

PR #6325, opened 2026-07-16
Do not require bug ID on docker compose up

evgenyrp:
This fails other agents that don't require `BUG_ID` when running from a root compose that nicely loads `.env`.
```
FAILURE_TASKS='{"build-linux":"GFjLPsWwRDiMhtFprcwyGQ"}' docker compose up build-repair-agent --build
```

### DIFF
--- a/agents/bug-fix/compose.yml
+++ b/agents/bug-fix/compose.yml
@@ -17,7 +17,7 @@ services:
       target: agent
     environment:
       - RUN_ID
-      - BUG_ID=${BUG_ID:?error}
+      - BUG_ID=${BUG_ID:-}
       - BUGZILLA_MCP_URL=http://bug-fix-broker:8765/mcp
       - SOURCE_REPO=/workspace/firefox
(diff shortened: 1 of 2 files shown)


suhaibmujahid:
What if we drop the linking to the root compose file? It might be a better route long run. WDYT?

evgenyrp:
> What if we drop the linking to the root compose file? It might be a better route long run. WDYT?

We could; it's a matter of convenience. Then we should update the README to use docker compose with `-f` flag to point to specific compose files. And also, a developer would need to load .env manually, which is a bit less convenient. Maybe we could point each individual compose to .env

suhaibmujahid:
> We could; it's a matter of convenience. Then we should update the README to use docker compose with `-f` flag to point to specific compose files. And also, a developer would need to load .env manually, which is a bit less convenient. Maybe we could point each individual compose to .env

The agent's compose could be loaded from inside the agent's directory. With that, you do not need `-f`.

For the `.env`, we could make it point to the root `.env` and adjacent `.env` at the same time. I think this need be don from the agent's compose file itself.

PR #6334, opened 2026-07-18
Fix Phabricator URL action references

### LINKED ISSUE #6330: Hackbot does not substitute action references when commenting on Bugzilla.
Example: https://bugzilla.mozilla.org/show_bug.cgi?id=2033910#c6

ShreyanshGoyal:
## Summary

- expose the canonical `url` field in successful Phabricator submission results
- retain `revision_url` for compatibility with existing consumers
- cover the documented action-reference contract in the handler test

## Rationale

The `phabricator.submit_patch` action tells agents to reference a submitted revision with `{{actions.<ref>.url}}`. The apply handler returned only `revision_url`, so the placeholder resolver could not find `url` and left the literal reference in Bugzilla comments. Returning the documented alias makes those references resolvable without breaking consumers of `revision_url`.

Fixes #6330.

## Test plan

- [x] Confirmed the updated handler assertion fails before the source change
- [x] `python -m pytest libs/hackbot-runtime/tests/test_phabricator_handler.py libs/hackbot-runtime/tests/test_phabricator_actions.py -q` (18 passed)
- [x] `python -m ruff check` and `python -m ruff format --check` on the changed Python files
- [x] `python -m compileall -q libs/hackbot-runtime/hackbot_runtime/actions/handlers/phabricator_handler.py`

### DIFF
--- a/libs/hackbot-runtime/hackbot_runtime/actions/handlers/phabricator_handler.py
+++ b/libs/hackbot-runtime/hackbot_runtime/actions/handlers/phabricator_handler.py
@@ -275,11 +275,11 @@ class SubmitPatchHandler:
             log.exception("Failed to submit Phabricator diff for bug %s", bug_id)
             return ActionResult.failed(str(exc))
 
+        revision_url = _revision_url(new_revision_id) if new_revision_id else None
         return ActionResult.ok(
             {
                 "revision_id": new_revision_id,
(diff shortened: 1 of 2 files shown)


suhaibmujahid (approved):

suhaibmujahid:
Thank you!

PR #6353, opened 2026-07-20
Support triggering the bug-fix agent for follow-ups when mentioning `@hackbot`

suhaibmujahid:
> [!Note]
This PR car be reviewed commit by commit

This PR will add support to trigger the bug-fix agent for follow-ups when mentioning `@hackbot` in a comment on a Phabricator revision.


agents/bug-fix/hackbot_agents/bug_fix/agent.py
```diff
@@ -147,7 +148,21 @@ async def run_bug_fix(
...
+        user_prompt = (
+            f"Follow up on Phabricator revision D{revision_id} for bug {bug}.\n\n"
+            f"A reviewer left a comment (see the instructions in your system "
+            f"prompt). Address it: investigate, make the necessary source "
+            f"changes, and verify the fix. When you are done, submit your "
+            f"changes by calling submit_patch with revision_id={revision_id} so "
+            f"the existing revision D{revision_id} is updated — do not create a "
+            f"new revision.\n\n"
+            f"Consult the relevant rules in {rules_path} if they apply."
+        )
```
marco-c: it would also be nice to support the use case where the comment is just a question for hackbot to answer. I can imagine a reviewer asking a question about a patch (e.g. "verify this can't do bla bla bla"), either from hackbot or also from another engineer.

suhaibmujahid: Yes, I will update it to support that.

suhaibmujahid: Done in 9d608dd6efe4affede55d98a4efabde9569ec337

marco-c (approved):
LGTM overall. I didn't review all the code, but the design looks good.

As discussed today, let's think if we also want to trigger hackbot automatically on all comments on hackbot-owned revisions.

PR #6354, opened 2026-07-20
Add regression-range hackbot agent (mozregression --prompt bisection)

### LINKED ISSUE #6259: Create hackbot agent to automatically try to find a regression range for regression bugs
The agent could use mozregression with https://github.com/mozilla/mozregression/pull/2197.

msujaws:
Resolves #6259 — a hackbot agent that automatically finds a regression range for a regression bug.

## What it does

Given a Bugzilla bug id, the agent:
1. Fetches the bug (fields + comments) and decides whether it is an **automatable regression** (declines cleanly otherwise: `not_a_regression` / `not_automatable`).
2. Determines **good/bad bounds** — using a suspected regressor's landing (via `mozilla_vcs`) when available, anchoring `good` just before it.
3. Resolves any required **prefs/feature-gating via searchfox** (never guesses pref names) and authors a natural-language good/bad reproduction directive.
4. Runs **`mozregression --prompt`** ([mozilla/mozregression#2197](https://github.com/mozilla/mozregression/pull/2197)) to drive a real bisection — each candidate build is classified GOOD/BAD by an agent driving the Firefox DevTools MCP (the natural-language equivalent of `git bisect run`).
5. Reports the resulting changeset/pushlog range and **records** a Bugzilla comment plus, at high confidence, proposed field updates (`cf_has_regression_range`, `regressed_by`, clearing `regressionwindow-wanted`). Actions are recorded for human review — never auto-applied.

## Key components

- `libs/agent-tools/agent_tools/mozregression.py` — framework-neutral `run_mozregression` tool: an async subprocess wrapper that builds the CLI invocation and parses the final range. Handles both regression (`Last good`/`First bad`) and `--find-fix` (`First good`/`Last bad`) wording, takes the last pushlog URL, and falls back to the pushlog `fromchange`/`tochange`. Always returns a dict; never raises.
- `agents/regression-range/` — the agent, mirroring the `frontend-triage` structure (brokered Bugzilla read + searchfox + mozilla_vcs + the mozregression tool + recorded actions). The system prompt scopes to reproductions that are **deterministic in a fresh headless profile** (including `about:`/chrome pages and browser UI — menus, tabs, sidebars, toolbars), resolves prefs via searchfox, anchors bounds to a suspected regressor's landing, and runs scroll/layout checks at a realistic window size.
- `services/hackbot-api` — registers the agent (`RegressionRangeInputs` + `AGENT_REGISTRY`).
- `services/hackbot-ui` — label overrides for the new finding fields.
- Tests for the tool's argv construction / range parsing and the registry entry.

## Verification (local)

- The `run_mozregression` tool drove a real `#2197 --prompt` bisection to a changeset/pushlog range.
- The full agent ran end-to-end via `docker compose` on real bugs: correct scoping decisions, searchfox-verified prefs, regressor-anchored bounds, and honest `inconclusive` reporting when a symptom wasn't observable headless (while still identifying the likely regressor).

## Notes

- The agent image installs mozregression from the fork carrying #2197 (`MOZREGRESSION_REF` build arg). Pin to a specific commit once #2197 lands / is released.
- For local `docker compose` runs, `BUGZILLA_API_URL` must include the `/rest` suffix.

### DIFF
--- a/agents/regression-range/Dockerfile
+++ b/agents/regression-range/Dockerfile
@@ -1,0 +1,80 @@
+FROM python:3.12 AS builder
+
+COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/
+
+ENV UV_PROJECT_ENVIRONMENT=/opt/venv
+
+WORKDIR /app
+
+# Install external deps without building workspace members.
+RUN --mount=type=cache,target=/root/.cache/uv \
+    --mount=type=bind,source=pyproject.toml,target=pyproject.toml \
+    --mount=type=bind,source=uv.lock,target=uv.lock \
+    --mount=type=bind,source=VERSION,target=VERSION \
+    uv sync --frozen --no-dev --no-install-workspace --package hackbot-agent-regression-range
+
+RUN --mount=type=cache,target=/root/.cache/uv \
+    --mount=type=bind,target=/app,rw \
+    uv sync --locked --no-dev --no-editable --package hackbot-agent-regression-range
+
+# mozregression is installed from the fork carrying PR #2197 (the --prompt mode
+# is not yet in a release), so it lives outside the workspace lockfile. Pin to a
+# specific commit once the PR settles; the branch tracks the PR for now.
+ARG MOZREGRESSION_REF="git+https://github.com/msujaws/mozregression.git@prompt-devtools-mcp-verdict"
+RUN --mount=type=cache,target=/root/.cache/uv \
+    uv pip install --python /opt/venv "mozregression @ ${MOZREGRESSION_REF}"
+
+FROM python:3.12 AS base
+
+COPY --from=builder /opt/venv /opt/venv
+WORKDIR /app
+
+ENV PYTHONUNBUFFERED=1
+ENV PYTHONDONTWRITEBYTECODE=1
+ENV PATH="/opt/venv/bin:$PATH"
+
+FROM base AS agent
+
+# mozregression's --prompt mode (mozilla/mozregression#2197) shells out to the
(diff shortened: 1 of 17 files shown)


agents/regression-range/Dockerfile
```diff
@@ -0,0 +1,80 @@
...
+RUN --mount=type=cache,target=/root/.cache/uv \
+    --mount=type=bind,target=/app,rw \
+    uv sync --locked --no-dev --no-editable --package hackbot-agent-regression-range
+
+# mozregression is installed from the fork carrying PR #2197 (the --prompt mode
+# is not yet in a release), so it lives outside the workspace lockfile. Pin to a
+# specific commit once the PR settles; the branch tracks the PR for now.
+ARG MOZREGRESSION_REF="git+https://github.com/msujaws/mozregression.git@prompt-devtools-mcp-verdict"
+RUN --mount=type=cache,target=/root/.cache/uv \
+    uv pip install --python /opt/venv "mozregression @ ${MOZREGRESSION_REF}"
```
suhaibmujahid, line 25: Can we instead use `[tool.uv.sources]` in `pyproject.toml` so it gets installed with other deps without a custom step in Dockerfile?

marco-c:
This is great, I wanted to talk to you about doing this, but you did it earlier than I reached out :)

What are your thoughts on the split between this and frontend-triage? Should frontend-triage use this as a sub-agent instead of them running separately?

suhaibmujahid (commented):
You might want to add the new agent to the `CODEOWNERS` file.

agents/regression-range/Dockerfile
msujaws: Done

PR #6365, opened 2026-07-21
Filter hackbot UI runs by selected agent and drop localStorage

suhaibmujahid:
## What

The hackbot UI recent-runs panel now shows only runs from the agent selected in the trigger form, and no longer relies on the browser's `localStorage` for the list of runs.

## Why

Previously the panel merged the hackbot-api `GET /runs` response with a `localStorage` store (`lib/store.ts`), and it showed runs from all agents regardless of which agent was selected. Selecting an agent had no effect on the panel.

## Changes

- **hackbot-api**: `GET /runs` accepts an optional `agent` query param and filters server-side (the `agent` column is already indexed).
- **UI `lib/hackbot.ts`**: `listRuns(limit, agent?)` forwards the filter.
- **UI `/api/runs`**: passes the `agent` query param through.
- **UI `lib/agents.ts`** (new): shared agent list + `parseAgent`, used by both the trigger form and the panel so they agree on the selected agent.
- **UI `TriggerForm`**: selecting an agent writes `?agent=` to the URL.
- **UI `RecentRuns`**: reads the selected agent from `?agent=`, fetches only that agent's runs from the API, and refetches on change / while polling. Wrapped in `Suspense` (uses `useSearchParams`).
- **UI**: removed `lib/store.ts` and all `localStorage` usage (also dropped the now-unused `updateRunStatus`/`saveRun` calls in `RunDetail`/`TriggerForm`).
- **README**: updated the stale note about there being no upstream list-runs endpoint.

## Notes

- With no `?agent=` param, both the form and the panel default to `bug-fix`, so they stay consistent.
- Could not run `next build`/`tsc` in this environment (no `node_modules`, registry not reachable); the pre-commit prettier hook over the TS/TSX files passed.


suhaibmujahid:
Abandoning in favour of #6370

PR #6375, opened 2026-07-23
Test repair agent: Part1

evgenyrp:
Analyzes test failure, identifies the culprit (including previous pushes), proposes a fix if it can and gives a recommendation what to do about it (back out, land fix, rerun etc.).

It follows the architecture of Build Repair Agent.

The most interesting parts are in "deterministic" code that prepares the data for the agent, as there might be bugs there. Still, we want to narrow it down for the agent to simplify it, make it less fragile, and spend fewer tokens.

The listener (Part 2) will filter non-regressions, intermittent failures etc., so that the API is triggered only for useful investigations.

I tested it on a few test failures locally, but we should deploy it and observe how it works with real-time failures in different environments.

[Test runs in Weave](https://wandb.ai/moz-bugbug/hackbot-test/weave/agents/conversations/f6fef743-790b-4f0a-bb15-d830dc668658?convTurn=1&chatTrace=33ffc2bcf2f71b35b748d410cf59f318) (the last one with all the fixes)

https://github.com/mozilla/bugbug/issues/6399

### DIFF
--- a/agents/test-repair/hackbot_agents/test_repair/config.py
+++ b/agents/test-repair/hackbot_agents/test_repair/config.py
@@ -1,0 +1,48 @@
+# -*- coding: utf-8 -*-
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+"""Models and tool allowlist for the test-repair agent."""
+
+ANALYSIS_MODEL = "claude-opus-4-8"
+FIX_MODEL = "claude-opus-4-8"
+
+# Bugzilla MCP tool names as exposed to the agent (mcp__<server>__<tool>).
+BUGZILLA_READ_TOOLS = [
+    "mcp__bugzilla__search_bugs",
+    "mcp__bugzilla__get_bugs",
+    "mcp__bugzilla__get_bug_comments",
+    "mcp__bugzilla__get_bug_attachments",
+    "mcp__bugzilla__download_attachment",
+]
+
+# In-process Firefox build/test MCP tools (reused from agent-tools). The agent
+# uses these to reproduce a test failure and verify a proposed fix.
+BUILD_TOOL = "mcp__firefox__build_firefox"
+FIREFOX_TOOLS = [
+    BUILD_TOOL,
+    "mcp__firefox__bootstrap_firefox",
+    "mcp__firefox__evaluate_testcase",
+    "mcp__firefox__evaluate_js_shell",
+]
+
+# Built-in tools the agent may call alongside the MCP servers. The agent runs in
+# an isolated container (permission_mode="bypassPermissions"), and reads the
+# candidate commit diffs via Bash `git show`.
+ALLOWED_TOOLS = [
+    "Read",
+    "Grep",
+    "Glob",
+    "Bash",
+    "Edit",
+    "Write",
+    "MultiEdit",
+    "WebFetch",
+    "WebSearch",
+]
+
+ADDITIONAL_DIRS = [
+    "~/.mozbuild",
+    "~/.cache/uv/",
+]
--- a/agents/test-repair/hackbot_agents/test_repair/config.py
+++ b/agents/test-repair/hackbot_agents/test_repair/config.py
@@ -17,15 +17,12 @@ BUGZILLA_READ_TOOLS = [
     "mcp__bugzilla__download_attachment",
 ]
 
-# In-process Firefox build/test MCP tools (reused from agent-tools). The agent
-# uses these to reproduce a test failure and verify a proposed fix.
+# In-process Firefox build MCP tool (reused from agent-tools). Only the build
+# tool: evaluate_testcase / evaluate_js_shell do not run the harnesses CI runs,
+# so the agent runs the failing test itself with mach over Bash, and bootstrap is
+# deterministic prep rather than something the agent has to decide to call.
 BUILD_TOOL = "mcp__firefox__build_firefox"
-FIREFOX_TOOLS = [
-    BUILD_TOOL,
-    "mcp__firefox__bootstrap_firefox",
-    "mcp__firefox__evaluate_testcase",
-    "mcp__firefox__evaluate_js_shell",
-]
+FIREFOX_TOOLS = [BUILD_TOOL]
 
 # Built-in tools the agent may call alongside the MCP servers. The agent runs in
 # an isolated container (permission_mode="bypassPermissions"), and reads the
--- a/agents/test-repair/hackbot_agents/test_repair/config.py
+++ b/agents/test-repair/hackbot_agents/test_repair/config.py
@@ -5,8 +5,8 @@
 
 """Models and tool allowlist for the test-repair agent."""
 
-ANALYSIS_MODEL = "claude-opus-4-8"
-FIX_MODEL = "claude-opus-4-8"
+ANALYSIS_MODEL = "claude-opus-5"
+FIX_MODEL = "claude-opus-5"
 
 # Bugzilla MCP tool names as exposed to the agent (mcp__<server>__<tool>).
 BUGZILLA_READ_TOOLS = [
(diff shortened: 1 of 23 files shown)


agents/test-repair/hackbot_agents/test_repair/config.py
```diff
@@ -0,0 +1,48 @@
...
+]
+
+# In-process Firefox build/test MCP tools (reused from agent-tools). The agent
+# uses these to reproduce a test failure and verify a proposed fix.
+BUILD_TOOL = "mcp__firefox__build_firefox"
+FIREFOX_TOOLS = [
+    BUILD_TOOL,
+    "mcp__firefox__bootstrap_firefox",
+    "mcp__firefox__evaluate_testcase",
+    "mcp__firefox__evaluate_js_shell",
```
marco-c: These two likely won't be useful for this agent, they are not running the same test cases we run in CI.
If we want the agent to be able to run Firefox tests, we either need to add other tools, or we just let it discover by itself how to run them using bash and the AGENTS.md/CLAUDE.md in the Firefox repo.

Take into account many test failures are also platform specific, so for example we can't reproduce a win or mac failure if we are only running on linux (unless we run the tests on taskcluster)

agents/test-repair/hackbot_agents/test_repair/prompts.py
```diff
@@ -0,0 +1,56 @@
...
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+"""Prompt templates for the test-repair agent."""
+
+ANALYSIS_TEMPLATE = """\
+You are investigating a failing Firefox CI test to find the commit that broke it.
+Treat this as a genuine regression: a commit that landed since the test was last
+green introduced the failure.
+
+Failing test groups (manifests) and a representative failing test in each:
```
marco-c: Why not passing all failing tests?

agents/test-repair/hackbot_agents/test_repair/prompts.py
```diff
@@ -0,0 +1,56 @@
...
+1. Read the sanitized failure lines to understand exactly how the test failed.
+2. Use `git show <commit>` on the candidate commits above to inspect their diffs
+   and identify the single commit that most plausibly introduced the failure. You
+   may search Bugzilla for a related bug.
+3. Write these files to {scratch_out}:
+   - summary.md: a short (2-4 sentence) verdict.
+   - analysis.md: the detailed reasoning, with evidence from the logs and diffs.
+   - verdict.json: an object with keys "culprit_commit" (a full sha from the
+     candidates above, or null if none is convincing), "culprit_bug" (integer or
+     null), "recommendation" ("backout" or "land_fix") and "confidence"
```
marco-c: Should we also have a recommendation to "rerun" in case the agent is not sure it is actually a real failure? Or just rely on confidence being low?

agents/test-repair/hackbot_agents/test_repair/resolve.py
```diff
@@ -0,0 +1,259 @@
...
+        return None
+    if summary.status == Status.PASS:
+        return "passed"
+    if summary.status == Status.FAIL:
+        return "failed"
+    # INTERMITTENT or still running: can't anchor last-green here.
+    return None
+
+
+def _last_green(branch: str, rev: str, group: str) -> str | None:
```
marco-c: We might want to do this at the test level instead of the group level, also we need to take into account the platform where the test run (e.g. if the test is only failing on linux, we risk last_green to find a commit where the test run and passed on win but didn't run on linux)

agents/test-repair/hackbot_agents/test_repair/config.py
evgenyrp: I see, I wasn't sure about these either

agents/test-repair/hackbot_agents/test_repair/prompts.py
evgenyrp: That's an interesting idea

agents/test-repair/hackbot_agents/test_repair/prompts.py
```diff
@@ -0,0 +1,114 @@
...
+{last_green_line}
+Failure logs (start with the sanitized failures file; fall back to the full log):
+{failure_logs}
+
+Do the following:
+1. Read the sanitized failure lines to understand exactly how the test failed.
+2. Enumerate the candidates with `git log --oneline {commit_range}`. Narrow them
+   before reading diffs -- `git log --oneline {commit_range} -- <path>` on the
+   failing test's directory and on the source it exercises is usually enough to
+   get to a handful. Then `git show <commit>` those to identify the single commit
```
marco-c: I wonder if the narrowing could exclude valid culprits

agents/test-repair/hackbot_agents/test_repair/prompts.py
```diff
@@ -0,0 +1,114 @@
...
+   - verdict.json: an object with keys "classification" ("regression" or
+     "intermittent"), "culprit_commit" (a full sha copied verbatim from the
+     candidate list above, or null if none is convincing), "culprit_bug" (integer
+     or null), "intermittent_bug" (integer or null; the bug already tracking this
+     intermittent, if you found one), "recommendation" ("backout", "land_fix",
+     "do_not_backout" or "rerun") and "confidence" (0.0-1.0).
+
+Use "rerun" when you cannot tell whether the failure is real -- infrastructure
+noise, a timeout under load, or a failure no candidate plausibly explains -- and
+a retrigger would settle it. Prefer it over blaming a commit you are unsure of.
```
marco-c, line 70: It'd be nice, when we can't find a 100% culprit, to return a list of potential culprits (so sheriffs can retrigger only on those instead of backfilling completely)

agents/test-repair/hackbot_agents/test_repair/resolve.py
```diff
@@ -0,0 +1,322 @@
...
+    not anchor a last-green for a Linux-only failure. None means non-decisive
+    (the group did not run here, or is intermittent/unfinished).
+    """
+    summary = push.group_summaries.get(group)
+    if summary is None:
+        return None
+    wanted = set(tests)
+    statuses = set()
+    for task in summary.tasks:
+        if not _same_platform(task, platform):
```
marco-c, line 180: The platform might not necessarily be just "windows", but also a sub-configuration (e.g. we used to have "windows with multiprocessing" or "windows without multiprocessing").
If we want to have exactly the same tests on the same configuration, we can use some of the "tags" from the task definition or the mozci "get_configuration" function.

However, we might also want to try to give the full list to the model, and then it can figure it out by itself (e.g. there might be cases where the test would fail the same way between different windows configurations, and cases where it would only fail on one specific windows configuration).

marco-c: Ah, I see you're using "test-platform", maybe that's enough then.

However, we're missing the other case where the failure might not be specific to a config, but just to the OS for example. But to identify this, we'd have to give everything to the model. Maybe a follow-up to investigate?

agents/test-repair/hackbot_agents/test_repair/resolve.py
```diff
@@ -0,0 +1,322 @@
...
+            continue
+        groups.append(FailingGroup(group=group, tests=[test for test, _type in fails]))
+    return groups
+
+
+def _same_platform(task, platform: str) -> bool:
+    return task.platform == platform or platform in (task.label or "")
+
+
+def _test_status(push: Push, group: str, tests: list[str], platform: str) -> str | None:
```
marco-c, line 165: Note not all test suites are group-based, e.g. gtest

evgenyrp:
I think some complexity in deterministic code can go away by using Treeherder CLI and/or API, as it provides some of the things we do manually out of the box. I've already switched to the API in the listener. I propose doing it as a follow-up, though, as a part of #6417

agents/test-repair/hackbot_agents/test_repair/resolve.py
evgenyrp: I wouldn't dig too deep into this for now. I think this will change significantly if we migrate to Treeherder CLI/API.

marco-c (approved):

PR #6380, opened 2026-07-24
Add `SubmitTestPlan` action to submit to Test plans to TestRails

### LINKED ISSUE #6414: Integrate the test-plan-generator agent with the QA workflow on TestRail
The agent already landed in #6237

jpangas:
Resolves #6414 

This pull request introduces support for recording and applying TestRail actions for generated test plans. It adds a new action type for submitting generated test cases to TestRail, implements the handler to make the necessary API calls, and ensures these actions are properly registered and tested. Additionally, the test plan generator agent is updated to automatically record and apply these actions, and new tests are included to verify the integration.


libs/hackbot-runtime/hackbot_runtime/actions/handlers/testrail_handler.py
```diff
@@ -0,0 +1,193 @@
...
+
+def _project_id() -> int:
+    return _required_int_env("TESTRAIL_PROJECT_ID")
+
+
+def _api_url(endpoint: str) -> str:
+    return f"{_base_url()}/index.php?/api/v2/{endpoint.lstrip('/')}"
+
+
+def _api_request(
```
suhaibmujahid: Let us move the part to communicate with TestRail api into its own lib (i.e., `libs/testrail-client`) like we do for Phabricator (`libs/phabricator-client`). We could follow that as a reference design. 

We keep here the business logic only.

libs/hackbot-runtime/hackbot_runtime/actions/handlers/testrail_handler.py
```diff
@@ -0,0 +1,193 @@
...
+        except Exception as exc:
+            log.exception(
+                "Failed to submit test plan to TestRail for run %s", ctx.run_id
+            )
+            return ActionResult.failed(str(exc))
+
+        return ActionResult.ok(
+            {
+                "suite_id": suite_id,
+                "suite_url": _suite_url(suite_id),
```
suhaibmujahid: Using `url` instead of `suite_url` would avoid know issues, see https://github.com/mozilla/bugbug/issues/6330
```suggestion
                "url": _suite_url(suite_id),
```

libs/hackbot-runtime/hackbot_runtime/actions/testrail.py
```diff
@@ -0,0 +1,28 @@
...
+from __future__ import annotations
+
+from typing import Any
+
+from hackbot_runtime.actions.recorder import ActionsRecorder
+
+ACTION_TYPE = "testrail.submit_test_cases"
+
+
+def record_test_plan(
```
suhaibmujahid, line 110: Is this working? I expect not.

Could you please structure it like `libs/hackbot-runtime/hackbot_runtime/actions/phabricator.py`?

libs/hackbot-runtime/hackbot_runtime/actions/testrail.py
```diff
@@ -0,0 +1,28 @@
...
+    recorder: ActionsRecorder,
+    test_plan: dict[str, Any],
```
suhaibmujahid, line 112: Please use `pydantic` to annotate the inputs like in `actions/phabricator.py`. That will be provided to the agent to know ho to use the tool.

libs/hackbot-runtime/hackbot_runtime/actions/testrail.py
```diff
@@ -0,0 +1,28 @@
...
+    *,
+    reasoning: str = "Upload the generated test cases to TestRail.",
```
suhaibmujahid: No point in having a default value here. If we have them test case by test case, then having the agent providing reasoning would make sense, but for the test plan itself, it is kind of useless, and we could drop it.

libs/hackbot-runtime/hackbot_runtime/actions/handlers/testrail_handler.py
```diff
@@ -0,0 +1,193 @@
...
+    return payload
+
+
+def _suite_url(suite_id: int) -> str:
+    return f"{_base_url()}/index.php?/suites/view/{suite_id}"
+
+
+class SubmitTestCasesHandler:
+    async def apply(self, params: dict[str, Any], ctx: ApplyContext) -> ActionResult:
+        feature = str(params.get("feature") or "").strip()
```
suhaibmujahid, line 88: This is a test case handler, but it's currently handling test plans too. :p

I still think it'd be better to allow adding cases one by one. We could have a separate method to create a suite, and another to create a section; each taking a ref. Test cases would then reference that ref, making it available when the cases get applied, so we can retrieve their IDs once they're created and use them to create the cases.

suhaibmujahid (changes requested):
Nice! Thank you! Please see my comments, and we can discuss details.

libs/testrail-client/testrail_client/client.py
```diff
@@ -0,0 +1,76 @@
...
+        return result
+
+    async def get_case_types(self) -> dict[str, Any] | list[Any]:
+        return await self.request("GET", "get_case_types")
+
+    async def get_templates(
+        self, project_id: int | None = None
+    ) -> dict[str, Any] | list[Any]:
+        return await self.request(
+            "GET", f"get_templates/{project_id or self.settings.project_id}"
```
suhaibmujahid: The `project_id` value should be either explicit (by parameter) or implicit (by attribute). Having both could be confusing when we debugging things.

libs/testrail-client/testrail_client/config.py
```diff
@@ -0,0 +1,22 @@
...
+from __future__ import annotations
+
+from pydantic import BaseModel, Field
+from pydantic_settings import BaseSettings, SettingsConfigDict
+
+
+class TestRailSettings(BaseModel):
+    username: str = Field(min_length=1)
+    api_key: str = Field(min_length=1)
+    project_id: int
```
suhaibmujahid, line 12: why do we need `project_id` in the config?

libs/hackbot-runtime/hackbot_runtime/actions/handlers/testrail_handler.py
```diff
@@ -0,0 +1,139 @@
...
+                            test_case,
+                            case_type_id,
+                            template_id,
+                        ),
+                    ),
+                    "case",
+                )
+                created_case_ids[generated_id] = case_id
+
+        except Exception as exc:
```
suhaibmujahid, line 126: I do not like big try-catch blocks, and I do not like the generic exceptions, but we already have them in other places. I will file an issue to refactor them. (No action needed here)

libs/testrail-client/testrail_client/config.py
jpangas: I went with keeping`project_id `in the config because we currently submit to one fixed TestRail project. Since it is deployment level confg, I think it’s better we configure it once instead of passing it through every method call. If we placed it outside the client settings, we would either need to add `project_id` to the action schema and make Hackbot provide it every time, or do another env lookup in the handler. 
Also, since the same `project_id` is needed by multiple TestRail calls, keeping it in settings gives us one source. 
Based off your previous comment, I would remove the `project_id` parameter so it is always implicit.
But, do you think we shall we support selecting a project later in hackbot in the future? in that case we would think of designing it in a way that it is resolved by hackbot.

libs/testrail-client/testrail_client/client.py
jpangas: I want to   remove explicit parameter, `project_id` parameter so it is always implicit. However, see my next comment and let me know what you think.

suhaibmujahid: Let us keep it in the setting and remove it from here. We can change that later when we want to support selecting the project.

libs/testrail-client/testrail_client/config.py
suhaibmujahid: Ok! Let us keep it here for now.

suhaibmujahid (changes requested):

suhaibmujahid (approved):
LGTM! Thank you!

PR #6384, opened 2026-07-26
frontend-triage: pin the model to Opus 5 and read files before naming them

crythms:
- Nothing sets a model to use, so every run used whatever the Claude Code CLI defaulted to. Override it per run with MODEL or the model field in the UI. Default to Opus 5  as the agent is producing better triage results with.

- Tell the agent to open the files it cites rather than trust a search hit: a search shows what is there, never what is missing.

- Condense the README (333 -> 134 lines) by cutting tool lists and a file tree that duplicate the code and go stale.


msujaws (approved):

msujaws:
Thank you!

PR #6386, opened 2026-07-26
Taskcluster v102 no longer supports normal priority - change to lowest

petemoore:
Prior to Taskcluster v102.0.1, task priority `normal` was internally [mapped](https://github.com/taskcluster/taskcluster/blob/05bfba092ae07bff31b232a1c502ee23ba08a46a/services/queue/src/api.js#L687) to `lowest`.

This sets it explicitly, since `normal` priority was removed in https://github.com/taskcluster/taskcluster/pull/8865 which was released in [Taskcluster v102.0.1](https://github.com/taskcluster/taskcluster/releases/tag/v102.0.1).

### DIFF
--- a/infra/taskcluster-hook-data-pipeline.json
+++ b/infra/taskcluster-hook-data-pipeline.json
@@ -44,7 +44,7 @@
       "image": "mozilla/bugbug-spawn-pipeline",
       "maxRunTime": 7200
     },
-    "priority": "normal",
+    "priority": "lowest",
     "provisionerId": "proj-bugbug",
     "retries": 5,
     "routes": [
--- a/infra/taskcluster-hook-check-models-start.json
+++ b/infra/taskcluster-hook-check-models-start.json
@@ -38,7 +38,7 @@
       "image": "mozilla/bugbug-spawn-pipeline",
       "maxRunTime": 7200
     },
-    "priority": "normal",
+    "priority": "lowest",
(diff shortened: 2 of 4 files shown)


infra/taskcluster-hook-data-pipeline.json
```diff
@@ -57,7 +57,7 @@
     "workerType": "batch"
   },
   "triggerSchema": {
-    "additionalProperties": false,
+    "additionalProperties": true,
```
suhaibmujahid: Is this related?

petemoore: Good call. It is fixing a different issue, I could put it in a separate PR. There is a bug in taskcluster which is just about to be fixed in https://github.com/taskcluster/taskcluster/pull/8915. The bug meant that the trigger schemas were not being applied to hooks which were fired by pulse events. This particular trigger schema, specified in this hook, says to block hooks for pulse messages which contain any properties - i.e. would block task creation for all pulse events. I don't think that is intentional, and the only reason is currently works is due to the bug in the hooks service which is about to be fixed. If you want it in a different PR, let me know. Thanks.

petemoore: I've moved this commit to a new PR (#6390), as indeed it was unrelated.

marco-c (approved):

PR #6391, opened 2026-07-27
hackbot-ui: add local stub API with a sample run dataset

sylvestre:
A dependency-free stdlib HTTP server that stands in for hackbot-api during local UI development, so the dashboard can be exercised without Cloud SQL or GCP. Serves GET /agents, GET /runs (honoring agent/status/author/limit/offset) and GET /runs/{id} from a seeded set of runs spanning several agents, statuses, and authors (including unattributed automation runs).

### DIFF
--- a/.claude/settings.local.json.tmp.430234.9e28f84d0377
+++ b/.claude/settings.local.json.tmp.430234.9e28f84d0377
@@ -1,0 +1,41 @@
+{
+  "permissions": {
+    "allow": [
+      "Bash(uv run *)",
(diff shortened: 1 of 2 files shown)


.claude/settings.local.json.tmp.430234.9e28f84d0377
suhaibmujahid, line 1: I expect this was committed by mistake.

services/hackbot-ui/dev/stub_api.py
suhaibmujahid, line 1: It would be nice to move it under `hackbot-api`.

suhaibmujahid (changes requested):

PR #6416, opened 2026-07-28
frontend-triage: link source files as Searchfox permalinks

msujaws:
> **Depends on #6425** (action hooks) — the placeholder expansion is registered as an `ActionHook`, so this is branched on that PR and should land after it. The commit above it is #6425's and will drop out of this diff once it merges.

## What

The triage comment cited source files as bare or backticked repo-relative paths, so a developer reading the bug had to go find the file themselves. Every source-file reference now becomes an inline Markdown link to a **revision-pinned** searchfox.org URL:

> The group line is drawn by the `.tab-group-label` rule in [browser/themes/shared/tabbrowser/tabs.css#1240-1252](https://searchfox.org/firefox-main/rev/d573d849c063353bda3a67541ab219c8a548773a/browser/themes/shared/tabbrowser/tabs.css#1240-1252), which sets `border-inline-start` unconditionally…

Because the revision is pinned, the link keeps pointing at the code the agent actually read after the tree moves on.

## How

The agent never handles the revision. It writes `{{searchfox.permalink}}/<path>#<line>` — the same placeholder convention as `{{actions.<ref>.url}}` — and an action hook on `bugzilla.add_comment` expands the prefix as the comment is recorded. Expanding at *record* rather than apply time means the comment waiting in `summary.json` for review is already clickable, which a prior action's result inherently could not be.

New `libs/hackbot-runtime/hackbot_runtime/searchfox.py` holds the prefix builder, the revision lookup, and the hook.

**A path that is not in the checkout is deliberately not linked.** The hook unwraps the Markdown link to its backticked label and logs why. Models do invent a plausible path now and then, and a permalink that 404s for whoever clicks it is worse than plain text. The check is a heuristic — the checkout and the pinned revision are hours apart, so a file added in between exists locally but not on Searchfox — but it catches the common case.

## Review feedback

Thanks — three of the four are done as suggested, and the placeholder idea in particular deleted a lot of code:

- **Placeholder instead of detection** — done, and it removed the whole heuristic: the top-level-directory list, the extension allowlist, the path regex, the fenced-block skipping. The agent now can't mangle a SHA either.
- **Not really a tool** — moved to `hackbot-runtime`; `agent-tools` is untouched by this PR.
- **Fits as a hook** — done, which also let me drop my `ActionsRecorder.metadata` addition and my edit to `actions/bugzilla.py` entirely.
- **Use the checkout's base commit** — I tried this and measured against it, and I don't think it can work; details below. Happy to be overruled.

### Why not the checkout's base commit

Searchfox only serves `/rev/<sha>` for revisions it has **indexed**, and its index trails the tip of `firefox.git`. Measured while writing this:

| | |
|---|---|
| `firefox.git` `main` tip | `878b64a4` (2026-07-30 13:27Z) |
| Searchfox indexed revision | `d573d849` (2026-07-29 17:08Z) |
| Gap | **83 commits / ~20 hours** |
| `GET /firefox-main/rev/878b64a4/toolkit/moz.build` | **HTTP 500** |

The agent clones `main` with no pinned ref, so its base commit is at or near the tip — i.e. not yet servable. Links pinned to it would 500 for roughly a day and then start working, which is exactly backwards: the window in which they're broken is the window in which someone reads a freshly-posted triage comment.

So the revision is read from the permalink Searchfox itself publishes on a file page (its "Create a revision-specific link" panel item), which is by construction a revision Searchfox can serve. `commit-info/HEAD` — what `searchfox-cli --permalink` uses — turned out not to work either: it only reports a usable revision when HEAD isn't a merge, and HEAD is a Lando merge much of the time. If no revision resolves at all, the placeholder expands to the revision-agnostic `/source/` prefix: still a working link, just unpinned.

## Also: `is_markdown`

Comments were posted without `is_markdown`, so Bugzilla rendered the markup literally — including the existing `*This is an automated analysis result…*` footer, which showed its asterisks. `AddCommentHandler` and the coalesced update+comment path now set it. Only `frontend-triage` and `bug-fix` record `bugzilla.add_comment`, and both author LLM prose, so the blast radius is small.

## Testing

- `libs/hackbot-runtime/tests/test_searchfox.py`: prefix building, revision resolution and every failure mode, placeholder expansion (multiple occurrences, inner whitespace, not clobbering `{{actions.*}}`, prefix-as-backreference), and each degradation path for a missing file, a missing bare placeholder, a shortened label, a directory, and no `repo_root`.
- Full suite green (190 tests on top of #6425), pre-commit clean.
- Ran a seeded local run end to end through the real recorder, hook and footer: all generated links verified HTTP 200 including the `#412` and `#1240-1252` anchor forms, an invented path correctly degraded to plain text, and the trailing plan block kept its bare paths. Rendered it through the console's own `Markdown.tsx` to confirm it displays as links there too.

Not included: bare-URL autolinking, and `AddAttachmentHandler`, whose `comment` is a plain string with no markdown flag on that endpoint.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/agents/frontend-triage/README.md
+++ b/agents/frontend-triage/README.md
@@ -111,8 +111,36 @@ Two things to know before acting on a plan:
   the agent pinned a root cause in the code, never whether the fix works — it
   cannot run anything. Read `high` as "trust the diagnosis, still review the
   patch."
-- **Line numbers are model-asserted and drift.** Trust the files, functions and
+- **Line numbers are model-asserted.** The permalink pins the revision, so a
+  cited line can't drift out from under the link — but the number is still the
+  model's claim about where the problem is. Trust the files, functions and
   selectors it names; confirm exact lines against the source.
+
+Source files in the comment are inline Markdown links to **permalinked**
+Searchfox URLs (`https://searchfox.org/firefox-main/rev/<sha>/<path>#<line>`), so
+they keep pointing at the code the agent read.
+
+The agent never handles the revision: it writes
+`{{searchfox.permalink}}/<path>#<line>`, the same placeholder convention as
+`{{actions.<ref>.url}}`, and an action hook on `bugzilla.add_comment` expands the
+prefix as the comment is recorded — so what lands in `summary.json` for review is
+already clickable. `agent.py` resolves the revision once per run with
+`hackbot_runtime.searchfox.resolve_index_revision()`.
+
+A path the agent names that isn't in the checkout is **not** linked: the hook
+unwraps the Markdown link to its backticked label and logs a warning. Models do
+occasionally invent a plausible path, and a permalink that 404s for whoever
+clicks it is worse than plain text. The check is a heuristic — the checkout and
+the pinned revision are hours apart, so a file added in between exists locally
+but not on Searchfox — but it catches the common case.
+
+That reads the permalink Searchfox publishes on a file page, rather than the
+checkout's base commit, because Searchfox only serves `/rev/<sha>` for revisions
+it has **indexed** — and its index trails the tip of `firefox.git` by hours (83
+commits / ~20h when measured), so a link pinned to the checkout would 500 for
+about a day, exactly the window in which someone reads the comment. If the
+revision can't be resolved at all, the placeholder expands to Searchfox's
+revision-agnostic `/source/` prefix: still a working link, just unpinned.
 
 ## Tuning
 
--- a/agents/frontend-triage/hackbot_agents/frontend_triage/agent.py
+++ b/agents/frontend-triage/hackbot_agents/frontend_triage/agent.py
@@ -32,6 +32,14 @@ from hackbot_runtime import ActionsRecorder, AgentError, HackbotAgentResult
 from hackbot_runtime.actions import ACTIONS_SERVER_NAME
 from hackbot_runtime.actions.claude_sdk import actions_server_for, actions_to_tool_names
 from hackbot_runtime.claude import Reporter
+from hackbot_runtime.searchfox import (
+    PLACEHOLDER as SEARCHFOX_PLACEHOLDER,
+)
+from hackbot_runtime.searchfox import (
+    permalink_hook,
+    permalink_prefix,
+    resolve_index_revision,
+)
 from searchfox import AsyncSearchfoxClient
 
 from .config import (
@@ -67,12 +75,40 @@ class FrontendTriageResult(HackbotAgentResult):
     result: str | None = None
 
 
+# How to cite a source file in the recorded comment. Injected into system.md
+# rather than written there literally, because the template is run through
+# str.format and the placeholder's braces would need doubling twice over in the
+# prompt file; a substituted value passes through untouched.
+_EXAMPLE_PATH = "browser/components/tabbrowser/content/tabgroup.js"
+SEARCHFOX_LINKS_PROMPT = (
+    "Every source file you reference in your Bugzilla comment must be an inline "
+    f"Markdown link built from the `{SEARCHFOX_PLACEHOLDER}` placeholder, which "
+    "is expanded into a revision-pinned Searchfox URL when your comment is "
+    "recorded:\n\n"
+    f"    [{_EXAMPLE_PATH}]({SEARCHFOX_PLACEHOLDER}/{_EXAMPLE_PATH})\n\n"
+    "- Write the placeholder **literally**. Do not put a revision, `tip` or "
+    "`HEAD` in it, and do not write a searchfox.org URL yourself — you do not "
+    "know which revision is being linked.\n"
+    "- After it, give the repo-relative path, plus a line anchor when you know "
+    "the line: `#1234`, or `#1234-1250` for a range. No `L` prefix.\n"
+    "- Use the path as the link text, without backticks — backticked text does "
+    "not render as a link.\n"
+    "- Leave the paths in the trailing ```json plan block as **bare paths** — "
+    "that block is parsed by a downstream tool, and a link there would corrupt "
+    "it.\n"
+    "- Only cite a path you have confirmed exists (Read/Glob it, or take it from "
+    "a Searchfox result). A path that is not in the checkout is stripped back to "
+    "plain text rather than linked, so guessing costs you the link."
+)
+
+
 def load_system_prompt(rules_dir: Path, extra: str) -> str:
     tmpl = (HERE / "prompts" / "system.md").read_text()
 
     return tmpl.format(
         rules_dir=str(rules_dir.resolve()),
         extra_instructions=extra or "(none)",
+        searchfox_links=SEARCHFOX_LINKS_PROMPT,
     )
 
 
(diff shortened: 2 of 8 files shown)


crythms (dismissed):
👍

libs/agent-tools/agent_tools/searchfox_links.py
```diff
@@ -0,0 +1,239 @@
...
+async def resolve_index_revision(*, client=None, timeout: float = 10.0) -> str | None:
+    """Return the git SHA Searchfox's index is pinned to, or ``None``.
+
+    Reads the permalink Searchfox publishes on a file page, so the revision is
+    by construction one Searchfox can serve.
+
+    Never raises: a caller that cannot get a revision should degrade to bare
+    paths rather than fail its run. ``client`` accepts an httpx-like async
+    client for testing; when omitted, ``httpx`` is imported lazily.
+    """
```
suhaibmujahid: Can we instead relay on the base commit of the checked out repo?

libs/agent-tools/agent_tools/searchfox_links.py
```diff
@@ -0,0 +1,239 @@
...
+# alternative, the commit-info/HEAD endpoint, only reports a usable revision
+# when HEAD is not a merge — and HEAD is a Lando merge much of the time.)
+_PROBE_PATH = "toolkit/moz.build"
+_PROBE_URL = f"{_BASE_URL}/source/{_PROBE_PATH}"
+_PANEL_PERMALINK_RE = re.compile(
+    rf'<a id="panel-permalink" href="/{TREE}/rev/(?P<rev>[0-9a-f]{{40}})/'
+)
+
+# Top-level mozilla-central directories. Requiring one of these as a prefix is
+# what keeps prose ("about:preferences", "Bug 1234") out of the linkifier.
```
suhaibmujahid: What of in the prompt we ask the agent to add a template prefix that we could just replace? So we do not need to figure out if this should be a Searchfox link or not.

Example: `{{searchfox.permalink}}/browser/components/tabbrowser/content/tabgroup.js`

This is consistent with what we have for referencing other actions (i.e., `{{actions.<ref>.url}}`)

libs/agent-tools/agent_tools/searchfox_links.py
suhaibmujahid, line 1: This is not really a tool; we could move it to the place where we will use it, e.g., hackbot-runtime.

libs/hackbot-runtime/hackbot_runtime/actions/bugzilla.py
```diff
@@ -95,6 +96,14 @@ async def add_comment(
...
     summary for human review — does not post to Bugzilla.
     """
+    # Safety net for agents that pinned a searchfox revision: any source path
+    # the agent left bare becomes a permalink. A no-op when the run has no
+    # revision (see ActionsRecorder.metadata).
+    text = linkify_source_paths(
+        text,
+        recorder.metadata.get("searchfox_rev"),
+        repo_root=recorder.metadata.get("source_repo"),
+    )
```
suhaibmujahid: This could fit as hook, see https://github.com/mozilla/bugbug/pull/6425

suhaibmujahid (changes requested):

suhaibmujahid:
I think this will be a nice feature that would make it much easier to follow up on triage results. Thank you for working on this!

libs/agent-tools/agent_tools/searchfox_links.py
msujaws: Done — this was a much better idea than what I had. The agent now writes `{{searchfox.permalink}}/browser/components/tabbrowser/content/tabgroup.js#412` and a hook expands the prefix, which deleted the entire heuristic: the top-level-directory list, the extension allowlist, the path regex and the fenced-block skipping all went away. It also means the agent never touches the SHA, so it cannot mangle it.

One deliberate addition: the hook still checks the path against the checkout, and if it is not there it unwraps the link to its backticked label rather than emitting a link. Without that, an invented path becomes a permalink that 404s for whoever clicks it — I hit exactly that while previewing a seeded run.

libs/agent-tools/agent_tools/searchfox_links.py
msujaws: Moved to `libs/hackbot-runtime/hackbot_runtime/searchfox.py`. `agent-tools` is untouched by this PR now.

libs/hackbot-runtime/hackbot_runtime/actions/bugzilla.py
msujaws: Done — thanks for the pointer, this is a much better fit. It is now an `ActionHook` registered for `bugzilla.add_comment`, which let me drop both the `ActionsRecorder.metadata` field I had added and my edit to `actions/bugzilla.py`.

This PR is branched on #6425 as a result, so it should land after it (the hooks commit will drop out of this diff when that merges).

libs/agent-tools/agent_tools/searchfox_links.py
msujaws: I tried this and I do not think it can work, though I would happily be overruled.

Searchfox only serves `/rev/<sha>` for revisions it has **indexed**, and its index trails the tip of `firefox.git`. Measured today:

| | |
|---|---|
| `firefox.git` `main` tip | `878b64a4` (2026-07-30 13:27Z) |
| Searchfox indexed revision | `d573d849` (2026-07-29 17:08Z) |
| Gap | **83 commits / ~20 hours** |
| `GET /firefox-main/rev/878b64a4/toolkit/moz.build` | **HTTP 500** |

The agent clones `main` with no pinned ref, so its base commit sits at or near the tip and is not yet servable. A link pinned to it would 500 for roughly a day and then start working — backwards, since the broken window is exactly when someone reads a freshly-posted triage comment.

So the revision is read from the permalink Searchfox itself publishes on a file page (its "Create a revision-specific link" panel item), which is by construction one it can serve. For what it is worth, `commit-info/HEAD` — what `searchfox-cli --permalink` reads — does not work either: it only reports a usable revision when HEAD is not a merge, and HEAD is a Lando merge much of the time.

The tradeoff is that the pinned revision is ~a day older than the bytes the agent read, so a line anchor can be slightly off in a file touched in that window. I think that beats a link that does not resolve, but if you would rather have exactness and accept the delay, the change is a one-liner.

suhaibmujahid (approved):
LGTM! Thank you! 

I landed #6425. ~~Could you please rebase?~~

Edit: I enabled the "update branch" button for this repo to make things smother :)

libs/agent-tools/agent_tools/searchfox_links.py
suhaibmujahid: I think this might be tricky since it will point to a version that is not what the agent saw, and it defeats the whole purpose of a permalink (pointing exactly to what we saw). We could try mitigating that on the Searchfox side (e.g., redirecting or linking to GitHub if the revision is not there yet).

BTW, as now, I see `878b64a4` on both, GitHub main and Searchfox.

PR #6430, opened 2026-07-30
Preserve Phabricator diff author on patch updates

### LINKED ISSUE #6411: Hackbot should not change the commit author when updating a patch
The author info in the new diff should match what we have already on the previous diff.

ayoubdiourin7:
## Summary

Preserve the original patch author when Hackbot updates an existing Phabricator revision.

Before this change, updating a patch could replace the reconstructed commit author with Hackbot's synthetic author. Now, updates keep the author name and email from the previous diff instead.

## Tests

Added a test that covers updating an existing Phabricator revision and verifies the new diff keeps the previous author metadata.

Resolves #6411

### DIFF
--- a/libs/hackbot-runtime/hackbot_runtime/actions/handlers/phabricator_handler.py
+++ b/libs/hackbot-runtime/hackbot_runtime/actions/handlers/phabricator_handler.py
@@ -127,24 +127,23 @@ async def _diff_commits(diff_id: Any) -> list[dict]:
     return commits if isinstance(commits, list) else []
 
 
-def _preserve_local_commit_authors(
-    local_commits: dict, previous_commits: list[dict]
-) -> None:
-    """Copy the previous diff's commit author identity onto new commit metadata."""
+def _local_commit_author_fields(previous_commits: list[dict]) -> dict[str, str]:
+    """Return the previous diff's commit author identity."""
     if not previous_commits:
-        return
+        return {}
 
     previous_author = previous_commits[-1].get("author") or {}
-    author_fields = {}
-    if previous_author.get("name"):
-        author_fields["author"] = previous_author["name"]
-    if previous_author.get("email"):
-        author_fields["authorEmail"] = previous_author["email"]
-    if not author_fields:
-        return
+    if not previous_author.get("name") or not previous_author.get("email"):
+        log.warning(
+            "Could not preserve local commit author: previous diff author metadata "
+            "is incomplete"
+        )
+        return {}
 
-    for commit_info in local_commits.values():
-        commit_info.update(author_fields)
+    return {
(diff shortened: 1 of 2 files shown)


libs/hackbot-runtime/hackbot_runtime/actions/handlers/phabricator_handler.py
```diff
@@ -265,6 +307,20 @@ async def apply(self, params: dict[str, Any], ctx: ApplyContext) -> ActionResult
...
+                _preserve_local_commit_authors(
+                    submission["local_commits"],
+                    await _diff_commits(previous_diff_id),
+                )
+            except Exception:
+                log.warning(
+                    "Could not preserve local commit author from previous diff %s",
+                    previous_diff_id,
+                    exc_info=True,
+                )
```
suhaibmujahid: I would avoid nested try-except.

libs/hackbot-runtime/hackbot_runtime/actions/handlers/phabricator_handler.py
```diff
@@ -108,6 +108,45 @@ async def _revision_fields(revision_id: int) -> dict:
...
+    if not data:
+        return []
+
+    commits = data[0].get("attachments", {}).get("commits", {}).get("commits", [])
+    return commits if isinstance(commits, list) else []
+
+
+def _preserve_local_commit_authors(
+    local_commits: dict, previous_commits: list[dict]
+) -> None:
```
suhaibmujahid: I would prefer returning the final values as a dict instead of mutating in place. That would be better for debugging.

libs/hackbot-runtime/hackbot_runtime/actions/handlers/phabricator_handler.py
```diff
@@ -108,6 +108,45 @@ async def _revision_fields(revision_id: int) -> dict:
...
+        return
+
+    previous_author = previous_commits[-1].get("author") or {}
+    author_fields = {}
+    if previous_author.get("name"):
+        author_fields["author"] = previous_author["name"]
+    if previous_author.get("email"):
+        author_fields["authorEmail"] = previous_author["email"]
+    if not author_fields:
+        return
```
suhaibmujahid: We should avoid partial info updates. It is either all or nothing. 

I do not think that would happen, but since this is a definitive measure, I would log a warning for that.

suhaibmujahid (changes requested):
Thank you! Please see my comments. 

BTW. you could test it live with the dev instance (i.e., https://phabricator-dev.allizom.org)

suhaibmujahid (approved):
LGTM! Thank you!

PR #6435, opened 2026-07-30
Make code review agent aware of the current date

### LINKED ISSUE #6376: The code review agent is not aware of the current date
Example https://phabricator.services.mozilla.com/D310592#inline-1686152

ayoubdiourin7:
Resolves  #6376

This PR makes the code review agent include the current UTC date in its prompt on every review call.

The change applies to both review flows:
the main code review prompt
the patch-scope review prompt

### DIFF
--- a/bugbug/tools/code_review/prompts.py
+++ b/bugbug/tools/code_review/prompts.py
@@ -60,8 +60,6 @@ Do not write comments that:
 
 PATCH_SCOPE_PROMPT = """You are an expert {target_software} engineer assessing whether a pull request is too large or unfocused to be reviewed well.
 
-Current date: {current_date}
-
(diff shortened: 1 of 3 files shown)


bugbug/tools/code_review/prompts.py
```diff
@@ -60,6 +60,8 @@
 
 PATCH_SCOPE_PROMPT = """You are an expert {target_software} engineer assessing whether a pull request is too large or unfocused to be reviewed well.
 
+Current date: {current_date}
+
```
suhaibmujahid: I do not think we need it twice, in the system problem and the message as well.

suhaibmujahid (changes requested):

tests/test_code_review.py
suhaibmujahid, line 1: I do not think this needs tests.

suhaibmujahid (approved):
Thank you!

PR #6439, opened 2026-07-30
Auto-apply high-confidence frontend-triage results

msujaws:
`frontend-triage` records its root-cause analysis and fix plan as Bugzilla actions, but
nothing applied them, so every triage needed a person to click Apply in the hackbot UI.
This closes that loop for the results the agent is confident about.

## How the decision is made

A run reports `findings.auto_apply` from the two values only it has: `confidence: high`,
and `actionable` not false. Both arrive in the trailing JSON block *after* every action
is already recorded, which is why this can't be a record-time check and isn't
hackbot-api's call to make. `may_apply_unattended()` in `agent.py` is that policy in
full.

It fails closed, since a plan that didn't parse has no `confidence`. And because
`confidence` is read out of the agent's free-form JSON, `parse_confidence` normalizes
case and whitespace, so `"High"` doesn't quietly read as not-high.

Medium and low are unchanged from today: still recorded, still visible in the UI, still
appliable by hand. So are all the other agents — `test_other_agents_do_not_auto_apply`
asserts exactly who auto-applies.

## What the agent may record

Confidence gates how sure the agent is, not how far its actions can go. An action's
params are model output however sure it is, and the run reads bug comments nobody
controls. Two hooks in `hooks.py` refuse, at record time:

- `add_comment_hook` — one comment, public, on the bug being triaged.
- `update_bug_hook` — one field change, add-only, on the bug being triaged, and only
  `keywords`/`severity` with values from `TRIAGE_SEVERITIES`/`TRIAGE_KEYWORDS` in
  `config.py`.

They refuse at record time rather than holding at apply time because `ActionsRecorder`
runs hooks before appending: the reason goes back to the agent as a tool error it can
correct in the same run, and the out-of-bounds action never reaches `summary.json` at
all. The vocabulary lives in the agent's config, next to the rules that would widen it —
`TRIAGE_SEVERITIES` is exactly the `S1`–`S4` that `rules/severity-assessment.md`
defines.

The action *type* needs no check: `ENABLED_ACTION_TYPES` already decides which tools the
actions server exposes, so this agent has no way to record a `bugzilla.create_bug` or a
Phabricator action in the first place.

## The hackbot-api side

`AgentSpec.auto_apply_requires_consent` makes `auto_apply_actions` conditional on
`findings.auto_apply`, and fails closed when a run reports nothing. That's the whole
service-side change; no policy about what an agent may write lives there. That belongs
with the agent, which knows its own rules and can explain a rejection to the model
rather than silently parking a run for a human.

## What the prompt now tells the agent

`confidence` used to be advisory metadata a human read off the run. It now decides
whether a comment reaches a real bug unreviewed, and the agent was being asked to
self-report it without being told what it does, which invites grading on a curve. The
system prompt now says so, and states what the action tools will accept, so the model
learns the shape before a hook has to reject it.

The hooks are now the only thing bounding what an unattended run writes. The apply step
dispatches whatever it finds in `summary.json` against the runtime's global handler
registry, which is far wider than the tools this agent was given. That's a real change
in where the boundary sits.

`bug-fix` and `test-repair` auto-apply everything they record and are untouched here.
Bounding them is a decision about those agents, so `test_which_agents_auto_apply_without_asking_for_consent`
records the gap rather than closing it, and fails if a new agent opts in without
bounding itself.

## Testing

- `agents/frontend-triage`: **27 passed** — a new suite, since the agent had none. Two
  files: `tests/test_hooks.py` covers what the hooks refuse, `tests/test_plan.py` covers
  `may_apply_unattended` and the confidence normalization that feeds it.
- `libs/hackbot-runtime`: **203 passed**.
- `services/hackbot-api`: **3 failed, 95 passed, 8 errors**. That failure set is
  pre-existing, not introduced here — the parent commit `59137c9b` alone gives **3
  failed, 84 passed, 8 errors**, the same 3 and the same 8, verified by running its suite
  unmodified in a clean worktree. The 3 are `comment.is_markdown` missing from
  expectations in `test_actions_applier.py`; the 8 are `fixture 'client' not found` in
  `test_list_runs_api.py` / `test_create_run_api.py`.

Note that neither `services/hackbot-api` nor the agent suites run in Taskcluster today —
`.taskcluster.yml` covers `tests/`, `http_service/`, `libs/agent-tools` and
`libs/hackbot-runtime` only. CI won't exercise any of the above.

## Deploying

No new configuration. `BUGZILLA_API_KEY` is already set on the service and working
(`hackbot@mozilla.tld` has been applying comments), so nothing new is needed for the
apply path.

`bugzilla.update_bug` needs `editbugs` on the apply account. The apply step coalesces a
same-bug field change with the nearest comment into one PUT, so losing that privilege
would take the analysis comment down with the rejected field change.

## Follow-up

- mozilla/bugbot will start triggering these runs automatically for newly filed
  `Firefox :: New Tab Page` bugs from staff. **This should land first**, so a triggered
  run auto-applies.
- Slack reporting is out of this PR; #6523 has since landed and is the right vehicle.

### DIFF
--- a/agents/frontend-triage/README.md
+++ b/agents/frontend-triage/README.md
@@ -76,8 +76,11 @@ Each run writes to `~/hackbot/artifacts/<run_id>/`:
 
 - **`summary.json`** — `findings` holds the structured plan (`root_cause`,
   `proposed_fix`, `target_files`, `confidence`) plus the executor handoff fields
-  `actionable`, `regressor_node` and `relevant_tests`. `actions` holds the
-  **recorded** Bugzilla comment — written here for review, not posted.
+  `actionable`, `regressor_node` and `relevant_tests`, plus `auto_apply` — the
+  run's own verdict on whether it may be posted without review. `actions` holds
+  the **recorded** Bugzilla comment (and, at high confidence, possibly a field
+  change). Recording is not posting, but see below: an `auto_apply` run's actions
+  do reach the bug unattended.
 - **`logs/agent.log`** — the streamed reasoning and every tool call, and the only
   record of which model actually ran.
 - **No `changes/` directory.** Its absence confirms the run stayed read-only.
@@ -99,11 +102,33 @@ Two caveats before acting on a plan:
 `bugzilla.add_comment` and `bugzilla.update_bug`, but those come from an
 in-process actions server that appends to `summary.json` and makes no network
 calls. The only Bugzilla access the agent has is through the broker sidecar,
-which exposes five read tools and holds the API key. Applying a recorded action
-is a separate, human step from the Hackbot UI — unlike `bug-fix`, this agent
-does not set `auto_apply_actions`.
+which exposes five read tools and holds the API key.
 
-Two hooks shape the comment as it is recorded:
+**Afterwards, though, a confident run posts itself.** When the run reports
+`confidence: high` and does not report `actionable: false`, it sets
+`findings.auto_apply`, and hackbot-api applies the recorded actions to the real
+bug with nobody in between. `may_apply_unattended()` in `agent.py` is that
+decision in full. Medium and low are held for a person to apply from the Hackbot
+UI, as before.
+
+Judgement and reach are bounded separately, because an action's params are model
+output no matter how sure the agent is — and the agent spends the run reading bug
+comments nobody controls. Two hooks in `hooks.py` refuse an action outright as it
+is recorded, and they are the only thing bounding what an unattended run writes:
+hackbot-api applies whatever it finds in `summary.json`, dispatching it against a
+handler registry far wider than the tools this agent was given.
+
+- `add_comment_hook` — one comment, public, on the bug being triaged.
+- `update_bug_hook` — one field change, add-only, on the bug being triaged, and
+  only `keywords`/`severity` with values from `TRIAGE_SEVERITIES` /
+  `TRIAGE_KEYWORDS` in `config.py`. Widen those sets alongside the rule that needs
+  the new value.
+
+A refusal reaches the agent as a tool error it can correct in the same run, and the
+action never lands in `summary.json`. The action _type_ needs no check:
+`ENABLED_ACTION_TYPES` decides which tools the actions server exposes at all.
+
+Two further hooks shape the comment text as it is recorded:
 
 - `permalink_hook` expands `{{searchfox.permalink}}/<path>#<line>` into
   `https://searchfox.org/firefox-main/rev/<sha>/…`, so every source reference
--- a/services/hackbot-api/app/actions_applier.py
+++ b/services/hackbot-api/app/actions_applier.py
@@ -2,12 +2,12 @@
 
 On run completion the recorded actions from `summary["actions"]` are always
 upserted as `run_actions` rows (one per entry) so they're visible and
-manageable in the UI. Whether they're then applied *automatically* depends on
-the agent's `auto_apply_actions` opt-in (see `app/agents.py`); either way they
-can be applied on demand (manual apply-all from the UI). Application runs each
-pending row through the handler registry in `hackbot_runtime.actions.handlers`
-and is idempotent per action — an already-`applied` row is never re-applied, so
-Pub/Sub retries and repeated manual applies are safe.
+manageable in the UI. Whether they're then applied *automatically* is decided by
+`_auto_apply_blocker` (see `app/agents.py`); either way they
+can be applied on demand (manual apply-all from the UI). Application runs each pending
+row through the handler registry in `hackbot_runtime.actions.handlers` and is
+idempotent per action — an already-`applied` row is never re-applied, so Pub/Sub
+retries and repeated manual applies are safe.
 """
 
 from __future__ import annotations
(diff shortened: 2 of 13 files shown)


evgenyrp:
I'm adding Slack support here, so I hope you'll be able to use it instead of emails when it's landed https://github.com/mozilla/bugbug/pull/6523

msujaws:
> I'm adding Slack support here, so I hope you'll be able to use it instead of emails when it's landed #6523

Thanks! I have removed the Slack notification commit from this PR since your approach will be much better.

services/hackbot-api/app/auto_apply.py
```diff
@@ -0,0 +1,111 @@
...
+    # An int or a plain run of digits, nothing looser: the handler interpolates this
+    # raw value into the REST path, whereas `int()` would also accept `"2_014_702"`,
+    # signs, whitespace and non-ASCII digits — validating a different string than the
+    # one sent.
+    if isinstance(value, bool):
+        return False
+    return isinstance(value, int) or (isinstance(value, str) and value.isdigit())
+
+
+def _field_change(field: str, value: Any) -> str | None:
```
crythms: Nice work on this!

One thing that looks left open is the values themselves. `_field_change` accepts any non-empty string for severity and any keyword for keywords. Doesn't seem to fire today for triage but may be worth bounding both to known lists in case that changes.

- for severity https://bugzilla.mozilla.org/rest/field/bug/bug_severity has the legal values
- for keywords  https://bugzilla.mozilla.org/describekeywords.cgi

crythms (approved):

libs/hackbot-runtime/hackbot_runtime/actions/bugzilla.py
suhaibmujahid, line 1: I think the instruction for the Bugzilla action would better fit in the agent's system prompt.

agents/frontend-triage/README.md
```diff
@@ -57,78 +77,87 @@ Three bugs that exercise the classes this agent handles:
...
+Three things to know before acting on a plan:
 
+- **`confidence` decides whether the actions are posted.** A `high`-confidence
+  run has its recorded actions applied to the real bug automatically; `medium`
+  and `low` are held for a human to apply from the hackbot UI. Only the fields
+  the agent is trusted with (`keywords`, `severity`) can go up unattended, and
+  only against the bug the run was asked about — anything else holds the whole
+  run for review. See `auto_apply_confidence` in
+  `services/hackbot-api/app/agents.py` and `frontend_triage_guard` in
+  `services/hackbot-api/app/auto_apply.py`.
```
suhaibmujahid: The agent doesn’t have access to that. And even if it has access, it might not be the best way.

services/hackbot-api/app/actions_applier.py
```diff
@@ -86,6 +86,58 @@ def _sub(match: re.Match) -> str:
...
 
 
+def _reported_confidence(run: Run) -> Confidence | None:
+    """The run's self-reported confidence, or None if it didn't report a usable one."""
+    return parse_confidence(
+        ((run.summary or {}).get("findings") or {}).get("confidence")
+    )
+
+
+def _should_auto_apply(
```
suhaibmujahid: Why not do that in the agent itself?

services/hackbot-api/app/auto_apply.py
suhaibmujahid: We could do this validation in an action hook for now. I'm planing to do more generic validation through pydantic. Once that in place, we could drop the or simplify the hook if needed.

services/hackbot-api/app/actions_applier.py
msujaws: Moved this to the agent, which will work better as each agent might want to determine its own criteria

services/hackbot-api/app/auto_apply.py
msujaws: Moved to the action hook

msujaws: 6b7680fa landed severity assessment and this now is limited to S1-S4. There's also six keywords that are seeded for now that are specific to frontend.

services/hackbot-api/app/agents.py
```diff
@@ -27,6 +27,12 @@ class AgentSpec:
     # succeeds. Off by default: actions are still recorded and can always be
     # applied manually from the UI; only opted-in agents auto-apply.
     auto_apply_actions: bool = False
+    # Whether auto-apply additionally needs the run's own say-so: `findings.auto_apply`
+    # must be True, or the actions are recorded and held. Set this for agents that
+    # judge their results one run at a time, since only the agent knows how sure it
+    # was; this is where that verdict is honored. Fails closed, so a run that reports
+    # no verdict never qualifies.
+    auto_apply_requires_consent: bool = False
```
suhaibmujahid, line 35: Instead of having two options, we could replace them with something like `action_apply_mode`. We need a clearer design for this config, but we can go ahead and merge with this PR for now and improve the config later. We do not want to delay this further.

suhaibmujahid (approved):
Thank you! LGTM!

services/hackbot-api/app/agents.py
msujaws: Thanks for filing https://github.com/mozilla/bugbug/issues/6575. I'll resolve this now. I can pick up https://github.com/mozilla/bugbug/issues/6575 later this week if time allows.

PR #6440, opened 2026-07-30
hackbot-ui: link to the bug from the run detail view

sylvestre:


services/hackbot-ui/components/RunDetail.tsx
```diff
@@ -173,6 +177,20 @@ export function RunDetail({ runId }: { runId: string }) {
...
+          {bugId !== null && (
+            <>
+              <dt>Bug</dt>
+              <dd>
+                <a
+                  href={`${BUGZILLA_URL}${bugId}`}
+                  target="_blank"
+                  rel="noreferrer"
+                >
+                  Bug {bugId}
```
suhaibmujahid, line 189: It would be redundant 
```suggestion
                 {bugId}
```

msujaws:
Can this please also show the link while it is in the pending state? I haven't tested this to see if it does this already or not, just wanted to ask as this has come up in feedback I've received also. Thanks for the PR!

suhaibmujahid:
> Can this please also show the link while it is in the pending state? I haven't tested this to see if it does this already or not, just wanted to ask as this has come up in feedback I've received also. Thanks for the PR!

@msujaws This is for when it is pending. When the result is ready, we already have it shown in the finding in the friendly view.

PR #6444, opened 2026-07-31
Restrict Hackbot triggers to authorized user with editbugs permissions

### LINKED ISSUE #6423: Control who can trigger Hackbot from a Phabricator comment
Since Hackbot would respond to any comment with the `@hackbot` tag, we need to make sure only users who are authorized to do edit the revision can instruct the agent and influence its actions.

ayoubdiourin7:
## Changes

- Check the author of each `@hackbot` comment.
- Process comments only from users authorized to edit revisions.
- Ignore comments from unauthorized users before creating a Hackbot run.
- Add test coverage for authorized, unauthorized

Resolves #6423

### DIFF
--- a/libs/phabricator-client/phabricator_client/client.py
+++ b/libs/phabricator-client/phabricator_client/client.py
@@ -122,11 +122,8 @@ class PhabricatorClient:
             constraints={"phids": [project_phid]},
             attachments={"members": True},
         )
-        data = result.get("data") or []
-        if not data:
-            return frozenset()
-        members = data[0].get("attachments", {}).get("members", {}).get("members", [])
-        return frozenset(member["phid"] for member in members if member.get("phid"))
+        members = result["data"][0]["attachments"]["members"]["members"]
+        return frozenset(member["phid"] for member in members)
 
     async def query_latest_diff(self, revision_id: int) -> PhabricatorDiff | None:
         """The most recent diff for a revision, or ``None`` if it has none.
--- a/libs/phabricator-client/phabricator_client/client.py
+++ b/libs/phabricator-client/phabricator_client/client.py
@@ -115,6 +115,19 @@ class PhabricatorClient:
             if user.get("phid")
         }
 
+    async def get_project_members(self, project_phid: str) -> frozenset[str]:
+        """Return the user PHIDs belonging to a Phabricator project."""
+        result = await self.conduit_request(
+            "project.search",
+            constraints={"phids": [project_phid]},
+            attachments={"members": True},
+        )
+        data = result.get("data") or []
+        if not data:
+            return frozenset()
+        members = data[0].get("attachments", {}).get("members", {}).get("members", [])
+        return frozenset(member["phid"] for member in members if member.get("phid"))
+
     async def query_latest_diff(self, revision_id: int) -> PhabricatorDiff | None:
         """The most recent diff for a revision, or ``None`` if it has none.
 
--- a/services/hackbot-api/app/config.py
+++ b/services/hackbot-api/app/config.py
@@ -45,7 +45,7 @@ class Settings(BaseSettings):
     # Phabricator Conduit connection config, embedded as a nested model and
     # populated in this single settings parse from PHABRICATOR_URL /
     # PHABRICATOR_API_KEY / PHABRICATOR_TIMEOUT_SECONDS (see env_nested_delimiter
-    # below). Injected directly as PhabricatorClient(settings.phabricator).
+    # below). Constructed once at application startup and injected as a dependency.
     # Required, so a missing/invalid api_key fails at startup.
     phabricator: PhabricatorSettings
 
--- a/services/hackbot-api/app/config.py
+++ b/services/hackbot-api/app/config.py
@@ -42,11 +42,6 @@ class Settings(BaseSettings):
     # API auth
     external_api_key: str = ""
 
-    # Phabricator Conduit connection config, embedded as a nested model and
-    # populated in this single settings parse from PHABRICATOR_URL /
-    # PHABRICATOR_API_KEY / PHABRICATOR_TIMEOUT_SECONDS (see env_nested_delimiter
-    # below). Constructed once at application startup and injected as a dependency.
-    # Required, so a missing/invalid api_key fails at startup.
     phabricator: PhabricatorSettings
 
     # Inbound webhook receiver config, embedded as a nested model and populated
--- a/services/hackbot-api/app/main.py
+++ b/services/hackbot-api/app/main.py
@@ -3,10 +3,15 @@ from contextlib import asynccontextmanager
 
 import sentry_sdk
 from fastapi import FastAPI
+from phabricator_client import PhabricatorClient
 
 from app import __version__
 from app.config import settings
 from app.database.connection import close_db, init_db
+from app.phabricator_authorization import (
+    AUTHORIZED_GROUP_PHID,
+    PhabricatorAuthorizer,
+)
 from app.routers import events_router, runs_router, webhooks_router
 
 if settings.sentry_dsn:
@@ -37,6 +42,11 @@ app = FastAPI(
     description="Agent orchestration platform that runs agents as Cloud Run Jobs",
     version=__version__,
     lifespan=lifespan,
+)
+app.state.phabricator_client = PhabricatorClient(settings.phabricator)
+app.state.phabricator_authorizer = PhabricatorAuthorizer(
+    app.state.phabricator_client,
+    AUTHORIZED_GROUP_PHID,
 )
 
 app.include_router(runs_router)
--- a/services/hackbot-api/app/main.py
+++ b/services/hackbot-api/app/main.py
@@ -3,15 +3,10 @@ from contextlib import asynccontextmanager
 
 import sentry_sdk
 from fastapi import FastAPI
-from phabricator_client import PhabricatorClient
 
 from app import __version__
 from app.config import settings
 from app.database.connection import close_db, init_db
-from app.phabricator_authorization import (
-    AUTHORIZED_GROUP_PHID,
-    PhabricatorAuthorizer,
-)
 from app.routers import events_router, runs_router, webhooks_router
 
 if settings.sentry_dsn:
@@ -42,11 +37,6 @@ app = FastAPI(
     description="Agent orchestration platform that runs agents as Cloud Run Jobs",
     version=__version__,
     lifespan=lifespan,
-)
-app.state.phabricator_client = PhabricatorClient(settings.phabricator)
-app.state.phabricator_authorizer = PhabricatorAuthorizer(
-    app.state.phabricator_client,
-    AUTHORIZED_GROUP_PHID,
 )
 
 app.include_router(runs_router)
--- a/services/hackbot-api/app/phabricator_authorization.py
+++ b/services/hackbot-api/app/phabricator_authorization.py
@@ -1,0 +1,69 @@
+"""Authorization checks for Phabricator webhook authors."""
+
+from __future__ import annotations
+
+import asyncio
+import time
+from typing import TYPE_CHECKING
+
+from cachetools import TTLCache
+
+if TYPE_CHECKING:
+    from phabricator_client import PhabricatorClient
+
+
+class PhabricatorAuthorizer:
+    """Cache-backed authorization checks against a Phabricator project."""
+
+    def __init__(
+        self,
+        group_phid: str,
+        *,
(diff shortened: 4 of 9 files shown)


services/hackbot-api/app/phabricator_webhook.py
```diff
@@ -57,7 +65,12 @@ def find_hackbot_mentions(
         for comment in transaction.get("comments") or []:
             raw = (comment.get("content") or {}).get("raw") or ""
             if token in raw:
-                matches.append(raw)
+                matches.append(
+                    HackbotMention(
+                        raw=raw,
+                        author_phid=transaction.get("authorPHID", ""),
```
suhaibmujahid: I do not think there is a comment transaction without `authorPHID`, if that is a possibility, we should skip the transaction altogether.

services/hackbot-api/app/phabricator_webhook.py
```diff
@@ -111,20 +124,35 @@ async def detect_mention_and_revision(
...
+    if not mentions:
         log.warning(
             "No %s mention found in triggering transactions %s on %s",
             webhook.mention_token,
             triggering_phids,
             object_phid,
         )
         return None
+
+    authorized_members = await client.get_project_members(EDITBUGS_GROUP_PHID)
```
suhaibmujahid: We might want to implement TTL cache, and re-fetch if a user is not exist in the list. So we optimize for the happy path. We could do it now or file a follow up issue and we do it later.

services/hackbot-api/app/phabricator_webhook.py
```diff
@@ -111,20 +124,35 @@ async def detect_mention_and_revision(
...
+            comments.append(mention.raw)
+        else:
+            log.warning(
+                "Ignoring %s mention from non-editbugs user %s on %s",
+                webhook.mention_token,
+                mention.author_phid or "<missing PHID>",
+                object_phid,
+            )
+    if not comments:
+        return None
```
suhaibmujahid, line 156: I would move this before `No %s mention found in triggering`, so we have only one check for the comments.

suhaibmujahid (changes requested):

services/hackbot-api/app/phabricator_webhook.py
ayoubdiourin7: updated in [ef9701c](https://github.com/mozilla/bugbug/pull/6444/commits/ef9701cdd7247d9acd80a5e96b61b6b166e6815f)

services/hackbot-api/app/phabricator_webhook.py
ayoubdiourin7: I added a TTL cache for it in  [8ae7728](https://github.com/mozilla/bugbug/pull/6444/commits/8ae772898e9a9e3cb737d423897c8f0eb3668cc5)

libs/phabricator-client/phabricator_client/client.py
```diff
@@ -115,6 +115,19 @@ async def search_users(self, phids: list[str]) -> dict[str, dict]:
...
+        """Return the user PHIDs belonging to a Phabricator project."""
+        result = await self.conduit_request(
+            "project.search",
+            constraints={"phids": [project_phid]},
+            attachments={"members": True},
+        )
+        data = result.get("data") or []
+        if not data:
+            return frozenset()
+        members = data[0].get("attachments", {}).get("members", {}).get("members", [])
```
suhaibmujahid: This pattern can easily introduce silent bugs. I would prefer failing load, so we could fix it. We have other places where we do this, but I would like to reactivate that to remove them, not introduce more.

If it a value that I expect to always be there, I would access it directly, without `git()`. If it could be missed, I would log that if it something that would need attention.

services/hackbot-api/app/phabricator_webhook.py
```diff
@@ -20,6 +25,22 @@
...
+EDITBUGS_GROUP_PHID = "PHID-PROJ-njo5uuqyyq3oijbkhy55"
+_MEMBERSHIP_CACHE_TTL_SECONDS = 60
+_MISSING_MEMBER_REFRESH_COOLDOWN_SECONDS = 30
+
+_editbugs_members_cache: TTLCache[str, frozenset[str]] = TTLCache(
+    maxsize=1,
+    ttl=_MEMBERSHIP_CACHE_TTL_SECONDS,
+)
+_editbugs_members_lock = asyncio.Lock()
+_last_editbugs_members_refresh = 0.0
```
suhaibmujahid: This can grow to be a bit ugly. I would encapsulate this in a class that exposes a simple method to check if the user is authorized or not.

ayoubdiourin7: Fixed  in [c755905](https://github.com/mozilla/bugbug/pull/6444/commits/c755905b1f1666a3f1014359cf4b5d3a208ace64)

services/hackbot-api/app/phabricator_authorization.py
```diff
@@ -0,0 +1,69 @@
...
+            maxsize=1,
+            ttl=cache_ttl_seconds,
+        )
+        self._members_lock = asyncio.Lock()
+        self._last_members_refresh = 0.0
+        self._missing_member_refresh_cooldown_seconds = (
+            missing_member_refresh_cooldown_seconds
+        )
+
+    async def is_authorized(self, client: PhabricatorClient, author_phid: str) -> bool:
```
suhaibmujahid: I would pass the client to the constructor instead of here.

services/hackbot-api/app/phabricator_authorization.py
```diff
@@ -0,0 +1,69 @@
...
+if TYPE_CHECKING:
+    from phabricator_client import PhabricatorClient
+
+
+class PhabricatorAuthorizer:
+    """Cache-backed authorization checks against a Phabricator project."""
+
+    def __init__(
+        self,
+        group_phid: str,
```
suhaibmujahid:
```suggestion
        authorized_group_phid: str,
```

services/hackbot-api/app/phabricator_authorization.py
```diff
@@ -0,0 +1,69 @@
...
+
+            members = await client.get_project_members(self._group_phid)
+            self._members_cache[self._group_phid] = members
+            self._last_members_refresh = time.monotonic()
+            return author_phid in members
+
+
+# Members of this project are authorized to trigger Hackbot.
+AUTHORIZED_GROUP_PHID = "PHID-PROJ-njo5uuqyyq3oijbkhy55"  # bmo-editbugs-team
+phabricator_authorizer = PhabricatorAuthorizer(AUTHORIZED_GROUP_PHID)
```
suhaibmujahid: Instead of using it as a singleton here, I would pass it as a dependency to the FastAPI route. The dependency injection with FastAPI makes it easier to avoid singletons.

services/hackbot-api/app/phabricator_webhook.py
```diff
@@ -22,6 +25,12 @@
 _COMMENT_TYPES = frozenset({"comment", "inline"})
 
 
+@dataclass(frozen=True)
+class HackbotMention:
+    raw: str
```
suhaibmujahid: The name `raw` is not clear, I would use something like `comment`, `text`, or `content`.

services/hackbot-api/app/phabricator_authorization.py
ayoubdiourin7: good idea ! Addressed in d1521fc. The app creates one authorizer on app.state, and the webhook route receives it through a FastAPI dependency.

services/hackbot-api/app/phabricator_authorization.py
ayoubdiourin7: Addressed in d1521fc.

services/hackbot-api/app/main.py
```diff
@@ -38,6 +43,11 @@ async def lifespan(app: FastAPI):
     version=__version__,
     lifespan=lifespan,
 )
+app.state.phabricator_client = PhabricatorClient(settings.phabricator)
+app.state.phabricator_authorizer = PhabricatorAuthorizer(
+    app.state.phabricator_client,
+    AUTHORIZED_GROUP_PHID,
+)
```
suhaibmujahid: This pattern would make server startup a bit slower, plus it is not utilizing the clean dependency injection pattern in FastAPI.

services/hackbot-api/app/routers/webhooks.py
```diff
@@ -25,16 +26,21 @@
 router = APIRouter(prefix="/webhooks")
 
 
-def get_phabricator_client() -> PhabricatorClient:
-    """Dependency: a Conduit client built from the service's Phabricator config."""
-    return PhabricatorClient(settings.phabricator)
+def get_phabricator_client(request: Request) -> PhabricatorClient:
+    """Dependency: the app-scoped Conduit client."""
+    return request.app.state.phabricator_client
```
suhaibmujahid: We do not need to change this; instead, we could use it as it was as a dependency in `get_phabricator_authorizer()`.

services/hackbot-api/app/routers/webhooks.py
```diff
@@ -25,16 +26,21 @@
...
 
 
 def get_hackbot_client() -> HackbotClient:
     """Dependency: a client for triggering runs over the public hackbot API."""
     return HackbotClient(settings.hackbot_api_url, settings.external_api_key)
 
 
+def get_phabricator_authorizer(request: Request) -> PhabricatorAuthorizer:
+    """Dependency: the app-scoped authorizer with its shared member cache."""
+    return request.app.state.phabricator_authorizer
```
suhaibmujahid: I expect this tow work (please verify):
```suggestion
def get_phabricator_authorizer(phab_client: PhabricatorClient = Depends(get_phabricator_client)) -> PhabricatorAuthorizer:
    return PhabricatorAuthorizer(phab_client, AUTHORIZED_GROUP_PHID)
```

services/hackbot-api/app/config.py
```diff
@@ -45,7 +45,7 @@ class Settings(BaseSettings):
     # Phabricator Conduit connection config, embedded as a nested model and
     # populated in this single settings parse from PHABRICATOR_URL /
     # PHABRICATOR_API_KEY / PHABRICATOR_TIMEOUT_SECONDS (see env_nested_delimiter
-    # below). Injected directly as PhabricatorClient(settings.phabricator).
+    # below). Constructed once at application startup and injected as a dependency.
```
suhaibmujahid: The whole comment is not necessary in the first place.

services/hackbot-api/app/phabricator_authorization.py
suhaibmujahid: I would use dependency injection instead of that. I left some comments regarding that.

services/hackbot-api/app/routers/webhooks.py
ayoubdiourin7: Since `PhabricatorAuthorizer` owns the TTL cache, creating it in this dependency would create a new empty cache for every request. I kept one app-scoped instance and inject it through the FastAPI dependency so the cache is shared across webhook requests.

suhaibmujahid: Good catch! But what if we cache an instance of `PhabricatorAuthorizer` in `request.app.state` from within `get_phabricator_authorizer()`, so that would lazy load it only when the endpoint is called.

services/hackbot-api/app/routers/webhooks.py
```diff
@@ -35,6 +39,18 @@ def get_hackbot_client() -> HackbotClient:
     return HackbotClient(settings.hackbot_api_url, settings.external_api_key)
 
 
+def get_phabricator_authorizer(
+    request: Request,
+    phab_client: PhabricatorClient = Depends(get_phabricator_client),
+) -> PhabricatorAuthorizer:
+    """Dependency: lazily create the app-scoped authorizer and its member cache."""
+    authorizer = getattr(request.app.state, "phabricator_authorizer", None)
```
suhaibmujahid: a bit cleaner:
```suggestion
    authorizer = request.app.state.get("phabricator_authorizer")
```

suhaibmujahid (dismissed):

services/hackbot-api/app/routers/webhooks.py
ayoubdiourin7: That's what i implemented in [2d334db](https://github.com/mozilla/bugbug/pull/6444/commits/2d334dbcf16b384124d70be3627c95cd3d967588)

suhaibmujahid (approved):

PR #6448, opened 2026-08-02
Make creation of Puppeteer script part of reproduction step for autowebcompat

ksy36:


agents/autowebcompat-repro/Dockerfile
```diff
@@ -47,19 +47,23 @@ RUN apt-get update \
...
+# package.json/lockfile. PUPPETEER_SKIP_DOWNLOAD stops Puppeteer's install
+# script from fetching its own bundled Chrome: the browsers this agent drives
+# are downloaded at startup instead (see browser.py).
 COPY agents/autowebcompat-repro/package.json \
      agents/autowebcompat-repro/package-lock.json \
-     /app/node/
-RUN cd /app/node && npm ci --omit=dev
+     agents/autowebcompat-repro/repro_reference.mjs \
+     /app/repro/
+RUN cd /app/repro && PUPPETEER_SKIP_DOWNLOAD=1 npm ci --omit=dev
```
ksy36, line 59: `/app/repro` contains node_modules, reference Puppeteer script and the script that agent writes (so it can access Puppeteer)

agents/autowebcompat-repro/repro_reference.mjs
```diff
@@ -0,0 +1,68 @@
+// REFERENCE — write your own script in this shape; don't copy this comment.
```
ksy36: The reference script is the same shape as these [scripts](https://github.com/jrmuizel/webcompat-reproduction-scripts/blob/main/1457335.js) from @jrmuizel

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/agent.py
```diff
@@ -671,18 +754,25 @@ async def next_repro_task(
...
+                    initial_repro.script_path, browser, chrome_browser.stable
+                )
+                if script_result is not None:
+                    repro_results.set_result(channel, extra, script_result)
+                    return
+                logger.info(
+                    "Confirmation script did not reach a verdict in %s; "
+                    "falling back to the reproduction steps",
+                    channel,
+                )
```
ksy36, line 772: After a script has been created in BugReproduction, this part runs the script and if Firefox failure is confirmed, doesn't proceed to reproducing it using the agent for a given channel - sets result with `reproduced: true`, `confirmed_by_script: true`.

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/agent.py
```diff
@@ -371,14 +413,26 @@ def system_prompt(self) -> str:
      web-compat issue; refine the steps and re-check before concluding.
    - If the reported broken behaviour reproduces in both browsers, it is not a
      Firefox web-compat issue: set `failure_reason` to `non_compat`.
-4. If the issue reproduces AND the breakage is visual in nature (incorrect
+4. If the issue reproduces, write and run a Puppeteer script that drives the real
```
jgraham, line 416: Hmm, so I guess a question is whether we want one script that launches both browsers, or a single script that can either launch Firefox or Chrome and produces different output (either exit code or stdout text) depending on whether the issue was reproduced or not. The advantage of the latter is that for future steps we want to be able to reuse the generated script where possible but not always require running in Chrome too (e.g. if we were using the script as the condition when running mozregression).

jgraham (changes requested):

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/agent.py
ksy36: Sounds good, I've changed it to launch one browser at a time and the script is checking expected behavior (and Firefox to fail and Chrome pass).

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/agent.py
ksy36: This runs the script only in Firefox (basically what the agent has been doing as well)

agents/autowebcompat-repro/repro_reference.mjs
```diff
@@ -1,33 +1,43 @@
...
-// Exit code:
-// 0 = Firefox-specific breakage reproduced: broken in Firefox, working in Chrome.
-// 1 = no Firefox-specific breakage: it worked in both, broke in both, or only Chrome broke.
-// 2 = no verdict — a browser wouldn't launch, or the script itself broke.
+// Run the script once per browser:
+//   BROWSER=firefox BROWSER_BIN=/path/to/firefox node this-script.mjs
+//   BROWSER=chrome  BROWSER_BIN=/path/to/chrome  node this-script.mjs
+//
+// Exit codes:
+//   0 = worked as expected
```
jgraham: "as expected" might be ambiguous here, because maybe it's not clear whether in Firefox we "expect" it to fail. Maybe "as intended by the site author" or something?

jgraham (dismissed):

jgraham (approved):

PR #6449, opened 2026-08-02
Support re-triggering a failed run from the Hackbot UI

### LINKED ISSUE #6413: Support re-triggering a faild run from Hackbot UI
We could trigger a new run with the same inputs as the failed run. That might be simpler than supporting re-executing the same run.

suhaibmujahid:
Resolves #6413

### DIFF
--- a/services/hackbot-ui/app/api/runs/[runId]/retrigger/route.ts
+++ b/services/hackbot-ui/app/api/runs/[runId]/retrigger/route.ts
@@ -24,7 +24,9 @@ export async function POST(
     const doc = await getRun(runId);
     if (!isFailed(doc.status)) {
       return NextResponse.json(
-        { error: `Only failed runs can be re-run (this one is ${doc.status})` },
+        {
+          error: `Only failed runs can be retriggered (this one is ${doc.status})`,
+        },
         { status: 409 }
       );
     }
--- a/services/hackbot-ui/app/api/runs/[runId]/retrigger/route.ts
+++ b/services/hackbot-ui/app/api/runs/[runId]/retrigger/route.ts
@@ -1,0 +1,37 @@
+import { NextResponse } from "next/server";
+
+import { createRun, getRun, HackbotError } from "@/lib/hackbot";
+import { getAuthedEmail } from "@/lib/session";
+import { isFailed } from "@/lib/types";
+
+export const dynamic = "force-dynamic";
+
+// POST /api/runs/:runId/retrigger: start a new run with the same inputs as a
+// failed one. Inputs are read server-side, so the browser only sends a run id.
+// Like POST /api/runs, the new run is attributed to the signed-in user rather
+// than to the original run's requester.
+export async function POST(
+  _req: Request,
+  { params }: { params: Promise<{ runId: string }> }
+) {
+  const email = await getAuthedEmail();
+  if (!email) {
+    return NextResponse.json({ error: "Unauthorized" }, { status: 401 });
+  }
+
+  const { runId } = await params;
+  try {
+    const doc = await getRun(runId);
+    if (!isFailed(doc.status)) {
+      return NextResponse.json(
+        { error: `Only failed runs can be re-run (this one is ${doc.status})` },
+        { status: 409 }
+      );
+    }
+    const run = await createRun(doc.agent, doc.inputs, email);
+    return NextResponse.json(run, { status: 201 });
+  } catch (err) {
+    const status = err instanceof HackbotError ? err.status : 500;
+    return NextResponse.json({ error: (err as Error).message }, { status });
+  }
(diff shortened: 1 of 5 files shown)


services/hackbot-ui/app/api/runs/[runId]/retrigger/route.ts
```diff
@@ -0,0 +1,37 @@
...
+  if (!email) {
+    return NextResponse.json({ error: "Unauthorized" }, { status: 401 });
+  }
+
+  const { runId } = await params;
+  try {
+    const doc = await getRun(runId);
+    if (!isFailed(doc.status)) {
+      return NextResponse.json(
+        { error: `Only failed runs can be re-run (this one is ${doc.status})` },
```
Comment (padenot) [inline services/hackbot-ui/app/api/runs/[runId]/retrigger/route.ts]: Why? What if you're not happy with the outcome and want to roll the dice again.

This is probably better served by a "Clone" action that let you edit it (change prompt, change model, etc.)

services/hackbot-ui/components/RunDetail.tsx
```diff
@@ -159,12 +187,27 @@ export function RunDetail({ runId }: { runId: string }) {
             </span>
           )}
         </div>
-        <Link href="/" className="muted">
-          ← all runs
-        </Link>
+        <div style={{ display: "flex", alignItems: "center", gap: 12 }}>
```
padenot: Do we typically do inline style here? I know some framework somehow like that.

services/hackbot-ui/components/RunDetail.tsx
```diff
@@ -159,12 +187,27 @@ export function RunDetail({ runId }: { runId: string }) {
...
+          {isFailed(run.status) && (
+            <button
+              type="button"
+              className="secondary"
+              onClick={retrigger}
+              disabled={retriggering}
+            >
+              {retriggering ? "Re-running…" : "Re-run with same inputs"}
+            </button>
+          )}
```
padenot: It's nice when UI is always the same look, and you can understand things without thinking. Here, we could have:

- Enabled, "Re-run with same inputs"
- Disabled, "Currently re-running"
- Disabled, "Run succeeded, cannot rerun"

GitHub CI, TreeHerder and others allow to retrigger a run regardless of its final or current state

services/hackbot-ui/components/RunDetail.tsx
```diff
@@ -159,12 +187,27 @@ export function RunDetail({ runId }: { runId: string }) {
...
+          <Link href="/" className="muted">
+            ← all runs
+          </Link>
+        </div>
       </div>
 
       {error && <div className="error-banner">Refresh error: {error}</div>}
+      {retriggerError && (
+        <div className="error-banner">Re-run failed: {retriggerError}</div>
```
padenot: It's the retrigger that failed, not the re-run, right?

padenot (dismissed):
Some comments, this largely does what it says it does and what is needed.

services/hackbot-ui/app/api/runs/[runId]/retrigger/route.ts
Comment (suhaibmujahid) [reply services/hackbot-ui/app/api/runs/[runId]/retrigger/route.ts]: You can still trigger it again manually. But supporting that with a button could be tricky without accommodating for cases where re-triggering have side effects that might create noise, like double posting on Bugzilla, or submitting two patches for the same bug.


> This is probably better served by a "Clone" action that let you edit it (change prompt, change model, etc.)

I agree!

services/hackbot-ui/components/RunDetail.tsx
suhaibmujahid: Next.js supports CSS-in-JS nicely, but I will refactor that at some point to make a more consistent styling, maybe with Tailwind.

services/hackbot-ui/components/RunDetail.tsx
suhaibmujahid: Good catch, this pr mixes re-run and retrigger, it should be retrigger since it is more accurate.

services/hackbot-ui/components/RunDetail.tsx
suhaibmujahid: Done in 017d7db2199216495f2763bd3b476b37dd8ce4fc

> GitHub CI, TreeHerder and others allow to retrigger a run regardless of its final or current state

We could do that with the "Clone" workflow that you suggested in https://github.com/mozilla/bugbug/pull/6449#discussion_r3713663346

services/hackbot-ui/components/RunDetail.tsx
suhaibmujahid: Fixed in 5a06970ef4341efe5ded5d8cecc600d01bed5fd3

padenot (approved):

PR #6450, opened 2026-08-02
Add expected results for verifiable steps in the `test-plan-generator` agent

### LINKED ISSUE #6432: Generate test cases in test rail with test steps in the field on the left and expectations in the field on the right
TestRail test have a test steps field and an associated test expectations field.  The bug right now is that both test steps and test expectations are in line as Steps in the left fields.  Use the expectation field that is available for every step. However, for readability, we don't need happy path test step expectations.  When it comes to the actual verification of what the test summary describes, that test step should have expectation description in the expectation field.

johnpangas:
Fixes #6432 

This pull request introduces significant improvements to how test case steps and their expected results are handled and validated throughout the test plan generation and submission process.

### DIFF
--- a/agents/test-plan-generator/hackbot_agents/test_plan_generator/result.py
+++ b/agents/test-plan-generator/hackbot_agents/test_plan_generator/result.py
@@ -16,6 +16,14 @@ RESULT_SERVER_NAME = "test-plan-generator"
 SUBMIT_RESULT_TOOL = f"mcp__{RESULT_SERVER_NAME}__submit_result"
 
 
+class GeneratedTestStep(BaseModel):
+    action: str = Field(description="The action a QA engineer should perform.")
+    expectation: str | None = Field(
+        default=None,
+        description=("Expected result for this step."),
+    )
+
+
 class GeneratedTestCase(BaseModel):
     id: int = Field(description="Sequential case id starting at 1.")
     title: str
@@ -27,12 +35,10 @@ class GeneratedTestCase(BaseModel):
         )
     )
     preconditions: str | None = None
-    steps: list[str] = Field(description="Concise ordered test steps for this case.")
-    step_expectations: list[str | None] = Field(
+    steps: list[GeneratedTestStep] = Field(
         description=(
-            "Expected results aligned by index with steps when present. Use null "
-            "for setup, navigation, or happy-path steps that do not directly "
-            "verify the case title."
+            "Concise ordered test steps. Each step has an action and an optional "
+            "expectation."
         )
     )
 
@@ -40,7 +46,7 @@ class GeneratedTestCase(BaseModel):
     def _validate_steps(self) -> "GeneratedTestCase":
         if not self.steps:
             raise ValueError("each generated test case must have at least one step")
-        if not any(self.step_expectations):
+        if not any(step.expectation for step in self.steps):
             raise ValueError(
                 "each generated test case must include an expected result on at "
                 "least one verification step"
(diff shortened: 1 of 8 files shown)


agents/test-plan-generator/hackbot_agents/test_plan_generator/result.py
```diff
@@ -22,14 +27,24 @@ class GeneratedTestCase(BaseModel):
         )
     )
     preconditions: str | None = None
-    steps: list[str] = Field(
-        description="Concise test steps for this case; between 1 and 6 steps."
+    steps: list[str] = Field(description="Concise ordered test steps for this case.")
+    step_expectations: list[str | None] = Field(
+        description=(
```
suhaibmujahid, line 39: I do not have a problem with that, if it works well with the mode, but I would expect having each step and expectation on one item would be less errorprone. E.g., `steps: list[Step]`, where the Step has two proprieties, action and exception.

libs/hackbot-runtime/hackbot_runtime/actions/handlers/testrail_handler.py
```diff
@@ -70,19 +70,32 @@ def _case_payload(
...
+    for step in test_case.get("steps", []):
+        expectation = step.get("expectation")
+        separated_steps.append(
+            {
+                "content": str(step["action"]),
+                "expected": "" if expectation is None else str(expectation),
+            }
+        )
+
+    return separated_steps
```
suhaibmujahid: This can be simplified

services/hackbot-ui/components/TestPlanView.tsx
```diff
@@ -27,6 +31,17 @@ function isStatus(value: unknown): value is TestCaseStatus {
...
+  if (!isPlainObject(value) || typeof value.action !== "string") {
+    return null;
+  }
+  return { action: value.action };
+}
+
+function isTestStep(value: TestStep | null): value is TestStep {
+  return value !== null;
+}
+
```
suhaibmujahid, line 54: Why we need these?

agents/test-plan-generator/hackbot_agents/test_plan_generator/result.py
suhaibmujahid, line 1: This makes all the work duplicated. We do things here and in the actions as well. Please fill a follow-up issue to drop it and use actions only.

suhaibmujahid (changes requested):

agents/test-plan-generator/hackbot_agents/test_plan_generator/result.py
suhaibmujahid: I just noticed that action is not used as it was supposed to be used: https://github.com/mozilla/bugbug/blob/beb81a21b489b2147f0c6b2645a01485eb17033e/agents/test-plan-generator/hackbot_agents/test_plan_generator/__main__.py#L36

The agent should use it directly instead of submitting through results, then copying it here.

suhaibmujahid: I filed https://github.com/mozilla/bugbug/issues/6508 to fix that.

services/hackbot-ui/components/TestPlanView.tsx
jpangas: `parseTestPlan` returns null when the result does not match the TestPlan shape. Then FindingsView does not render TestPlanView, it falls back to the generic findings rendering path.
So any malformed test plan data should not crash the custom UI. It just won’t get the specialized TestPlan view.

Then for `isTestStep` lets the UI reject malformed steps while proving  that the remaining parsed steps are safe to render in the custom `TestPlan` view

suhaibmujahid (approved):

PR #6488, opened 2026-08-03
code_review: default review_context_repo from the patch's own repo

padenot:
Previously review_context_repo had to be passed explicitly to every review. Add Patch.github_repo()/github_repo_branch() hooks (default None/"main") so each platform can answer this from what it already knows. PhabricatorPatch resolves it by mapping the revision's repository callsign to its GitHub mirror and branch. The agent only consults this when the caller didn't pass review_context_repo explicitly, so existing callers are unaffected.

When we're ready to make this work on github, it will be quite natural and quick to do.

### DIFF
--- a/bugbug/tools/core/platforms/phabricator.py
+++ b/bugbug/tools/core/platforms/phabricator.py
@@ -568,7 +568,7 @@ class PhabricatorPatch(Patch):
             return None
 
     @alru_cache
-    async def _github_mirror(self) -> Optional[tuple[str, str]]:
+    async def github_repo_ref(self) -> Optional[tuple[str, str]]:
         repository_phid = self._revision_metadata["fields"].get("repositoryPHID")
         if not repository_phid:
             return None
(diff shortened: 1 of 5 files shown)


padenot:
@suhaibmujahid implemented it as you said, with a set of defaults.

bugbug/tools/core/platforms/phabricator.py
```diff
@@ -524,6 +567,30 @@ def _fetch() -> Optional[str]:
...
+
+        return PHABRICATOR_REPO_TO_GITHUB.get(callsign)
+
+    async def github_repo(self) -> Optional[str]:
+        mirror = await self._github_mirror()
+        return mirror[0] if mirror else None
+
+    async def github_repo_branch(self) -> str:
+        mirror = await self._github_mirror()
+        return mirror[1] if mirror else "main"
```
suhaibmujahid: If we do not know the repo, why not return `None`?

suhaibmujahid (dismissed):

bugbug/tools/core/platforms/phabricator.py
padenot: The code was not leveraging the type system enough (wild thing to say about python heh). If we have a `github_repo`, it's a programming mistake to not have a `github_repo_branch`, so I tied both together.

suhaibmujahid (approved):

PR #6490, opened 2026-08-03
code_review: include the commit message in the review prompt

padenot:
The agent only ever saw an LLM-generated patch summary, never the author's actual title/description, so it couldn't flag unclear or misleading commit messages, or cross-check the description against the diff. Add the verbatim commit message as its own prompt section, and instruct the agent to compare it against the diff and flag commit-message issues alongside code issues.

This quite important for us, since we spend quite some time doing code archeology. Besides, an incorrect commit message can mislead or confused the (human) reviewer, and we'd like to catch this early.

### DIFF
--- a/bugbug/tools/code_review/agent.py
+++ b/bugbug/tools/code_review/agent.py
@@ -212,10 +212,15 @@ class CodeReviewTool(GenerativeModelTool):
     ) -> str:
         created_before = patch.date_created if self.is_experiment_env else None
 
+        commit_message = patch.patch_title
+        if patch.patch_description:
+            commit_message += f"\n\n{patch.patch_description}"
+
         return FIRST_MESSAGE_TEMPLATE.format(
             current_date=current_date_for_prompt(),
             patch=format_patch_set(patch.patch_set),
             patch_summarization=patch_summary,
+            commit_message=commit_message,
             external_context=external_context,
             comment_examples=self._get_comment_examples(patch, created_before),
             approved_examples=self._get_generated_examples(patch, created_before),
--- a/bugbug/tools/code_review/prompts.py
+++ b/bugbug/tools/code_review/prompts.py
@@ -20,11 +20,13 @@ Follow this systematic approach to review the patch:
 - Understand what the patch is trying to accomplish
 - Use the patch summary for context, but focus primarily on what you can see in the actual diff
 - Identify the intent and structure of the changes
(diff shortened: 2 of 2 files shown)


bugbug/tools/code_review/prompts.py
```diff
@@ -20,11 +20,13 @@
 - Understand what the patch is trying to accomplish
 - Use the patch summary for context, but focus primarily on what you can see in the actual diff
 - Identify the intent and structure of the changes
+- Compare the commit message against the diff: does it accurately describe what changed, and does it explain why?
```
marco-c, line 23: I wonder if we should add this as a separate prompt instead of bundling it with the main one

bugbug/tools/code_review/agent.py
```diff
@@ -212,10 +212,15 @@ def generate_initial_prompt(
     ) -> str:
         created_before = patch.date_created if self.is_experiment_env else None
 
+        commit_message = patch.patch_title
+        if patch.patch_description:
+            commit_message += f"\n\n{patch.patch_description}"
+
```
suhaibmujahid, line 218: We do that in the patch summary. We could keep it there and make it expose inconsistencies as note in the summary so the main review agent would add a comment about that.

bugbug/tools/code_review/prompts.py
padenot: For me as a reviewer, this is fundamental to the review process. It happens that I reject a patch just because I can't understand what is being done by reading the message.

bugbug/tools/code_review/agent.py
padenot: I'm not convinced the patch summary is any useful (but this is best adjusted separately)

This deserves to be in its own fake XML tags per the prompting guidelines: https://platform.claude.com/docs/en/build-with-claude/prompt-engineering/claude-prompting-best-practices#structure-prompts-with-xml-tags

bugbug/tools/code_review/prompts.py
marco-c: Yes, I'm not saying we shouldn't do it, I'm just wondering if it's better to do it as part of the same prompt or to have a separate prompt specifically for this. The result would still be posted together.
I'm wondering about this because having it in the same prompt could reduce (and so slightly worsen) the LLM's focus on the diff review, but it's cheaper because it's fewer tokens.

PR #6505, opened 2026-08-04
Add components of paths mentioned in comments as features of the component model

marco-c:
Train on Taskcluster: component

### DIFF
--- a/bugbug/bug_features.py
+++ b/bugbug/bug_features.py
@@ -620,6 +620,54 @@ def get_author_ids():
     return author_ids
 
 
+PATH_RE = re.compile(r"(?:[\w.-]+/)+[\w.-]+\.[a-zA-Z0-9]+")
+
+
+def find_component(path, component_mapping):
+    # A path mentioned in a comment might not match a repository path exactly
+    # (e.g. it could have extra leading directories, like a local checkout
+    # path), so try the path itself and then progressively shorter suffixes
+    # of it, keeping the longest one that matches.
+    parts = path.split(b"/")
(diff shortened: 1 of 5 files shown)


bugbug/bug_features.py
```diff
@@ -620,6 +620,54 @@ def get_author_ids():
...
+    component_mapping = repository.get_component_mapping()
+    components = (find_component(path, component_mapping) for path in paths)
+    return [component for component in components if component is not None]
+
+
+# Components of paths mentioned in bug comments.
+# In case of stack traces, the position is important (e.g. component of the first path is more indicative than component of the last path,
+# which tends to be a "main"-like entry point common to most stack traces). We weight each mention by the reciprocal of its rank among all
+# paths mentioned in the comments, and sum the weights of repeated mentions of the same component.
+
```
suhaibmujahid, line 655: What if we make the component and the position as one feature, and then the model will learn the importance based on the position, which could learn to detect outlier patterns.

suhaibmujahid (commented):
The test task is failing.

bugbug/bug_features.py
marco-c: What do you mean by "one feature"? Do you mean one-hot COMPONENT-POSITION? That would explode quickly with too many combinations, no?

suhaibmujahid: Yes, we would try that and see (we could limit the number of components through the mapping). Then, if it does not work, we can get one for the first line, another for the last line, and then for all things in the middle, we could use weights.

PR #6516, opened 2026-08-05
Enrich Phabricator webhook comment context

### LINKED ISSUE #6434: Enrich the comment context when triggering a run through a Phabricator comment
When the webhook triggers a run, it might be helpful for the agent if we provide extra identifying context, like the commit id, whether it is an inline or normal comment, and we might even provide the diff context for inline comments.

ayoubdiourin7:
**Summary**

- Structure webhook-triggered Phabricator comments as XML with `comment_id`, `type`, and, for inline comments, `diff_id`.
- Add coverage for normal comments, inline comments, mixed comments, and XML escaping.

Resolve #6434


agents/bug-fix/hackbot_agents/bug_fix/prompts/follow-up.md
```diff
@@ -6,7 +6,9 @@ Respond only to the comments quoted below. Ignore any earlier mentions of you el
 {comment}
 </comments>
 
-A quoted comment rarely stands alone: an inline one only makes sense next to the code it sits on. Before acting, use the `phabricator` tools to read D{revision_id} and its thread, and to locate the comment context so you can understand it. Your tree is at the revision's latest diff, so a comment on an older `diff_id` may point at code that has since changed.
+The block contains one service-generated `<comment>` element for each triggering comment. Its `comment_id` and `type` attributes match the names returned by the Phabricator tools; inline comments also have a `diff_id`.
+
+A quoted comment rarely stands alone: an inline one only makes sense next to the code it sits on. Before acting, use the `phabricator` tools to read D{revision_id} and its thread, and locate the relevant comment context so you can understand it. For an inline comment, use the supplied `diff_id` to inspect its diff. Your tree is at the revision's latest diff, so a comment on an older `diff_id` may point at code that has since changed.
```
suhaibmujahid: I would not force the behaviour of fetching the diff since we do not know if it is a problem. The agent might perform worse when you give it strict instruction.

agents/bug-fix/hackbot_agents/bug_fix/prompts/follow-up.md
```diff
@@ -6,7 +6,9 @@ Respond only to the comments quoted below. Ignore any earlier mentions of you el
 {comment}
 </comments>
 
-A quoted comment rarely stands alone: an inline one only makes sense next to the code it sits on. Before acting, use the `phabricator` tools to read D{revision_id} and its thread, and to locate the comment context so you can understand it. Your tree is at the revision's latest diff, so a comment on an older `diff_id` may point at code that has since changed.
+The block contains one service-generated `<comment>` element for each triggering comment. Its `comment_id` and `type` attributes match the names returned by the Phabricator tools; inline comments also have a `diff_id`.
```
suhaibmujahid: The agent will see that, we do not need to describe it.

services/hackbot-api/app/phabricator_webhook.py
```diff
@@ -39,6 +43,46 @@ def triggering_transaction_phids(payload: dict) -> list[str]:
...
+    return HackbotMention(
+        comment=comment["content"]["raw"],
+        author_phid=author_phid,
+        comment_id=comment["id"],
+        comment_type="inline" if transaction_type == "inline" else "comment",
+        diff_id=diff_id,
+    )
+
+
+def _get_hackbot_mention(
```
suhaibmujahid: Having `_build_hackbot_mention()`, `_get_hackbot_mention()`, and `find_hackbot_mentions()` is a bit confusing, especially that the names are similar. I would simplify that.

services/hackbot-api/app/phabricator_webhook.py
```diff
@@ -58,35 +102,39 @@ def find_hackbot_mentions(
...
+        f'comment_id="{mention.comment_id}"',
+        f'type="{mention.comment_type}"',
+    ]
+    if mention.comment_type == "inline":
+        attributes.append(f'diff_id="{mention.diff_id}"')
+    body = "\n".join(f"    {line}" for line in escape(mention.comment).splitlines())
+    return f"  <comment {' '.join(attributes)}>\n{body}\n  </comment>"
+
+
+def _join_comments(mentions: list[HackbotMention]) -> str:
```
suhaibmujahid: I do not think we need a function just for this.

services/hackbot-api/app/phabricator_webhook.py
```diff
@@ -28,6 +29,9 @@
 class HackbotMention:
     comment: str
     author_phid: str
+    comment_id: int
+    comment_type: Literal["comment", "inline"]
```
suhaibmujahid:
```suggestion
    comment_type: Literal["regular", "inline"]
```

services/hackbot-api/app/phabricator_webhook.py
```diff
@@ -28,6 +29,9 @@
...
+    diff_id: int | None = None
```
suhaibmujahid, line 34: If we have it all the time, it would be nice. But if it will increase complexity (e.g., extra http requests), let us file a follow-up issue and do it later, not part of this PR.

services/hackbot-api/app/phabricator_webhook.py
```diff
@@ -39,6 +43,46 @@ def triggering_transaction_phids(payload: dict) -> list[str]:
...
+    transaction: dict, comment: dict, author_phid: str
+) -> HackbotMention:
+    """Build the minimal agent context for a matching comment transaction."""
+    transaction_type = transaction["type"]
+    if transaction_type == "inline":
+        fields = transaction["fields"]
+        diff = fields["diff"]
+        diff_id = diff["id"]
+    else:
+        diff_id = None
```
suhaibmujahid:
```suggestion
    diff_id = transaction["fields"]["diff"]["id"] if transaction["type"] == "inline" else None
```

suhaibmujahid (changes requested):

services/hackbot-api/app/phabricator_webhook.py
ayoubdiourin7: All metadata added here is already returned by the existing transaction.search call, so no additional HTTP request or lookup is needed.

suhaibmujahid: I mean to have it even for regular comments as well.

ayoubdiourin7: addressed it in [bea1a18](https://github.com/mozilla/bugbug/pull/6516/commits/bea1a18258a76f1673d819d1dde65ae76a8be81d)

services/hackbot-api/app/phabricator_webhook.py
```diff
@@ -60,33 +63,45 @@ def find_hackbot_mentions(
             continue
         if transaction.get("type") not in _COMMENT_TYPES:
             continue
+
         author_phid = transaction.get("authorPHID")
-        if not author_phid:
-            continue
-        if bot_phid and author_phid == bot_phid:
+        if not author_phid or (bot_phid and author_phid == bot_phid):
```
suhaibmujahid: Why not simplify it to this? 
```suggestion
        if not author_phid or author_phid == bot_phid:
```

services/hackbot-api/app/phabricator_webhook.py
suhaibmujahid: Comment bea1a18258a76f1673d819d1dde65ae76a8be81d does the opposite, instead of adding `diff_id` for all comments, it removes it for all.

ayoubdiourin7: I misunderstood your comment at first, so I reverted [bea1a18](https://github.com/mozilla/bugbug/pull/6516/commits/bea1a18258a76f1673d819d1dde65ae76a8be81d)
Inline comments now have `diff_id` again. For regular comments, we can handle it in a follow-up issue since it requires an additional lookup.

services/hackbot-api/app/phabricator_webhook.py
ayoubdiourin7: updated in dd86f99

services/hackbot-api/app/phabricator_webhook.py
suhaibmujahid: WFM. Could you please file an issue for that.

suhaibmujahid (approved):

services/hackbot-api/app/phabricator_webhook.py
ayoubdiourin7: Done in #6572

PR #6519, opened 2026-08-05
frontend-triage: invite feedback via Bugzilla reactions and tags

### LINKED ISSUE #6489: Remove the comment footer in Hackbot's comments
This:

> This is an automated analysis result. If this result is incorrect please add a needinfo and feel free to correct the error.

The comment is posted by a Hackbot account. It is clear enough that this is automated analysis. We could add more clarification on the account itself, but no need to mention this in every comment.

crythms:
We have no way to get feedback from engineers on the Hackbot comments posted to Bugzilla.. Bugzilla already has
comment reactions and tags that can be used to gather feedback, so this just adds a line pointing readers at them.

Fixes #6489 

Triage comments now end:

    ---

    If you'd like to provide feedback on this comment, please use the 👍 or 👎
    reaction. If you want to categorize your feedback you can add one of the
    following tags: ai-triage-wrong-file, ai-triage-wrong-cause,
    ai-triage-hallucination, ai-triage-out-of-scope.

### DIFF
--- a/agents/frontend-triage/hackbot_agents/frontend_triage/agent.py
+++ b/agents/frontend-triage/hackbot_agents/frontend_triage/agent.py
@@ -56,6 +56,22 @@ HERE = Path(__file__).resolve().parent
 # machine-consumable for downstream handoff (summary.json -> execution agent).
 _JSON_BLOCK = re.compile(r"```json\s*(\{.*?\})\s*```", re.DOTALL)
 
+_FEEDBACK_TAGS = (
+    "If you want to categorize your feedback you can add one of the following "
+    "tags: ai-triage-wrong-file, ai-triage-wrong-cause, ai-triage-hallucination, "
+    "ai-triage-out-of-scope."
+)
+
+
+def feedback_tags_hook(action: dict) -> None:
+    """Offer the triage-specific feedback tags below the runtime's footer."""
+    params = action.get("params")
+    if not isinstance(params, dict):
+        return
+    text = params.get("text")
+    if isinstance(text, str):
+        params["text"] = f"{text.rstrip()}\n{_FEEDBACK_TAGS}"
+
 
 class FrontendTriageResult(HackbotAgentResult):
     bug_id: int
@@ -233,6 +249,7 @@ async def run_frontend_triage(
         "bugzilla.add_comment",
         permalink_hook(permalink_prefix(searchfox_rev), source_repo.resolve()),
     )
+    actions_recorder.add_hook("bugzilla.add_comment", feedback_tags_hook)
 
     system_prompt = load_system_prompt(rules_dir, instructions)
 
--- a/libs/hackbot-runtime/hackbot_runtime/actions/bugzilla.py
+++ b/libs/hackbot-runtime/hackbot_runtime/actions/bugzilla.py
@@ -21,8 +21,8 @@ from pydantic import Field
 from hackbot_runtime.actions.recorder import ActionsRecorder
 
 _COMMENT_FOOTER = (
-    "*This is an automated analysis result. If this result is incorrect "
-    "please add a needinfo and feel free to correct the error.* "
+    "If you'd like to provide feedback on this comment, please use the 👍 or 👎 "
+    "reaction."
 )
 _ATTACHMENT_COMMENT_FOOTER = (
     "*This is the analysis tool's suggested fix. Feel welcome to adopt "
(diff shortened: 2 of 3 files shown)


agents/frontend-triage/hackbot_agents/frontend_triage/agent.py
```diff
@@ -233,6 +259,7 @@ async def run_frontend_triage(
         "bugzilla.add_comment",
         permalink_hook(permalink_prefix(searchfox_rev), source_repo.resolve()),
     )
+    actions_recorder.add_hook("bugzilla.add_comment", feedback_footer_hook)
```
msujaws: I think we'll end up with two footers (the 'add a needinfo' footer and this new footer). That needinfo footer is defined in actions/bugzilla.py:110. 

From Claude,

> Recommended fix: strip the shared constant, don't re-spell it
> 
> Make the runtime's footer public so the agent can strip exactly what the runtime added:
> 
> `libs/hackbot-runtime/hackbot_runtime/actions/bugzilla.py`
> ```
> COMMENT_FOOTER = (
>     "*This is an automated analysis result. If this result is incorrect "
>     "please add a needinfo and feel free to correct the error.* "
> )
> ```
> 
> (rename the two internal uses at bugzilla.py:110 and the phabricator test's sibling is separate — leave _ATTACHMENT_COMMENT_FOOTER alone; frontend-triage records no attachments.)
> 
> Then in the PR's hook:
> 
> from hackbot_runtime.actions.bugzilla import COMMENT_FOOTER
> 
> def feedback_footer_hook(action: dict) -> None:
>     """Replace the runtime's needinfo footer with a feedback invitation.
> 
>     Strips ``COMMENT_FOOTER`` rather than appending past it: ``add_comment``
>     adds that footer before hooks run, so appending would leave the comment
>     asking for feedback twice, two different ways.
>     """
>     params = action.get("params")
>     if not isinstance(params, dict):
>         return
>     text = params.get("text")
>     if not isinstance(text, str):
>         return
>     body = text.rstrip()
>     if body.endswith(COMMENT_FOOTER.rstrip()):
>         body = body[: -len(COMMENT_FOOTER.rstrip())].rstrip()
>     params["text"] = f"{body}\n\n---\n\n{_FEEDBACK_FOOTER}"
> 
> The .rstrip() on both sides matters — COMMENT_FOOTER ends in a trailing space that add_comment preserves.
> 
> Importing the constant rather than copying the string means a future reword of the runtime footer keeps stripping correctly. Worth pairing with a test asserting "needinfo" not in text on a recorded triage comment, since the failure mode is silent (footer comes back, nothing errors).
> 
> Alternative, if you'd rather fix the layering
> 
> Move footer application out of add_comment into a hook that actions_server_for registers by default, with a comment_footer= parameter agents can override. Then frontend-triage passes its own string and deletes the hook entirely — no string surgery. Downside: "every recorded comment carries a footer" stops being a property of add_comment and starts depending on how the recorder was constructed; direct ActionsRecorder() callers (the existing tests do this) would silently get no footer. Bigger diff, and it's a runtime change riding on an agent-scoped PR.
> 
> One judgment call for whoever lands it: the new footer drops the "add a needinfo to correct it" affordance from triage comments. If reactions/tags are meant to supplement rather than replace that, folding a needinfo clause into _FEEDBACK_FOOTER keeps it in one paragraph.

crythms: Good catch. I'll replace the original footer with the new one

msujaws (dismissed):
Thanks!

agents/frontend-triage/hackbot_agents/frontend_triage/agent.py
suhaibmujahid: @crythms we have an issue filed to remove that comment footer altogether: https://github.com/mozilla/bugbug/issues/6489 

That will make things simpler.

msujaws: We can fix this issue and #6489 if we remove the text
> "*This is an automated analysis result. If this result is incorrect please add a needinfo and feel free to correct the error.* "

and keep the rest of the patch.

crythms: I'll remove it

crythms: Alright, its removed

msujaws (approved):

agents/frontend-triage/hackbot_agents/frontend_triage/agent.py
```diff
@@ -56,6 +56,22 @@
...
+)
+
+
+def feedback_tags_hook(action: dict) -> None:
+    """Offer the triage-specific feedback tags below the runtime's footer."""
+    params = action.get("params")
+    if not isinstance(params, dict):
+        return
+    text = params.get("text")
+    if isinstance(text, str):
```
suhaibmujahid, line 72: We should improve the hook types so we do not need to do these checks here. I will file a follow-up issue for this.

suhaibmujahid (approved):

PR #6523, opened 2026-08-05
Slack notifications for test repair agent

### LINKED ISSUE #6510: Support Slack notifications
Agents like test-repair need to notify a specific group (slack #sheriff-notifications in this case)

It should be an action and handler (similar to Bugzilla and Phabricator).

It should authenticate through a Slack API app and use the `chat.postMessage` API.

evgenyrp:
- add support for runtime following Phab pattern
- send notifications to sheriffs' channel from the build repair agent

fixes #6510

### DIFF
--- a/agents/test-repair/hackbot_agents/test_repair/__main__.py
+++ b/agents/test-repair/hackbot_agents/test_repair/__main__.py
@@ -3,11 +3,13 @@ import tempfile
 from pathlib import Path
 
 from hackbot_runtime import HackbotContext, run_async
+from hackbot_runtime.actions.slack import ACTION_TYPE, record_message, run_link_hook
 from pydantic_settings import BaseSettings, SettingsConfigDict
 
 from .agent import TestRepairResult
-from .config import SKIP_FIREFOX_BUILD
+from .config import SKIP_FIREFOX_BUILD, SLACK_CHANNEL
 from .logs import download_failure_logs
+from .notify import build_message
 from .resolve import Investigation, resolve_investigation
 
 logger = logging.getLogger(__name__)
@@ -60,7 +62,7 @@ async def main(ctx: HackbotContext) -> TestRepairResult:
     logger.info("Pinning checkout to %s with depth %s", ref, depth)
     source_repo = await ctx.prepare_repo(ref=ref, depth=depth)
 
-    return await run_test_repair(
+    result = await run_test_repair(
         bugzilla_mcp_server=bugzilla_mcp_server,
         source_repo=source_repo,
         fx_ctx=ctx.firefox,
@@ -74,6 +76,11 @@ async def main(ctx: HackbotContext) -> TestRepairResult:
         verbose=True,
         publish_file=ctx.publish_file,
     )
+
+    # Notifications are active for all runs
+    ctx.actions.add_hook(ACTION_TYPE, run_link_hook(ctx.run_id))
+    record_message(ctx.actions, SLACK_CHANNEL, build_message(result, investigation))
+    return result
 
 
 if __name__ == "__main__":
--- a/agents/test-repair/hackbot_agents/test_repair/__main__.py
+++ b/agents/test-repair/hackbot_agents/test_repair/__main__.py
@@ -3,13 +3,13 @@ import tempfile
 from pathlib import Path
 
 from hackbot_runtime import HackbotContext, run_async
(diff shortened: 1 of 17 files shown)


agents/test-repair/hackbot_agents/test_repair/__main__.py
```diff
@@ -72,6 +77,16 @@ async def main(ctx: HackbotContext) -> TestRepairResult:
         publish_file=ctx.publish_file,
     )
 
+    # Recorded here rather than decided by the agent: a sheriff acts on every
+    # verdict, including the ones that recommend leaving the tree alone.
+    record_message(
+        ctx.actions,
+        inputs.slack_channel,
+        build_message(result, investigation),
+        reasoning="Sheriffs decide whether to back out the culprit or retrigger.",
```
suhaibmujahid: You could keep thin `None`, it meant to be used by the agent for debugging purposes.

libs/hackbot-runtime/hackbot_runtime/actions/handlers/slack_handler.py
```diff
@@ -0,0 +1,112 @@
...
+    """Map a recorded audience onto the channel the deployment routes it to.
+
+    An audience with no entry of its own falls back to ``default`` when one is
+    configured -- a channel belongs to the team that owns the subject, so the map
+    is per-audience rather than one address. With nothing configured (or no
+    fallback) the recorded value is the channel, so a plain ``#channel`` works
+    out of the box.
+    """
+    channels = _channels()
+    return channels.get(channel) or channels.get(_DEFAULT_CHANNEL_KEY) or channel
```
suhaibmujahid: We could do that as a hook instead, so it lives in the agent itself instead of the handler.

libs/hackbot-runtime/hackbot_runtime/actions/handlers/slack_handler.py
```diff
@@ -0,0 +1,112 @@
...
+    return {
+        "Authorization": f"Bearer {token}",
+        "Content-Type": "application/json; charset=utf-8",
+    }
+
+
+def _post(body: dict[str, Any]) -> dict[str, Any]:
+    response = requests.post(
+        _POST_MESSAGE_URL, json=body, headers=_headers(), timeout=_TIMEOUT_SECONDS
+    )
```
suhaibmujahid: Let us use the official Slack SDK: https://pypi.org/project/slack-sdk

suhaibmujahid (commented):
Looks nice and clean! Thank you!

libs/hackbot-runtime/hackbot_runtime/actions/handlers/slack_handler.py
```diff
@@ -0,0 +1,80 @@
...
+    """Map a recorded audience onto the channel the deployment routes it to.
+
+    An audience with no entry of its own falls back to ``default`` when one is
+    configured -- a channel belongs to the team that owns the subject, so the map
+    is per-audience rather than one address. With nothing configured (or no
+    fallback) the recorded value is the channel, so a plain ``#channel`` works
+    out of the box.
+    """
+    channels = _channels()
+    return channels.get(channel) or channels.get(_DEFAULT_CHANNEL_KEY) or channel
```
suhaibmujahid: I would suggest dropping `_resolve_channel()`, this can be handled by the agent itself. IF we need any advanced routing, we could do it in a followup.

libs/hackbot-runtime/hackbot_runtime/actions/slack.py
```diff
@@ -0,0 +1,80 @@
...
+
+_CHANNEL_DESCRIPTION = (
+    'Where to post: a Slack channel ("#example-channel", or a channel '
+    'ID), or an audience key the deployment maps to one ("example", '
+    '"<Product> :: <Component>").'
+)
+_TEXT_DESCRIPTION = (
+    "Message body in Slack mrkdwn: *bold*, `code`, and <url|label> for links. "
+    "Keep it to the few lines the channel needs, with links out for detail."
+)
```
suhaibmujahid: I would embed these directly in `Field(description=...)`

suhaibmujahid (approved):
Looks good! Just minor comments.

evgenyrp:
Done, I also changed the prompt so that the summary message is smaller

PR #6524, opened 2026-08-05
Test repair agent improvements

evgenyrp:
This addresses feedback from @Archaeopteryx and @marco-c (the agent side of it)

- do not build Firefox by default to save time
- widen commit range
- inspect a known bug for the intermittent issue
- remove land-fix action as sheriffs don't do this
- simplify prompt and docs


agents/test-repair/README.md
```diff
@@ -1,68 +1,67 @@
...
+The agent gets the range (`base..head`), not a list of shas, and narrows it itself
+with `git log`. When the range is not known to reach a green run it is passed as
+`HEAD~N..HEAD` and the prompt stops asserting the culprit is in it.
+
+## Stages
+
+Stage 1 (analysis) is read-only and always runs. Stage 2 (fix) runs whenever
+stage 1 blamed a commit.
+
+By default the fix stage does not build (`SKIP_FIREFOX_BUILD`): the patch is
```
marco-c, line 33: I would actually split the two stages, so we can send a notification about backouts as soon as possible, and we can work on the fix with full verification (build + running the test that failed to verify it's actually fixed). With no verification, it would be very noisy

evgenyrp: We can land the current improvements and do that as a follow up as it also requires splitting the fix planning stage and a mechanism to trigger the second part of the work. Disabling firefox build makes fixing time negligible. Also it already doesn't run when it can't find the culprit.

evgenyrp: Also, we don't notify the developers yet, so the patch will be just for reference, and I doubt it will add much noise beyond the analysis at this point

agents/test-repair/hackbot_agents/test_repair/prompts.py
```diff
@@ -5,138 +5,127 @@
...
+2. Enumerate the candidates with `git log --oneline {commit_range}`. Path
+   filtering on the failing test and the source it exercises tells you what to
+   `git show` first, but never clears anyone: build config, shared headers,
+   toolchain bumps, manifest and harness changes, and changes that only shift
+   timing all break tests they do not touch. Work through the rest of the list
+   before concluding nothing explains the failure.
+3. Search Bugzilla for an intermittent bug tracking this failure -- the test name,
+   or its file name plus "intermittent", finds most, including per-test "single
+   tracking bug" entries. If one matches, the classification is "intermittent" and
+   its id goes in "intermittent_bug", however suspicious a commit looks. Search
```
marco-c, line 42: in some cases a patch might be causing an intermittent to become a real failure, maybe the agent should be able to suggest "rerun" in those cases to confirm? (follow-up)

agents/test-repair/hackbot_agents/test_repair/prompts.py
```diff
@@ -5,138 +5,127 @@
...
-`{commit_range}` is the {span} most recent commits. This range is NOT known to
-reach back to a run where the test was green, so the culprit may predate it -- if
-nothing in it plausibly caused the failure, say so in your analysis, set
-"culprit_commit" to null and leave "candidate_commits" empty rather than naming
-the least implausible commit in the range."""
-
-# Explicitly an hg revision: everything else in the prompt is a git hash, and an
-# unlabelled hg node just makes the agent try `git` commands that exit 128.
+`{commit_range}` is the {span} most recent commits, and is NOT known to reach back
+to a green run, so the culprit may predate it. If nothing in it plausibly caused
```
marco-c, line 78: What if we always give a range up to the last passing run?

marco-c (approved):

agents/test-repair/hackbot_agents/test_repair/prompts.py
evgenyrp: Interesting. We'll need to analyze how often that happens and adjust.

agents/test-repair/hackbot_agents/test_repair/prompts.py
evgenyrp: Good question. I think we still need some limits because walking pushes and checking group_summaries takes time (can be minutes) + we do the shallow clone manually, so if it's infinite, the clone can take a while. We value fast processing here to notify sheriffs early. Let's bump the numbers significantly and see how it works in practice and if we ever hit the situation when we couldn't find the last green in these limits.

evgenyrp: Bumped significantly, let's start retesting it and see if it ever doesn't reach the last green and how much time it really takes to walk and clone

PR #6525, opened 2026-08-06
Improve Phabricator base-commit resolution

### LINKED ISSUE #6415: Failing to resolve the full commit hash for checkout on follow-up
This is slimier to #6412, but here it is the first patch in the stack, so it should not be a problem. However, it seems to be failing to resolve the commit hash to the full one.
```
CalledProcessError: Command '['git', '-C', '/workspace/firefox', 'fetch', '--depth=2', 'origin', '0397cc0f5dca']' returned non-zero exit status 128.
```

suhaibmujahid:
Fixes #6415

### DIFF
--- a/agents/bug-fix/hackbot_agents/bug_fix/broker.py
+++ b/agents/bug-fix/hackbot_agents/bug_fix/broker.py
@@ -74,12 +74,12 @@ def _patch_endpoint(client: PhabricatorClient):
             )
         raw_diff = await client.get_raw_diff(diff.id)
         # The recorded base is often an abbreviated hash; git can only fetch a
-        # full object id, so expand it here (falling back to the raw value).
+        # full object id, so expand it here.
         try:
             base_commit = await client.resolve_commit(diff.base_commit)
         except UnresolvedCommitError as exc:
-            log.warning("Base commit of D%s stays unexpanded: %s", revision_id, exc)
-            base_commit = diff.base_commit
+            return JSONResponse({"error": str(exc)}, status_code=422)
+
         return JSONResponse({"base_commit": base_commit, "raw_diff": raw_diff})
 
     return get_patch
--- a/agents/bug-fix/hackbot_agents/bug_fix/broker.py
+++ b/agents/bug-fix/hackbot_agents/bug_fix/broker.py
@@ -71,7 +71,15 @@ def _patch_endpoint(client: PhabricatorClient):
         raw_diff = await client.get_raw_diff(diff.id)
         # The recorded base is often an abbreviated hash; git can only fetch a
         # full object id, so expand it here (falling back to the raw value).
-        base_commit = await client.resolve_commit(diff.base_commit) or diff.base_commit
+        base_commit = await client.resolve_commit(diff.base_commit)
+        if not base_commit:
+            log.warning(
+                "Phabricator could not expand base commit %s of D%s; "
+                "returning it unexpanded",
+                diff.base_commit,
+                revision_id,
+            )
+            base_commit = diff.base_commit
         return JSONResponse({"base_commit": base_commit, "raw_diff": raw_diff})
 
     return get_patch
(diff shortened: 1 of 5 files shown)


libs/phabricator-client/phabricator_client/client.py
```diff
@@ -157,8 +157,24 @@ async def resolve_commit(self, ref: str) -> str | None:
...
+
+        identifiers = {
+            commit["identifier"]
+            for commit in commits
+            if commit["identifier"].startswith(ref)
+            and _is_full_commit(commit["identifier"])
+        }
+
+        if len(identifiers) != 1:
             return None
```
marco-c: Can't we just use the first one if there's more than one?

agents/bug-fix/hackbot_agents/bug_fix/broker.py
```diff
@@ -71,7 +71,15 @@ async def get_patch(request):
         raw_diff = await client.get_raw_diff(diff.id)
         # The recorded base is often an abbreviated hash; git can only fetch a
         # full object id, so expand it here (falling back to the raw value).
-        base_commit = await client.resolve_commit(diff.base_commit) or diff.base_commit
+        base_commit = await client.resolve_commit(diff.base_commit)
+        if not base_commit:
+            log.warning(
```
marco-c: if not expanding is going to break things, maybe we can make it an error?
or, we could use the lando api to expand (e.g. https://lando.moz.tools/api/git2hg/firefox/0397cc0f5dca).

libs/phabricator-client/phabricator_client/client.py
marco-c: And post a warning so we can see what's wrong on phab

libs/phabricator-client/tests/test_client.py
```diff
@@ -248,6 +248,122 @@ async def test_resolve_commit_expands_abbreviated_hash(monkeypatch):
...
+            }
+        },
+    )
+    # The unfetchable short id is dropped, leaving one usable expansion.
+    assert await _client().resolve_commit("0397cc0f5dca") == full
+
+
+async def test_resolve_commit_returns_none_when_ambiguous(monkeypatch):
+    # Two different commits share the abbreviation: expanding to either could
+    # check the agent out on the wrong tree, so refuse to guess.
```
marco-c, line 356: This is the case I'm talking about: maybe better to guess, post a warning so we can inspect the case (which should never happen)

marco-c (dismissed):

libs/phabricator-client/phabricator_client/client.py
suhaibmujahid: This is a rare case, and it might never happen. But when it happens, using a random hash would make things harder to catch; the patch will be applied of the wrong parent commit.

A warning would be helpful here.

libs/phabricator-client/tests/test_client.py
suhaibmujahid: I agree with the warning, but guessing here would make it hard to catch the bug if happened.

agents/bug-fix/hackbot_agents/bug_fix/broker.py
suhaibmujahid: Make since! 

For the Lando part, I don't anticipate a case where expansion would fail with Phab but succeed with Lando. So, for now, I'd suggest we simply return an error in that scenario. If we do encounter that error in practice, we can then check whether it's expandable via Lando and address it at that point by adding Lando as a fallback. This way, we avoid introducing extra complexity for a case that may never actually occur.

libs/phabricator-client/phabricator_client/client.py
suhaibmujahid: Improved in f3c32dd47ed85ec5eb37730b70c97bd1be3ccf1f, we raise an error now, which is more obvious than a warning. And we will get a useful error message instead of `exit status 128`.

agents/bug-fix/hackbot_agents/bug_fix/broker.py
suhaibmujahid: Improved in 66f9b73a2237b0553f5283f530dab422eaadd127, the broker now return 422 error.

marco-c (approved):

PR #6530, opened 2026-08-06
frontend-triage: simplify and refresh the README

msujaws:
The `frontend-triage` README had grown into a narrative doc. Roughly 30 lines
explained why Searchfox permalinks resolve from the *index* revision rather
than the checkout — which the module docstring in
`libs/hackbot-runtime/hackbot_runtime/searchfox.py` already covers in full —
plus similar standalone rationale for model pinning and the read-only design.
This trims it toward the terser style of the `build-repair` and `test-repair`
READMEs and folds the safety bullets into a single section about what actually
reaches Bugzilla.

It also fixes three stale points:

- **Feedback channel.** The README predates #6519, so it never mentioned the
  runtime's 👍/👎 reaction footer or the four `ai-triage-*` tags appended by
  `feedback_tags_hook`. Every needinfo reference is gone.
- **`BROKER_URL`** was missing from the inputs table even though `AgentInputs`
  requires it. The table now also makes clear that `BUGZILLA_API_URL` /
  `BUGZILLA_API_KEY` are consumed by the broker container only.
- **Recorded actions are not auto-applied** for this agent — a human applies
  them from the Hackbot UI.

Smaller corrections: `rules/` and `prompts/` live under
`hackbot_agents/frontend_triage/`, not at the agent root, which the old text
left ambiguous; and a closing line notes the agent has no unit tests of its own.

Net 65 insertions / 81 deletions, one file, prettier-clean.

## Conflicts with #6439

#6439 ("Auto-apply high-confidence frontend-triage results and report them to
Slack") edits the same README section and inverts the auto-apply claim above.
Whichever lands second will need a fixup — the wording here is correct against
`master` today, and #6439's own README hunk already describes the
`auto_apply_confidence` behavior that supersedes it.

## Verification

Documentation only. Claims were checked against `config.py`, `__main__.py`,
`agent.py` (`feedback_tags_hook`), `hackbot_runtime/actions/bugzilla.py`
(`_COMMENT_FOOTER`), `compose.yml` (`BROKER_URL`), and
`services/hackbot-api/app/agents.py`. The agent was not run end to end, so the
footer and tag line are confirmed by reading the hooks rather than by
inspecting a real `summary.json`.

### DIFF
--- a/agents/frontend-triage/README.md
+++ b/agents/frontend-triage/README.md
@@ -5,50 +5,30 @@ analysis plus a proposed fix plan**. It reads the source tree, navigates the
 codebase with Searchfox, and inspects regressor changesets on hg.mozilla.org. It
 does **not** build Firefox, edit source, reproduce the bug, or write to Bugzilla.
 
-Think of it as first-pass triage on a UI papercut: read the bug, find the
-responsible code, explain the likely cause, propose a fix — then hand that to a
-human or an execution agent to implement and verify. It deliberately stops at a
-plan, because visual and interaction bugs can't be verified by the
-crash-reproduction loop the [`bug-fix`](../bug-fix/) agent relies on.
+It deliberately stops at a plan: visual and interaction bugs can't be verified by
+the crash-reproduction loop the [`bug-fix`](../bug-fix/) agent relies on, so a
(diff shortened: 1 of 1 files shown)


crythms (approved):
Thanks

PR #6531, opened 2026-08-06
hackbot: add a try-server push action

### LINKED ISSUE #6528: Add a new agent action to push to try server
This should be an action like we do with Bugzilla and Phabricator.

We should use Lando API for this, with that we need to prepare the patch in the format that the API expects. We could do it in the agent itself if needed, similar to how we do with when we want to submit a patch to Phabricator. Then the submission happens in hackbot-api.

Here's the doc for the try_api https://lando.moz.tools/api/try/docs#/ and here is a dirty way to create a patch from a diff: https://github.com/mozilla-conduit/lando/blob/af02dce67bc4c5906ecaeb703d8a0b7f30c62592/src/lando/conftest.py#L360-L386

suhaibmujahid:
Resolves #6528 

Depends on https://github.com/mozilla-conduit/lando/pull/1408

### DIFF
--- a/libs/hackbot-runtime/hackbot_runtime/actions/handlers/try_server_handler.py
+++ b/libs/hackbot-runtime/hackbot_runtime/actions/handlers/try_server_handler.py
@@ -31,11 +31,6 @@ _COMMIT_BODY = "Pushed via hackbot."
 # the push to the authenticated user), but it keeps the commit identifiable.
 _AUTHOR = "Hackbot <hackbot@mozilla.tld>"
 
-# `git format-patch`'s version-info trailer. Lando's parser needs it: it finds
-# the end of the diff by scanning back for the "--" barrier and raises
-# "Malformed patch" without one. Built by concatenation because the barrier
-# carries a trailing space that an editor or linter would strip from a literal.
-_PATCH_TRAILER = "-- " + "\n2.51.0\n"
 
 # One `git format-patch` email adding `try_task_config.json`. The diffstat block
 # a real format-patch puts after "---" is omitted: Lando skips everything
@@ -55,6 +50,7 @@ new file mode 100644
 +++ b/{filename}
 @@ -0,0 +1,{line_count} @@
 {added_lines}
+--
 """
 
 
@@ -157,7 +153,7 @@ def try_task_config_patch(
         line_count=len(added_lines),
         added_lines="\n".join(f"+{line}" for line in added_lines),
     )
-    return (patch + _PATCH_TRAILER).encode()
+    return patch.encode()
 
 
 def _commit_title(title: str | None) -> str:
--- a/libs/hackbot-runtime/hackbot_runtime/actions/handlers/try_server_handler.py
+++ b/libs/hackbot-runtime/hackbot_runtime/actions/handlers/try_server_handler.py
@@ -29,7 +29,7 @@ _COMMIT_BODY = "Pushed via hackbot."
 
 # Author of the generated try_task_config.json commit. Cosmetic (Lando attributes
 # the push to the authenticated user), but it keeps the commit identifiable.
-_AUTHOR = "Hackbot Agent <hackbot@mozilla.tld>"
+_AUTHOR = "Hackbot <hackbot@mozilla.tld>"
 
 # `git format-patch`'s version-info trailer. Lando's parser needs it: it finds
 # the end of the diff by scanning back for the "--" barrier and raises
--- a/libs/hackbot-runtime/hackbot_runtime/actions/handlers/try_server_handler.py
+++ b/libs/hackbot-runtime/hackbot_runtime/actions/handlers/try_server_handler.py
@@ -1,0 +1,170 @@
+"""Apply-side try-server action: push an already-built patch series to Lando."""
+
+from __future__ import annotations
+
+import json
+import logging
+from datetime import datetime, timezone
+from email.utils import format_datetime
+from functools import lru_cache
+from typing import Any
+
+from lando_client import LandoClient, encode_patch
+
+from hackbot_runtime.actions.handlers.base import ActionResult, ApplyContext
+
+log = logging.getLogger(__name__)
+
+_TRY_PUSH_ARTIFACT_KEY = "changes/try_push.json"
+
+# Which Try repository the push lands on, and the Treeherder repo it shows up
+# under. Firefox's plain "try"; a Thunderbird agent would need
+# "try-comm-central" here (and a matching Lando permission).
+_TRY_REPO_NAME = "try"
+
+_TRY_CONFIG_FILENAME = "try_task_config.json"
+
+_DEFAULT_TITLE = "Hackbot try push"
+_COMMIT_BODY = "Pushed via hackbot."
+
+# Author of the generated try_task_config.json commit. Cosmetic (Lando attributes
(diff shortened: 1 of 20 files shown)


libs/hackbot-runtime/hackbot_runtime/actions/try_server.py
```diff
@@ -0,0 +1,95 @@
...
+        Field(
+            description=(
+                "Treeherder task labels to run, e.g. ['build-linux64/opt', "
+                "'test-linux2404-64/opt-mochitest-browser-chrome-1']. Only the "
+                "tasks that actually exercise your change: every extra label "
+                "spends build machine time. Must not be empty — there is no "
+                "'run everything' shorthand."
+            )
+        ),
+    ],
```
marco-c: Might be better to have the agent select tests and not tasks, or use "mach try auto" (maybe in combination with agent's selected tests).

libs/hackbot-runtime/hackbot_runtime/actions/try_server.py
```diff
@@ -0,0 +1,95 @@
...
+    you cannot build, or a test suite you cannot run. It does not deliver a fix:
+    to submit code for review, use ``submit_patch`` (a try push and a revision
+    are independent, so a run may reasonably do both).
+
+    You do not supply a patch file, and you do not need to touch
+    ``try_task_config.json``: your final code changes in the working directory
+    are pushed as-is and the task selection is built from ``tasks``, so make and
+    verify all your edits first, then call this once you are done. Calling it
+    records the push as a proposed action for review; nothing is pushed during
+    the run, so you will not see the results — a human reads them on Treeherder.
```
marco-c: This is until we have a way to have hackbot continue after a try push (or other action). Are you planning to do that after as a follow-up?

suhaibmujahid: Yes, I'm planning to do that. I will file some issues in that regard.

libs/hackbot-runtime/hackbot_runtime/actions/try_server.py
suhaibmujahid: Done in 4fd3c9fc205d66d7316b2d7f251cbbdb5b833f07

libs/hackbot-runtime/hackbot_runtime/actions/try_server.py
```diff
@@ -0,0 +1,152 @@
...
+
+TRY_PUSH_ACTION_TYPE = "try_server.push"
+
+# Anything gated on "this run pushes to try" — today the patch-series artifact
+# built in ``context.publish_changes`` — keys off this set, for symmetry with
+# ``phabricator.PATCH_ACTION_TYPES``.
+TRY_ACTION_TYPES = frozenset({TRY_PUSH_ACTION_TYPE})
+
+
+def validate_test_paths(tests: dict[str, list[str]]) -> dict[str, list[str]]:
```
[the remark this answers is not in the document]
suhaibmujahid: Fixed in b7dfdbef494f97b004f03baf5e384747eff41b7e

libs/hackbot-runtime/hackbot_runtime/actions/handlers/try_server_handler.py
```diff
@@ -0,0 +1,229 @@
...
+
+# Author of the generated try_task_config.json commit. Cosmetic (Lando attributes
+# the push to the authenticated user), but it keeps the commit identifiable.
+_AUTHOR = "Hackbot <hackbot@mozilla.tld>"
+
+# `git format-patch`'s version-info trailer. Lando's parser needs it: it finds
+# the end of the diff by scanning back for the "--" barrier and raises
+# "Malformed patch" without one. Built by concatenation because the barrier
+# carries a trailing space that an editor or linter would strip from a literal.
+_PATCH_TRAILER = "-- " + "\n2.51.0\n"
```
evgenyrp: What is this hardcoded version? Should it be the version of the actual git tool? I assume Lando needs it for a reason.

libs/hackbot-runtime/hackbot_runtime/actions/handlers/try_server_handler.py
```diff
@@ -0,0 +1,229 @@
...
+
+# One `git format-patch` email adding `try_task_config.json`. The diffstat block
+# a real format-patch puts after "---" is omitted: Lando skips everything
+# between the commit message and the first "diff " line, and `git apply` derives
+# the stat from the diff itself.
+_PATCH_TEMPLATE = """\
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
```
evgenyrp, line 40: This is weird. Are there some assumptions that this is not used at the moment? Why not to put something meaningful there?

libs/lando-client/lando_client/client.py
```diff
@@ -0,0 +1,121 @@
...
+        """
+        payload = {
+            "repo_name": repo_name,
+            "base_commit": base_commit,
+            "base_commit_vcs": base_commit_vcs,
+            "patch_format": patch_format,
+            "patches": patches,
+        }
+        async with httpx.AsyncClient(timeout=self.settings.timeout_seconds) as client:
+            response = await client.post(
```
evgenyrp, line 80: I'm just curious: could we reuse the code from Firefox itself that does `./mach try` here instead of reimplementing it?

suhaibmujahid: I discussed that with the team. What I understood was that `./mach try` needs a checked-out repo, which we do not have here. The solution was to craft the submission in our code.

evgenyrp: That's what I thought, but I thought maybe there's a way to extract the lib from there. Ok.

libs/hackbot-runtime/hackbot_runtime/actions/handlers/try_server_handler.py
suhaibmujahid: What do you mean? Are you referring to `Mon Sep 17 00:00:00 2001`? That's a hardcoded value in git patches, if you run git format-patch -1 HEAD in any repo, you'll get the same date. The whole "From" line isn't actually used; it's just there to make the patch conform to the mbox format.

evgenyrp: Here's what I see `From 59137c9b2d8fb29ede7d377ae2a06ff331887fee Mon Sep 17 00:00:00 2001`. I guess I meant zeros as well. Ok, but not obvious it should be like that.

libs/hackbot-runtime/hackbot_runtime/actions/handlers/try_server_handler.py
suhaibmujahid: It is not used by Lando, dropped in 513913257a2f64e75d6eecd67fb74857af870532.

evgenyrp (approved):
Looks good! To double-check: will we be able to run it from the deterministic part of the agent as well?

suhaibmujahid:
> will we be able to run it from the deterministic part of the agent as well?

Yes, same as you do with Slack notifications.

BTW, this will not work before having https://github.com/mozilla-conduit/lando/pull/1408 deployed in production.

PR #6551, opened 2026-08-07
Autowebcompat diagnosis agent

ksy36:
This agent has the same browser / MCP setup as the autowebcompat-repro agent. 
I'm going to move the browser installation and MCP servers into a shared directory in a separate PR.

The agent has 3 tasks:

1. DiagnosisPlan
Creates a plan for diagnosis, picks the browser channel to install, locates the Puppeteer script if available, records url and STR.

2. ReproScript (If Puppeteer script exists and passes in Chrome/ fails in Firefox this step is skipped)
If the script doesn't exist or doesn't work as expected, it tries to reproduce the bug again and create the script.

3. Diagnosis
Does root-cause diagnosis, creates minimal testcase.

I've added reproduction and puppeteer script creation mainly so we could use it for opt-in diagnosis on bugzilla bugs that don't have prior runs of the repro agent. But we can certainly make it diagnosis only and run the repro agent first when diagnosis is requested. 
Also when sites change and previously created puppeteer script does not work, this agent would create it's own revised version of the script and have one more attempt to reproduce the bug, so it might be good to keep ReproScript anyway.

### DIFF
--- a/agents/autowebcompat-diagnosis/hackbot_agents/autowebcompat_diagnosis/agent.py
+++ b/agents/autowebcompat-diagnosis/hackbot_agents/autowebcompat_diagnosis/agent.py
@@ -1,0 +1,649 @@
+"""Firefox web-compatibility diagnosis agent."""
+
+from __future__ import annotations
+
+import logging
+import os
+import subprocess
+import tempfile
+from abc import ABC, abstractmethod
+from collections.abc import Callable
+from dataclasses import dataclass
+from datetime import datetime
+from enum import Enum
+from pathlib import Path
+from typing import Any, Generic, Literal
+
+from claude_agent_sdk import (
+    ClaudeAgentOptions,
+    ClaudeSDKClient,
+    McpServerConfig,
+    ResultMessage,
+)
+from hackbot_runtime import AgentError
+from hackbot_runtime.claude import Reporter
+from pydantic import BaseModel
+
+from .browser import ChromeBrowsers, FirefoxBrowsers
+from .config import BUGZILLA_READ_TOOLS, CHROME_DEVTOOLS_TOOLS, DEVTOOLS_TOOLS
+from .mcp_servers import build_chrome_devtools_server, build_firefox_devtools_server
+from .result import (
+    RESULT_SERVER_NAME,
+    SUBMIT_RESULT_TOOL,
+    DiagnosisPlanResult,
+    DiagnosisResult,
+    ReproScriptResult,
+    ResultCollector,
+    ResultT,
+    build_result_server,
+)
+
+HERE = Path(__file__).resolve().parent
+
+# Where the pinned npm deps (puppeteer, the DevTools MCP servers) are installed
+# in the image; the agent runs the reproduction script with this on NODE_PATH so
+# its `import puppeteer` resolves.
+WORK_DIR = Path("/app/diagnosis")
+NODE_MODULES = WORK_DIR / "node_modules"
+
+logger = logging.getLogger("autowebcompat-diagnosis")
+
+PublishFile = Callable[[str, Path, str | None], str]
+
+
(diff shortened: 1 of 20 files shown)


agents/autowebcompat-diagnosis/hackbot_agents/autowebcompat_diagnosis/agent.py
```diff
@@ -0,0 +1,651 @@
...
+   When a Bugzilla bug id is passed, request it explicitly: `cf_user_story` is not
+   in the default field set. Prefer `nightly` if it is in the list,
+   otherwise pick first listed channel. Default to `nightly` if there is no evidence
+   of the affected channel in the report.
+
+3. If a Puppeteer reproduction script is attached to the bug (an mjs file,
+   typically named `Reproduction script generated by autowebcompat bot`),
+   download it to exactly:
+     {self.script_path}.
+
```
ksy36, line 359: I'm thinking the bugzilla attachment link to the script (i.e. https://bug1952830.bmoattachments.org/attachment.cgi?id=9623008) could also be passed by the caller as an additional parameter instead of asking the agent to locate it.

jrmuizel:
@ksy36 let me know that this duplicates a bunch of code with the reproduction agent. She's going to factor that out next week when she gets back. I don't think we need to block landing this on that.

jrmuizel:
I looked over a diff of this vs. repro agent and it seemed fine. I tested it and it looks good. People are asking for it and I'm eager to get it deployed.

jrmuizel (approved):

suhaibmujahid (approved):

suhaibmujahid:
> @ksy36 let me know that this duplicates a bunch of code with the reproduction agent. She's going to factor that out next week when she gets back. I don't think we need to block landing this on that.

I believe some parts could be even facilitated by Hackbot directly, so we could move it there. Other things could be in reusable components in agent-tools. We should coordinate that once she is back.

jgraham (commented):
I don't mind if we merge this, if we want to get some diagnosis task deployed asap.

However reading this, I think we've picked the wrong high level approach by putting this in a separate agent.

One problem is that we're sharing a lot of library-type code (the Task class, the code to set up Chrome and Firefox, etc.). That could obviously be fixed by moving those things out into a shared library that could be accessed by both agents.

However, even with that refactor the requirement to handle bugs that don't have an existing repro run means that we're also putting a lot of the actual reproduction agent into this agent. That's more worrying because it means any changes we make to the approach to reproduction need to be duplicated across both agents.

As Ksenia mentions we could fix that at the orchestration layer: have the code first schedule a reproduction job, and if that's a success schedule a followup task to perform the diagnosis. That separation has some advantages e.g. each task only affects certain parts of the bug. But it also requires more complexity in the scheduling.

It seems like it might be easier to fold this into a single agent and have it take a "fill in the blanks" approach: if it's been asked for a diagnosis and there isn't already STR it runs through the reproduction tasks and then uses the outputs of those directly for diagnosis, otherwise it uses the cached diagnosis artifacts from the input data.

I don't think that's a huge refactor; as I say the code here is nearly all the same as the other agent anyway. But the scheduling logic will also be different for one agent vs two, so merging this as-is is going to cost us some additional work vs refactoring and landing the better long-term design.

jgraham:
It's also worth knowing that there isn't a PR up for the scheduling part of this, so even if we deploy this change today the agent won't get invoked from bugzilla; you'd have to use the webform to invoke it directly.

agents/autowebcompat-diagnosis/hackbot_agents/autowebcompat_diagnosis/agent.py
jrmuizel: We can do this as a follow up.

PR #6563, opened 2026-08-10
Add performance regresssion predictor inference feature

Ali-Sayed-Salehi:
Bugzilla:
https://bugzilla.mozilla.org/show_bug.cgi?id=2055853

# Add Performance Regression Predictor (Autoland push-based inference)

## Summary

Adds an inference-only model, an HTTP service endpoint, and a CLI that predict
whether a commit is likely to introduce a **performance regression**. The
service takes an Autoland/try push identified by `(branch, rev)` — the same
contract as the test-selection endpoint — and resolves everything server-side
from its own local Mercurial clone. Each commit in the push is scored
independently; the push-level risk is the **max** across commits.

## Motivation

Give CI/taskgraph an early signal for pushes that are likely to regress
performance, without plumbing diffs or credentials from Phabricator. Reusing the
`(branch, rev)` + local-clone approach that `schedule_tests` already uses keeps
the integration surface minimal and consistent.

## How it works

The worker (`classify_perf_regression(branch, rev)`) mirrors `schedule_tests`:

1. `repository.pull(REPO_DIR, branch, rev)` into the service's local hg clone.
2. `get_hgmo_stack(branch, rev)` loads the full push stack via
   `automationrelevance` (with `try`'s draft/empty/try-config filtering).
3. Each commit is exported from the clone (`hg export --git`) to get its message
   and diff.
4. Each commit is scored **one at a time** (CPU inference — avoids multiplying
   peak memory by stack size).
5. The response contains per-commit scores plus the max as the top-level
   `risk_score`.

The model is a HuggingFace CodeBERT sequence classifier (inference-only,
512-token window, binary). The input is the cleaned commit message plus the diff
rendered into a structured `<FILE>/<ADDED>/<REMOVED>` representation matching
training. The `risk_score` is the raw (uncalibrated) softmax probability of the
positive class.

## API

```
GET /perfregressionpredictor/predict/push/<branch>/<rev>
X-Api-Key: ...
```

`branch` is an hg.mozilla.org path (`integration/autoland`, `try`; `autoland` is
accepted as an alias). The first call returns `202 {"ready": false}`; poll until
`200`. A missing push returns `{"available": false}`.

Example result:

```json
{
  "branch": "integration/autoland",
  "rev": "76383a875678",
  "risk_score": 0.75,
  "commits": [
    {"node": "…", "prob": [0.25, 0.75], "class": 1, "risk_score": 0.75}
  ],
  "extra_data": {
    "model_name": "Perf Regression Predictor",
    "max_length": 512,
    "calibrated": false,
    "commit_count": 1
  }
}
```

## Also included

- **CLI** (`bugbug-predict-perf-regression`) for local inference against a
  checkpoint + patch file, with no service/Redis/clone required.
- **Model distribution**: the model is trained outside bugbug, so it declares a
  public `artifact_url`; `download_models` now supports fetching a model from a
  URL instead of the Taskcluster index.
- **Dependencies**: new optional `perf-regression-predictor` extra
  (`torch==2.13.0`, `transformers==5.15.0`, pinned to the CPU wheel index).
- **Docs**: model docs (`docs/models/perf-regression-predictor.md`) and the
  HTTP-service local-dev guide
  (`http_service/README.perf-regression-predictor.md`).

## Testing

Unit test:
```sh
cd /home/ali/repos/bugbug
python -m pytest tests/test_perf_regression_predictor.py http_service/tests/test_perf_regression_predictor.py
```

Via docker compose:
```sh
cd /bugbug/http_service

docker compose up --build \
  redis \
  bugbug-http-service \
  bugbug-http-service-bg-worker

docker compose exec bugbug-http-service-bg-worker \
  python -c "from bugbug import utils; from bugbug.models.perf_regression_predictor import PerfRegressionPredictorModel as M; utils.download_model_from_url('perfregressionpredictor', M.artifact_url)"


docker compose exec bugbug-http-service-bg-worker \
  test -f /code/perfregressionpredictormodel/config.json \
  && echo "Model is mounted"


docker compose logs -f bugbug-http-service-bg-worker   # wait for "Worker boot done"


curl --compressed -sS -w '\nHTTP %{http_code}\n' \
  -H "X-Api-Key: local-test" \
  "http://localhost:8000/perfregressionpredictor/predict/push/autoland/540bd3cda62fd69f9e38ed8053d938a5aeed21c0"


docker compose down
```


http_service/bugbug_http/app.py
```diff
@@ -522,6 +537,77 @@ def model_prediction(model_name, bug_id):
...
+          content:
+            application/json:
+              schema: NotAvailableYet
+        401:
+          description: API key is missing
+          content:
+            application/json:
+              schema: UnauthorizedError
+    """
+    if not request.headers.get(API_TOKEN):
```
Ali-Sayed-Salehi, line 599: Here we only check the existence of the API token and we are not doing any serious checks. This is the pattern other models use in the repo so the assumption is that when deployed in GCP, there is some other security feature in front of it. If not this may be a problem.

http_service/README.md
```diff
@@ -1,23 +1,207 @@
-### Local development
+# HTTP Service Local Development
```
gmierz: Undo these changes, and put the updates to the docs in a separate PR. Take the model-specific things, and add them to your models .md doc file.

http_service/docker-compose.yml
```diff
@@ -39,6 +39,8 @@ services:
       - REDIS_URL=redis://redis:6379/0
       - BUGBUG_ALLOW_MISSING_MODELS
       - BUGBUG_REPO_DIR
+      - PHABRICATOR_API_KEY
```
Ali-Sayed-Salehi, line 42: This is the first time we have a Phabricator API key for a HTTP service model. We already have PHABRICATOR_TOKEN but that is used by Taskcluster models and those models are deployed differently than http model. We also have PHABRICATOR_API_KEY in hackbot agents but then again those ones are deployed differently in Google Cloud Run. So for this feature which is a http model, we need to provide PHABRICATOR_API_KEY in the prod environment.

http_service/README.md
Ali-Sayed-Salehi: fixed

http_service/docker-compose.yml
Ali-Sayed-Salehi: Not relevant anymore since the API has changed and we don't use Phabricator anymore.

bugbug/models/perf_regression_predictor.py
```diff
@@ -0,0 +1,360 @@
...
+            0
+        ].strip()
+        message = "\n\n".join(part for part in (subject, body) if part)
+        return message or None
+
+    return None
+
+
+def diff_to_structured_text(diff_string: str) -> str:
+    """Convert a Git or Mercurial diff to the model's structured format.
```
suhaibmujahid: I would suggest a class fore each feature here, slimier to what we do with other features, example: https://github.com/mozilla/bugbug/blob/5bb16a1f19a5a36f87da1a2e97c2eaa95d6eddc1/bugbug/bug_features.py#L367-L369

http_service/bugbug_http/app.py
```diff
@@ -522,6 +547,90 @@ def model_prediction(model_name, bug_id):
     return compress_response(data, status_code)
 
 
+@application.route("/perfregressionpredictor/predict/push/<path:branch>/<rev>")
+@cross_origin()
+def perf_regression_prediction(branch: str, rev: str):
+    """
```
suhaibmujahid: why do we need a separate endpoint here, why not using an existing endpoint?

bugbug/tools/core/platforms/phabricator.py
suhaibmujahid, line 1: I do not think we use this in the http service?

http_service/bugbug_http/download_models.py
```diff
@@ -6,15 +6,21 @@
...
 def download_models():
-    for model_name in MODELS_NAMES:
-        utils.download_model(model_name)
+    for model_name in MODELS_TO_DOWNLOAD:
+        # Some models are published outside the Taskcluster index.
+        artifact_url = getattr(get_model_class(model_name), "artifact_url", None)
+        if artifact_url:
+            utils.download_model_from_url(model_name, artifact_url)
+        else:
+            utils.download_model(model_name)
```
suhaibmujahid: we could have the support for this in `utils.download_model()` instead of here. So we keep the `MODELS_NAMES` as it is, just adding the new model.

suhaibmujahid (changes requested):

bugbug/models/perf_regression_predictor.py
Ali-Sayed-Salehi: Fixed. I kept the newly-added classes stay beside the model because their output format is validated against this checkpoint's tokenizer, and they'll be extracted to a shared module if a second model ever needs them.

http_service/bugbug_http/app.py
Ali-Sayed-Salehi: No existing endpoint could serve a prediction by push keys, but the route is now generic over the model name so future push-based models plug into the registry with no new endpoints.

bugbug/tools/core/platforms/phabricator.py
Ali-Sayed-Salehi: Reverted these changes. This was remaining from the previous Phabricator logic that I had refactored.

http_service/bugbug_http/download_models.py
Ali-Sayed-Salehi: Moved the `artifact_url ` fallback into `utils.download_model()` so the loop here stays as it was. I didn't add the model to `MODELS_NAMES ` since that list also gates the bug endpoints (it would expose `/perfregressionpredictor/predict/<bug_id>`, which can't work), instead `MODELS_TO_DOWNLOAD` is now derived as `[*MODELS_NAMES, *PUSH_CLASSIFIERS]`.

bugbug/models/perf_regression_predictor.py
```diff
@@ -0,0 +1,366 @@
...
+
+        self.current_file = None
+        self.pending_binary_status = None
+        self.rename_from = None
+        self.rename_to = None
+        self.pending_rename = False
+
+    def __call__(self, diff_string: str) -> str:
+        self._reset()
+        started = False
```
suhaibmujahid: Why is this stateful? The same class will be used across different commits. It should be stateless, unless we need something as a cache, which is not really a state.

bugbug/models/perf_regression_predictor.py
```diff
@@ -0,0 +1,366 @@
...
+
+    def __init__(self, tokenizer: Any = None, transformer_model: Any = None) -> None:
+        super().__init__()
+        self.tokenizer = tokenizer
+        self.transformer_model = transformer_model
+        self.calculate_importance = False
+        self.model_directory: str | None = None
+        self.model_metadata: dict[str, Any] = {}
+        self.commit_message_cleaner = CommitMessageCleaner()
+        self.diff_structurer = DiffStructurer()
```
suhaibmujahid: Can we have the structure similar to the other models? The only difference here is that we do not need to train.

bugbug/models/perf_regression_predictor.py
Ali-Sayed-Salehi: fixed.

bugbug/models/perf_regression_predictor.py
Ali-Sayed-Salehi: Aligned it with the other models: plain no-arg __init__ (state is assigned by load()), an explicit train() raising NotImplementedError, and the same classify/load/get_extra_data signatures as the base class. The custom load/classify bodies have to stay, though since this is a HF checkpoint, not a pickled sklearn pipeline, so the base implementations (model.pkl, extraction_pipeline/clf) don't apply.

PR #6571, opened 2026-08-11
Support listing and deleting recorded agent actions

### LINKED ISSUE #6513: Support managing the reported agent actions
Add tools to view or remove any of the existing agent actions, so the agent can manage it's reported actions if needed. For example, if the agent later finds that a proposed action is inaccurate, it should be able to delete that action. Also, in a long context, some details might get lost, having a way to view the registered action helps with context retention.

ayoubdiourin7:
## Summary

- Add tools for agents to list and remove actions recorded during the current run.
- Assign stable action IDs so recorded actions can be referenced safely.
- Enable the new `recorded_actions` tools for the bug-fix and frontend-triage agents.
- Update prompts to explain when agents should review or retract recorded actions.
- Add tests covering action listing, removal, unknown IDs, attachments, and final runtime summaries.

Fixes #6513


agents/frontend-triage/hackbot_agents/frontend_triage/prompts/system.md
```diff
@@ -80,7 +80,9 @@ When you spawn an investigator via the Task tool, write a complete, self-contain
 
 # Recording actions
 
-The `actions` MCP tools (`bugzilla_add_comment`, `bugzilla_update_bug`) do **not** mutate Bugzilla directly. They record an intended action into the run's `summary.json` for a human reviewer (or a downstream apply step) to enact. Treat each recorded action as a final, irrevocable proposal.
+The `actions` MCP tools (`bugzilla_add_comment`, `bugzilla_update_bug`) do **not** mutate Bugzilla directly. They record an intended action for the run's `summary.json`, where a human reviewer or downstream apply step can enact it.
+
+Recorded actions remain proposals until the run finishes. Use `recorded_actions_list_actions` when you need to recover or verify their complete details. If you discover that a proposal is inaccurate, retract exactly that action with `recorded_actions_remove_action`, then record a corrected action if needed.
```
suhaibmujahid: Let us not use this in any agent yet; we could enable it later based on the need.

ayoubdiourin7: Adressed in [c582040](https://github.com/mozilla/bugbug/pull/6571/commits/c582040c521d9be962071b151a191f3523e06b90)

libs/hackbot-runtime/hackbot_runtime/actions/recorder.py
```diff
@@ -106,15 +117,49 @@ def record(
...
+        self._actions[action_id] = action
+        self._last_action_id = action_id
         return action
 
+    def list_actions(self) -> list[dict]:
+        """Return complete copies of the current actions with stable in-run IDs."""
+        return [
+            {**copy.deepcopy(action), "action_id": action_id}
+            for action_id, action in self._actions.items()
+        ]
```
suhaibmujahid, line 126: I would return MD table, with ID, Action (based on the tool name), and reasoning.

libs/hackbot-runtime/hackbot_runtime/actions/recorder.py
```diff
@@ -106,15 +117,49 @@ def record(
...
-                    f"attachments/{idx}/{name}",
+                    f"attachments/{sequence}/{name}",
                     path,
                 )
                 recorded_attachments.append({"name": name, "uploaded_key": key})
             action["attachments"] = recorded_attachments
 
-        self._actions.append(action)
+        self._actions[action_id] = action
+        self._last_action_id = action_id
```
suhaibmujahid: Using `_last_action_id` is a bit fragile, race conditions could miss up with that.

libs/hackbot-runtime/hackbot_runtime/actions/recorder.py
```diff
@@ -88,7 +97,9 @@ def record(
         recording leaves nothing behind: the action the hooks see carries no
         ``attachments`` key yet.
         """
-        idx = len(self._actions)
+        sequence = self._next_action_sequence
```
suhaibmujahid: Also, sequential IDs would be easy for the agent to guess and might remove things by accident. We could use UUIDs, which would be harder to guess.

libs/hackbot-runtime/hackbot_runtime/actions/bugzilla.py
```diff
@@ -27,7 +27,7 @@
 
 
 def _confirm(recorder: ActionsRecorder, action_type: str) -> str:
-    return f"Recorded {action_type} (#{len(recorder.actions) - 1})."
+    return f"Recorded {action_type} as {recorder.last_action_id}."
```
suhaibmujahid:
```suggestion
    return f"Recorded {action_type} (ID: {recorder.last_action_id})."
```

libs/hackbot-runtime/hackbot_runtime/actions/phabricator.py
```diff
@@ -33,7 +33,7 @@
 
 
 def _confirm(recorder: ActionsRecorder, action_type: str) -> str:
-    return f"Recorded {action_type} (#{len(recorder.actions) - 1})."
+    return f"Recorded {action_type} as {recorder.last_action_id}."
```
suhaibmujahid: Same as https://github.com/mozilla/bugbug/pull/6571/changes#r3797340769

libs/hackbot-runtime/hackbot_runtime/actions/recorded_actions.py
```diff
@@ -0,0 +1,47 @@
...
+
+@tool
+async def remove_action(
+    recorder: ActionsRecorder,
+    action_id: Annotated[
+        str,
+        Field(
+            description=(
+                "Exact stable ID returned when the action was recorded or by "
+                "list_actions (for example, action-2)."
```
suhaibmujahid: I would simplify it. Also, we do not need an example; it makes it a bit confusing since it does not look like the action IDs that will be in real cases.

libs/hackbot-runtime/hackbot_runtime/actions/recorded_actions.py
```diff
@@ -0,0 +1,47 @@
...
+            )
+        ),
+    ],
+) -> dict:
+    """Retract one proposed action from this agent run.
+
+    The removed action will not appear in the final run summary and cannot be
+    applied. This operation accepts exactly one action ID and has no cascade or
+    force mode.
```
suhaibmujahid: The shorter the better.

libs/hackbot-runtime/hackbot_runtime/actions/recorded_actions.py
```diff
@@ -0,0 +1,47 @@
...
+
+@tool
+async def list_actions(recorder: ActionsRecorder) -> list[dict]:
+    """List every action currently proposed by this agent run.
+
+    Returns each action's stable in-run ID and its complete recorded payload,
+    including parameters, reasoning, references, and attachment metadata. Use
+    this when earlier action details are no longer present in your context or
+    before deciding whether a proposal needs to be retracted.
+    """
```
suhaibmujahid: The shorter the better.

suhaibmujahid (changes requested):

libs/hackbot-runtime/hackbot_runtime/actions/recorder.py
```diff
@@ -106,15 +114,36 @@ def record(
...
+        return _detach(action_id, action)
+
+    def list_actions(self) -> list[dict]:
+        """Return complete copies of the current actions with stable in-run IDs."""
+        return [
+            _detach(action_id, action) for action_id, action in self._actions.items()
+        ]
+
+    def remove_action(self, action_id: str) -> dict:
+        """Remove and return an action."""
```
suhaibmujahid: The docstrings are not clear here.
```suggestion
        """Remove the action with the given ID and return a copy of it."""
```

libs/hackbot-runtime/hackbot_runtime/actions/recorded_actions.py
```diff
@@ -0,0 +1,52 @@
...
+
+@tool
+async def remove_action(
+    recorder: ActionsRecorder,
+    action_id: Annotated[
+        str,
+        Field(description="ID of the action to remove."),
+    ],
+) -> dict:
+    """Remove and return a recorded action."""
```
suhaibmujahid: The docstrings here targets the agent, so it should provide what should help the agent only. In this case, no need to mention the return part, and the returned value should be something that confirms the what happened instead of the removed action which could be a bit confusing.

libs/hackbot-runtime/hackbot_runtime/actions/slack.py
```diff
@@ -64,8 +64,8 @@ async def post_message(
 
     Recorded into the run summary for human review -- does not post to Slack.
     """
-    recorder.record(ACTION_TYPE, _params(channel, text), reasoning=reasoning)
-    return f"Recorded {ACTION_TYPE} (#{len(recorder.actions) - 1})."
+    action = recorder.record(ACTION_TYPE, _params(channel, text), reasoning=reasoning)
+    return f"Recorded {ACTION_TYPE} (ID: {action['action_id']})."
```
suhaibmujahid, line 74: We should file a follow-up issue to centralize the confirmation message.

libs/hackbot-runtime/hackbot_runtime/actions/action_records.py
suhaibmujahid, line 1: What if we call it `actions_manager` or something related instead of  `recorded_actions`? WDYT?

libs/hackbot-runtime/hackbot_runtime/actions/recorder.py
```diff
@@ -62,17 +67,18 @@ def record(
...
     ) -> dict:
-        """Record an intended action.
+        """Record an action and return a detached copy with its ID.
 
         ``action_type`` uses ``<domain>.<verb>`` (e.g. ``bugzilla.update_bug``,
         ``phabricator.create_revision``). ``params`` is action-specific data
         the apply step will need. ``attachments`` maps a logical name to a
         local file path; each file is preserved under the stable key
-        ``attachments/<action_index>/<name>``: uploaded via the runtime
+        ``attachments/<action_sequence>/<name>``: uploaded via the runtime
```
suhaibmujahid: It should be action id, we do not want to use sequence.

suhaibmujahid (commented):
I think you are still working on this, but here are some comments anyway.

libs/hackbot-runtime/hackbot_runtime/actions/action_records.py
ayoubdiourin7: `actions_manager` feels a little generic. What about `recorded_actions_manager` or `action_records`? However, I’m also fine with `actions_manager`.

libs/hackbot-runtime/hackbot_runtime/actions/recorded_actions.py
ayoubdiourin7: updated in  [051c813](https://github.com/mozilla/bugbug/pull/6571/commits/051c8138fc9410bcecf5b83ec9c0fb2caff59920)

libs/hackbot-runtime/hackbot_runtime/actions/recorder.py
ayoubdiourin7: Updated in [0da937d](https://github.com/mozilla/bugbug/pull/6571/commits/0da937d4715f781d62608324f8c8236e2fa6b1f4)

libs/hackbot-runtime/hackbot_runtime/actions/slack.py
ayoubdiourin7: Filed #6697

libs/hackbot-runtime/hackbot_runtime/actions/action_records.py
suhaibmujahid: I agree. Let's go with `action_records`.

suhaibmujahid: BTW, the file name should match to keep things consistent.

ayoubdiourin7: Done in [26d71c4](https://github.com/mozilla/bugbug/pull/6571/commits/26d71c48ea933c4ada691860bdbb32477d8cd734)

suhaibmujahid (approved):

PR #6577, opened 2026-08-12
Report an auto-applied frontend triage to the owning team's Slack channel

msujaws:
`5cf03b94` made frontend-triage apply its own results at `confidence: high`, so a triage comment and a `severity`/`keywords` change now reach a real bug with nobody in between. Nothing tells the team that owns the component it happened. Bugmail says the bug changed; it does not say an unreviewed agent is what changed it, which is the case worth watching while auto-apply is new.

## Change

An auto-applied run reports two lines to the channel of the team that owns the bug's component:

```
*<https://bugzilla.mozilla.org/show_bug.cgi?id=1968342|Bug 1968342> — Weather widget shows the previous city after changing location*
<https://hackbot.moz.tools/runs/abc123|frontend-triage run details>
```

An `S1` assessment adds a `:red_circle:` and names the level next to it, so it still reads as S1 in a client that does not render the emoji:

```
:red_circle: *<…|Bug 1971004> — New Tab crashes on startup when a pinned tile has no favicon* (S1)
<…|frontend-triage run details>
```

Nothing else is in the message. The channel already says which product and component this is, the analysis is on the bug, and the detail is in the run. Confidence is left out too, since only a `high` run gets this far. The severity reported is the one the run *judged*, which is not always the one it wrote to the bug — `rules/severity-assessment.md` holds the field change back when severity confidence is not high, and an S1 the agent only suspects is still worth flagging.

Only an auto-applied run reports. A medium or low result wrote nothing to Bugzilla, so there is nothing to tell anyone; it stays recorded and visible in the Hackbot UI as before.

## Reuse

Recorded as a `slack.post_message` action rather than posted from the run, so this reuses everything #6523 built for test-repair: visible in the UI before it lands, delivered at most once even though Pub/Sub is at-least-once, and no new secret or service code. **Nothing outside `agents/frontend-triage/` changes** — no `libs/`, no `services/`, no `pyproject.toml` (`slack-sdk` is already a `hackbot-runtime` dependency and the recording side does not import it).

`slack.post_message` is deliberately left out of `ENABLED_ACTION_TYPES`. The wording is code, in `notify.py`, so the agent is never given the tool and has no say in what is said or where. `record_message` goes through the recorder directly.

The existing `hooks.py` limits need no change: hooks are registered per action type, so `_check_only_one` cannot see a Slack action.

## Routing

`SLACK_CHANNELS` in `config.py` maps `"<Product> :: <Component>"` to a channel — today only `Firefox :: New Tab Page` → `#hnt-dev`. A component that is not listed sends nothing, and there is deliberately no default: posting one team's triage into another team's channel is worse than silence. Other frontend components stay silent until someone adds a line, which is an opt-in rather than a rollout to teams that have not asked for it.

Product and component come from the agent's plan block, because nothing else carries them out of a run whose only input is a bug id — `FrontendTriageInputs` is just `bug_id`, and the auto-apply decision point sees neither. `system.md` tells the agent to copy both verbatim from Bugzilla, and `parse_plan` drops a non-string rather than failing the whole result over a field that only picks a channel. A garbled value therefore matches no team and notifies nobody.

## Ordering

The Slack action is recorded last, so it has the highest `idx` and applies after the Bugzilla writes it reports. Its type differs from theirs, so `plan_coalesced_groups` will not pull it into the single `PUT /bug/{id}` that merges the comment with the field change — the one-bugmail behavior is unaffected.

No `{{actions.<ref>.<field>}}` placeholder is used. `AddCommentHandler` returns only `{bug_id, url}` where `url` is the plain `show_bug.cgi?id=<id>` page, which the message already builds itself, so a `ref` would buy nothing — and `_apply_pending_rows` drops a coalescing group whose rows carry a `ref`, so it would cost the merged PUT.

## Also fixed

Two stale README claims that predate this, fixed because the new section lands next to them: the summary said the agent never writes to Bugzilla, which the section 100 lines below it contradicts, and Registration said there are no unit tests, which stopped being true when `test_hooks.py` and `test_plan.py` landed.

## Testing

39 pass in `agents/frontend-triage/tests`, up from 28. New `tests/test_notify.py` has 11: the two message shapes line by line, a lowercase `s1` still marked (the value comes out of the agent's free-form JSON block), S2/S3/S4 and a null assessment unmarked, a missing summary degrading to the bug link alone, the channel lookup including whitespace and each half being absent, and the three cases that must record nothing — not auto-applied, unowned component, null component. `tests/test_plan.py` gains 2 for `product`/`component` round-tripping and for a non-string being dropped.

```sh
uv run --package hackbot-agent-frontend-triage pytest agents/frontend-triage/tests -q   # 39 passed
uv run --package hackbot-runtime pytest libs/hackbot-runtime/tests/test_slack_actions.py \
                                       libs/hackbot-runtime/tests/test_slack_handler.py -q   # 11 passed
```

`services/hackbot-api/tests/test_actions_applier.py` has 3 failures — `test_coalesces_update_and_comment_into_one_put`, `test_extra_comments_applied_separately`, `test_backward_placeholder_resolves_in_coalesced_comment`. They fail identically with this branch's changes stashed, so they are pre-existing on master and nothing here touches `services/`.

I have not yet confirmed a live post to `#hnt-dev`. The app has been invited to the channel, which settles `chat:write.public` and `not_in_channel`; what is untested is whether the channel restricts who may post.

A rejected Bugzilla `PUT` still notifies. `_apply_pending_rows` applies each action independently, so if the write is refused — lost `editbugs`, a bug that went security-restricted mid-run — the message still goes out and reads as though triage landed. A reader clicking through sees no comment, and the run page shows the failed action. Making one action conditional on an earlier one's result is not something the applier can express today, and I did not think it worth building for this.

A manual apply posts too. Only a `high`-confidence run records the action, but if such a run's actions were held (auto-apply switched off for the agent) or retried by hand, the post happens then. The message says what was triaged, not that nobody reviewed it, so it stays accurate.

`SLACK_BOT_TOKEN` is not configured anywhere in this repo; it is attached to the hackbot-api Cloud Run service out of band and test-repair already depends on it. If it is unset the Slack row fails, visibly, without touching the Bugzilla rows.

`test-repair` and `bug-fix` are untouched. So is the severity-confidence gap noted in `rules/severity-assessment.md` — it predates this and belongs with the hooks, not the notification.

### DIFF
--- a/agents/frontend-triage/README.md
+++ b/agents/frontend-triage/README.md
@@ -3,7 +3,9 @@
 Triages a Firefox desktop frontend bug from Bugzilla and produces a **root-cause
 analysis plus a proposed fix plan**. It reads the source tree, navigates the
 codebase with Searchfox, and inspects regressor changesets on hg.mozilla.org. It
-does **not** build Firefox, edit source, reproduce the bug, or write to Bugzilla.
+does **not** build Firefox, edit source, or reproduce the bug. It writes to
+Bugzilla only after the run, and only when it rated itself confident — see [What
+it writes back to Bugzilla](#what-it-writes-back-to-bugzilla).
 
 It deliberately stops at a plan: visual and interaction bugs can't be verified by
 the crash-reproduction loop the [`bug-fix`](../bug-fix/) agent relies on, so a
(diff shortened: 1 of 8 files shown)


crythms (approved):
Looks good!

PR #6583, opened 2026-08-13
Sanitize Phabricator summary headers

### LINKED ISSUE #6547: Hackbot is failing to submit a patch because of the commit message content
Example https://hackbot.moz.tools/runs/e6a23751-69dc-4a60-a60a-6d18ea717cae
```
Conduit error ERR-CONDUIT-CORE: Validation errors: - The value you have entered in "Summary" can not be parsed unambiguously when rendered in a commit message. Edit the message so that keywords like "Summary:" and "Test Plan:" do not appear at the beginning of lines. Parsed keys: testPlan.
```

ayoubdiourin7:
## Bug

Hackbot fails to submit a patch when the summary contains a line starting with `Tests:`. Phabricator treats `Tests` as an alias for the `Test Plan` field, so the summary is parsed ambiguously and rejected.

## Reproduction

Reproduced on `phabricator-dev.allizom.org`:

- `Tests: ...` is rejected with `ERR-CONDUIT-CORE ...`.
- ` Tests: ...` is accepted.

## Solution

Add one leading space to lines starting with a Test Plan alias, such as Tests:. This prevents Phabricator from parsing them as fields. Apply it to revision summaries and generated commit metadata.

Fixes #6547

### DIFF
--- a/libs/hackbot-runtime/hackbot_runtime/actions/handlers/phabricator_handler.py
+++ b/libs/hackbot-runtime/hackbot_runtime/actions/handlers/phabricator_handler.py
@@ -93,9 +93,23 @@ Bug #: {bug_id}
 # in the visible title only adds cleanup work when promoting to review.
 _WIP_PREFIX_RE = re.compile(r"^(?:WIP[: ]|WIP$)", re.IGNORECASE)
 
+# Phabricator reparses summaries as commit messages. Test Plan aliases at the
+# start of a line are interpreted as fields, making the summary ambiguous.
+_PHABRICATOR_TEST_PLAN_HEADER_RE = re.compile(
+    r"^(?=(?:Test Plan|Testplan|Tested|Tests):)",
+    re.IGNORECASE | re.MULTILINE,
+)
+
 
 def _revision_title(title: str) -> str:
     return _WIP_PREFIX_RE.sub("", title).strip()
+
+
+def _sanitize_summary(summary: str | None) -> str | None:
+    """Indent lines Phabricator would parse as Test Plan field headers."""
+    if not summary:
+        return summary
+    return _PHABRICATOR_TEST_PLAN_HEADER_RE.sub(" ", summary)
 
 
 async def _revision_fields(revision_id: int) -> dict:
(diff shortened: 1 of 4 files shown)


libs/hackbot-runtime/hackbot_runtime/actions/handlers/phabricator_handler.py
```diff
@@ -152,7 +166,7 @@ def _arc_commit_message(title: str, summary: str | None, bug_id: Any, url: str)
     reconstructed commit reads identically to a moz-phab submission. Reviewers
     are always empty: hackbot never assigns them (draft submissions omit them).
     """
-    body = summary or ""
+    body = _sanitize_summary(summary) or ""
```
suhaibmujahid: We might do this validation when recording the action with a useful error message, so the agent can correct that. That way, the content should always be good at the apply time.

suhaibmujahid (changes requested):

libs/hackbot-runtime/hackbot_runtime/actions/phabricator.py
```diff
@@ -48,7 +67,14 @@ async def submit_patch(
...
-        Field(default=None, description="Revision summary/description."),
+        Field(
+            default=None,
+            description=(
+                "Revision summary/description. Do not start a line with a Test "
+                "Plan header or alias such as Tests:, Test Plan:, Testplan:, or "
+                "Tested:; indent or rephrase it."
+            ),
+        ),
     ] = None,
```
suhaibmujahid, line 71: I would not edit the description with extra instructions unless it happens frequently.

libs/hackbot-runtime/hackbot_runtime/actions/phabricator.py
```diff
@@ -25,11 +26,29 @@
...
+def _validate_summary(summary: str | None) -> None:
+    if not summary:
+        return
+
+    match = _PHABRICATOR_TEST_PLAN_HEADER_RE.search(summary)
+    if match:
+        raise ToolError(
+            f'Invalid Phabricator summary: "{match.group()}" at the beginning of '
+            "a line is interpreted as a Test Plan field. Add a leading space or "
+            "rephrase that line, then call submit_patch again."
```
suhaibmujahid: I would not suggest a specific solution; the agent would most probably have a better solution by rewording the summary.

I would have something along these lines: 
```suggestion
            "a line is interpreted as a Test Plan field. Call submit_patch again with that fixed."
```

libs/hackbot-runtime/hackbot_runtime/actions/phabricator.py
ayoubdiourin7: updated in [49ef4c6](https://github.com/mozilla/bugbug/pull/6583/commits/49ef4c66a701b64318019f74ca66f4edb3d79980)

suhaibmujahid (approved):

PR #6585, opened 2026-08-13
Route Android, installer, and updater bugs to their own channels and widen the frontend-triage prompt past the desktop frontend

msujaws:
## Problem

bugbot is about to start sending `Firefox for Android :: History`, `Toolkit :: Application Update`, and `Firefox :: Installer` bugs to this agent (bugbot PR to follow — **this should land and deploy first**). Neither half of the agent was ready for them.

`SLACK_CHANNELS` held only `Firefox :: New Tab Page`, and `channel_for` fails closed on an unlisted component. That silences the notification but not the run: `record_notification` returns `None` while the Bugzilla comment and the severity/keyword change still apply. So the three new components would have got unattended triage on their bugs with nobody told.

The bigger gap is that none of the three is the JS/CSS frontend the prompt and rulesets describe:

| Component | Language | Where |
| --- | --- | --- |
| `Firefox for Android :: History` | Kotlin, Fragment/Store/Middleware/View | `mobile/android/fenix/app/src/main/java/org/mozilla/fenix/library/history/` |
| `Firefox :: Installer` | NSIS | `browser/installer/windows/nsis/` |
| `Toolkit :: Application Update` | `.sys.mjs`, IDL, C++ | `toolkit/mozapps/update/` |

`prompts/system.md` named only `browser/`, `toolkit/`, and `devtools/` and called the agent's remit "Firefox desktop frontend bugs", while `rules/frontend-triage.md` told it to stop when the bug "is not a frontend bug". A Fenix or NSIS bug would likely have been declared out of scope rather than triaged.

## Change

- `config.py` — three `SLACK_CHANNELS` entries. `#android-core-dev` for Android History; `#installer-updater-bug-triage` for both Application Update and Installer, which the product-and-component key already allows without either entry knowing about the other.
- `prompts/system.md` — remit restated as user-facing Firefox, and the single source-roots sentence replaced with a per-area layout of where the code lives and what language to expect. Test-discovery guidance now names the real harness per area rather than only browser-chrome mochitests.
- `rules/frontend-triage.md`, `rules/scoping.md` — the three components added to the in-scope lists, and the "not a frontend bug → stop" instruction narrowed so it no longer catches them.

All three areas are already in the checkout `hackbot.toml` clones, so nothing needed widening on the source side.

## Two calibration changes worth a reviewer's attention

These matter more than they look, because this agent writes to bugs unattended at high confidence.

1. `rules/frontend-triage.md` asserted that in-scope bugs are "usually documented with a video or screenshot" and are "not crashes, hangs". An update that did not apply arrives as an error code and an `update.log` excerpt with no steps to reproduce — the normal shape of a bug in that area, which must not read as grounds to skip it. Now spelled out, with "the installer failed" called out as not being a crash report.
2. `rules/severity-assessment.md` said papercuts default to S3. Applied to a user whose update does not install, that understates it: they are left on an unpatched build with no in-product workaround. Install and update failures now start from **S2** and are judged on reach from there.

## Testing

`pytest tests/` in `agents/frontend-triage`: 40 passed. New cases in `tests/test_notify.py` cover a mapping per new component, both installer/updater keys resolving to the same channel, and `channel_for("Firefox", "History") is None` — pinning that a component name is only owned within its own product. A key that is off by a character notifies nobody rather than failing, so each mapping is asserted rather than assumed from its neighbour.

Every path named in the prompt was checked against a mozilla-firefox checkout. That caught two errors in my first draft: the uninstaller is `uninstaller.nsi`, not `uninstall.nsi`, and the updater's xpcshell directories are `unit_aus_update`/`unit_background_update`/`unit_update_binary`, not the `unit_base_updater`/`unit_service_updater` I had guessed.

The prompt and ruleset changes are prose, so the tests do not cover them. They are unverified against a real run.

## Residual risk

**The prompt widening for Kotlin and NSIS is a guess until a run is observed, and feedback will be slow.** Those three components saw 11 staff-filed defects in the last 90 days combined (Installer 6, Application Update 4, Android History 1), against 191 for New Tab Page — roughly one Fenix History bug per quarter. Worth reading the first Fenix and Installer comments by hand before trusting them.

**Installer coverage is thin, by construction.** The only automated test is `browser/installer/windows/nsis/test/xpcshell/test_stub_installer.js`, which drives `test_stub.nsi` and covers the stub installer only; nothing exercises `installer.nsi` or the uninstaller. The prompt says so plainly rather than implying a search failure, so an empty `relevant_tests` on an Installer bug reads as the correct answer instead of a lazy one.

**Cheap lever if it goes badly:** drop a component's `SLACK_CHANNELS` entry to silence it, or drop it from bugbot's `TRIAGED_COMPONENTS` to stop the runs at the source.

### DIFF
--- a/agents/frontend-triage/README.md
+++ b/agents/frontend-triage/README.md
@@ -177,7 +177,7 @@ Routing is `SLACK_CHANNELS` in `config.py`, keyed by `"<Product> :: <Component>"
 
 | Product :: Component             | Channel                         |
 | -------------------------------- | ------------------------------- |
-| `Firefox :: New Tab Page`        | `#hnt-dev`                      |
+| `Firefox :: New Tab Page`        | `#hnt-dev-triage`               |
 | `Firefox for Android :: History` | `#android-core-dev`             |
 | `Toolkit :: Application Update`  | `#installer-updater-bug-triage` |
 | `Firefox :: Installer`           | `#installer-updater-bug-triage` |
--- a/agents/frontend-triage/README.md
+++ b/agents/frontend-triage/README.md
@@ -1,6 +1,6 @@
 # frontend-triage agent
 
-Triages a Firefox desktop frontend bug from Bugzilla and produces a **root-cause
+Triages a user-facing Firefox bug from Bugzilla and produces a **root-cause
 analysis plus a proposed fix plan**. It reads the source tree, navigates the
 codebase with Searchfox, and inspects regressor changesets on hg.mozilla.org. It
 does **not** build Firefox, edit source, or reproduce the bug. It writes to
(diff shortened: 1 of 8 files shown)


crythms (commented):
looks great!

crythms (approved):

PR #6586, opened 2026-08-13
Update diagnosis prompts

jgraham:
Don't ask for a testcase if there's no platform difference. Try to get spec links if possible.

### DIFF
--- a/agents/autowebcompat-diagnosis/hackbot_agents/autowebcompat_diagnosis/result.py
+++ b/agents/autowebcompat-diagnosis/hackbot_agents/autowebcompat_diagnosis/result.py
@@ -166,11 +166,15 @@ class DiagnosisResult(BaseModel):
 
     root_cause: str = Field(
         description=(
-            "Your root-cause hypothesis for why the site behaves differently in "
-            "Firefox: what the page does, which behavior it depends on, and why "
-            "that produces the reported breakage in Firefox but not Chrome. Be "
-            "specific about the mechanism (e.g. the API, CSS property, or "
-            "user-agent check involved). Do not propose a fix."
+            """Your root-cause hypothesis for why the site behaves differently in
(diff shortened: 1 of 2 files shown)


agents/autowebcompat-diagnosis/hackbot_agents/autowebcompat_diagnosis/result.py
```diff
@@ -166,11 +166,15 @@ class DiagnosisResult(BaseModel):
...
-            "specific about the mechanism (e.g. the API, CSS property, or "
-            "user-agent check involved). Do not propose a fix."
+            """Your root-cause hypothesis for why the site behaves differently in
+            Firefox: what the page does, which behavior it depends on, and why
+            that produces the reported breakage in Firefox but not Chrome. Be
+            specific about the mechanism (e.g. the API, CSS property, or
+            user-agent check involved). Where the behavior is related to a browser
+            engine difference covered by a web-standard such as HTML or CSS
+            then if possible provide links to the relevant parts of the specification
+            document that define the behaviour. Skip these links if you don't know the
```
jrmuizel, line 176: I feel like this would probably work better if we allowed the web search tool.

jrmuizel (approved):

suhaibmujahid:
Rebased since the CI was broken.

agents/autowebcompat-diagnosis/hackbot_agents/autowebcompat_diagnosis/result.py
jgraham: Yeah, agreed. For diagnosis that might be generally less problematic than reproduction, but we could also have a specific "find the relevant spec" task that is given web search access if we want to more tightly control what the diagnosis does. But let's consider that for a followup.

PR #6589, opened 2026-08-13
Create Hackbot docs

### LINKED ISSUE #6578: Document Hackbot
It's very hard to grasp how it works overall and each particular functionality as well. We should generate concise, human-readable docs that explain the main functions and design decisions. Ideally under `/docs/hackbot`. Markdown files would be fine initially; we can convert them later to a website, readthedocs, or anything. We should update those on every PR to keep them actual. This would greatly simplify reviewing PRs with AI-generated code.

evgenyrp:
I tried to make them human-readable, concise, structured, and deduplicated. 

The goal is to be able to look at a doc for a specific part of Hackbot to understand how it works faster. Another goal is to maintain those together with the code, which is essentially free now with AI, and help with reviewing PRs, as the change in the docs will often describe the code changes and how they fit the overall architecture.

We still have a couple of readme files on Hackbot outside of the `/docs`, but maybe it's better to leave them there for discoverability until we set up a documentation website.

fixes #6578


docs/hackbot/actions.md
```diff
@@ -0,0 +1,155 @@
...
+
+```python
+recorder.record(
+    "phabricator.submit_patch",              # <domain>.<verb>
+    {"bug_id": ..., "title": ...},           # params the apply step will need
+    reasoning="why the agent is doing this", # audit trail
+    attachments={"log": Path(...)},          # optional files, published as artifacts
+    ref="patch",                             # optional label, see cross-references below
+)
+```
```
suhaibmujahid: This is kind of an implementation detail.

docs/hackbot/actions.md
```diff
@@ -0,0 +1,155 @@
...
+
+The recorded list becomes `summary.json`'s `actions` array. Attachments are published
+under the stable key `attachments/<action_index>/<name>` and referenced by that key — the
```
suhaibmujahid: The `action_index` part will be outdated soon. And it does not serve a purpose.

docs/hackbot/actions.md
```diff
@@ -0,0 +1,155 @@
...
+local path disappears with the container.
+
+### The catalog
+
+Nine action types, each with a declaration the agent calls and a handler that applies it.
```
suhaibmujahid: The "Nine action types" part is a moving target.

docs/hackbot/api.md
```diff
@@ -0,0 +1,137 @@
...
+
+Authenticated by Phabricator's own HMAC signature over the raw body, so it sits on its own
+router without the API-key dependency. See [triggers.md](triggers.md).
+
+### Internal events — Google OIDC token
+
+| POST | `/internal/events/agent-run-finished` | An execution reached a terminal state |
+| POST | `/internal/events/apply-run-actions` | Consumer of `run.completed` |
+
+Named for the domain outcome or the job they do, not the GCP mechanism that feeds them.
```
suhaibmujahid: Not needed since it is completely internal and kind of implementation details.

docs/hackbot/deployment.md
suhaibmujahid, line 1: This is section is not needed for now. This is still WIP.

docs/hackbot/triggers.md
```diff
@@ -0,0 +1,111 @@
...
+`middleware.ts` is only an optimistic cookie-presence guard — it redirects to `/login`, or
+returns 401 JSON for `/api/*`, but is not the authority.
+
+Runs are attributed to the signed-in user via `X-On-Behalf-Of`, including retriggers
+(attributed to whoever clicked, not the original requester).
+
+Adding an agent to the UI means adding it to `lib/agents.ts`, the shared list behind both
+the trigger form and the run filter.
+
+## hackbot-pulse-listener — Taskcluster CI
```
suhaibmujahid, line 42: Maybe we keep this out since it will be migrated and we are not advertising for others to use for now.

suhaibmujahid (approved):
I did a quick pass, looks good overall.

For each section, worth asking: does this serve agent builders/maintainers, or Hackbot users? And could someone answer that just by reading the code? If yes, we should probably drop or shorten the section; we always could reference the code instead if needed, so we're not duplicating content we then have to keep in sync.

docs/hackbot/deployment.md
evgenyrp: I would keep it for us and update later because I struggled to find where all the deploys live

docs/hackbot/triggers.md
evgenyrp: I think it's fine to document the current state of the codebase, but we can make a note that this will be migrated to Treeherder. When it's migrated, we remove it from the docs.

evgenyrp:
> I did a quick pass, looks good overall.
> 
> For each section, worth asking: does this serve agent builders/maintainers, or Hackbot users? And could someone answer that just by reading the code? If yes, we should probably drop or shorten the section; we always could reference the code instead if needed, so we're not duplicating content we then have to keep in sync.

My goal (and prompt) was to create docs for people who work on this repo. This includes us and agent builders who occasionally touch runtime as well. This should be an entry point to quickly understand the architecture and functionality without reading the code. I agree we could just link to the code in some cases to minimize syncing in the future.

docs/hackbot/triggers.md
suhaibmujahid: no objection

evgenyrp:
I addressed the comments and also fixed the code links. There are still some details about the code. They can be helpful; let's see if maintaining them becomes a problem.

PR #6595, opened 2026-08-14
Set UTF-8 charset for agent logs (GCS)

### LINKED ISSUE #6588: Agent logs served with `Content-Type: text/plain` missing `charset=utf-8` causing garbled display when viewed via GCS link
From chat with @jrmuizel:
> Can we change the agent logs so that they have `Content-Type: text/plain; charset=utf-8` instead of `Content-Type: text/plain`?

ayoubdiourin7:
Fixes #6588

### DIFF
--- a/libs/hackbot-runtime/hackbot_runtime/runtime.py
+++ b/libs/hackbot-runtime/hackbot_runtime/runtime.py
@@ -132,7 +132,7 @@ def _load_hackbot(entrypoint: Callable, config: ConfigArg) -> HackbotContext | N
 def _publish_log(ctx: HackbotContext) -> None:
     """Publish the run log under the canonical key, if the agent wrote one."""
     if ctx.log_path.exists():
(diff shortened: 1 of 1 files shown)


suhaibmujahid (approved):

jrmuizel:
This doesn't seem to be working. The log from https://hackbot.moz.tools/runs/9db674e5-7cdb-435c-b37f-58a1cdfc44ef is still served without `; charset=utf-8`

suhaibmujahid:
> This doesn't seem to be working.

@jrmuizel it is working. You can see it here: https://hackbot.moz.tools/runs/45765ced-4b5e-4a54-8cac-e4099d1ffd3f

The `autowebcompat-diagnosis` agent was not deployed after merging this PR, that’s why you do not see the impact. I will deploy both `autowebcompat` agents shortly.

jrmuizel:
Ah great.

PR #6610, opened 2026-08-17
Route Site Permissions triage to #privacy-team-automation and tell the agent where permission code lives

msujaws:
## Problem

bugbot is about to start sending `Firefox :: Site Permissions` bugs to this agent (mozilla/bugbot#2991 — **this should land and deploy first**). `SLACK_CHANNELS` has no entry for the component, and `channel_for` fails closed on an unlisted one. That silences the notification but not the run: `record_notification` returns `None` while the Bugzilla comment and the severity change still apply, so the component would have got unattended triage on its bugs with nobody told.

Unlike the Kotlin and NSIS components added in #6585, Site Permissions is desktop JS, so the prompt was not wrong about it — just incomplete in the place that matters most. The permission a user grants is stored by `nsIPermissionManager`, implemented in C++ at `extensions/permissions/PermissionManager.cpp`, which is outside the `browser/`, `toolkit/`, `devtools/` list `prompts/system.md` gives for the desktop frontend. "I allowed the camera and it asked again after restart" is a scope, expiry, or store bug, and `rules/frontend-triage.md` tells the agent to stop when a bug "is not in user-facing Firefox at all" — so an agent told the frontend lives in three directories could reasonably read that as a Core bug and skip it.

The component also splits across three layers, which the component name alone does not tell you:

| Layer | Where |
| --- | --- |
| Permission state, scopes, defaults | `browser/modules/SitePermissions.sys.mjs` |
| Doorhanger prompts, one subclass per type | `browser/modules/PermissionUI.sys.mjs` |
| Camera / microphone / screen sharing (own path) | `browser/actors/WebRTCParent.sys.mjs` |
| Backing store, persistence and expiry | `extensions/permissions/PermissionManager.cpp` |

## Change

- `config.py` — one `SLACK_CHANNELS` entry, `Firefox :: Site Permissions` → `#privacy-team-automation`.
- `prompts/system.md` — a Site Permissions bullet in **Source repository** naming those four entry points and saying the C++ store is in scope, plus a matching three-way split in the test-discovery list (`browser/base/content/test/permissions/`, `browser/modules/test/browser/`, `extensions/permissions/test/`).
- `rules/frontend-triage.md`, `rules/scoping.md`, `README.md` — `Site Permissions` added to the component lists. Those are prompt-facing scope statements rather than routing; they already name Address Bar and Theme, neither of which has a channel.

The second commit fixes a separate staleness the first one surfaced: `TRIAGE_TASK` in `__main__.py` still opened "Triage this Firefox desktop frontend bug", which is the per-run task string handed to the agent and has been wrong for Android and NSIS since #6585. The registry `description` in `services/hackbot-api/app/agents.py` said the same. Both now say "user-facing Firefox" and name the four areas.

## Testing

`uv run --package hackbot-agent-frontend-triage pytest agents/frontend-triage/tests` — 40 pass. The routing guard is one added assertion in `test_the_channel_belongs_to_the_component` (`tests/test_notify.py`); `test_an_unowned_component_has_no_channel` needed no change, since none of its cases collide with the new key.

`services/hackbot-api/tests` gives 3 failed, 95 passed, 8 errors — pre-existing. I ran that suite on a clean `upstream/master` and got the identical counts.

Volume is low: `Firefox :: Site Permissions` saw two staff-filed defects in the last 90 days, against 90 for New Tab Page over the same window. So the prompt widening will not be exercised much for a while, and the same caveat #6585 gave applies — worth reading the first few Site Permissions comments by hand before trusting them.

Still not done, and out of scope here: nothing checks that this map and bugbot's `TRIAGED_COMPONENTS` agree. A component added to one and not the other is silently either triaged with nobody told, or configured and never triggered. Worth building, but larger than adding a component.


suhaibmujahid (approved):
r+ for changes in `services/hackbot-api/app/agents.py`

crythms (approved):

PR #6617, opened 2026-08-17
Route four more components for frontend triage, and stop writing the component list in four places

msujaws:
Four components join the frontend-triage agent's routing: `Firefox for Android :: Toolbar` and `Firefox for Android :: Homepage` to `#android-core-dev`, `Firefox :: IP Protection` to `#team-eng-ip-protection-triage`, and `Firefox :: Sharing` to `#content-sharing-automation`. All four channels exist and are public.

Adding one component used to touch six files, and #6610 (Site Permissions) is the worked example: one line of routing in `config.py`, then the same component name written into `rules/frontend-triage.md`, `rules/scoping.md`, two places in `README.md`, and one assertion in `tests/test_notify.py`. Nothing kept those in step, and the failure is silent in the direction that matters — `channel_for` fails closed, so a component missing from `SLACK_CHANNELS` still gets its comment and severity change applied unattended, with nobody told. Rather than pay that four more times, the first two commits collapse it.

## The registry

`TRIAGE_SCOPE` in `config.py` now holds each routed component once, as a `ScopedComponent(product, component, area, channel)`. `SLACK_CHANNELS` is derived from it, so `notify.py` has **zero diff** — it keeps its `.strip()`, its no-default `.get()`, and one flat mapping to look up. `render_scope` in `agent.py` renders the same tuple into a new `Components in scope` section of `system.md`, which was already a `str.format` template.

`channel` is required rather than optional. An entry without one would be a component under unattended triage with nobody told, which is what failing closed already produces by accident; there is no reason to be able to express it on purpose. So the registry is exactly the routed set, and it is exactly bugbot's `TRIAGED_COMPONENTS` today.

Only the enumeration is generated. The per-area guidance under `Source repository` stays hand-authored markdown — that prose is the payload, not the tax. The Site Permissions bullet is six clauses of careful hedging; moving it into an implicitly-concatenated literal inside a NamedTuple, reflowed by `ruff-format`, where `{` needs no doubling but the surrounding `system.md` does, would make authoring worse to save an edit that only happens when a component's code layout is genuinely new.

## What the components needed told

Three of the four are not just a routing line, and that is most of this diff.

**`Firefox :: Sharing`** is the one area whose code reaches outside `browser/`, `toolkit/`, `devtools/`. The frontend is `browser/modules/SharingUtils.sys.mjs` and `browser/components/contentsharing/`, but the share sheet itself is `widget/cocoa/nsMacSharingService.mm` and `shareUrl` on `widget/nsIWindowsUIUtils.idl`. So "Share does nothing on macOS" is a per-OS bug in Objective-C++ that the prompt would otherwise imply is out of scope — the same trap #6610 fixed for the C++ permission store. The bullet also separates the two unrelated things this tree calls sharing: WebRTC screen/camera capture dominates a grep for `sharing` and belongs to site permissions.

**`Firefox :: IP Protection`** has two state machines that both have a `READY`, and conflating them is the mistake worth pre-empting. `IPProtectionStates` in `IPProtectionService.sys.mjs` is entitlement and sign-in; `IPPProxyStates` in `IPPProxyManager.sys.mjs` is the connection. "It showed connected when it was not" is the second and lives in `toolkit/`, even though every visible trace of it is in the panel — so an agent that starts where the screenshot points lands in the wrong tree. The bullet also sends it to `toolkit/components/ipprotection/docs/`, which is real in-tree prose documentation no other area here has, and corrects the prefs path (`browser/app/profile/firefox.js`, not `modules/libpref/init/all.js`: 25 hits against 0).

**`Firefox for Android :: Toolbar` and `:: Homepage`** are in the existing Android area, so they needed no new bullet — but both carry a stub-versus-full-installer style trap. There are two toolbars (`fenix/components/toolbar/` and the homepage's own `fenix/home/toolbar/`) over two generations of the android-components widget, so a fix planned against the retired one reads correct and changes nothing. And the homepage is one package per section, so a top-sites bug is not in `Homepage.kt` — with sibling components (`Top Sites`, `Stories`, `Collections`) able to receive the same bug, and `Stories` being `home/pocket/` in the tree because the rename never happened.

`severity-assessment.md` gains one paragraph, for IP Protection only: turning the VPN off is not a workaround for the VPN not working, so start from S2 rather than the S3 a papercut gets. It also asks for a distinction the reports will not make on their own — state merely *displayed* wrong is a UI bug, state actually wrong means traffic is unproxied and belongs above S2. Android Toolbar and Homepage bugs are ordinary papercuts and keep the S3 default. Sharing has an in-product workaround (copy the link), so it does too.

## Volume

Open defects filed in the last 90 days, and the subset whose reporter contains `@mozilla.com` — the same approximation that reproduced the "90 for New Tab Page" figure in #6610:

| Component | Open defects / 90d | Staff-filed |
| --- | --- | --- |
| `Firefox :: Sharing` | 38 | 34 |
| `Firefox for Android :: Homepage` | 35 | 24 |
| `Firefox :: IP Protection` | 32 | 22 |
| `Firefox for Android :: Toolbar` | 18 | 2 |

For comparison, New Tab Page is 148/90 and Site Permissions is 2.

## Testing

`uv run --package hackbot-agent-frontend-triage pytest agents/frontend-triage/tests` — **46 pass**, and they pass at every commit in the series, not just at the tip.

Four new tests, all in service of making the remaining per-component cost visible rather than forgettable:

- `test_every_area_has_prompt_guidance` (`tests/test_plan.py`) is the load-bearing one. It fails when a `TRIAGE_SCOPE` area has no bolded bullet under `Source repository` — which is exactly how a component gets routed with the agent having no idea where its code lives. I verified it fails by adding a component with area `PDF viewer` and no bullet: `AssertionError: PDF viewer`.
- `test_the_registry_names_each_component_once` covers the duplicate key that would collapse silently in the derived dict — the failure mode a growing registry actually has, and one that is invisible in a diff that only adds a line.
- `test_every_channel_is_a_channel_name` covers the missing `#` or stray capital that becomes `channel_not_found` at apply time, after the Bugzilla writes have already landed.
- `test_the_scope_says_it_is_neither_a_limit_nor_a_vocabulary` covers both ways a component list in a system prompt gets misread (below).

`test_the_channel_belongs_to_the_component` now loops `TRIAGE_SCOPE` instead of asserting a line per component, so a new component needs no test edit. The invariants a loop cannot express stay written out: whitespace stripping, fail-closed on an unlisted or garbled pair, `Firefox :: History` being nothing at all (it is not a real BMO component — the desktop one is `Bookmarks & History`), and two components sharing a channel, each asserted separately.

I also rendered the system prompt before and after the registry commit at an unchanged component set and diffed it, to confirm the commit moves no prose and rewords nothing.

Unrelated and pre-existing: `services/hackbot-api/tests` has 3 failures and 8 errors on clean master, as ecea6ca6's message records. This PR does not touch that service.

**The README routing table is gone.** The section above it already said "Routing is `SLACK_CHANNELS` in `config.py`", so the table told the reader to open a file and then copied it. It is replaced by the four things about routing that are *not* obvious from reading the registry, plus one worked example. The cost is real: to see where one component routes you now open `config.py`. I considered keeping the table behind a test that regexes its rows back out and compares them to the registry, and decided a six-row list did not earn the machinery — and prettier realigns every row when the widest cell changes, so "add one row" is a six-row diff either way. Happy to put it back if it is load-bearing for someone.

**The component list moved into the always-present system prompt**, out of `rules/*.md`, which the agent globs and reads only when it judges a ruleset relevant. That is a deliberate strengthening, and `frontend-triage.md` already deferred upward for the per-area layout — but it does mean the rendered model context is not byte-identical before and after, and it is worth knowing that.

**A list of components in a system prompt invites two expensive misreadings**, and the generated section says so outright. Read as exhaustive it declares an in-scope bug out of scope, which is the ecea6ca6 mistake; `scoping.md`'s rule has always been that *any* user-facing Firefox defect is in scope, and its list ended in an ellipsis. Read as a vocabulary it gets a bug's component adjusted to match, and since the component is also the routing key, `notify.py` then tells nobody.

**This does not make every future component a one-liner, and I want to be straight about that.** A component in an existing area is now genuinely one file and one line — I verified it by adding `Firefox :: Downloads Panel` under `Desktop frontend`: 1 file, 46 tests still pass, routing and the rendered prompt both pick it up. But of the nine components now in scope, six needed a new area, and the two Android ones needed prose anyway. The refactor removed the toil — five duplicated bookkeeping edits that carried no information and could silently drift — not the work, which is deciding whether the agent needs to be told where this component's code lives. What it did buy, as a side effect rather than a goal, is that the expensive part is now impossible to skip by accident.

## Landing order

**This needs to be deployed before the matching bugbot change** (mozilla/bugbot#2996, which adds the same four pairs to `TRIAGED_COMPONENTS`). Without an entry here, `channel_for` fails closed: the notification is silenced but the run is not, so the analysis and the severity change still reach the bug with nobody told. bugbot's cron checks out `releases/latest` rather than master, so that PR is inert until a release is cut — the gate is "this rolled out before that release", not merge order.

Nothing here is keyed by component outside `agents/frontend-triage/`. I checked `TRIAGE_TASK` in `__main__.py` and the `frontend-triage` description in `services/hackbot-api/app/agents.py`: both already read "user-facing Firefox" after ecea6ca6, which covers all four, so neither is touched.


crythms (approved):
looks good!

PR #6652, opened 2026-08-18
Include revision diff ID in regular comments

### LINKED ISSUE #6572: Add diff ID context to regular Phabricator comments
Inline `@hackbot` comments include their `diff_id`, but regular comments do not. It may also be helpful to include a `diff_id` in regular Phabricator comment context.

ayoubdiourin7:
In Phabricator, regular comments aren’t tied to a specific diff. Since Hackbot runs when tagged, the current `diffID` is used, as the latest diff is likely the most relevant to the comment.

Fixes #6572


suhaibmujahid:
See https://github.com/mozilla/bugbug/issues/6572#issuecomment-5344542560

PR #6653, opened 2026-08-18
Preserve requested URL through Google sign-in

### LINKED ISSUE #6645: Fix hackbot-ui redirect after login
`services/hackbot-ui/middleware.ts` redirects unauthenticated requests to a fixed `/login` with no `callbackUrl`/`next` parameter, and `app/login/page.tsx` hardcodes `callbackURL: "/"` on sign-in. A deep link visited while signed out always lands on the home page after authentication instead of back where the user started.

ayoubdiourin7:
Tested locally ; users return to their original page after signing in.

FIxes #6645


services/hackbot-ui/app/login/page.tsx
```diff
@@ -4,18 +4,44 @@ import { useState } from "react";
...
 import { signIn } from "@/lib/auth-client";
 
+const DEFAULT_PATH = "/";
+
+function safeRedirectPath(raw: string | null): string {
+  if (
+    !raw ||
+    !raw.startsWith("/") ||
+    raw.startsWith("//") ||
+    raw.includes("\\")
```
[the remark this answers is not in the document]
ayoubdiourin7: Addressed in [d24f940](https://github.com/mozilla/bugbug/pull/6653/commits/d24f940f291091cc3ea2b66d7022c4d41793d901)

services/hackbot-ui/app/login/page.tsx
```diff
@@ -4,18 +4,47 @@ import { useState } from "react";
...
+    return DEFAULT_PATH;
+  }
+
+  const path = raw.split(/[?#]/)[0];
+  if (path === "/login" || path.startsWith("/login/")) {
+    return DEFAULT_PATH;
+  }
+
+  return raw;
+}
```
suhaibmujahid: I would use `encodeURIComponent()` instead, which is a global JS function.

services/hackbot-ui/app/login/page.tsx
```diff
@@ -4,18 +4,47 @@ import { useState } from "react";
...
+
 export default function LoginPage() {
```
suhaibmujahid: I would avoid using `window.location.search`, instead using nextjs native props:
```suggestion
export default function LoginPage(props: {
  searchParams: Promise<{ callbackUrl?: string}>;
}) {
  const searchParams = await props.searchParams;
  const callbackURL = searchParams.callbackUrl || "/";
```

services/hackbot-ui/app/login/page.tsx
```diff
@@ -4,18 +4,47 @@ import { useState } from "react";
...
   const [loading, setLoading] = useState(false);
   const [error, setError] = useState<string | null>(null);
 
   async function onGoogle() {
     setError(null);
     setLoading(true);
+    const params = new URLSearchParams(window.location.search);
+    const next = safeRedirectPath(params.get("next"));
```
suhaibmujahid: Why do we need `safeRedirectPath()`? I do not think it is needed.

services/hackbot-ui/middleware.ts
```diff
@@ -16,7 +16,12 @@ export function middleware(req: NextRequest) {
     return NextResponse.json({ error: "Unauthorized" }, { status: 401 });
   }
 
+  // Remember the requested URL so the login page can send the user back here
   const loginUrl = new URL("/login", req.url);
+  loginUrl.searchParams.set(
+    "next",
```
suhaibmujahid: I would use `callbackURL` instead of `next`.

services/hackbot-ui/middleware.ts
```diff
@@ -16,7 +16,12 @@ export function middleware(req: NextRequest) {
...
+    `${req.nextUrl.pathname}${req.nextUrl.search}`
```
suhaibmujahid: why not `req.nextUrl` directly?

suhaibmujahid (changes requested):

services/hackbot-ui/app/login/page.tsx
ayoubdiourin7: Yes, i updated to use Next.js `searchParams`. cause this is a client component, I unwrap it with React’s `use()`.

services/hackbot-ui/app/login/page.tsx
ayoubdiourin7: Added it to prevent unsafe external redirects, but Better Auth already handles this, so I dropped it. [851c867](https://github.com/mozilla/bugbug/pull/6653/commits/851c8676dbae8eda10aa333d0baa1f9039dc883e)

services/hackbot-ui/middleware.ts
ayoubdiourin7: Addressed in [851c867](https://github.com/mozilla/bugbug/pull/6653/commits/851c8676dbae8eda10aa333d0baa1f9039dc883e)

services/hackbot-ui/middleware.ts
ayoubdiourin7: yes we can , fix it  in [851c867](https://github.com/mozilla/bugbug/pull/6653/commits/851c8676dbae8eda10aa333d0baa1f9039dc883e)

suhaibmujahid (approved):
Thank you!

Issue #6511, opened 2026-08-04
The test-plan-generator agent should better utilize actions in reporting the results


suhaibmujahid:
Currently, the execution results are reported in separate lists. It would be cleaner to report the execution result along with the test case itself.

Issue #6436, opened 2026-07-30
Include `test-plan-generator` agent's execution results in TestRail


jpangas:
While we are submitting the generated test cases to the test suite, we could submit a test run and result containing the agent's execution results.
The execution results would appear under "Test Runs and Results" on TestRail. We could also add the label `Hackbot`. It should show what passed or failed etc.
Our agent's labels .e.g. `passed` or `failed` would need to match the labels used in TestRail.

Issue #6576, opened 2026-08-12
Validate tool inputs against the tool schema in agent-tools


suhaibmujahid:
Tool arguments coming from the model are never checked against the tool's own schema, so malformed input reaches handlers and fails in ways the agent cannot act on.

`ToolDefinition.args_model` (in `libs/agent-tools/agent_tools/registry.py`) already builds a pydantic model from each tool's signature and caches it, and `input_schema` is derived from it and handed to the SDK. But no adapter validates with it — `build_sdk_server` calls the handler directly:

```python
@sdk_tool(mcp_name, defn.description, defn.input_schema)
async def run(args):
    try:
        result = await defn.handler(ctx, **args)
    except ToolError as e:
        ...
```

The schema therefore only *guides* the model; nothing enforces it.

### Why it matters

Two concrete failure modes, using `try_server.push`'s `tests: dict[str, list[str]]`:

- A bare string (`{"xpcshell": "dom/base/test"}`) is iterated **character by character**, silently producing `['', 'a', 'b', 'd', 'e', 'm', 'o', 's', 't']` as "paths". Handlers currently need hand-rolled `isinstance` guards to avoid this.
- A wrong element type (`{"xpcshell": [42]}`) gets past a container check and raises `AttributeError` inside the handler. `run()` only catches `ToolError`, so it escapes as an unhandled exception rather than becoming feedback the agent could correct.

Pydantic already rejects both, in default (lax) mode, with no configuration:

```
bare string  -> REJECTED list_type: Input should be a valid list
int in list  -> REJECTED string_type: Input should be a valid string
tuple        -> OK  (normalised to list)
```

Related symptoms of the same shape elsewhere: #6517 (an LLM returned a list where a string was expected and it crashed on `.rfind`) and possibly #6493 (unhelpful errors when the model malforms tool arguments).

### Proposal

Validate with the existing `args_model` before dispatching, and turn failures into structured tool errors. Put the helper on `ToolDefinition` (e.g. `defn.validate(args)`) rather than inline in the claude-sdk adapter, so a future non-claude-sdk adapter reuses it — `registry.py` is already the framework-neutral home for `args_model`.

Handlers then drop their hand-rolled type guards and keep only the semantic checks a type cannot express.

### Decisions to make

[...]

PR #6657, opened 2026-08-19
Use the TestRail action directly in test-plan-generator

### LINKED ISSUE #6508: The test-paln-generator agent should use the TestRail action directly
I just noticed that action is not used as it was supposed to be used: https://github.com/mozilla/bugbug/blob/beb81a21b489b2147f0c6b2645a01485eb17033e/agents/test-plan-generator/hackbot_agents/test_plan_generator/__main__.py#L36

The agent should use it directly instead of submitting through results, then copying it here.

_Originally posted by @suhaibmujahid in https://github.com/mozilla/bugbug/pull/6450#discussion_r3713445510_

johnpangas:
Fixes #6508

## What changed

`testrail.submit_test_plan` is now exposed to the agent as a recordable action, and the `result` module is gone: the action schema is the only definition of a generated test plan.

- The per-case `context` label goes with it. It was guidance for tool selection, so it moves into the prompt where that judgment belongs, rather than being carried as data nothing consumed.
- The case-count bound and the sequential-id check move from `TestPlanResult` onto `SubmitTestPlanInput`, so they are enforced for every caller instead of only the ones going through the old result model.
- Recording twice would create two TestRail suites on apply, so a second `submit_test_plan` call is rejected.
- The agent's findings become its closing prose, so the UI reads the plan from the recorded action's params, falling back to `findings.result` for runs recorded before this change.

## Note

Between this PR and the next one in the chain, the action does not yet carry execution results, so the UI renders test cases without pass/fail badges. That is restored by #6658, which nests the result inside each case. Merging the two together avoids exposing the gap on `master`.

## Stack

Please merge in this order  each builds on the one before:

1. **This PR** — #6508
2. #6658 — #6511, nest each case's execution result
3. #6659 — #6436, submit execution results to TestRail


libs/hackbot-runtime/hackbot_runtime/actions/testrail.py
```diff
@@ -34,9 +29,6 @@ class TestRailStepInput(BaseModel):
 class TestRailCaseInput(BaseModel):
     id: int
     title: str = Field(description="TestRail test case title.")
-    context: str | None = Field(
-        default=None, description="Optional context for where this case applies."
-    )
```
suhaibmujahid, line 39: Why are we removing this? What was it used for?

libs/hackbot-runtime/hackbot_runtime/actions/testrail.py
```diff
@@ -86,6 +80,14 @@ def feature_must_not_be_blank(cls, value: str) -> str:
             raise ValueError("feature must not be blank")
         return value
 
+    @model_validator(mode="after")
+    def case_ids_must_be_sequential(self) -> "SubmitTestPlanInput":
+        case_ids = [case.id for case in self.generated_test_cases]
+        expected_ids = list(range(1, len(self.generated_test_cases) + 1))
+        if case_ids != expected_ids:
+            raise ValueError("test case ids must be sequential starting at 1")
```
suhaibmujahid: We should raise ToolError here instead of ValueError.

libs/hackbot-runtime/hackbot_runtime/actions/testrail.py
```diff
@@ -118,30 +120,30 @@ async def submit_test_plan(
     ],
     generated_test_cases: Annotated[
         list[TestRailCaseInput],
-        Field(description="Generated test cases to upload together to TestRail."),
+        # Bounds are repeated from SubmitTestPlanInput: this copy is the agent's
+        # schema, that one is the enforcement.
```
suhaibmujahid, line 124: This should be enforced once we implement https://github.com/mozilla/bugbug/issues/6576. So good that we enforce it manually for now.

services/hackbot-ui/components/RunDetail.tsx
```diff
@@ -251,7 +260,9 @@ export function RunDetail({ runId }: { runId: string }) {
         </div>
       )}
 
-      {hasFindings && <FindingsView findings={findings} agent={run.agent} />}
+      {(hasFindings || hasTestPlanAction) && (
+        <FindingsView findings={findings} agent={run.agent} actions={actions} />
+      )}
```
suhaibmujahid, line 265: I would simplify it here and use the `TestPlanView` instead of `FindingsView` if the agent is `test-plan-generator`, with that we have the view logic for thr QA agent in one place. 

Yes, we will not have the `JSON` for that, but it is OK; the finding is empty anyway, and you could access the full thing from the `summary.json` file.

suhaibmujahid (changes requested):
Thank you! We need some minor changes.

libs/hackbot-runtime/hackbot_runtime/actions/testrail.py
jpangas: We werent using it for anything useful by recording it. It was used to log if the agent used `chrome` or `content` context tools to run a test case.

services/hackbot-ui/components/RunDetail.tsx
```diff
@@ -162,6 +163,12 @@ export function RunDetail({ runId }: { runId: string }) {
   const log = extractLog(run);
   const findings = run.summary?.findings ?? {};
   const hasFindings = Object.keys(findings).length > 0;
+  // The QA agent gets its own purpose-built view; its plan lives on the
+  // TestRail action, so findings are usually empty (raw data is in summary.json).
+  const testPlan =
+    run.agent === "test-plan-generator"
+      ? parseTestPlan(findings, actions)
+      : null;
```
suhaibmujahid, line 171: This is not the cleanest way to route to the view, but it is okay for now. We should have a structure why to customize the views based on the agent and have it in UX.

suhaibmujahid (approved):
Thank you! LGTM!

services/hackbot-ui/components/RunDetail.tsx
suhaibmujahid: I field https://github.com/mozilla/bugbug/issues/6684 as a follow-up

PR #6658, opened 2026-08-19
Report each TestRail case result with the case itself

### LINKED ISSUE #6511: The test-plan-generator agent should better utilize actions in reporting the results

johnpangas:
Fixes #6511

Depends on #6657, please merge that first.

## What changed

The result is nested inside each generated test case. Every case carries a required `result` with its status, summary and failure reason, which makes an unexecuted case unrepresentable rather than merely absent from a second list.

The run wide write up moves to a `summary` on the action, where it describes the plan as a whole instead of competing with the per-case summaries. It becomes the TestRail run description in #6659.

The UI reads the nested result and still falls back to the old `results` list, so previously recorded runs render unchanged.

## Stack

1. #6657 — #6508, agent uses the action directly
2. **This PR** — #6511
3. #6659 — #6436, submit execution results to TestRail

I couldn't use the GitHub stack since it requires push access to bases branches on `mozilla/bugbug`). So These target `master` rather than each other (no push access for base branches on `mozilla/bugbug`), so Files-changed is cumulative until #6657 merges.

### DIFF
--- a/agents/test-plan-generator/hackbot_agents/test_plan_generator/prompts/system.md
+++ b/agents/test-plan-generator/hackbot_agents/test_plan_generator/prompts/system.md
@@ -2,8 +2,8 @@ You are a Firefox QA test-plan generation and execution agent.
 
 Generate test cases from the provided Firefox feature name, feature description,
 and test scope, run them in Firefox with the available DevTools MCP tools, and
-record the generated test plan for TestRail. Do not try to fix, patch or make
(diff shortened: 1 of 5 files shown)


suhaibmujahid (approved):
LGTM! Thanks!

Issue #6515, opened 2026-08-05
We should keep a record of the IDs of created items when submitting TestRail


suhaibmujahid:
This will help track items and collect signals to be used as feedback on agent performance, and on how the generated test cases were used or discarded.

jpangas:
https://github.com/mozilla/bugbug/blob/62d76c848b75852c6217be095acb5a4493c60b73/libs/hackbot-runtime/hackbot_runtime/actions/handlers/testrail_handler.py#L141-L147

We are currently recording the case_ids submitted to TestRail after the action succeeds. 

Although if it fails, we are only returning the error:
https://github.com/mozilla/bugbug/blob/62d76c848b75852c6217be095acb5a4493c60b73/libs/hackbot-runtime/hackbot_runtime/actions/handlers/testrail_handler.py#L135-L139
 
and we dont record what may have been already submitted to TestRails if it fails completely or partially. Should we patch this now to also include `case_ids` already submitted incase of a partial failure or we wait for the fix that will introduction actions to submit each case one by one? 

Considering someone may also be able to search  manually on TestRail by feature name to discover if anything exists at all but that would hard for some of the users to tell that some cases were uploaded.

jpangas:
We are already recording IDs using `case_ids` as mentioned above. Commenting so issue can be closed.

PR #6659, opened 2026-08-19
Submit the agent's execution results to TestRail

### LINKED ISSUE #6436: Include `test-plan-generator` agent's execution results in TestRail

johnpangas:
Fixes #6436

Depends on #6657 and #6658 — please merge those first.

## What changed

After the cases are created, the handler creates a run over exactly those case ids and posts one result per case.

- **Status mapping.** `passed` and `failed` match TestRail's own status names; `unsuitable` posts as **Blocked**, the closest status a result can carry. Statuses are matched on the lowercase `name`, not the display `label`.
- **Fail before creating anything.** Status ids are resolved before the suite exists, so an unmapped status fails the submission rather than leaving a half-populated suite behind.
- The action's `summary` becomes the run description, and each case's summary and failure reason become the result comment.
- The run id joins the other created ids on the action result, extending the record from #6515.

## NOte

If run creation fails after the cases are created, the whole submission is marked failed, so a retry duplicates the suite. There is a test pinning that behavior (`test_submit_test_plan_fails_when_the_run_cannot_be_created`). Worth a follow-up issue on de-duplicating retries rather than blocking this.

## Stack

1. #6657 — #6508, agent uses the action directly
2. #6658 — #6511, nest each case's execution result
3. **This PR** — #6436


suhaibmujahid (approved):
LGTM!

PR #6674, opened 2026-08-20
Add Slack interactivity webhook support

suhaibmujahid:
Resolves #6670

Add a dedicated Slack interactions endpoint with request signature verification.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

PR #6677, opened 2026-08-20
Add Block Kit layout to frontend-triage notifications

### LINKED ISSUE #6673: Use Slack Block Kit layout for frontend-triage agent messages
This will improve the readability of the messages.

suhaibmujahid:
Resolves #6673

Refactors notification rendering into reusable helpers and adds `build_blocks` for richer Slack messages.

---

### DIFF
--- a/agents/frontend-triage/hackbot_agents/frontend_triage/notify.py
+++ b/agents/frontend-triage/hackbot_agents/frontend_triage/notify.py
@@ -54,29 +54,39 @@ def channel_for(product: str | None, component: str | None) -> str | None:
     return SLACK_CHANNELS.get(f"{product.strip()} :: {component.strip()}")
 
 
-def build_message(result: FrontendTriageResult, *, run_id: str) -> str:
-    """Render the notification for an auto-applied run."""
-    assessment = result.severity_assessment
+def _bug_link(result: FrontendTriageResult) -> str:
+    return _link(BUG_URL.format(bug_id=result.bug_id), f"Bug {result.bug_id}")
+
+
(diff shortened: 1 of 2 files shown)


crythms:
@suhaibmujahid  
#6679 and this PR conflict. I think #6679 would need to land first. it targets `master` directly while this sits on top of other unmerged PRs (#6674 → #6675 → #6676), and it changes whether the severity in these notifications was actually
applied to the bug.

#6679 removes the triage agent's ability to write `severity` to Bugzilla at all. Severity just becomes a suggestion in the triage comment, and is left out below medium confidence.

After #6679 lands, this would need a rebase plus a couple of changes:

1. `_severity_field` adds the word "suggested" only when confidence is below high. That's right today, because a high-confidence severity really does get written to  the bug. After #6679 nothing is ever written, so a plain `S3` would look like the bug was changed when it wasn't. The  `if` could go and the label could say it instead  -- `*Suggested severity*` with the level underneath, every time. The `build_blocks` headline has the same problem: it shows a bare `*S1*`.

2. `_is_urgent` and `_severity_field` both ignore confidence. #6679 only reports severity at high or medium (`REPORTABLE_SEVERITY_CONFIDENCES` in `config.py`), so these would need to match, otherwise Slack flags an S1 on a bug whose comment doesn't mention severity at all.

A few tests in this PR assume the old behaviour too, but those should fail loudly on the rebase so they're easy to spot.

msujaws (approved):
I tested this out and it looks good. We may want to change how some of the blocks are positioned and what is in them after we live with this for a while but I don't want to force perfection on the first pass here. Thank you!

PR #6679, opened 2026-08-20
Frontend triage severity as comment suggestion

crythms:
The agent used to write severity straight to the bug's `severity` field. At `confidence: high` that happened with no human in between. It still works out a severity — it just suggests it at the end of the comment now, before the footer:

    ---

    Suggested severity: S4
    Purely cosmetic, no functional impact. Only visible on Nightly.

The block is left out entirely when the agent isn't confident about the impact. A level it's unsure of still reads like a judgement someone might act on so we avoid that. 

## Other relevant changes

**Which confidence decides whether the block sev appears.** The agent reports two, and they measure different things:
- `confidence` — how sure it is that it found the cause in the code
- `severity_assessment.confidence` — how sure it is about the user impact

The block is gated on the second one. They're independent: the agent can be certain a bug is cosmetic while having no idea which file is at fault, and that bug should still get a severity suggestion. The rules didn't distinguish the two before; they
do now.

`notify.py` uses the same threshold for its Slack S1 marker, so Slack can't flag an S1 on a bug whose comment says nothing about severity.

**`severity_assessment` is now normalized in `parse_plan`.** It came straight from the model's JSON before, so `"High"` instead of `"high"` would have silently dropped both the comment block and the Slack marker.

### DIFF
--- a/agents/frontend-triage/README.md
+++ b/agents/frontend-triage/README.md
@@ -132,9 +132,8 @@ Two caveats before acting on a plan:
 ## What it writes back to Bugzilla
 
 **Nothing, during a run.** `ENABLED_ACTION_TYPES` in `config.py` allows
-`bugzilla.add_comment` and `bugzilla.update_bug`, but those come from an
-in-process actions server that appends to `summary.json` and makes no network
-calls. The only Bugzilla access the agent has is through the broker sidecar,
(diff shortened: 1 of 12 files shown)


msujaws (approved):

msujaws:
Thanks!

PR #6680, opened 2026-08-20
frontend-triage: flag possible duplicate bugs in the triage comment

crythms:
The agent never checked whether a bug had already been filed. It now does, before it starts investigating the code. When it finds a likely match, the comment opens with one line:

    **Possible duplicate:** [1998432](https://bugzilla.mozilla.org/show_bug.cgi?id=1998432)

It goes first because that's the feedback we've had; an engineer wants to know about dups before reading anything else. When nothing is found, nothing is written and the comment starts with the analysis as it does today.

## It only suggests

Finding a duplicate never changes the triage. The agent still investigates, still writes the fix plan, and the run is still eligible to post itself. An engineer decides whether the bugs are really the takes action as they see fit

## Why a separate subagent

The existing `investigator` has 18 tools and runs at the main agent's model and effort. Reusing it measured roughly 7x the cost for the same search. `duplicate_hunter` gets three Bugzilla tools, low effort, a 6-turn cap, and nothing that can write.

## One shared-library change

`get_bug_comments` takes an optional `limit`. Checking a candidate only needs the original report, but the tool returned the entire comment thread — 11KB for the longer ones, 32KB at worst. Bugzilla ignores `limit` on that endpoint, so the tool truncates after fetching; the point is keeping it out of the agent's context, not saving bandwidth. The default is unchanged, so the five other agents calling it are unaffected.


agents/frontend-triage/hackbot_agents/frontend_triage/prompts/system.md
```diff
@@ -109,6 +110,36 @@ Use it when:
 
 When you spawn an investigator via the Task tool, write a complete, self-contained prompt: what to look at, what question to answer, what format to return. The investigator has no memory of previous spawns.
 
+# Checking for a duplicate
+
+Before you localize, spawn the `duplicate_hunter` subagent once. Give it the bug id, its
+product and component, and the discriminating signal you took from the summary and comment 0
```
suhaibmujahid, line 118: Instead of retrieving the comments, we could simplify by instructing the agent to retrieve `description` (through `include_fields`), which is comment 0. That would be less confusing for the agent and can be retrieved in the same tool call instead of requiring a separate one for comments.

With that, we do not even need to update `libs/agent-tools`.

crythms: done

agents/frontend-triage/hackbot_agents/frontend_triage/prompts/system.md
```diff
@@ -109,6 +110,36 @@ Use it when:
 
 When you spawn an investigator via the Task tool, write a complete, self-contained prompt: what to look at, what question to answer, what format to return. The investigator has no memory of previous spawns.
 
+# Checking for a duplicate
```
suhaibmujahid, line 115: For a system prompt, it is already too long, and this makes it longer. It might be okay for now, but this will not scale nicely. It might be better to rely more on pointing to rules so the agent can read them as needed.

suhaibmujahid (approved):

agents/frontend-triage/hackbot_agents/frontend_triage/prompts/system.md
crythms: You're right.  There's a lot of bloat is all the per-component source tree notes too. I'll clean that up in a separate PR.

PR #6685, opened 2026-08-21
Handle background tasks in agent response loop

### LINKED ISSUE #6656: The bug-fix agent generated a fix but did not use the submit patch tool
The agent produced a patch but did not submit it to Phabricator, examples: 

- https://hackbot.moz.tools/runs/69509c1c-2043-472a-91df-e6bb0e6ce9ce
```json
{
  "status": "ok",
  "error": null,
  "findings": {
    "num_turns": 42,
    "total_cost_usd": 3.5385792499999997,
    "bug_id": 2009361,
    "result": "I've implemented the fix and am waiting on an independent review pass of the patch (no JVM toolchain here, so a manual compile review stands in for `./mach gradle`)."
  },
  "actions": []
}
```
- https://hackbot.moz.tools/runs/b063af3f-33a1-4d48-b7bc-f0b4b23d9b96
```json
{
  "status": "ok",
  "error": null,
  "findings": {
    "num_turns": 45,
    "total_cost_usd": 4.427090499999999,
    "bug_id": 2009361,
    "result": "Patch is in place. Summary of changes so far:\n\n| File | Change |\n|---|---|\n| `browser/state/.../BrowserAction.kt` | new `DefaultDesktopModeAction.ScreenSizeChanged` |\n| `browser/state/.../DesktopModeReducer.kt` | handles it as a side-effect-only action |\n| `fenix/.../DesktopModeRepository.kt` | screen-size lookup injected + evaluated per read (was `by lazy`, cached for process lifetime) |\n| `fenix/.../DesktopModeMiddleware.kt` | re-reads the default on `ScreenSizeChanged`, not just `InitAction` |\n| `fenix/.../HomeActivity.kt` | dispatches `ScreenSizeChanged` when `smallestScreenWidthDp` changes |\n| `fenix/.../ScreenMetricsCompat.kt` | corrects the KDoc that wrongly claimed the value can't change |\n| 3 test files + AC changelog | regression tests for fold/unfold and for \"user preference wins\" |\n\nI'm waiting on the adversarial review of the patch before recording the action."
  },
  "actions": []
}
```

On the same bug before deploying the latest update, it tried to submit a patch: https://hackbot.moz.tools/runs/735cf449-aeee-4f70-b98c-f3a21dab0fe5
```json
{
  "status": "ok",
  "error": null,
  "findings": {
    "num_turns": 81,
    "total_cost_usd": 10.847795750000007,
    "bug_id": 2009361,
    "result": "Recorded.\n\n## Summary\n\n**Bug 2009361** — *Opening and closing the screen does not adjust the \"desktop mode\" on foldable* (Firefox for Android :: Settings, S3, blocks the foldable meta bug 1852822).\n\n**Triage:** No ruleset in `rules/` applies — `unsupported-config.md` is about non-default prefs, and this bug specifies none. Proceeded with the fix directive.\n\n**Root cause** (found in source, deterministic):\n\nThe global desktop-mode default is `persistedPreference ?: isLargeScreenSize()`, but the screen-size half was computed exactly once:\n- `DefaultDesktopModeRepository.defaultDesktopMode` was `by lazy` → memoized for the process lifetime.\n- `DesktopModeMiddleware` only read the repository on `InitAction`.\n- `HomeActivity.onConfigurationChanged()` never re-evaluated it (it only dispatched `OrientationChange`).\n\n`HomeActivity` declares `smallestScreenSize|screenLayout|screenSize` in `configChanges`, so folding resizes the screen without recreating the activity or process — hence \"stale until I close and open it again\", exactly as reported. The reporter's note that a restart gives the right answer also confirms the underlying screen-size read *is* fold-aware, so caching was the whole defect.\n\n**Fix:** stop caching the default; add `DefaultDesktopModeAction.ScreenSizeChanged` dispatched from `onConfigurationChanged()`; the middleware re-reads the preference off the main thread and only dispatches `DesktopModeUpdated` when the effective value changed. An explicitly-set user preference still wins. Also serialized preference reads/writes with a `Mutex` (a fold arriving right after a toggle could otherwise revert the user's choice), and made the Settings switch refresh on resume. Corrected the `isLargeScreenSize()` KDoc that wrongly claimed the value never changes — that claim is what invited the `by lazy`.\n\nScoped deliberately to the *default for new tabs*: already-open tabs keep their per-tab mode, since re-toggling them would force reloads and stomp explicit per-tab menu choices.\n\n**Verification — and its limits, which I flagged in the revision:** this environment has no JDK/Android SDK/Kotlin compiler, and the `firefox` MCP harness runs desktop Firefox testcases under grizzly, which cannot exercise Fenix Android UI state. So I could not compile, run the tests, or reproduce on a device. What I did instead: traced the root cause in source, ran an independent review pass over the diff and fixed all three findings (ktfmt width, the toggle/fold race, a vacuous test replaced with a `CaptureActionsMiddleware` assertion), and added tests at the repository, middleware, and reducer levels that fail before / pass after by construction. QA had already confirmed reproduction on two foldables in comment 1, so the diagnosis does not rest on my own reproduction — but a reviewer should compile, run the unit tests, and confirm on real hardware before landing."
  },
  "actions": [
    {
      "type": "phabricator.submit_patch",
      "params": {
        "bug_id": 2009361,
        "title": "Bug 2009361 - Re-evaluate the default desktop mode when the device screen size changes",
        "summary": "The global \"Desktop site\" default is `preference ?: isLargeScreenSize()`, but the\nscreen-size half of that was only ever computed once:\n\n* `DefaultDesktopModeRepository.defaultDesktopMode` was `by lazy`, so the value\n  was memoized for the whole process lifetime.\n* `DesktopModeMiddleware` only read the repository on `InitAction`.\n* `HomeActivity.onConfigurationChanged()` did not re-evaluate it.\n\nOn a foldable, folding/unfolding changes the screen size but keeps the process\nand the activity alive (HomeActivity declares `smallestScreenSize|screenLayout|\nscreenSize` in `configChanges`), so the default stayed stale until Firefox was\nrestarted — which is what the reporter and QA observed.\n\nChanges:\n\n* `defaultDesktopMode` is no longer cached, and the screen-size source is\n  injectable so it can be varied in tests. The misleading `isLargeScreenSize()`\n  KDoc claiming the value never changes during the app's lifecycle is corrected.\n* New `DefaultDesktopModeAction.ScreenSizeChanged`, dispatched from\n  `HomeActivity.onConfigurationChanged()`. It does not touch state itself;\n  `DesktopModeMiddleware` re-reads the preference off the main thread and only\n  dispatches `DesktopModeUpdated` when the effective value actually changed.\n* A preference the user has set still wins: the read is `preference ?: default`,\n  so folding only moves the default when the user never touched the switch.\n* Reads and writes of the preference are now serialized with a `Mutex`, and the\n  toggled value is captured synchronously in `invoke()`. Without this, a\n  configuration change arriving right after a toggle could read the pre-toggle\n  DataStore value and revert the user's choice.\n* `SiteSettingsFragment` refreshes the switch in `setupPreferences()` instead of\n  binding `isChecked` only once, so it isn't stale after a fold.\n\nAlready-open tabs deliberately keep their own `ContentState.desktopMode`: only\nthe default for newly opened tabs follows the screen size. Re-toggling live tabs\nwould reload them and would override explicit per-tab choices made through the\n\"Desktop site\" menu item.\n\nTests: added coverage for the non-cached default and for the user preference\nwinning over a changed default (`DesktopModeRepositoryTest`), for the store\nbeing updated on `ScreenSizeChanged` and no redundant update being dispatched\n(`DesktopModeMiddlewareTest`), and for the reducer leaving state untouched\n(`DesktopModeReducerTest`).\n\nNote on verification: this was written and reviewed against the source, but the\nenvironment it was produced in has no JDK/Android SDK, so it has not been\ncompiled and the new tests have not been executed. Please run\n`./mach gradle ... testDebugUnitTest` and confirm the behaviour on a real\nfoldable before landing."
      },
      "reasoning": "Bug 2009361: on foldables, the default desktop mode does not follow fold/unfold until Firefox is restarted. Root cause found in source and is unambiguous: DefaultDesktopModeRepository.defaultDesktopMode was memoized with `by lazy`, and the screen-size derived default was only read on InitAction, with no recomputation in HomeActivity.onConfigurationChanged. The reporter's own observation that a restart fixes it confirms the app-context screen-size read is fold-aware and that caching is the defect. Fix removes the cache, adds a ScreenSizeChanged action dispatched on configuration change so the middleware re-reads the preference, keeps an explicitly set user preference authoritative, guards the new read against racing with a concurrent preference write, and refreshes the Settings switch. Unit tests added at repository, middleware and reducer level. No triage ruleset applies (unsupported-config.md concerns prefs; this bug specifies none). Not compiled or run: no JDK/Android SDK in this environment and the firefox MCP harness only runs desktop Firefox testcases, which cannot exercise Fenix Android UI state; this limitation is stated in the revision summary. QA already confirmed reproduction on two foldables (comment 1), so the diagnosis does not rest on my own reproduction."
    }
  ]
}
```

suhaibmujahid:
Fixes #6656

### DIFF
--- a/agents/bug-fix/hackbot_agents/bug_fix/agent.py
+++ b/agents/bug-fix/hackbot_agents/bug_fix/agent.py
@@ -19,12 +19,15 @@ from claude_agent_sdk import (
     ClaudeAgentOptions,
     ClaudeSDKClient,
     McpServerConfig,
-    ResultMessage,
 )
 from hackbot_runtime import ActionsRecorder, AgentError, HackbotAgentResult
 from hackbot_runtime.actions import ACTIONS_SERVER_NAME
 from hackbot_runtime.actions.claude_sdk import actions_server_for, actions_to_tool_names
-from hackbot_runtime.claude import Reporter
+from hackbot_runtime.claude import (
+    Reporter,
+    UnsettledResponseError,
+    receive_settled_response,
+)
 
 from .config import (
     BUGZILLA_READ_TOOLS,
@@ -98,6 +101,7 @@ async def run_bug_fix(
     verbose: bool = False,
     log: Path | None = None,
     actions_recorder: ActionsRecorder | None = None,
+    background_task_timeout_s: float = 3 * 60 * 60,
 ) -> BugFixResult:
     """Triage and fix a single Bugzilla bug with a claude-agent-sdk agent.
 
@@ -165,18 +169,22 @@ async def run_bug_fix(
         setting_sources=[],
     )
 
-    result_msg: ResultMessage | None = None
     with Reporter(verbose=verbose, log_path=log) as reporter:
         reporter.header(f"bug {bug}")
         async with ClaudeSDKClient(options=options) as client:
             await client.query(user_prompt)
-            async for msg in client.receive_response():
-                reporter.message(msg)
-                if isinstance(msg, ResultMessage):
-                    result_msg = msg
+            try:
+                result_msg = await receive_settled_response(
+                    client,
+                    on_message=reporter.message,
+                    timeout_s=background_task_timeout_s,
+                )
+            except UnsettledResponseError as exc:
+                raise AgentError(
+                    f"bug {bug}: agent run did not settle "
+                    f"(a build/test it started may still have been running): {exc}"
+                ) from exc
 
-    if result_msg is None:
-        raise AgentError(f"bug {bug}: agent produced no result message")
     if result_msg.is_error:
         raise AgentError(
             f"bug {bug} triage failed: {result_msg.result or result_msg.subtype}"
--- a/agents/bug-fix/hackbot_agents/bug_fix/agent.py
+++ b/agents/bug-fix/hackbot_agents/bug_fix/agent.py
@@ -180,10 +180,7 @@ async def run_bug_fix(
                     timeout_s=background_task_timeout_s,
                 )
             except UnsettledResponseError as exc:
-                raise AgentError(
-                    f"bug {bug}: agent run did not settle "
-                    f"(a build/test it started may still have been running): {exc}"
-                ) from exc
+                raise AgentError(f"bug {bug}: agent run did not settle: {exc}") from exc
 
     if result_msg.is_error:
         raise AgentError(
--- a/libs/hackbot-runtime/hackbot_runtime/claude.py
+++ b/libs/hackbot-runtime/hackbot_runtime/claude.py
@@ -10,19 +10,31 @@ Requires the ``claude-sdk`` optional extra of hackbot-runtime.
 
 from __future__ import annotations
 
+import asyncio
 import json
+from collections.abc import Callable
 from pathlib import Path
 
 from claude_agent_sdk import (
+    TERMINAL_TASK_STATUSES,
     AssistantMessage,
+    ClaudeSDKClient,
+    Message,
     ResultMessage,
     SystemMessage,
+    TaskNotificationMessage,
+    TaskStartedMessage,
+    TaskUpdatedMessage,
     TextBlock,
     ThinkingBlock,
     ToolResultBlock,
     ToolUseBlock,
     UserMessage,
 )
+
+
+class UnsettledResponseError(RuntimeError):
+    """Used for when ``receive_settled_response`` gave up before the agent's turn settled."""
 
 
 def _truncate(s: str, n: int = 500) -> str:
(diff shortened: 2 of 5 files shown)


libs/hackbot-runtime/hackbot_runtime/claude.py
```diff
@@ -123,3 +135,75 @@ def message(self, msg) -> None:
...
+
+    try:
+        async with asyncio.timeout(timeout_s):
+            async for msg in client.receive_messages():
+                if on_message is not None:
+                    on_message(msg)
+
+                if isinstance(msg, TaskStartedMessage):
+                    pending[msg.task_id] = msg.description
+                elif isinstance(msg, (TaskNotificationMessage, TaskUpdatedMessage)):
```
ayoubdiourin7, line 208: It would be better to limit pending-task tracking only to `local_agent` and `local_workflow` , since these are the tasks whose completion runs a follow-up. ( Background shells and monitors may run indefinitely). Found it here https://github.com/anthropics/claude-agent-sdk-python/blob/main/src/claude_agent_sdk/_internal/query.py#L38-L52

agents/bug-fix/hackbot_agents/bug_fix/agent.py
```diff
@@ -98,6 +101,7 @@ async def run_bug_fix(
     verbose: bool = False,
     log: Path | None = None,
     actions_recorder: ActionsRecorder | None = None,
+    background_task_timeout_s: float = 3 * 60 * 60,
```
ayoubdiourin7, line 104: 3 hours seems too long for the entire agent run.

suhaibmujahid: I agree, but I think about it as a safeguard, that should never happen, but if it happened this is to make it less bad. We should have a sign when this happens and act on it by solving the underlying issues instead of just reducing the general timeout.

libs/hackbot-runtime/hackbot_runtime/claude.py
suhaibmujahid: Nice catch! Fixed in b871759a09dab51085986fa4484652f487729224.

libs/hackbot-runtime/hackbot_runtime/claude.py
```diff
@@ -10,13 +10,21 @@
 
 from __future__ import annotations
 
+import asyncio
 import json
+from collections.abc import Callable
 from pathlib import Path
 
 from claude_agent_sdk import (
+    TERMINAL_TASK_STATUSES,
```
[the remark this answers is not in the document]
suhaibmujahid: Fixed in a54ad72b050a638acb6fcecfc480d9fc5bcd0759

evgenyrp (approved):

PR #6686, opened 2026-08-21
Make autowebcompat repro script easier to use with mozregression

jgraham:
Switch to puppeteer-core since we aren't using the browser download functionality anyway.

Use MOZREGRESSION_ environment variables to determine the browser name/path.

Together this means that one can investigate a regression locally using:

```
npm i puppeteer-core
./mach mozregression --command="node test.mjs"
```


ksy36 (approved):

ksy36:
we might need to update package-lock as well?

jgraham:
I'm sort of surprised that it worked without the package-lock.json update, but it did.

Anyway I updated it now and added an extra comment to the script template, and checked it works locally.

ksy36:
Thanks!

PR #6687, opened 2026-08-21
Install ffmpeg into a few hackbot agent definition

padenot:
This adds two related things in a few agent dockerfiles:
- the Firefox/Chrome in the docker will be able to play h264 and AAC, those are very popular codecs and important for e.g. web compat or just authoring patches
- `ffprobe` and `ffmpeg` will be available as CLI programs. Those are the industry standard tools to poke at and generate media files, for diagnosis and/or test case authoring, testcase reduction, etc.

tested by booting the image and playing an h264/aac file in a real firefox there


jgraham (approved):
autowebcompat changes seem fine

suhaibmujahid (approved):
LGTM! Thank you!

PR #6689, opened 2026-08-21
Support non-headless mode for autowebcompat-repro

jgraham:
This uses fluxbox running in Xvfb as the WM/display, similar to the configuration we have for web-platform-tests


agents/autowebcompat-repro/Dockerfile
```diff
@@ -63,7 +67,14 @@ COPY agents/autowebcompat-repro/hackbot.toml /app/hackbot.toml
 
 RUN useradd --create-home --shell /bin/bash agent \
     && mkdir -p /workspace \
-    && chown agent:agent /workspace /app/repro
+    && chown agent:agent /workspace /app/repro \
+    && usermod -a -G sudo agent \
+    && echo 'ALL ALL = (ALL) NOPASSWD: ALL' >> /etc/sudoers
```
ksy36: Wonder if there's a reason to give the agent passwordless root? fwiw I tried it without that, dropped the sudo package, the usermod/sudoers lines, and the two sudo prefixes in start_xvfb() and the display works fine (took a screenshot from inside the container to confirm).

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/agent.py
```diff
@@ -702,6 +709,40 @@ def into_result(self) -> AutowebcompatReproResult:
...
+                "0",
+                "%sx%sx%s"
+                % (
+                    os.environ["SCREEN_WIDTH"],
+                    os.environ["SCREEN_HEIGHT"],
+                    os.environ["SCREEN_DEPTH"],
+                ),
+            ]
+        )
+        self.start(["sudo", "fluxbox", "-display", os.environ["DISPLAY"]])
```
ksy36: This won't be a problem on production, only for development, but I've noticed if the container is reused (docker compose up on a stopped-but-not-removed container rather than a fresh one), the second run fails to get a display:

```
autowebcompat-repro-agent-1  | (EE) 
autowebcompat-repro-agent-1  | Fatal server error:
autowebcompat-repro-agent-1  | (EE) Server is already active for display 99
autowebcompat-repro-agent-1  |  If this server is no longer running, remove /tmp/.X99-lock
autowebcompat-repro-agent-1  |  and start again.
autowebcompat-repro-agent-1  | (EE) 
autowebcompat-repro-agent-1  | Warning: Failed to open file(fluxbox.cat)
autowebcompat-repro-agent-1  | for translation, using default messages.
autowebcompat-repro-agent-1  | Error: Couldn't connect to XServer:99.0
```

agents/autowebcompat-repro/Dockerfile
jgraham: Apparently not! I thought this was needed, but now removed since it doesn't seem to be.

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/agent.py
jgraham: I've tried to make the shutdown cleaner, I think this will help.

ksy36 (approved):

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/agent.py
ksy36: Thanks!

jrmuizel:
We should port this over to the diagnosis agent so there's not a divergence there.

ksy36:
> We should port this over to the diagnosis agent so there's not a divergence there.

Ah yes, I'll do that

PR #6692, opened 2026-08-22
frontend-triage: load component guidance as needed

crythms:
## Why

The system prompt described where the code lives for all eight areas — Android, the installer, IP Protection, and so on — on every run. A bug only ever needs one. `system.md` also grew every time a component was added. The prompt has doubled in the past  7 weeks.

## What

Each area is now its own file under `rules/areas/`. Before the run starts we look up the bug's component and put in just that area.

| bug component | prompt | vs today |
|---|---:|---:|
| New Tab Page | 19,768 | −39% |
| Application Update | 20,143 | −38% |
| IP Protection | 22,901 | −29% |
| Sharing | 23,169 | −28% |
| unrecognized component | 33,360 | +3% |

The prompt no longer grows as components are added — a new one costs a file, not a tax on every other run.

## Three things stop it being worse than today

1. **When we're not sure, we send everything.** An unrecognized component, or a failed lookup, gets all eight areas — exactly today's prompt. That's the +3% row, and the only case that grows.
2. **Sharing gets site permissions too.** A "stop sharing" bug is filed under Sharing but the code belongs to site permissions, so both go in.
3a. **If the agent works out mid-run that the bug is in another area (component), it can't finish without reading that area.** The comment is refused and told which one to load to continue triage.
3b. **If there is no existing guidance to load**  — fall back to Searchfox as it does today

## Tested on real bugs

Three runs against live Bugzilla, all of which produced a usable fix plan.

| bug | filed under | what happened |
|---|---|---|
| [2065201](https://bugzilla.mozilla.org/show_bug.cgi?id=2065201) | New Tab Page | got Desktop frontend, nothing else |
| [2040869](https://bugzilla.mozilla.org/show_bug.cgi?id=2040869) | Sharing | got Sharing + site permissions, then pulled Desktop frontend itself |
| [2057762](https://bugzilla.mozilla.org/show_bug.cgi?id=2057762) | Installer | worked out it was really the updater and pulled that guidance itself |

**Quality held.** 2065201 was also run with all eight areas, so the two are directly comparable: same three target files, same confidence, and the one-area run found more of the existing tests.

Two bugs turned up, both fixed here:

- **The lookup never worked.** It asked Bugzilla for the component without asking for the bug id, which made the call fail, so every run quietly fell back to all eight areas. It failed the safe way — the triage was still good, just with the old prompt. That is point 1 above working in the wild rather than in a test.
- **Sharing pointed at two paths that no longer exist.** `contentsharing/` and `browser/modules/SharingUtils.sys.mjs` moved into `browser/components/sharing/`, so nothing could ever match to Sharing. Those names have been wrong in the prompt since before this change; every path in `rules/areas/` is now checked against a real checkout.


msujaws (approved):

msujaws:
Thanks!

PR #6698, opened 2026-08-24
report unsubmitted agent patches to Sentry

### LINKED ISSUE #6683: Send a Sentry warning when an agent run has code changes but no patch submitted
This will help catch bugs like #6656

We could do this in the hackbot-api service.

ayoubdiourin7:
Fixes  #6683

### DIFF
--- a/services/hackbot-api/app/routers/runs.py
+++ b/services/hackbot-api/app/routers/runs.py
@@ -272,15 +272,14 @@ async def finalize_run(db: AsyncSession, run: Run) -> None:
         run.error = error
     run.finalized_at = datetime.now(timezone.utc)
 
-    run_id = run.run_id
-    agent = run.agent
     await db.commit()
+
     if _has_unsubmitted_patch(summary, artifacts):
         log.error(
             "Agent run produced code changes without submitting a patch "
             "(run_id=%s, agent=%s)",
-            run_id,
-            agent,
+            run.run_id,
+            run.agent,
         )
     await pubsub.publish_run_completed(str(run.run_id), run.agent, run.status)
 
--- a/services/hackbot-api/app/routers/runs.py
+++ b/services/hackbot-api/app/routers/runs.py
@@ -276,7 +276,7 @@ async def finalize_run(db: AsyncSession, run: Run) -> None:
     agent = run.agent
     await db.commit()
     _capture_unsubmitted_patch_warning(run_id, agent, new_status, summary, artifacts)
-    await pubsub.publish_run_completed(str(run_id), agent, new_status.value)
+    await pubsub.publish_run_completed(str(run.run_id), run.agent, run.status)
 
 
 def _capture_unsubmitted_patch_warning(
--- a/services/hackbot-api/app/routers/runs.py
+++ b/services/hackbot-api/app/routers/runs.py
@@ -290,14 +290,8 @@ def _capture_unsubmitted_patch_warning(
     summary: RunSummary | None,
     artifacts: list[ArtifactRef],
 ) -> None:
-    """Warn when a patch-submitting agent leaves source changes undelivered."""
-    spec = AGENT_REGISTRY.get(agent)
-    if (
-        run_status != RunStatus.succeeded
-        or summary is None
-        or spec is None
-        or not spec.warn_on_unsubmitted_patch
-    ):
+    """Warn when an agent leaves source changes undelivered."""
+    if run_status != RunStatus.succeeded or summary is None:
         return
 
     artifact_names = {artifact.name for artifact in artifacts}
--- a/services/hackbot-api/app/routers/runs.py
+++ b/services/hackbot-api/app/routers/runs.py
@@ -4,7 +4,6 @@ import uuid
 from datetime import datetime, timezone
 from typing import Annotated
 
-import sentry_sdk
 from fastapi import APIRouter, Depends, Header, HTTPException, Query, status
 from hackbot_runtime.actions.phabricator import PATCH_ACTION_TYPES
 from pydantic import BeforeValidator
@@ -34,9 +33,6 @@ log = logging.getLogger(__name__)
 router = APIRouter(dependencies=[Depends(require_api_key)])
 
 _PATCH_ARTIFACT = "changes/changes.patch"
-_UNSUBMITTED_PATCH_WARNING = (
-    "Agent run produced code changes without submitting a patch"
-)
 
 
 def _normalize_identity(email: str | None) -> str | None:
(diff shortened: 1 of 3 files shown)


services/hackbot-api/app/routers/runs.py
```diff
@@ -269,9 +277,47 @@ async def finalize_run(db: AsyncSession, run: Run) -> None:
     run.finalized_at = datetime.now(timezone.utc)
 
     await db.commit()
+    _capture_unsubmitted_patch_warning(run, new_status, summary, artifacts)
```
suhaibmujahid: Reading after `commit()` will trigger a new db query.

services/hackbot-api/app/routers/runs.py
```diff
@@ -269,9 +277,47 @@ async def finalize_run(db: AsyncSession, run: Run) -> None:
...
+        _UNSUBMITTED_PATCH_WARNING,
+        run.run_id,
+        run.agent,
+    )
+    with sentry_sdk.new_scope() as scope:
+        scope.set_context("run", {"run_id": str(run.run_id)})
+        sentry_sdk.capture_message(
+            _UNSUBMITTED_PATCH_WARNING,
+            level="warning",
+        )
```
suhaibmujahid: We do not need to use the Sentry SDK directly; logging an error should be enough.

services/hackbot-api/app/routers/runs.py
```diff
@@ -269,9 +277,47 @@ async def finalize_run(db: AsyncSession, run: Run) -> None:
...
+    summary: RunSummary | None,
+    artifacts: list[ArtifactRef],
+) -> None:
+    """Warn when a patch-submitting agent leaves source changes undelivered."""
+    spec = AGENT_REGISTRY.get(run.agent)
+    if (
+        run_status != RunStatus.succeeded
+        or summary is None
+        or spec is None
+        or not spec.warn_on_unsubmitted_patch
```
suhaibmujahid: We could keep it simple for now without `warn_on_unsubmitted_patch`. If that becomes a noise source for any reason, we could add it later.

suhaibmujahid (changes requested):

services/hackbot-api/app/routers/runs.py
ayoubdiourin7: `expire_on_commit=False` already prevents an extra query, but I moved the reads before commit() to be safe. I also fixed the next line because it did the same thing in [cff1400](https://github.com/mozilla/bugbug/pull/6698/commits/cff14009e8bd84cbba84e70e8c321812314bc061)

services/hackbot-api/app/routers/runs.py
ayoubdiourin7: updated in [2400eb0](https://github.com/mozilla/bugbug/pull/6698/commits/2400eb01fdfdc33d27686860bcab6074ec7e77b3)

services/hackbot-api/app/routers/runs.py
ayoubdiourin7: updated in [46498c4](https://github.com/mozilla/bugbug/pull/6698/commits/46498c49b043ab6fa329313ddbf2d63098857f2b)

services/hackbot-api/app/routers/runs.py
suhaibmujahid: You are right, thank you for pointing that out. I would not modify `pubsub.publish_run_completed()` since that is a bit out of the scope of this PR.

services/hackbot-api/app/routers/runs.py
```diff
@@ -268,8 +272,37 @@ async def finalize_run(db: AsyncSession, run: Run) -> None:
         run.error = error
     run.finalized_at = datetime.now(timezone.utc)
 
+    run_id = run.run_id
+    agent = run.agent
     await db.commit()
-    await pubsub.publish_run_completed(str(run.run_id), run.agent, run.status)
+    _capture_unsubmitted_patch_warning(run_id, agent, new_status, summary, artifacts)
```
suhaibmujahid: To make it clearer, this function should only return a boolean. Then the logging itself would happen here, that would be clearer. For sure, adjust the function name to fit what it is doing.

services/hackbot-api/app/routers/runs.py
```diff
@@ -268,8 +272,37 @@ async def finalize_run(db: AsyncSession, run: Run) -> None:
...
+
+def _capture_unsubmitted_patch_warning(
+    run_id: uuid.UUID,
+    agent: str,
+    run_status: RunStatus,
+    summary: RunSummary | None,
+    artifacts: list[ArtifactRef],
+) -> None:
+    """Warn when an agent leaves source changes undelivered."""
+    if run_status != RunStatus.succeeded or summary is None:
```
suhaibmujahid: If the summery is not there but we have change artifacts, I guess we still need warn.

services/hackbot-api/app/routers/runs.py
```diff
@@ -268,8 +272,37 @@ async def finalize_run(db: AsyncSession, run: Run) -> None:
...
+        return
+
+    artifact_names = {artifact.name for artifact in artifacts}
```
suhaibmujahid: We could do the check in one shot. Cleaner and more efficient. WDYT?

```suggestion
    has_patch_artifacts = any(artifact.name == _PATCH_ARTIFACT for artifact in artifacts)
```

services/hackbot-api/app/routers/runs.py
```diff
@@ -268,8 +272,37 @@ async def finalize_run(db: AsyncSession, run: Run) -> None:
...
+    action_types = {action.get("type") for action in summary.actions}
```
suhaibmujahid: Same here:
```suggestion
    has_patch_actions = any(action["type"] in PATCH_ACTION_TYPES for action in summary.actions)
```

services/hackbot-api/app/routers/runs.py
ayoubdiourin7: Agree, updated in [6584bae](https://github.com/mozilla/bugbug/pull/6698/commits/6584bae6b0a47ee81b9e555de382d0d8a4d5f11b)

services/hackbot-api/app/routers/runs.py
ayoubdiourin7: Yess, it is cleaner, updated in [6584bae](https://github.com/mozilla/bugbug/pull/6698/commits/6584bae6b0a47ee81b9e555de382d0d8a4d5f11b)

services/hackbot-api/app/routers/runs.py
```diff
@@ -268,10 +272,31 @@ async def finalize_run(db: AsyncSession, run: Run) -> None:
...
+    run_id = run.run_id
+    agent = run.agent
     await db.commit()
+    if _has_unsubmitted_patch(summary, artifacts):
+        log.error(
+            "Agent run produced code changes without submitting a patch "
+            "(run_id=%s, agent=%s)",
+            run_id,
+            agent,
+        )
```
suhaibmujahid: nit:
```suggestion
    await db.commit()

    if _has_unsubmitted_patch(summary, artifacts):
        log.error(
            "Agent run produced code changes without submitting a patch "
            "(run_id=%s, agent=%s)",
            run.run_id,
            run.agent,
        )
```

suhaibmujahid (approved):
Thank you!

suhaibmujahid (approved):

PR #6700, opened 2026-08-24
Keep inline comments that have no text and render their suggestions

### LINKED ISSUE #6699: get_phabricator_revision drops inline comments that have no text (suggestion-only)
### What happens

`get_phabricator_revision` (and the `phabricator://revision/D{id}` resource) omits inline review comments whose comment text is empty. The comment does not render as blank -- it is absent entirely, with no header, no file, and no line number. A reader of the tool output has no way to tell a review comment was left there at all.

This bites hardest on "Suggest Edit" inlines. When a reviewer puts everything into the suggestion and leaves the text empty, Phabricator stores `content.raw: ""`, so the entire comment disappears from the tool's output while the web UI shows a replacement.

### Why

`PhabricatorPatch._all_comments` in `bugbug/tools/core/platforms/phabricator.py` filters on comment text:

```python
@cached_property
def _all_comments(self) -> list:
    return [c for c in self.get_comments() if c.content.strip()]
```

For a general comment, empty text means there is nothing to show. For an inline, the location (`path`, `line`, `length`) and the suggestion are carried in `transaction["fields"]`, which `PhabricatorInlineComment.__init__` reads -- but the filter runs first, so those fields are never looked at.

### Reproducing

D319190 comment 1710710 is a suggestion-only inline on `browser/components/tabbrowser/docs/gbrowser.md` line 13. `transaction.search` returns it with `content.raw: ""`; `get_phabricator_revision(319190)` does not mention it.

### Related

[mozilla-conduit/phabricator#102](https://github.com/mozilla-conduit/phabricator/pull/102) ([bug 2024394](https://bugzilla.mozilla.org/show_bug.cgi?id=2024394)) adds `hasSuggestion`, `suggestionText` and `isNewFile` to the inline fields of `transaction.search`. It is merged but not yet deployed. Once it is, the suggestion text becomes available on the same transaction -- but only for comments this filter has not already discarded.

`isNewFile` also resolves the `on_removed_code = None` limitation noted at the same spot, which currently sends `get_matching_hunk` down its guess-the-side heuristic.

daogottwald:
Fixes #6699.

`_all_comments` filtered on `c.content.strip()`, so an inline comment whose text is empty was discarded before `PhabricatorInlineComment` ever read `transaction["fields"]`. A "Suggest Edit" inline where the reviewer put everything into the suggestion has `content.raw: ""`, so it disappeared from `to_md()` entirely -- no header, no file, no line -- and a reader of `get_phabricator_revision` could not tell a review comment was left there.

An inline comment points at a file and a line, so it is worth showing even with no text; a general comment with no text is not. That split is now `is_renderable`. Removed comments are dropped explicitly, since the text filter used to do that incidentally (Phabricator blanks the content of a removed comment).

While here, the two inline fields from mozilla-conduit/phabricator#102 are picked up: `suggestionText` renders as a fenced `suggestion` block, and `isNewFile` resolves `on_removed_code`, which currently sends `get_matching_hunk` down its guess-the-side heuristic. Both are read with `.get()`, so they stay inert until that change is deployed and nothing breaks on a Phabricator that lacks them.

`_sanitize_comments` clears `suggestion_text` alongside `content`, so an untrusted author's suggestion cannot reach a model prompt through the new field.

### Testing

`tests/test_phabricator.py` gains unit coverage for each branch plus an end-to-end `to_md()` test for the reported symptom; the latter fails without the fix.

Locally: 325 passed across `tests/` (excluding `test_nlp.py`, which needs the `nlp` extra). The `test_repository.py` failures on my machine are environmental -- no `hg`, no `rust-code-analysis-web` -- and reproduce identically on `master`.


bugbug/tools/core/platforms/phabricator.py
```diff
@@ -299,10 +307,26 @@ def __init__(self, transaction: dict):
...
+        self.suggestion_text: str | None = (
+            inline_fields.get("suggestionText")
+            if inline_fields.get("hasSuggestion")
+            else None
+        )
+
+        # `isNewFile` says which side of the changeset `start_line` indexes
+        # into. Older Phabricator versions omit it, leaving the side unknown.
+        is_new_file = inline_fields.get("isNewFile")
+        self.on_removed_code = None if is_new_file is None else not is_new_file
```
suhaibmujahid, line 319: Is it supported now?

tests/test_phabricator.py
suhaibmujahid, line 1: I do not think the tests here is very useful.

suhaibmujahid (approved):
Thank you! I left some nitpicks. Feel free mark them as resolved and the PR will be merged automatically.

bugbug/tools/core/platforms/phabricator.py
daogottwald: fixed in https://github.com/mozilla-conduit/phabricator/pull/102

PR #6705, opened 2026-08-24
Add Sentry triage agent & MCP

jpetto:
progress towards adding an agent to triage sentry alerts. stopped at MCP server implementation - do we create our own (as has been started here) or proxy sentry's?


libs/agent-tools/agent_tools/sentry.py
```diff
@@ -0,0 +1,44 @@
...
+def _sentry_error(e: Exception, what: str, http_status_error: str) -> ToolError:
+    """Render a sentry failure as a structured, machine-parseable error."""
+    return ToolError(
+        f"{what}: {e}",
+        payload={"error": http_status_error, "what": what, "message": str(e)},
+    )
+
+
+@tool
+async def get_issue_event(
```
suhaibmujahid, line 26: Why not using the official Sentry MCP instead of rebuilding it?

jpetto: yes, excellent question, and why i opened the issue attached here 😄. should we move discussion to the issue?

PR #6707, opened 2026-08-24
Gate Bugzilla attachment downloads behind a mimetype allowlist

### LINKED ISSUE #6701: frontend-triage (and potentially other agents) should not download and attach screencasts / screen recordings
https://hackbot.moz.tools/runs/5daab632-7b2a-41d6-ae17-7f57bb3bdb8c shows a run that was much more expensive than the others since there was a video attachment on the bug. In the triage results, there is this comment:
> One honest caveat carried into the comment: I could not decode the attached screencast, so this is reasoned from code, and if QA saw the panel body missing rather than the trigger controls, that's a separate issue.

It's important that we continue to include text and image attachments. We should exclude video, zip, and other more complex attachments that Claude can't interpret without specialized agents.

msujaws:
`get_bug_attachments` and `download_attachment` applied no mimetype filter, so a video attachment on a bug got downloaded, read, and billed as tens of thousands of tokens before the agent gave up on decoding it ([run 5daab632](https://hackbot.moz.tools/runs/5daab632-7b2a-41d6-ae17-7f57bb3bdb8c)). `download_attachment` now refuses anything outside `ALLOWED_ATTACHMENT_TYPES` plus every `text/*` subtype, and probes metadata with `exclude_fields=data` first so a refused attachment's bytes never leave Bugzilla. `get_bug_attachments(include_data=true)` is narrower still, since a tool result is text and a base64 image is therefore never an image block: it inlines only textual types under 256KiB, for the first 10 that qualify, and points at `download_attachment` for everything else.

`is_patch` resolves to `text/plain` and `application/octet-stream` resolves from Python's built-in mimetypes table rather than the host's `/etc/mime.types`, so an `update.log` behaves identically on a laptop and in the agent containers. The cost is `1 + K` requests for K inlined attachments, each wrapped so one inaccessible attachment sets `data_error` rather than discarding the ones already fetched. `libs/agent-tools/tests/test_bugzilla.py` goes from 4 tests to 26 and covers `download_attachment`'s body for the first time; `dest_path` is still an unconstrained agent-supplied string and belongs with the path-hardening work.

Fixes #6701

### DIFF
--- a/libs/agent-tools/agent_tools/bugzilla.py
+++ b/libs/agent-tools/agent_tools/bugzilla.py
@@ -9,13 +9,170 @@ denied) as a structured :class:`~agent_tools.registry.ToolError`.
 from __future__ import annotations
 
 import base64
+import mimetypes
 from dataclasses import dataclass
+from pathlib import PurePosixPath
 from typing import Annotated, Any
 
 import bugsy
 from pydantic import Field
 
 from agent_tools.registry import ToolError, tool, tools_in
+
+# What Claude can actually interpret once an attachment reaches it — as an image
+# block via the built-in Read tool, or as text. Deliberately narrower than what
+# Bugzilla accepts: a screencast costs a full download and tens of thousands of
+# tokens and still comes back undecodable, which is what
+# https://github.com/mozilla/bugbug/issues/6701 was filed about. Anything absent
+# is refused, so a new video or archive format needs no matching reject list.
+ALLOWED_ATTACHMENT_TYPES = frozenset(
+    {
+        # The four image formats the Anthropic API accepts as image blocks.
+        "image/png",
+        "image/jpeg",
+        "image/gif",
+        "image/webp",
+        # Not rendered as an image, but readable as markup.
+        "image/svg+xml",
+        "application/pdf",
+        # Textual application/* types Bugzilla serves for logs, testcases, configs.
+        "application/json",
+        "application/xml",
+        "application/javascript",
+        "application/x-javascript",
+    }
+)
+
+# Every text/* subtype is allowed on top of the set above — text/plain, text/html,
+# text/css, text/csv, text/x-log and text/x-phabricator-request are all just text,
+# and enumerating them would only mean missing one.
+_ALLOWED_TYPE_PREFIX = "text/"
+
+# Bugzilla records the type the uploader chose, so a plain log routinely arrives
+# as application/octet-stream, and so does a screencast saved without a type.
+# Resolve those from the file name instead.
+_OCTET_STREAM = "application/octet-stream"
+
+# Python's built-in table, with the system mime database deliberately left out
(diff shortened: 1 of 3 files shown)


libs/agent-tools/agent_tools/bugzilla.py
```diff
@@ -193,22 +321,47 @@ async def get_bug_attachments(
...
+    omitted = 0
+    for att in atts:
+        allowed, effective = attachment_type_allowed(att)
+        if not allowed:
+            att["data_omitted"] = (
+                f"{effective or 'unrecognized type'} is not readable by Claude"
+            )
+            omitted += 1
+            continue
+        one = _request(ctx, f"bug/attachment/{att['id']}")
```
crythms: `_request` raises on any failure, and this call is in a loop with nothing catching it. So
one inaccessible attachment fails the whole call and discards everything already
fetched. Before this change there was a single request - it either worked or it didn't,
it couldn't fail partway through.

this could be wrapped this in try/except, set a `data_error` on the attachment that failed,
and keep going. Same pattern as `comments_error`. 

edge case to consider: maybe worth capping K too. A bug with 40 screenshots is 41 sequential requests inside one call.

libs/agent-tools/agent_tools/bugzilla.py
```diff
@@ -193,22 +321,47 @@ async def get_bug_attachments(
         bool,
         Field(
             description=(
-                "If true, include base64-encoded attachment content. Default "
-                "false. Use sparingly — attachments can be large."
+                "If true, include base64-encoded content for the attachments "
+                "Claude can read. Default false. Use sparingly — attachments "
+                "can be large."
```
crythms: This filter checks the file type, but #6701 was about cost, and cost comes from size.

Bugzilla lets people attach files up to 10 MB. When we base64 an attachment it becomes text, and text costs tokens. A 10 MB attachment wouldn't fit into the context window. Even a  2 MB screenshot runs to hundreds of thousands of tokens and Claude can't see it as an image anyway as base64 in a text result is just a very long string.

`download_attachment` could handle this.  

Suggestion: in the loop, treat allowed-but-binary the same as refused,  skip the
fetch and set data_omitted to something like "image/png - use download_attachment". Inline
only the types that are actually text (text/*, JSON, XML, JavaScript, and SVG, if the
"readable as markup" reasoning above holds)

crythms (changes requested):
added a few suggested changes for review

crythms (approved):
Thanks for addressing the feedback. Looks great!

suhaibmujahid:
@mozilla/hackbot-webcompat is PR okay for you since it changes something you also use?

msujaws:
> @mozilla/hackbot-webcompat is PR okay for you since it changes something you also use?

I caught up with webcompat and they're not using video today in their agent so we can proceed here now.

crythms (approved):

suhaibmujahid (approved):

PR #6709, opened 2026-08-24
Add agents.md

evgenyrp:
This is what I use locally + something else related to conversations that I didn't include here. Many of those have been in my settings for a while, but some specifically aim to reduce verbosity and overengineering by Opus 5. Open to feedback. @padenot @suhaibmujahid

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,33 @@
+# Global
+
+Do not overengineer for the use cases you were not asked to handle.
+
+# Docs
+
+Double check if the docs should be updated in the `/docs` folder.
+Make sure the docs you write are well-structured, concise and human-readable.
+Link to the code where appropriate instead of repeating implementation details in the docs.
+Follow the existing style.
+
+# Comments
+
+Limit the amount of comments you put in the code to a strict minimum.
+You should almost never add comments, except sometimes on non-trivial code, function definitions if the arguments aren't self-explanatory, and class definitions and their members.
+Aim at the code being self-documented.
+Do not remove existing comments unless they are directly related to what you are changing.
+If you do write comments, be concise.
+Do not add a comment explanation to every thing you were asked to correct.
+
+# Code Style
+
+Our style guide forbids the use of emoji.
+Make sure the code is simple and concise.
+
+# Specs
+
+Be concise in writing specs so that they are easily human-readable in short amount of time and easily comprehensible.
+Do not repeat yourself.
+
+# Security
+
+NEVER read .env files. They might include secrets.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,1 @@
+AGENTS.md


AGENTS.md
```diff
@@ -0,0 +1,33 @@
...
+You should almost never add comments, except sometimes on non-trivial code, function definitions if the arguments aren't self-explanatory, and class definitions and their members.
+Aim at the code being self-documented.
+Do not remove existing comments unless they are directly related to what you are changing.
+If you do write comments, be concise.
+Do not add a comment explanation to every thing you were asked to correct.
+
+# Code Style
+
+Our style guide forbids the use of emoji.
+Make sure the code is simple and concise.
```
suhaibmujahid, line 24: Without special instruction, I do not have a problem with these.

AGENTS.md
```diff
@@ -0,0 +1,33 @@
...
+
+# Specs
+
+Be concise in writing specs so that they are easily human-readable in short amount of time and easily comprehensible.
+Do not repeat yourself.
```
suhaibmujahid, line 29: What do you mean with "Specs" here?

AGENTS.md
```diff
@@ -0,0 +1,33 @@
...
+
+# Security
+
+NEVER read .env files. They might include secrets.
```
suhaibmujahid, line 33: We could enforce it in the setting instead, this tends to encourage it to read it instead. I have it enforced globally in my local setup.

AGENTS.md
```diff
@@ -0,0 +1,33 @@
...
+Follow the existing style.
+
+# Comments
+
+Limit the amount of comments you put in the code to a strict minimum.
+You should almost never add comments, except sometimes on non-trivial code, function definitions if the arguments aren't self-explanatory, and class definitions and their members.
+Aim at the code being self-documented.
+Do not remove existing comments unless they are directly related to what you are changing.
+If you do write comments, be concise.
+Do not add a comment explanation to every thing you were asked to correct.
```
suhaibmujahid, line 19: Nice!

AGENTS.md
```diff
@@ -0,0 +1,33 @@
+# Global
+
+Do not overengineer for the use cases you were not asked to handle.
+
+# Docs
+
+Double check if the docs should be updated in the `/docs` folder.
+Make sure the docs you write are well-structured, concise and human-readable.
+Link to the code where appropriate instead of repeating implementation details in the docs.
+Follow the existing style.
```
suhaibmujahid, line 10: This might be expensive to force it. I would leave it out.

AGENTS.md
```diff
@@ -0,0 +1,33 @@
+# Global
+
+Do not overengineer for the use cases you were not asked to handle.
```
suhaibmujahid, line 3: I do not see it a problem in my personal use.

AGENTS.md
evgenyrp: The ones that it writes in planning mode mostly. We could drop it since it's related to my personal workflow, but it can become useful in case we want to submit those to the repo one day. This shouldn't hurt, but either way is fine.

AGENTS.md
evgenyrp: It would be helpful to include it in the repo so it never reads them for the devs who don't have it in their local setup. Do you mean adding settings here along with AGENTS.md? How does it look in your case?

AGENTS.md
evgenyrp: I think this wouldn't hurt, but we can drop it

AGENTS.md
evgenyrp: There were many complains on overengineerg by Opus 5 on the internet. I did feel it became worse in this regard, but it's very hard to measure.

AGENTS.md
evgenyrp: This was one of the main motivations to add AGENTS.md, as we want to make sure the docs are up to date so the agents should be aware of them. I think this will also help them understand the architecture faster without reading a lot of code, so it can become less expensive in the end.

AGENTS.md
padenot: I often see that, but mostly on Gecko.

AGENTS.md
padenot: There's some truth to that in some cases, if the docs are terse. Sometimes it's better handled by special LLM docs (rules, skills, potentially per subdir)

AGENTS.md
padenot: https://claude.com/blog/the-new-rules-of-context-engineering-for-claude-5-generation-models#:~:text=Unhobbling%20Claude is a good read here. I also use something like this for Gecko fwiw, I haven't reevaluated.

AGENTS.md
padenot: It used to be a big problem, but I agree with @suhaibmujahid that I don't see this anymore.

AGENTS.md
padenot: This is best done by sandboxing, I agree with @suhaibmujahid that this can be counter-productive. https://github.com/padenot/moz-local-sandbox works well with `bugbug`, and it can be made to simply block `.env` at the kernel level.

AGENTS.md
evgenyrp: We now see anecdotes about that in ai4dev, not only on the internet, so I would keep it. I've been working with this in my local settings for a while.

PR #6713, opened 2026-08-25
frontend-triage: do not triage a bug that already has a patch

crythms:
The run https://hackbot.moz.tools/runs/ecc7138a-e8c8-4a88-8815-c3a8f9282986,  the agent saw an attached Phabricator revision, said so, and investigated anyway — 13 turns and $1.21 to conclude the work was already done.

`rules/scoping.md` says to stop on these, but the ruleset is just advisory and `TRIAGE_TASK` outranks it.

`preflight.py` reads the bug's attachments before `prepare_repo` and ends the run when a non-obsolete Phabricator revision is attached: no checkout, no model turn, no comment. Everything it can't read (broker failure, dropped field) falls
through to a triage.

Not included: raw `is_patch` attachments (that flag is set by whoever attaches, so a reporter's speculative diff would suppress triage) or closed status.

This may not capture every single case, but it catches the majority of cases the agent will encounter.

### DIFF
--- a/agents/frontend-triage/hackbot_agents/frontend_triage/preflight.py
+++ b/agents/frontend-triage/hackbot_agents/frontend_triage/preflight.py
@@ -1,0 +1,85 @@
+"""Stop a run before it triages a bug whose fix is already written.
+
+``rules/scoping.md`` has always said to stop on these, but a ruleset is advisory:
+on bug 2066504 the agent read the attachment, named the revision, and investigated
+anyway. This is the same rule somewhere the model cannot outrank it.
+
+The gate is one-directional -- all it can do is stop a run -- so everything it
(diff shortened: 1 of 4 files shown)


msujaws (approved):

msujaws:
We could do a separate item that skips if a bug has a comment tagged with 'frontend-triage' so bugs don't get triaged twice

agents/frontend-triage/hackbot_agents/frontend_triage/preflight.py
```diff
@@ -0,0 +1,85 @@
+"""Stop a run before it triages a bug whose fix is already written.
```
suhaibmujahid, line 1: Is there a reason why we need to do this here? It seems that bugbot is a better fit for this.

PR #6717, opened 2026-08-26
Enable the bug-fix agent to push to try

### LINKED ISSUE #6672: Enable the bug-fix agent to push to try when needed
The agent should be able to push to the try server using the action introduced in https://github.com/mozilla/bugbug/pull/6531.  If a try push is submitted, we should link to the push URL in Treeherder within the Phabricator revision description.  

Hint: one way to pass the link for the try push is using the ref mechanism that we already using in Bugzilla comments to get link to Phabricator revisions.

ayoubdiourin7:
Resolve #6672

### DIFF
--- a/libs/hackbot-runtime/hackbot_runtime/actions/try_server.py
+++ b/libs/hackbot-runtime/hackbot_runtime/actions/try_server.py
@@ -123,6 +123,10 @@ async def push(
     Your working-directory changes are pushed as-is, so finish your edits first.
     This records the push for review rather than performing it, so you will not
     see the results.
+
+    Always set ``ref`` so ``submit_patch`` in the same run can reference the new
+    CI push's Treeherder URL as ``{{actions.<ref>.url}}`` in its ``summary``.
+    Record this push before ``submit_patch``.
     """
     cleaned_tasks = [task.strip() for task in (tasks or []) if task and task.strip()]
     if cleaned_tasks and auto:
--- a/libs/hackbot-runtime/hackbot_runtime/actions/try_server.py
+++ b/libs/hackbot-runtime/hackbot_runtime/actions/try_server.py
@@ -124,9 +124,9 @@ async def push(
     This records the push for review rather than performing it, so you will not
     see the results.
 
-    Always set ``ref`` so ``submit_patch`` in the same run can reference the new
-    CI push's Treeherder URL as ``{{actions.<ref>.url}}`` in its ``summary``.
-    Record this push before ``submit_patch``.
+    Always set ``ref`` so a later action in the same run can link the new CI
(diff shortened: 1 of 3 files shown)


libs/hackbot-runtime/hackbot_runtime/actions/try_server.py
```diff
@@ -123,6 +123,10 @@ async def push(
     Your working-directory changes are pushed as-is, so finish your edits first.
     This records the push for review rather than performing it, so you will not
     see the results.
+
+    Always set ``ref`` so ``submit_patch`` in the same run can reference the new
+    CI push's Treeherder URL as ``{{actions.<ref>.url}}`` in its ``summary``.
+    Record this push before ``submit_patch``.
```
suhaibmujahid: This is not the place for "Record this push before ``submit_patch``“. If we need this, it will be case-specific in the agent itself instead of here. An agent could be using try without `submit_patch`, this might encourage unnecessary `submit_patch` or be a bit confusing for the agent.

libs/hackbot-runtime/hackbot_runtime/actions/try_server.py
```diff
@@ -123,6 +123,10 @@ async def push(
     Your working-directory changes are pushed as-is, so finish your edits first.
     This records the push for review rather than performing it, so you will not
     see the results.
+
+    Always set ``ref`` so ``submit_patch`` in the same run can reference the new
```
suhaibmujahid: We can have this instruction in the system prompt, so we tell the agent that if it submits a patch, it should make sure to reference the try push URL in the summery if applicable.

libs/hackbot-runtime/hackbot_runtime/actions/try_server.py
suhaibmujahid: What if `phabricator.submit_patch` comes before `try_server.push`? If that does not work as now, we should file an issue to follow up on improving that so an item with a ref is always applied before the actions that use that ref.

suhaibmujahid (changes requested):

libs/hackbot-runtime/hackbot_runtime/actions/try_server.py
suhaibmujahid: @ayoubdiourin7 wdyt?

suhaibmujahid (approved):
Since it is working, and we see that it references the push in the patch summary, let us move forward with this and file a follow-up issue for the ordering restriction. 
```
Pushed to try for the test suites: {{actions.tryrun.url}}.
```

libs/hackbot-runtime/hackbot_runtime/actions/try_server.py
ayoubdiourin7: Filed a follow-up #6719

PR #6720, opened 2026-08-26
Derive frontend-triage's component docs from mozilla-central instead of restating them

msujaws:
The eight files under `rules/areas/` described where each triaged component's code lives, which mozilla-central already documents and documents better, and ours had drifted from it. This derives the links from the tree instead: one `git grep` for `SPHINX_TREES` finds each component's in-tree docs and the URL to cite, so adding a triage component is now a single entry in `config.py` rather than three edits in two places. The only parts that are kept are things not mentioned elsewhere, like that two unrelated things in this tree are called sharing.


agents/frontend-triage/hackbot_agents/frontend_triage/config.py
```diff
@@ -112,29 +144,78 @@ class Area(NamedTuple):
...
-    Area(
+    ScopedComponent(
+        "Firefox",
         "Sharing",
-        "sharing",
-        ("browser/components/sharing/", "widget/ (the per-OS half)"),
-        # Not `widget/`: that is the whole platform widget layer, and a bug in any
-        # other area citing a file there has nothing to do with sharing a URL out.
+        "#content-sharing-automation",
+        trees=("browser/components/sharing/", "widget/cocoa/", "widget/windows/"),
```
crythms: `trees` now drives doc lookup, and `widget/cocoa/` + `widget/windows/` resolve to generic platform docs. Drop them from trees so the line becomes ` trees=("browser/components/sharing/",),`

the specific files stay in `owns` so the guidance hook still fires on them.

agents/frontend-triage/hackbot_agents/frontend_triage/config.py
```diff
@@ -112,29 +144,78 @@ class Area(NamedTuple):
...
+        notes=(
+            '**Two unrelated things are called "sharing" in this tree.** This '
+            "component is sharing a URL _out_ to another app. Screen, camera and "
+            'microphone sharing (the sharing indicator, the "stop sharing" button, '
+            "per-tab sharing state) is WebRTC. It lives in "
+            "`browser/actors/WebRTCParent.sys.mjs` and belongs to site permissions. A "
+            'grep for `sharing` returns both, so a bug about an indicator or a "stop '
+            'sharing" control is almost certainly the WebRTC one.\n\n'
+            "**The platform half is in `widget/`**, per-OS, and is **not** out of scope "
+            'for being C++ or Objective-C++ rather than JS: "the Share menu is empty", '
```
crythms: Change to line 180 and 181 goes along with the `tree` change above.

```
"**The platform half is in `widget/`**: "
"`widget/nsIMacSharingService.idl` with "
"`widget/cocoa/nsMacSharingService.mm` for macOS, and "
"`widget/windows/nsSharePicker.cpp` for Windows. It is per-OS and **not** "
'out of scope for being C++ or Objective-C++ rather than JS: "the Share menu is empty", '
```

agents/frontend-triage/hackbot_agents/frontend_triage/config.py
```diff
@@ -56,51 +56,83 @@ class ScopedComponent(NamedTuple):
...
+    ScopedComponent(
+        "Firefox",
+        "New Tab Page",
+        "#hnt-dev-triage",
+        trees=("browser/extensions/newtab/", "browser/components/newtab/"),
+        owns=("browser/extensions/newtab/", "browser/components/newtab/"),
+        notes=(
+            "Two directories, and only the first is documented above. "
+            "`browser/components/newtab/` is the about: module and startup-cache glue "
+            "that serves about:home and about:newtab, and nothing documents it. "
```
crythms: Point directly to the startup cache doc.

```
"`browser/components/newtab/` is the about: module and startup-cache glue "
"that serves about:home and about:newtab. The cache it maintains is "
"documented in `browser/extensions/newtab/docs/v2-system-addon/"
"about_home_startup_cache.md` -- filed under the first directory rather "
"than beside the code it describes, which is why it is easy to miss. "
```

crythms (changes requested):
Thanks! This is way better than all the rule files.

I found a few minor things but otherwise all looks greast

crythms (approved):
Thanks!

PR #6725, opened 2026-08-27
Route Toolkit :: Data Sanitization bugs to #privacy-team-automation, and derive its docs from the tree that registers them rather than the tree its code is in

msujaws:
**This should land and deploy before mozilla/bugbot#3002.**


crythms (approved):
👍

PR #6742, opened 2026-08-31
Add a Shared Hackbot API Client lib

### LINKED ISSUE #6635: Create a client library for the Hackbot API in /libs
Add a typed, async `hackbot-client` package under `/libs`, alongside the existing `lando-client`, `phabricator-client`, and `testrail-client`.

ayoubdiourin7:
Resolves #6635


libs/hackbot-client/hackbot_client/models.py
```diff
@@ -0,0 +1,20 @@
...
+
+class RunStatus(str, Enum):
+    pending = "pending"
+    running = "running"
+    succeeded = "succeeded"
+    failed = "failed"
+    timed_out = "timed_out"
+
+
+class RunRef(BaseModel):
```
suhaibmujahid, line 19: Now we have redundant models: https://github.com/mozilla/bugbug/blob/02f19ef4a10b75e99535c9f209cae385e75aebb4/services/hackbot-api/app/schemas.py#L51

We should at least document that to keep them in sync, andfile a followup issue on this. 

To make sure that they keep in sync (out of the scope of this PR), we could use something like [datamodel-code-generator](https://pydantic.dev/docs/validation/latest/integrations/dev-tools/datamodel_code_generator/) to auto generate the models based on the API. Another idea, we could add as some project level tests/linters to compare the schema and make sure that they are identical.

suhaibmujahid (approved):

libs/hackbot-client/hackbot_client/models.py
ayoubdiourin7: Agreed. filed a follow-up issue #6757

PR #6746, opened 2026-08-31
Add link to action_url

### LINKED ISSUE #6473: Link to the TestRail test plan (and revision) from the hackbot UI once submitted
After a test plan is submitted to TestRail, show a link to it in the hackbot UI. Do the same for submitted Phabricator revisions.

johnpangas:
Fixes #6473 
This pull request enhances the `RunDetail` component by adding support for displaying external links associated with each action. Now, if an action contains a `result.url`, an "Open" link will be shown, allowing users to easily access the related external system (such as Bugzilla, TestRail).


services/hackbot-ui/components/RunDetail.tsx
```diff
@@ -271,11 +278,17 @@ export function RunDetail({ runId }: { runId: string }) {
           <ul className="action-list">
             {actions.map((a) => {
```
suhaibmujahid, line 279: We might want to file a follow-up issue to replace `a` with `action` to have better readability.

services/hackbot-ui/components/RunDetail.tsx
```diff
@@ -25,6 +25,13 @@ function commentPreview(a: RunAction): string | null {
   return typeof text === "string" && text.trim() ? text : null;
 }
 
+//Every action may have a result.url, which is a link to the external system (e.g. Bugzilla) where the action was applied.
+// We surface that as an "Open" link in the actions list.
+function actionUrl(a: RunAction): string | null {
```
suhaibmujahid, line 30: Why do we need this function? I guess we could use the URL as it is. WDYT?

suhaibmujahid (approved):

services/hackbot-ui/components/RunDetail.tsx
jpangas: Only some action types return a link, so without the check we would render a dead "Open" link. Keeping it as a named function matches `commentPreview` right above it and similar to the rest of the helpers in the file. I was going to make it inline, but I thought its better to keep it consistent with the rest.

suhaibmujahid: I mean, instead having function (i.e.,  `const url = actionUrl(a)`), simplify it to `const url = a.result?.url`

suhaibmujahid (approved):

services/hackbot-ui/components/RunDetail.tsx
```diff
@@ -271,11 +271,18 @@ export function RunDetail({ runId }: { runId: string }) {
           <ul className="action-list">
             {actions.map((a) => {
               const preview = commentPreview(a);
+              const url =
+                typeof a.result?.url === "string" ? a.result.url : null;
```
suhaibmujahid, line 275: Why do we need the `typeof` part? Do we expect anything other than a string?

johnpangas: We do expect a non string sometimes. E.g. `bugzilla.create_bug` returns a url as None when Bugzilla doesn't give back an id, so the field can be null. Also some actions have no url at all, so it's just missing.
Then TypeScript needsit, RunAction.result is Record<string, unknown>|null and each handler returns a different shape.

johnpangas: We can cast it to simplify it but I think it may be harmful going forward, we shall miss some errors.

suhaibmujahid: If the value is `null`, this check doesn’t change anything.  We already set the final value to `null` if it’s not a string. I’m asking because I can’t see the purpose of this check.

PR #6747, opened 2026-08-31
Update webcompat definition for Firefox unsupported cases

ksy36:
For bugs like https://bugzilla.mozilla.org/show_bug.cgi?id=1902392 / https://bugzilla.mozilla.org/show_bug.cgi?id=2066385 where the warning is shown in all browsers.


agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/prompts/system.md
```diff
@@ -27,6 +27,20 @@ some or all of a website, or which would cause the website to look or
...
 
+A site that steers users away from Firefox is a webcompat issue even if
+the page otherwise renders and works identically in every browser. This
+includes:
+
+- Telling users to switch to, or that the site is best viewed in, another
+  browser (e.g. "it is recommended to use Chrome").
+- Warning that Firefox is unsupported, or that some feature will not work
+  in Firefox.
+- Blocking Firefox outright, or serving it a degraded/legacy experience,
```
jgraham: I guess this would already be covered under the "behaves differently in Firefox compared to other browsers" condition; I might just delete it unless we have reason to believe it's needed here.

jgraham (approved):

agents/autowebcompat-repro/hackbot_agents/autowebcompat_repro/prompts/system.md
ksy36: Sounds good, I removed the third point

jgraham (approved):

PR #6756, opened 2026-09-02
Apply Hackbot actions in reference dependency order

### LINKED ISSUE #6719: Apply actions in reference-dependency order
Follow-up on https://github.com/mozilla/bugbug/pull/6717#discussion_r3863068720.

Action references currently depend on recording order. Actions defining a referenced value should always be applied before actions that use it.

ayoubdiourin7:
Fixes #6719.

- Apply ref producers before their consumers.
- Preserve recorded order for unrelated actions.
- Correctly order actions when one action uses multiple refs or multiple actions use the same ref.

### DIFF
--- a/docs/hackbot/actions.md
+++ b/docs/hackbot/actions.md
@@ -107,12 +107,15 @@ verified-good state is not wanted even if it recorded something before erroring.
 ## Cross-action references
 
 An action's result often isn't known until it's applied — a Phabricator revision has no URL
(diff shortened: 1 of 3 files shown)


suhaibmujahid (commented):
The code in `actions_applier.py` is already ugly, which makes this PR less of a starting point. We might need to do a bit of refactoring here before doing this.

PR #6761, opened 2026-09-02
Route Firefox :: Sidebar to #p10y-bots and Firefox :: Settings UI to #fx-recomp-bots

msujaws:
Neither component was in `TRIAGE_SCOPE`, so a run in either reported to nobody. Both entries get notes for the same reason: each component ships two implementations at once, `sidebar.revamp` inside `#ifdef NIGHTLY_BUILD` and `browser.settings-redesign.enabled` defaulting to true, so which one the reporter saw decides which of two copies of the same control a plan should be written against. Settings UI additionally has no registered `SPHINX_TREES` under anything it owns, so its notes are the whole of its guidance rather than a supplement to the docs.

`Toolkit :: Data Sanitization` gains `Firefox :: Settings UI` in `related` because its notes already name `browser/modules/SiteDataManager.sys.mjs` and `browser/components/preferences/dialogs/siteDataSettings.js` to say the "Manage Data" list is not the sanitizer, and those two paths were unowned until now.

## Testing

114 passed, 1 skipped; the skip is the real-checkout docs test, which passes with `SOURCE_REPO` set and confirms Sidebar resolves its own `docs/` while Settings UI resolves nothing rather than a sibling's tree.

mozilla/bugbot#3012 adds the matching pairs to `TRIAGED_COMPONENTS`. **This should deploy first**, since `channel_for` fails closed and a pair that arrives before the agent knows its channel gets unattended triage with nobody told.

### DIFF
--- a/agents/frontend-triage/README.md
+++ b/agents/frontend-triage/README.md
@@ -26,8 +26,10 @@ and the channel each reports to:
 | Component                         | Reports to                       |
 | --------------------------------- | -------------------------------- |
 | `Firefox :: New Tab Page`         | `#hnt-dev-triage`                |
+| `Firefox :: Sidebar`              | `#p10y-bots`                     |
 | `Firefox :: Site Permissions`     | `#privacy-team-automation`       |
 | `Toolkit :: Data Sanitization`    | `#privacy-team-automation`       |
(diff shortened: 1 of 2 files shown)


crythms (approved):
👍

PR #6762, opened 2026-09-02
Add ai-triage-wrong-fix and ai-triage-shallow-fix to the frontend-triage feedback tags

msujaws:
All four tags the triage comment offers, from `ai-triage-wrong-file` to `ai-triage-out-of-scope`, describe the diagnosis, so a reader who thinks the fix plan itself is wrong, or that it only patches the symptom, has nothing but an uncategorized thumbs-down. `ai-triage-wrong-fix` and `ai-triage-shallow-fix` give those two failure modes a name; the README change is incidental, correcting a claim that the 👍/👎 footer sits below the tags when `bugzilla.add_comment` appends it before the hook runs.


crythms (approved):
thanks!

Issue #6662, opened 2026-08-19
Reduce delay in Test Repair Agent


evgenyrp:
The delay can be substantial (varies from 10 to 40 min). It is critical for provide the infromation to sheriffs ASAP after the failure. The agent itslef typically takes 3-5 minutes + there are some infra delays we need to investigate. I think the main contribution is the listener which needs to resolve whether to run the agent based on intermittent and regression filters. This will likely improve a lot after migrating to Treeherder. See #6574

evgenyrp:
The triggering delay was reduced from ~12 min to 2-6 min, depending on the issue. Next, we need to look into the agent itself and the infra

evgenyrp:
The shallow Firefox cloning takes just about a couple of minutes, so the next major source of delay is the agent itself (the analysis stage).

PR #6782, opened 2026-09-03
Reduce listener delay for CI agents

evgenyrp:
Another iteration of listener improvements with a focus on reducing the delay to trigger the agent. This will increase the number of runs when the agent analyzes an intermittent failure (capped at 100 per 24 hours).

Specifically:
- do not perform lengthy extra checks if the test was never intermittent based on tests.firefox.dev data.
- do not wait for the parent build to finish
- some other fixes for the issues the agent found during the iterative process of dry runs and analysis of the logs

Stats from the last tested dry run:

### Dry run (16.5 h, autoland)

| Metric | Value |
|---|---|
| Window | Sep 2 17:07 → Sep 3 09:37 (16.5 h) |
| Distinct failing tasks seen | 1315 |
| Test-repair would-triggers | 23 (~33 / day) |
| Build-repair would-triggers | 4 |
| Fast path fired | 36 |
| Fast-path fires that reached trigger | 6 |
| Job end → trigger, fast path | 0.1 – 0.6 min |
| Job end → trigger, wait path | 10.2 – 17.1 min |
| History check → trigger gap | 1 – 5 s |
| Ancestor walk stepped over a pending parent | 35 |
| Ancestor walk fail-open after 600 s | 3 |
| Fast-path triggers the skipped wait would have dropped | 0 (all classified 7–9 h later) |
| Log-parse wait exhausted (60 s) | 2 |
| Cross-push dedupe skips | 11 |
| Test triggers later classified intermittent | 19 / 23 (fast path 5/6, wait path 14/17) |
| Budget hits (cap 100) | 0 |
| Errors / tracebacks | 0 |

Redelpoyed live for testing

related to #6662

### DIFF
--- a/services/hackbot-pulse-listener/README.md
+++ b/services/hackbot-pulse-listener/README.md
@@ -48,7 +48,7 @@ Failed **build** tasks go to `build-repair`; failed **test** tasks go to `test-r
    - _Both:_ a walk is abandoned as soon as another task triggers the run for that
      push. One push can emit dozens of failing tasks, and without this each one holds
      a worker for the full wait only to find the push already handed off.
-6. **Budget.** At most `MAX_TEST_REPAIRS_PER_DAY` test-repair runs (default 50) may
+6. **Budget.** At most `MAX_TEST_REPAIRS_PER_DAY` test-repair runs (default 100) may
    start in any rolling 24 hours. A slot is taken when a run is triggered and given
    back if the trigger fails, so only runs that really started count. Once the budget
    is spent, later test failures stop before any Treeherder work. Build-repair is not
--- a/services/hackbot-pulse-listener/README.md
+++ b/services/hackbot-pulse-listener/README.md
@@ -27,7 +27,11 @@ Failed **build** tasks go to `build-repair`; failed **test** tasks go to `test-r
    agent works from a single task but reads the push's other failures itself, so a
    second run for the same push would re-tread the same ground. A revision is recorded
    only once a run is actually triggered, so a task rejected as intermittent or
-   inherited leaves the push open for the next one.
+   inherited leaves the push open for the next one. A second cache spans pushes: a
+   failure already investigated on a recent push is skipped for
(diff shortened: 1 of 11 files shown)


marco-c:
> do not perform lengthy extra checks if the test was never intermittent based on tests.firefox.dev data.

How many tests do we have that were never intermittent? Maybe we can have a configurable threshold instead of 0, say start with a very low number like 0.1%.

evgenyrp:
> > do not perform lengthy extra checks if the test was never intermittent based on tests.firefox.dev data.
> 
> How many tests do we have that were never intermittent? Maybe we can have a configurable threshold instead of 0, say start with a very low number like 0.1%.

We do check the failure rate against the threshold; see https://github.com/mozilla/bugbug/pull/6782/changes#diff-c516aab8c954b9767286eefb7f8c2cb34ace3122326bbad4dd31760a5e28d413R250

Here's what we have in the logs now:

```
2026-09-03 14:10:21.024
2026-09-03 21:10:21 INFO    app.flakiness: Test devtools/client/inspector/rules/test/browser_rules_inherited-element-backed-pseudo-elements.js failed 5432 of its last 11690 runs (46.467%); it may be an intermittent
2026-09-03 14:04:08.844
2026-09-03 21:04:08 INFO    app.flakiness: Test toolkit/content/tests/chrome/test_findbar.xhtml failed 1166 of its last 13025 runs (8.952%); it may be an intermittent
2026-09-03 14:03:23.632
2026-09-03 21:03:23 INFO    app.flakiness: Test gfx/layers/apz/test/mochitest/test_group_touchevents-6.html failed 148 of its last 17102 runs (0.865%); it may be an intermittent
```

evgenyrp:
Trigger time has been reduced from 11-13 minutes to 2-6, so the agent is not often faster than the sheriffs. https://hackbot-ci-agents-analytics.quick.mozilla.cloud/ This one is good to go.

suhaibmujahid (approved):

PR #6789, opened 2026-09-04
Route the three Core :: Machine Learning components to #smart-window-bug-triage

msujaws:
Smart Window work is filed under three Core Machine Learning components the agent has no channel or trees for, so those bugs report to nobody. `moz.build` assigns `browser/components/aiwindow/models/` to the General component while the bugs about that directory are filed under Models, so both claim it and the registry reads the identical claim as co-ownership.

**Deploy before the matching bugbot change**: `channel_for` fails closed, so the reverse order posts unattended analyses with nobody told.


crythms (approved):
👍

PR #6802, opened 2026-09-07
Move action handlers to hackbot api

### LINKED ISSUE #6754: Move Hackbot action handlers to hackbot-api service
They are currently in `hackbot-runtime`, while these handlers are not a runtime related. The actions themselves are.

With that, the  `hackbot-api` service does not need to depend on `hackbot-runtime`, and we could eliminate some of the dependencies of `hackbot-runtime` itself as well.

ayoubdiourin7:
This PR is best reviewed commit by commit:

Fixes #6754


services/hackbot-api/tests/test_try_server_handler.py
```diff
@@ -9,16 +9,29 @@
...
+
+def _git(repo, *args: str) -> str:
+    """Run a git command in ``repo`` and return its stdout."""
+    return subprocess.run(
+        ["git", "-C", str(repo), *args],
+        check=True,
+        capture_output=True,
+        text=True,
+    ).stdout
+
```
suhaibmujahid, line 33: Why do we need this here? The hackbot-api service does not have a git repo!

services/hackbot-api/app/action_handlers/email_handler.py
```diff
@@ -26,12 +26,16 @@
...
-from hackbot_runtime.changes import PATCH_ARTIFACT
+from app.action_handlers.base import ActionResult, ApplyContext
 
 log = logging.getLogger(__name__)
 
+# Artifact containing a run's source-code patch.
+_PATCH_ARTIFACT = "changes/changes.patch"
+
+# Email-body token replaced with the source-code patch.
+PATCH_PLACEHOLDER = "{patch}"
```
suhaibmujahid, line 37: why are we adding this?

services/hackbot-api/app/action_handlers/registry.py
```diff
@@ -1,21 +1,24 @@
...
 )
-from hackbot_runtime.actions.handlers.slack_handler import PostMessageHandler
-from hackbot_runtime.actions.handlers.testrail_handler import SubmitTestPlanHandler
-from hackbot_runtime.actions.handlers.try_server_handler import PushHandler
+from app.action_handlers.slack_handler import PostMessageHandler
+from app.action_handlers.testrail_handler import SubmitTestPlanHandler
+from app.action_handlers.try_server_handler import PushHandler
+
+# Actions that submit source changes to Phabricator.
+PATCH_ACTION_TYPES = frozenset({"phabricator.submit_patch", "phabricator.update_patch"})
```
suhaibmujahid, line 21: Is this a duplicate now? If so, we should file a follow-up issue and link it here, so we could fix that later.

If we have other cases, we could use the same issue to track all of them.

ayoubdiourin7: This is duplicated because both packages need it.  Here is the  follow-up #6822

services/hackbot-api/app/action_handlers/email_handler.py
ayoubdiourin7: This is not new. It was temporarily in  [contract.py`](https://github.com/mozilla/bugbug/pull/6802/changes/10471ef7fc30839176a104618a2e56dfa1d691a3#diff-e0e835fd81973ec95f50fc5933bf24c73b868c40fe50e755a9a3128132dc87f5R7)  during the move and is now kept locally because only this handler uses it.

services/hackbot-api/tests/test_try_server_handler.py
ayoubdiourin7: This was already used by the try-server handler tests to verify that the generated patch can be applied with `git am`. I only moved the helper here to avoid depending on `hackbot-runtime`.

suhaibmujahid (approved):
Thank you! Please solve the conflicts an marge it.

PR #6803, opened 2026-09-07
Support service account authentication in the Hackbot API

### LINKED ISSUE #6472: Support authentication through service accounts
Allow the services to authenticate using service accounts.

ayoubdiourin7:
Resolves  #6472

### DIFF
--- a/libs/hackbot-client/hackbot_client/client.py
+++ b/libs/hackbot-client/hackbot_client/client.py
@@ -2,23 +2,57 @@
 
 from __future__ import annotations
 
+import asyncio
 from collections.abc import Mapping
 from typing import Any
 
 import httpx
+from google.auth.transport import requests as google_requests
+from google.oauth2 import id_token
 
 from hackbot_client.models import RunRef
 
 
 class HackbotClient:
-    """Call the public, API-key-authenticated Hackbot endpoints."""
+    """Call the public Hackbot API with either API key or service account auth."""
 
     def __init__(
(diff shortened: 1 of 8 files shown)


libs/hackbot-client/hackbot_client/client.py
```diff
@@ -2,24 +2,58 @@
...
 class HackbotClient:
-    """Call the public, API-key-authenticated Hackbot endpoints."""
+    """Call the public Hackbot API with either API key or service account auth."""
 
     def __init__(
-        self, base_url: str, api_key: str, timeout_seconds: float = 30.0
+        self,
+        base_url: str,
+        api_key: str = "",
+        audience: str = "",
```
suhaibmujahid: We could drop `audience` here and use `base_url` for that.

libs/hackbot-client/hackbot_client/client.py
```diff
@@ -2,24 +2,58 @@
...
 
 
 class HackbotClient:
-    """Call the public, API-key-authenticated Hackbot endpoints."""
+    """Call the public Hackbot API with either API key or service account auth."""
 
     def __init__(
-        self, base_url: str, api_key: str, timeout_seconds: float = 30.0
+        self,
+        base_url: str,
```
suhaibmujahid: We should have a default value here:

```suggestion
        base_url: str = "https://hackbot-api.moz.tools",
```

Issue #6642, opened 2026-08-18
Add a rule/guardrail mechanism for Hackbot actions


suhaibmujahid:
Add a generic policy layer that can enforce rules across actions and agents, for example: no two try pushes in the same run, no two patch submissions in the same run, and refusing to submit a patch to a bug other than the one the run is assigned to.

suhaibmujahid:
Another example, a Phabricator revision title should not select reviewers (e.g., `r=...` or `r?...`).

PR #6811, opened 2026-09-08
Prevent Hackbot from adding reviewers to Phabricator revision titles

### LINKED ISSUE #6778: Hackbot should not add a reviewer when submitting a patch to Phabricator
The reviewer selection should be handled by Herald rules on Phabricator side.

ayoubdiourin7:
Fixes #6778


suhaibmujahid (approved):
This is good. But it does not prevent. We could do the preventing part in #6642.

PR #6818, opened 2026-09-09
Keep the build-repair blame when the culprit landed in an earlier push

### LINKED ISSUE #6788: Build repair agent contradicts itself with the culprit commit
The blamed commit doesn't match the one it says is the culprit in the summary.

https://hackbot.moz.tools/runs/4702442a-32b3-4ec0-bfb9-e8a9d8dd69b2

evgenyrp:
- Write the correct blame commit to results when it's from another push
- Notify the author of the culprit from another push
- Communicate clearly in the email that the culprit is from an earlier push

fixes #6788


agents/build-repair/hackbot_agents/build_repair/config.py
```diff
@@ -12,6 +12,10 @@
 # developer about it is noise.
 NOTIFY_ONLY_WITH_PATCH = True
 
+# Commits the shallow checkout reaches back, so the agent can inspect the pushes
+# before the failing one when the job did not run on every push.
+CHECKOUT_DEPTH = 500
+
```
suhaibmujahid, line 18: At some point, we will have an optimized setup where we have the full repo. But for now, I'm curious why we need 500 commits?

suhaibmujahid (approved):

agents/build-repair/hackbot_agents/build_repair/config.py
evgenyrp: To find the real culprit in the recent history and notify the author

PR #6819, opened 2026-09-09
Ignore quoted Hackbot mentions in Phabricator replies

### LINKED ISSUE #6815: Hackbot should not be triggered by Phabricator mentions in replies
Example: https://phabricator.services.mozilla.com/D324282#11273578

ayoubdiourin7:
Fixes #6815

### DIFF
--- a/services/hackbot-api/app/phabricator_webhook.py
+++ b/services/hackbot-api/app/phabricator_webhook.py
@@ -61,13 +61,13 @@ def find_hackbot_mentions(
 ) -> list[HackbotMention]:
     """Return every triggering comment that mentions ``token``.
 
-    Only considers comment transactions named in this delivery and not authored
-    by the bot itself (loop prevention). Mentions that appear only in Remarkup
-    quote lines are ignored; a fresh, unquoted mention still counts. A single
-    review can leave several inline comments, each represented by its own
-    transaction, so all matches are returned in transaction order. At most one
-    match is returned per transaction because its ``comments`` list contains the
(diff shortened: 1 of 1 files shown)


services/hackbot-api/app/phabricator_webhook.py
```diff
@@ -25,6 +25,15 @@
 _COMMENT_TYPES = frozenset({"comment", "inline"})
 
 
+def _strip_quoted_lines(text: str) -> str:
+    """Return ``text`` with every line that starts with ``>`` removed.
+
+    Phabricator uses these lines for quoted replies. Excluding them prevents a
+    mention copied from an earlier comment from being treated as a new request.
+    """
+    return "\n".join(line for line in text.splitlines() if not line.startswith(">"))
```
suhaibmujahid, line 34: I'm not sure if `lstrip()` is needed here. We could verify that by looping over hustrcly comments in bugbug data (https://github.com/mozilla/bugbug/blob/master/docs/data.md#phabricator-revisions).
```suggestion
    return "\n".join(line for line in text.splitlines() if not line.lstrip().startswith(">"))
```

suhaibmujahid (approved):
LGTM! Thank you!

services/hackbot-api/app/phabricator_webhook.py
ayoubdiourin7: In the dataset, `>` appears at the beginning of the line, so I don’t think lstrip() is necessary here.

PR #6820, opened 2026-09-09
Revert "Handle background tasks in agent response loop"

### LINKED ISSUE #6812: Bug fix agent is not traced by Weave
The bug-fix agent uses `receive_settled_response`, which drains `client.receive_messages()`. Weave patches only `receive_response`.

suhaibmujahid:
This reverts commit e9a891a51393fbb000fe9a848cedb880e3de9307.

Fixes #6812


libs/agent-tools/pyproject.toml
```diff
@@ -15,7 +15,7 @@ bugzilla = [
     "six",
 ]
 firefox = ["grizzly-framework", "prefpicker"]
-claude-sdk = ["claude-agent-sdk>=0.2.30"]
+claude-sdk = ["claude-agent-sdk>=0.1.30"]
```
evgenyrp, line 18: This is probably fine to have it as a lower number, but we'll want to use the latest SDK most of the time as things improve fast

uv.lock
evgenyrp, line 1: If it's a pure revert commit, it's better to relock the uv not to lose the latest updates

evgenyrp (approved):
Approved with a few comments

uv.lock
suhaibmujahid: Looking at the diff, it does not seem to downgrade anything.

libs/agent-tools/pyproject.toml
suhaibmujahid: Yes, but we also did a cap on the upper version in another PR as a workaround to fix a bug until it is fixed upstream.

PR #6821, opened 2026-09-09
Submit build repair patch to Phabricator

### LINKED ISSUE #6661: Integrate Build Repair Agent with Phabricator
It should be possible to review and land the patch with Phabricator. There are a few things we need to clarify here based on developers' preferences.

evgenyrp:
- the agent can submit the patch to Phabricator with an action

fixes #6661


libs/hackbot-runtime/hackbot_runtime/changes.py
suhaibmujahid, line 1: In general, changes to the runtime or API should be abstract/generic and not tightly coupled to a specific agent. As such, these are better submitted for review as independent changes, a stacked PRs would be preferable here. 

In any case, minimizing changes to shared components as much as possible is strongly encouraged.

evgenyrp: This is needed for the agent to commit changes. Stacks would be nice, but they don't work from forks, unfortunately.

libs/hackbot-runtime/hackbot_runtime/context.py
```diff
@@ -140,12 +140,15 @@ async def prepare_repo(
...
         # Record where the agent starts editing, so publish_changes() can later
-        # diff the final tree against it. Best-effort: a failure here must not
-        # break the agent's access to source — it only disables change capture.
+        # diff the final tree against it, and give the checkout an identity so an
+        # agent that commits its own work can do so with a plain `git commit`.
+        # Both best-effort: a failure here must not break the agent's access to
+        # source — it only costs change capture or a nicer commit message.
         try:
             self._source_base = self._published_base = changes.base_commit(path)
+            changes.ensure_git_identity(path)
```
suhaibmujahid: I do not understand the use cases here?

services/hackbot-api/app/agents.py
```diff
@@ -99,9 +99,11 @@ def model_to_env(inputs: BaseModel) -> dict[str, str]:
         description="Analyze a Firefox build failure at a specific commit and produce a candidate fix patch.",
         job_name="hackbot-agent-build-repair",
         input_schema=BuildRepairInputs,
-        # Its only action is the failure-analysis email, which used to be sent
-        # unconditionally by the pulse listener.
+        # The failure-analysis email keeps going out on its own (it used to be
+        # sent unconditionally by the pulse listener), but the fix waits: a
+        # developer reads the patch in the UI before a revision is opened for it.
         auto_apply_actions=True,
+        never_apply_actions=frozenset({"phabricator.submit_patch"}),
```
suhaibmujahid: nit: I fee this clearer and does not even need comments:

```suggestion
        auto_apply_actions=False,
        always_apply_actions=frozenset({"email.send"}),
```

agents/build-repair/hackbot_agents/build_repair/prompts.py
```diff
@@ -173,3 +173,32 @@
...
+
+Pass bug_id={bug_id}, a `title` of the form "Bug {bug_id} - <what the fix does>",
+a `summary` naming the busted commit, the failing task, the root cause and the
+fix, and the `reasoning` it stores for the audit log. Your final working-tree
+edits are submitted as the diff, so make every edit first and pass no patch file.
+Do not attach the patch to the bug instead, and do not comment on the bug.
+
+If the fix does not build, or you are not confident in it, record nothing and say
+so in your final message.
+"""
```
suhaibmujahid, line 183: I do not think you need to explain how to use the `phabricator_submit_patch` tool; the tool description already takes care of that.

agents/build-repair/hackbot_agents/build_repair/prompts.py
```diff
@@ -173,3 +173,32 @@
...
+COMMIT_INSTRUCTIONS = """
+Once the build passes, commit the fix so the patch carries a real message. Stage
+only the files you edited (`git add <path> ...`, never `git add -A`) and commit
+with the subject "{bug_prefix}<what the fix does, imperative, under 72 chars>"
+and a body of one short paragraph: which commit broke the build, how it broke,
+and what this change does about it. An identity is already configured, so a plain
+`git commit` works. Leave nothing uncommitted -- anything you skip is squashed
+into a nameless "Uncommitted agent changes" commit instead. End the message after
+that paragraph: no Co-Authored-By, no Signed-off-by, no tool attribution.
+"""
```
suhaibmujahid: I do not know if we really need this, the Phab revision title will be used for the commit.

suhaibmujahid (changes requested):
I suggest taking out the UI changes in a separate PR since they are generic improvements, not tight to this PR, and we could discuss options there.

agents/build-repair/hackbot_agents/build_repair/prompts.py
evgenyrp: this now explains only what's relevant to the build repair agent

libs/hackbot-runtime/hackbot_runtime/context.py
evgenyrp: removed commiting to git

agents/build-repair/hackbot_agents/build_repair/prompts.py
evgenyrp: removed

evgenyrp:
> I suggest taking out the UI changes in a separate PR since they are generic improvements, not tight to this PR, and we could discuss options there.

Moved UI to a separate PR https://github.com/mozilla/bugbug/pull/6856 and removed commiting to Git and runtime changes.

suhaibmujahid (approved):

PR #6824, opened 2026-09-10
Route Firefox :: General, Toolkit :: General and Firefox :: Untriaged to #fx-toolkit-general-triage-notifications

msujaws:
`TRIAGE_SCOPE` covers 16 narrow components, each owned by a team, and none of the three buckets a reporter lands in when they cannot pick a component. Those are the filings that most need a first pass, so this adds all three to one channel for the people who work the unowned queue.

None of the three sets `owns`, which is the one decision here worth a look. `owns` is what `component_guidance_hook` refuses comments against and `owners_for_path` resolves longest-claim-wins, so claiming `browser/` for General would make every unclaimed file of desktop chrome General's and a Sidebar or Settings UI run would then have its comment refused for citing its own code. The README's own "Adding a triage component" recipe already names that trap.

`docs_for` keys off `trees`, and a bucket has no narrow tree to give it. Measured against a mozilla-central checkout, the three arrive with 28, 33 and 28 doc registrations out of 135 tree-wide, where every existing component gets between zero and three. `docs.py` says the fix for a broad tree is narrowing the entry rather than filtering in the lookup, and there is nothing to narrow to here, so this is a real cost rather than one I can design away.

## Testing

`uv run --package hackbot-agent-frontend-triage pytest agents/frontend-triage/tests` is 115 passed with `SOURCE_REPO` pointed at a real checkout, 114 passed and 1 skipped without it. No test file needed editing: the channel and registry-shape tests loop over `TRIAGE_SCOPE`.

**This should land and deploy before the matching mozilla/bugbot change**, which adds the same three pairs to `TRIAGED_COMPONENTS`. `channel_for` fails closed, so a pair that arrives before the agent knows its channel is silent on Slack but still comments on the bug and still assesses severity.

### DIFF
--- a/agents/frontend-triage/README.md
+++ b/agents/frontend-triage/README.md
@@ -23,24 +23,27 @@ component's code lives in. A bug handed to the agent by hand in some other compo
 (`Firefox :: Menus`, say) is triaged the same way and reports to nobody. The components,
 and the channel each reports to:
 
-| Component                            | Reports to                       |
-| ------------------------------------ | -------------------------------- |
-| `Firefox :: New Tab Page`            | `#hnt-dev-triage`                |
-| `Firefox :: Sidebar`                 | `#p10y-bots`                     |
-| `Firefox :: Site Permissions`        | `#privacy-team-automation`       |
-| `Toolkit :: Data Sanitization`       | `#privacy-team-automation`       |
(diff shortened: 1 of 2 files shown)


crythms (approved):
👍

PR #6826, opened 2026-09-11
Centralize recorded action confirmation messages

### LINKED ISSUE #6697: Centralize recorded action confirmation messages
Recorded action tools build the same confirmation message independently across Bugzilla, Phabricator, Slack ..., Centralise these confirmation messages in a shared helper

ayoubdiourin7:
Fixes #6697


suhaibmujahid (approved):
As a follow up, we could covert that action to be a class instead of dict (maybe Pydantic model), then the confirmation could be a method on that action class.

Issue #6629, opened 2026-08-18
Replace repeated `x.get(y) or {}` chaining with a safer accessor


suhaibmujahid:
This defensive-defaulting idiom is widespread, for example in `services/hackbot-pulse-listener/app/{taskcluster,github,notify}.py` and `services/hackbot-api/app/{actions_applier,routers/events,phabricator_webhook,routers/webhooks}.py`. Introduce a small helper, or normalize optional fields at the parsing boundary, instead of re-deriving `{}` fallbacks ad hoc throughout the codebase.

PR #6828, opened 2026-09-11
Use first public parent for unresolvable Phabricator bases (#6792)

### LINKED ISSUE #6792: Failing to checkout out a revision with an unknown base
Example: https://hackbot.moz.tools/runs/16d4c648-693f-4073-be6f-c5ce5aa42b25

In this case it is a stack where the first revision in the stack is based on a commit that seem to not be in the repo.

```
UnresolvedCommitError: Cannot expand 0a336b1a481b to a full commit hash: Diffusion does not know one; the commit may not be imported, e.g. an unlanded parent of a stacked patch.
```

Related: https://bugzilla.mozilla.org/show_bug.cgi?id=2068637

olliehcrook:
Fixes #6792 after [moz-phab patch](https://phabricator.services.mozilla.com/D324588) is rolled out. 
Safe to add before - falls back to previous behaviour if first public commit is not present.

related: 
-  https://github.com/mozilla/libmozdata/pull/304
- https://github.com/mozilla/code-review/pull/3626

### DIFF
--- a/libs/phabricator-client/phabricator_client/models.py
+++ b/libs/phabricator-client/phabricator_client/models.py
@@ -23,10 +23,21 @@ class PhabricatorDiff(BaseModel):
     base_commit: str | None = Field(default=None, alias="sourceControlBaseRevision")
     author_name: str | None = Field(default=None, alias="authorName")
     author_email: str | None = Field(default=None, alias="authorEmail")
+    properties: dict = Field(default_factory=dict)
 
     @property
     def author(self) -> str | None:
         """The author as git wants it, ``Name <email>``, or ``None`` if unknown."""
         if self.author_name and self.author_email:
             return f"{self.author_name} <{self.author_email}>"
+        return None
+
+    @property
+    def first_public_parent(self) -> str | None:
+        """The first landed parent recorded by mozphab, if present."""
+        local_commits = self.properties.get("local:commits") or {}
+        for commit in local_commits.values():
+            first_public_parent = commit.get("firstPublicParent")
+            if first_public_parent:
+                return first_public_parent
         return None
--- a/libs/hackbot-runtime/hackbot_runtime/revision.py
+++ b/libs/hackbot-runtime/hackbot_runtime/revision.py
@@ -7,7 +7,11 @@ import subprocess
 from pathlib import Path
 from typing import TYPE_CHECKING, NamedTuple
 
-from phabricator_client import PhabricatorClient, PhabricatorSettings
+from phabricator_client import (
+    PhabricatorClient,
+    PhabricatorSettings,
+    UnresolvedCommitError,
+)
 
 from hackbot_runtime import changes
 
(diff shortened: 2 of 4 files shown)


libs/phabricator-client/phabricator_client/models.py
```diff
@@ -25,10 +25,21 @@ class PhabricatorDiff(BaseModel):
...
     def author(self) -> str | None:
         """The author as git wants it, ``Name <email>``, or ``None`` if unknown."""
         if self.author_name and self.author_email:
             return f"{self.author_name} <{self.author_email}>"
         return None
+
+    @property
+    def first_public_parent(self) -> str | None:
+        """The first landed parent recorded by mozphab, if present."""
+        local_commits = self.properties.get("local:commits") or {}
```
suhaibmujahid, line 40: nit: This is not typical, and it is useful to know about it when debugging. Also, I do not like `or {}` (see #6629) ;)
```suggestion
        local_commits = self.properties.get("local:commits")
        if not local_commits:
            log.warning("The diff has no commit metadata")
            return None
```

libs/phabricator-client/phabricator_client/models.py
```diff
@@ -25,10 +25,21 @@ class PhabricatorDiff(BaseModel):
...
+        for commit in local_commits.values():
+            first_public_parent = commit.get("firstPublicParent")
```
suhaibmujahid, line 42: If it is expected to have `firstPublicParent` missing, it is ok to use `.get()`, otherwise, I would avoid that and use direct access (i.e., `commit["firstPublicParent"]`)

libs/phabricator-client/phabricator_client/models.py
```diff
@@ -25,10 +25,21 @@ class PhabricatorDiff(BaseModel):
     base_commit: Annotated[str | None, Field(alias="sourceControlBaseRevision")] = None
     author_name: Annotated[str | None, Field(alias="authorName")] = None
     author_email: Annotated[str | None, Field(alias="authorEmail")] = None
+    properties: dict = Field(default_factory=dict)
```
suhaibmujahid, line 28: Since this is not a regular class, direct assessment of `{}` should be okay.
```suggestion
    properties: dict = {}
```

libs/phabricator-client/phabricator_client/models.py
suhaibmujahid, line 40: What do we expect here?

suhaibmujahid (approved):
Thank you! LGTM, just left some minor comments.

libs/phabricator-client/phabricator_client/models.py
olliehcrook: Yes, it could be missing in the case where either sourceBaseRevision == public parent, or if moz-phab version doesn't have public parent yet.

libs/phabricator-client/phabricator_client/models.py
olliehcrook: yeah I agree that this is not typical, even before adding public parent, local:commits should always be there

PR #6856, opened 2026-09-15
Add a patch panel to preview a run's changes

evgenyrp:
The panel looks like this. I dropped the description because the agent doesn't commit the changes by design now. It will be a part of the Phabricator action.
Screenshot 2026-09-15 at 3 44 19 PM


services/hackbot-ui/app/api/runs/[runId]/artifacts/[...path]/route.ts
```diff
@@ -20,10 +30,25 @@ export async function GET(
...
+    if (!res.ok) {
+      throw new HackbotError(
+        `Could not read ${artifactName} (${res.status})`,
+        res.status
+      );
+    }
+    const buf = await res.arrayBuffer();
+    const truncated = buf.byteLength > RAW_MAX_BYTES;
+    const text = new TextDecoder().decode(buf.slice(0, RAW_MAX_BYTES));
+    return NextResponse.json({ text, truncated });
```
Comment (suhaibmujahid) [inline services/hackbot-ui/app/api/runs/[runId]/artifacts/[...path]/route.ts]: Why are we doing this here instead of on the client side? The redirect approach makes it free for us since the artifacts are served from GCS directly instead of us processing it in the middle.

services/hackbot-ui/components/PatchView.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+import { useEffect, useState } from "react";
+
+import { parsePatch, type PatchFile } from "@/lib/patch";
+
+// The artifact every agent that edits source publishes (see
+// HackbotContext.publish_changes).
+export const PATCH_ARTIFACT = "changes/changes.patch";
+
+export function PatchPanel({
```
suhaibmujahid, line 15: Do we need to implement this ourselves? It might be better to use a package for this.

In the past, I used `diff2html`, but I guess nowadays we might have more modern options, especially with React.

https://github.com/mozilla/bugbug/blob/df1218f262b816c17908538e0382eb49e8cbcb13/functions/diff2html/package.json#L8

services/hackbot-ui/app/api/runs/[runId]/artifacts/[...path]/route.ts
Comment (evgenyrp) [reply services/hackbot-ui/app/api/runs/[runId]/artifacts/[...path]/route.ts]: This requires setting a CORS policy on the results bucket

Comment (evgenyrp) [reply services/hackbot-ui/app/api/runs/[runId]/artifacts/[...path]/route.ts]: Either way works for me. All this code is expendable.

services/hackbot-ui/components/PatchView.tsx
evgenyrp: Good point, using diff2html now

services/hackbot-ui/app/api/runs/[runId]/artifacts/[...path]/route.ts
Comment (suhaibmujahid) [reply services/hackbot-ui/app/api/runs/[runId]/artifacts/[...path]/route.ts]: I added https://hackbot.moz.tools to the buket's CORS configs

Comment (evgenyrp) [reply services/hackbot-ui/app/api/runs/[runId]/artifacts/[...path]/route.ts]: Reworked that. Usually, it's a good idea to add localhost there too for testing. Like http://localhost:3000

services/hackbot-ui/components/PatchView.tsx
```diff
@@ -0,0 +1,109 @@
...
+            <p className="muted">
+              Only the first part of the patch is shown; download the artifact
+              below for the rest.
+            </p>
+          )}
+          {/* diff2html only emits markup for the diff it was given and escapes
+              its content, so this does not render arbitrary agent HTML. */}
+          <div
+            className="patch-diff"
+            dangerouslySetInnerHTML={{
```
suhaibmujahid, line 43: Some other alternative that support React would work nicer here than diff2html (like https://github.com/MrWangJustToDo/git-diff-view or https://github.com/otakustay/react-diff-view), but we could file a follow up issue to investigate that and move on with this PR for now.

services/hackbot-ui/app/globals.css
```diff
@@ -394,23 +396,6 @@ dl.kv dd {
...
-  justify-content: space-between;
-  padding: 8px 0;
-  border-bottom: 1px solid var(--border);
-  font-family: var(--mono);
-  font-size: 13px;
-}
-.artifact-list li:last-child {
-  border-bottom: none;
-}
-
```
suhaibmujahid, line 413: Why are we removing these here?

services/hackbot-ui/app/globals.css
```diff
@@ -1,3 +1,5 @@
+@import "diff2html/bundles/css/diff2html.min.css";
```
suhaibmujahid: It might be better to do `import "diff2html/bundles/css/diff2html.min.css";` in `layout.tsx`.

suhaibmujahid (approved):
Thank you! Thins mask things nicer!

services/hackbot-ui/app/globals.css
evgenyrp: Good catch! I thought it was something unused, but it's just AI went rogue, likely due to approaching context size

services/hackbot-ui/components/PatchView.tsx
evgenyrp: https://github.com/mozilla/bugbug/issues/6868

PR #6860, opened 2026-09-15
Timestamp every agent log record

evgenyrp:
This helps with investigating latency based on agent traces:

```
############################################################
# /css/css-grid/alignment: analysis
# started 23:09:28.821 UTC
############################################################
23:09:29.769 [system] session started (model=claude-opus-5)

23:09:31.406 --- turn 1 ---
23:09:31.406 [agent] I'll start by querying Treeherder for the actual failure details.

23:09:34.912 --- turn 2 ---
23:09:34.912 [agent→tool] Bash
{ ... }
23:09:37.260 [tool←ok]
Metadata saved to: ...

...

============================================================
23:14:02.368 [done] turns=11 cost=$0.3592 elapsed=43s

```


libs/hackbot-runtime/tests/test_claude.py
```diff
@@ -29,6 +31,54 @@ def test_header_always_prints_even_when_not_verbose(capsys):
     assert "# bug 999" in out
 
 
+def test_log_records_are_timestamped_and_done_carries_elapsed(tmp_path):
```
suhaibmujahid: nit: I would drop the test before landing.

libs/hackbot-runtime/hackbot_runtime/claude.py
```diff
@@ -29,13 +30,22 @@ def _truncate(s: str, n: int = 500) -> str:
     return s if len(s) <= n else s[:n] + f"... [{len(s) - n} more chars]"
 
 
+def _stamp() -> str:
+    """UTC wall clock, HH:MM:SS.mmm, prefixed to every log-file record."""
+    now = time.time()
+    return (
+        time.strftime("%H:%M:%S", time.gmtime(now)) + f".{int(now * 1000) % 1000:03d}"
```
suhaibmujahid: nit: For readability, I would assign `time.strftime("%H:%M:%S", time.gmtime(now))` to a variable and then use it instead of the f-string template.

suhaibmujahid: Do we really need the milliseconds?

suhaibmujahid (approved):

libs/hackbot-runtime/hackbot_runtime/claude.py
evgenyrp: why not? It seems to be a standard practice on GCP for example: `2026-09-15 14:05:55.511 PDT`

PR #6861, opened 2026-09-16
Add run dedupe key and idempotent create flow

### LINKED ISSUE #6279: Hackbot job deduplication
Sometimes external triggers can duplicate Hackbot jobs. For example, when 20 builds failed for the same revision. We need a way to deduplicate them based on an agent-specific set of inputs and currently running jobs.

suhaibmujahid:
Resolves #6279


services/hackbot-api/app/routers/runs.py
```diff
@@ -44,6 +46,20 @@ def _normalize_identity(email: str | None) -> str | None:
...
 
+DedupeKey = (
+    Annotated[
+        str,
+        StringConstraints(
+            strip_whitespace=True,
+            min_length=1,
+            max_length=200,
+            to_lower=True,
+            ascii_only=True,
```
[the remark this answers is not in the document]
suhaibmujahid: Fixed in 748ddf66ad221e7a34cb0bbc9d77d2f19e757ad7

services/hackbot-api/app/routers/runs.py
```diff
@@ -89,33 +116,57 @@ async def create_run(
...
+            run.dedupe_key,
+            run.run_id,
+            run.status,
+        )
+
+        return Response(
+            content=RunRef.model_validate(run).model_dump_json(),
+            media_type="application/json",
+            status_code=status.HTTP_200_OK,
+        )
```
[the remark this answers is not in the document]
suhaibmujahid: Fixed in 8bef4a48b0216aaf703f95a19795ac693835ae6d

services/hackbot-api/app/routers/runs.py
```diff
@@ -89,33 +129,58 @@ async def create_run(
...
-        "RUN_ID": str(run_id),
-        "RESULTS_BUCKET": settings.results_bucket,
-        "RESULTS_PREFIX": results_prefix,
-        "RESULTS_POLICY_URL": policy["url"],
-        "RESULTS_POLICY_FIELDS": json.dumps(policy["fields"]),
-        **(agent.build_env or model_to_env)(inputs),
-    }
+    try:
+        await db.commit()
+    except IntegrityError:
```
evgenyrp: Are we sure this happens because of the dedupe key always when it's not none?

services/hackbot-api/app/routers/runs.py
```diff
@@ -89,33 +129,58 @@ async def create_run(
...
+        if dedupe_key is None:
+            # Without a key there is no duplicate to read this as, so it is a
+            # real constraint fault.
+            raise
+
+        await db.rollback()
```
evgenyrp: I think there's a way to do a conditional insert in Postgres instead of doing commit/rollback. It's worth asking AI whether it would be safer.

services/hackbot-api/app/routers/runs.py
suhaibmujahid: If it is not, it will fail at line 156, where the trace will show both errors.

services/hackbot-api/app/routers/runs.py
suhaibmujahid: Yes, we could use `ON CONFLICT DO NOTHING`, but that means not utilizing the ORM for that transaction and braking the cross DB compatibility part. I would prefer this since it feels cleaner.

suhaibmujahid: I did that anyway in 483d57ec570427051895c126f80fe718364ff47f, we can revert it if you think the ORM approach is still better.

evgenyrp: I see your point, but the insert seems to be a bit safer to me compared to the rollback, so I'd keep it.

evgenyrp (approved):

PR #6864, opened 2026-09-16
Create default mozconfig before starting the agent

### LINKED ISSUE #6863: Bug-fix agent lacks the mozconfig required to build Firefox
The agent configures Firefox paths but does not create `.mozconfig`, causing `build_firefox()` to fail when no configuration already exists.

ayoubdiourin7:
fixes #6863


suhaibmujahid (approved):

agents/bug-fix/hackbot_agents/bug_fix/agent.py
```diff
@@ -122,6 +123,18 @@ def make_investigator() -> AgentDefinition:
     )
 
 
+def _write_mozconfig(fx_ctx: FirefoxContext) -> None:
+    """Write a normal optimized browser config, preserving any existing config."""
+    if fx_ctx.mozconfig.exists():
+        return
+    fx_ctx.mozconfig.write_text(
+        "ac_add_options --enable-application=browser\n"
```
padenot, line 136: this is the default, but doesn't hurt

agents/bug-fix/hackbot_agents/bug_fix/agent.py
```diff
@@ -122,6 +123,18 @@ def make_investigator() -> AgentDefinition:
...
+        "ac_add_options --disable-debug\n"
```
padenot: we should do `--enable-debug`, it's really important and will catch a bunch of stuff during development.

`--enable-clang-plugin` is super nice because it will turn on a _lot_ of very specialized, mozilla-specific static analysis. It make the tree compile a tad more slowly (which won't matter for long). My experience is that LLMs will write things that this will catch.

agents/bug-fix/hackbot_agents/bug_fix/agent.py
```diff
@@ -122,6 +123,18 @@ def make_investigator() -> AgentDefinition:
...
+        "ac_add_options --enable-optimize\n"
```
padenot, line 138: this is the default

agents/bug-fix/hackbot_agents/bug_fix/agent.py
```diff
@@ -122,6 +123,18 @@ def make_investigator() -> AgentDefinition:
...
+        f"mk_add_options MOZ_OBJDIR={quote(str(fx_ctx.objdir))}\n"
```
padenot: this isn't useful, there's a good default

padenot (approved):
Some recommendations to maximize the usefulness here. debug + opt is what most developers do.

agents/bug-fix/hackbot_agents/bug_fix/agent.py
ayoubdiourin7: Updated in  [cfd0286](https://github.com/mozilla/bugbug/pull/6864/commits/cfd02862d330e01827394764a86550f7c150d963)

Issue #6724, opened 2026-08-27
Hackbot API should keep a record of the service that triggered a run and who this service is acting on behalf


suhaibmujahid:
When a service triggers a run, we should keep record of both, the service and the actual user reported from that service.

PR #6869, opened 2026-09-17
Email the requester when a UI-launched run finishes

### LINKED ISSUE #6866: Allow Hackbot UI users to opt-in to receiving an email when the job has completed
These days, if I invoke Hackbot via the UI, I keep the tabs open in a group somewhere, and will periodically check in on them to see how they're doing. What'd be lovely is if I could just close the tab, and receive an email (and/or other notification) when the job completes so that I can then re-open the job to inspect what occurred.

ayoubdiourin7:
Fixes #6866


services/hackbot-api/app/routers/runs.py
```diff
@@ -363,6 +363,7 @@ async def finalize_run(db: AsyncSession, run: Run) -> None:
             run.agent,
         )
     await pubsub.publish_run_completed(str(run.run_id), run.agent, run.status)
+    await notifications.notify_requester(run)
```
suhaibmujahid: This should be in the internal API that receives the "run completed" events.

services/hackbot-api/app/config.py
```diff
@@ -104,6 +104,13 @@ class Settings(BaseSettings):
     push_auth_audience: str = ""
     push_auth_service_account: str = ""
 
+    # Run-completion email to the requester (see app/notifications.py). The
+    # override replaces the recipient so a dev deployment never mails real people.
+    ui_base_url: str = "http://localhost:3000"
+    sendgrid_api_key: str = ""
+    notification_sender: str = ""
+    notification_override_email: str = ""
```
suhaibmujahid, line 112: Why we need `notification_override_email`?

services/hackbot-api/app/notifications.py
```diff
@@ -0,0 +1,92 @@
...
+outcome and a link to the run page, so nobody has to keep a tab open. Runs
+with no requester (triggered by webhooks) are skipped.
+
+This module only composes and delivers the message. It keeps no state, so
+calling it once per run is the caller's job. Delivery is best-effort: a
+failed send is logged, never raised. Only the requester is addressed; this
+is a personal ping, not a report.
+"""
+
+from __future__ import annotations
```
suhaibmujahid: We do not need this, We are in the future :p

services/hackbot-api/app/notifications.py
```diff
@@ -0,0 +1,92 @@
...
+    outcome = run.status.replace("_", " ")
+    subject = f"[Hackbot] {run.agent} run {outcome}"
+
+    body = "\n".join(
+        [
+            f"Your **{run.agent}** run has **{outcome}**.",
+            "",
+            f"Open the run: {run_url(str(run.run_id))}",
+        ]
+    )
```
suhaibmujahid: We could use an template in a separate file for this.

services/hackbot-api/app/notifications.py
```diff
@@ -0,0 +1,92 @@
...
+_SEND_TIMEOUT_SECONDS = 10
+
+
+def run_url(run_id: str) -> str:
+    return f"{settings.ui_base_url.rstrip('/')}/runs/{run_id}"
+
+
+def build_message(run: Run) -> tuple[str, str]:
+    """Compose the subject and Markdown body of the notice for ``run``."""
+    outcome = run.status.replace("_", " ")
```
suhaibmujahid, line 40: I would mention the first part of the run ID, similar to the UI. If there is a bug id, I would mention it here.

services/hackbot-api/app/notifications.py
```diff
@@ -0,0 +1,92 @@
...
+    return settings.notification_override_email.strip() or run.requested_by
+
+
+def _send_sync(recipient: str, subject: str, body_md: str) -> int:
+    message = Mail(
+        From(settings.notification_sender),
+        To(recipient),
+        Subject(subject),
+        Content("text/plain", body_md),
+        HtmlContent(markdown2.markdown(body_md, extras=["fenced-code-blocks"])),
```
suhaibmujahid: What if the template is HTML in the first place? That would give us more flexibility in the design.

services/hackbot-api/app/notifications.py
```diff
@@ -0,0 +1,92 @@
...
+
+    No requester is a quiet no-op; a delivery failure is logged and swallowed.
+    """
+    if not run.requested_by:
+        return False
+
+    recipient = _recipient(run)
+    subject, body = build_message(run)
+    try:
+        status_code = await asyncio.to_thread(_send_sync, recipient, subject, body)
```
suhaibmujahid: Can we have a native way to send async?

services/hackbot-api/app/notifications.py
```diff
@@ -0,0 +1,92 @@
...
+    return response.status_code
+
+
+async def notify_requester(run: Run) -> bool:
+    """Mail the run's requester about its terminal state. Returns whether it sent.
+
+    No requester is a quiet no-op; a delivery failure is logged and swallowed.
+    """
+    if not run.requested_by:
+        return False
```
suhaibmujahid: We should file an issue to have a proper why to identify it was triggered by the UI (that issue would be blocked by #6724). That issue should be referenced here.

suhaibmujahid: It would be better to have this condition out of `notify_requester()`.

suhaibmujahid (changes requested):

services/hackbot-api/app/config.py
```diff
@@ -104,6 +104,13 @@ class Settings(BaseSettings):
     push_auth_audience: str = ""
     push_auth_service_account: str = ""
 
+    # Run-completion email to the requester (see app/notifications.py). The
+    # override replaces the recipient so a dev deployment never mails real people.
+    ui_base_url: str = "http://localhost:3000"
```
suhaibmujahid: We have `hackbot_api_url`, thous using `hackbot_ui_url` would be a better fit for consistency.

services/hackbot-api/app/config.py
ayoubdiourin7: It’s for testing. It redirects emails to a test inbox so we can verify that notifications work correctly.

services/hackbot-api/app/notifications.py
ayoubdiourin7: I checked, and SendGrid’s Python SDK doesn’t support native async sending.

services/hackbot-api/app/config.py
ayoubdiourin7: update in https://github.com/mozilla/bugbug/pull/6869/commits/2a05123628b1bf792d3434893f90b15803471421

services/hackbot-api/app/notifications.py
ayoubdiourin7: [4b03f3a](https://github.com/mozilla/bugbug/pull/6869/commits/4b03f3ad90b91b1477ec65a5d5079eddb0d27dbb)

services/hackbot-api/app/notifications.py
ayoubdiourin7: updated in[839d796](https://github.com/mozilla/bugbug/pull/6869/commits/839d7966caf6bcfc92245ef92916b4b6805c11a3)

services/hackbot-api/app/routers/runs.py
ayoubdiourin7: [87a177b](https://github.com/mozilla/bugbug/pull/6869/commits/87a177bd48d7a60fc0913af2c1a48e76314d507a)

suhaibmujahid: Having it with a separate subscriber and new endpoint is an overkill, let’s put it under where we call `finalize_run()` instead of inside `finalize_run()`.

services/hackbot-api/app/config.py
suhaibmujahid: Let we have a clearer name for that (e.g., `override_recipient_email`), and a clarifying comment here.

services/hackbot-api/app/config.py
```diff
@@ -106,6 +106,13 @@ class Settings(BaseSettings):
     push_auth_audience: str = ""
     push_auth_service_account: str = ""
 
+    # Run-completion email to the requester (see app/notifications.py). The
+    # override replaces the recipient so a dev deployment never mails real people.
+    hackbot_ui_url: str = "http://localhost:3000"
+    sendgrid_api_key: str = ""
+    notification_sender: str = ""
+    notification_override_email: str = ""
```
suhaibmujahid, line 114: If it is optional, the default should be `None`, not `""`. I think we have an issue about dropping `""` from settings.

services/hackbot-api/app/routers/runs.py
ayoubdiourin7: updated in [f6e569b](https://github.com/mozilla/bugbug/pull/6869/commits/f6e569b1dee1f4a65060a6c6e93f06a845139cc7) , is it good now ?

Issue #6865, opened 2026-09-16
Resolve uplift merge conflicts using a hackbot agent


cgsheeh:
See https://bugzilla.mozilla.org/show_bug.cgi?id=2051452 for more detail.

Initially we will add an agent to resolve merge conflicts for uplift jobs in Lando. In the future we will add agents to audit uplift stacks and perform other tasks in the workflow.

PR #6876, opened 2026-09-18
Bug 2051452: Add the uplift merge-conflict resolver agent

cgsheeh:
Adds the first uplift sub-agent: it reproduces a failed uplift cherry-pick on a
stable branch, resolves the merge conflicts, and returns the resulting patch
with a confidence level for review. Sources are git commits and/or Phabricator
revisions applied in order, so a stack is just several sources, and a caller can
pin both the diff to uplift and the commit to uplift onto. It never pushes or
lands anything, auto-applies no actions, and does not build Firefox — resolving
a conflict is a source-level judgment and a reviewer compiles the result.

`resolved` is not taken on trust. The agent grades its own work, so after the
session the run checks the checkout — no conflicted paths, no cherry-pick left
open, nothing uncommitted, and commits that actually collect to a patch — and
overrules the claim if they disagree, publishing the verified report alongside
what the agent wrote. Confidence stays the agent's own judgment. The Lando side
of this work (the uplift checkbox, waiting on a run, retrieving the result,
Phabricator submission and human approval) is not in this PR.

Issue: https://github.com/mozilla/bugbug/issues/6865


agents/uplift/hackbot_agents/uplift_merge_conflict_resolver/config.py
evgenyrp, line 1: This file doesn't look like a config to me :) I would expect just some constants here, maybe within a class.

agents/uplift/hackbot_agents/uplift_merge_conflict_resolver/prompts/system.md
evgenyrp, line 1: why do we need a system prompt at all?

agents/uplift/hackbot_agents/uplift_merge_conflict_resolver/prompts/system.md
```diff
@@ -0,0 +1,109 @@
...
+- `high`: conflicts were mechanical (imports, adjacent edits, context shifts)
+  and the resolution is unambiguous.
+- `medium`: you had to make a judgment call, but the bug/revision context
+  supports it.
+- `low`: the branch has diverged enough that you are unsure, or you could not
+  fully resolve a hunk. A human must look closely.
+
+## When you are done
+
+Leave all resolved patches committed in the working tree (the platform collects
```
evgenyrp, line 82: Per our latest discussion with @suhaibmujahid, it shouldn't commit anything (we don't even set git identity). The edits will be collected automatically in `changes.patch` (see build repair agent). Maybe it's different in other agents, not sure.

agents/uplift/README.md
```diff
@@ -0,0 +1,67 @@
+# Uplift Agents
+
+Home for the backport/uplift sub-agents that help land patches on Firefox's
```
evgenyrp, line 3: Technically, this is not a sub-agent. I'm not even sure we can change the name from `hackbot_agents` to `uplift`; I thought it was just a convention. See `agents/README.md`

agents/uplift/hackbot_agents/uplift_merge_conflict_resolver/prompts/system.md
```diff
@@ -0,0 +1,109 @@
...
+the final patch as an artifact.
+
+## How to apply each patch
+
+Apply the sources one at a time, in the order given. The checkout is shallow, so
+commits you need are not present locally until you fetch them.
+
+For a **Git** source (a commit SHA):
+
+1. Fetch the commit and its parent so cherry-pick can three-way merge:
```
evgenyrp, line 14: This is probably what you need here, but we prefetch 500 commits for the test-repair agent, for example. So if you're sure that the ones you need are easy to prefetch, it's another option.

services/hackbot-api/app/schemas.py
```diff
@@ -175,3 +175,72 @@ class TestPlanGeneratorInputs(BaseModel):
...
+    )
+
+
+# `kind` discriminates the two, so one run can mix both.
+UpliftSource = Annotated[
+    Union[GitUpliftSource, PhabricatorUpliftSource], Field(discriminator="kind")
+]
+
+
+class UpliftInputs(BaseModel):
```
evgenyrp, line 217: Do you really need such a complicated input? If things can be resolved at runtime, it's better to do that, in my experience, because later you might also want to reproduce the runs locally, run evals for the agent, implement deduplication, and many of those are easier with a simple input. I ended up using just the Taskcluster task ID for CI agents, and everything else is resolved from there.

agents/uplift/hackbot_agents/uplift_merge_conflict_resolver/prompts/system.md
cgsheeh: My understanding was that the system prompt contains the details of the agent that are constant across runs (which can be re-used by prompt caching), and the user prompt contains the per-run details. Sort of like a function, the system prompt is the implementation itself, and the user prompt is the arguments passed to the function. If my understanding is wrong or there are reasons we should consider a different split, I'm happy learn why and use only a user prompt. :)

agents/uplift/hackbot_agents/uplift_merge_conflict_resolver/prompts/system.md
cgsheeh: I think for this use case we will need to make a commit. We're not just creating a single patch, we're reconstructing a stack and resolving conflicts for the entire stack. For example, we may have a stack like `D1->D2->D3->D4` that we want to uplift to beta, but D2 has a conflict. We need to apply D2 and resolve the conflict, but then also apply D3 and D4 and resolve any conflicts there as well. Hackbot then needs to post each of the results as a separate patch so we can re-create the revisions in Lando. Without committing, `D2->D3->D4` are squashed into a single patch, which is not ideal from a code archaeology perspective.

I didn't realize we don't set a Git identity, I'll need to work around that or explicitly set one for the uplift agent. The author information isn't important from Lando's perspective (we'll set the original author when re-creating the patch anyways) but keeping each commit separate from others is ideal here.

agents/uplift/hackbot_agents/uplift_merge_conflict_resolver/models.py
cgsheeh: Agreed, I'll rename this to `models.py` instead. :)

agents/uplift/README.md
cgsheeh: That's fair, I'll change the "sub-agent" wording to just "agent".

agents/uplift/hackbot_agents/uplift_merge_conflict_resolver/prompts/system.md
cgsheeh: Most of the time the commits will be relatively fresh on autoland, however we need to support uplifting any existing commit. I think sticking with an explicit fetch, at least for the initial implementation, will give us the most flexibility.

services/hackbot-api/app/schemas.py
cgsheeh: Good question — let me split the fields, because I think two different things are bundled here.

`model`, `max_turns` and `effort` aren't uplift-specific: `BugFixInputs`, `BuildRepairInputs`, `TestRepairInputs` and `FrontendTriageInputs` all carry them, so dropping them here would make this agent the odd one out. They're all optional and exist so a run stays reproducible if we change a default later — pin them explicitly and you get the old behaviour back. `target_commit` is the same idea for the checkout: branch names move, so reproducing a specific failed uplift means pinning the commit it was attempted on. None of them are required, so the minimal request is just `target_branch` plus `sources`.

Those two I don't think we can resolve at runtime. There's no single upstream ID to derive them from the way CI agents derive everything from a Taskcluster task ID — an uplift request names the branch it's going to (release/beta/esrN, and a patch can target several), and the patches themselves come in two shapes: commits already on autoland, which we cherry-pick, and Phabricator revisions that haven't landed yet, which we apply as diffs. Lando already supports both, so Hackbot has to accept both to be usable from it.

That said, I'd like to keep the input as simple as it can be, and I take the point about evals and dedup. Is there a shape you had in mind that still covers multiple target branches and both source kinds? Happy to change it if so.

agents/uplift/compose.yml
```diff
@@ -0,0 +1,60 @@
...
+
+  uplift-merge-conflict-resolver:
+    build:
+      context: ../..
+      dockerfile: agents/uplift/Dockerfile
+      target: agent
+    environment:
+      - RUN_ID
+      - TARGET_BRANCH=${TARGET_BRANCH:-}
+      - SOURCES=${SOURCES:-}
```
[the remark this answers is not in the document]
cgsheeh: Good catch — fixed. The service now forwards `TARGET_COMMIT=${TARGET_COMMIT:-}` alongside `TARGET_BRANCH`, so a local run pins the commit the same way a deployed one does.

agents/uplift/hackbot_agents/uplift_merge_conflict_resolver/models.py
```diff
@@ -0,0 +1,240 @@
...
+    """The agent's ``report.json``, in the shape the system prompt specifies.
+
+    Every field defaults, so a partial report still parses. Validation is what
+    catches a plausible but wrong value: `"false"` is a string `bool()` reads
+    as `True`, and a caller gates on `confidence`.
+    """
+
+    model_config = ConfigDict(extra="ignore")
+
+    resolved: bool = False
```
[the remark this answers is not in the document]
cgsheeh: I'd rather leave this one. Pydantic maps `"true"` to `True` and `"false"` to `False`, so a string-valued `resolved` still means what it says — the docstring's worry is about plain `bool("false")`, not about validation. `resolved` isn't trusted either way: `verify_uplift` checks the checkout and forces it to `False` when the two disagree, and that forced value is what `report.json` publishes.

agents/uplift/hackbot_agents/uplift_merge_conflict_resolver/verify.py
```diff
@@ -0,0 +1,80 @@
...
+    ):
+        if (git_dir / marker).exists():
+            problems.append(f"{operation} left unfinished in the checkout")
+            break
+
+    # Uncommitted work is swept into a container-authored commit, losing the
+    # original author.
+    dirty = git_output(repo, "status", "--porcelain")
+    if dirty:
+        problems.append(f"{len(dirty.splitlines())} path(s) left uncommitted")
```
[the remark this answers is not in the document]
cgsheeh: Not needed here, and the prompt no longer asks for `--author` at all. Lando re-attributes each patch when it re-creates the revisions from the mbox, so who the container committed as is not what lands — I've dropped that instruction and the `render_author` machinery behind it.

The dirtiness check stays, but for a different reason: uncommitted work gets swept into one synthetic commit, which collapses the stack into a single patch.

agents/uplift/hackbot_agents/uplift_merge_conflict_resolver/verify.py
```diff
@@ -0,0 +1,80 @@
...
+    if report.resolved:
+        if git_output(repo, "merge-base", base_commit, "HEAD") != base_commit:
+            problems.append(
+                f"HEAD no longer descends from the commit the run started at "
+                f"({base_commit[:12]})"
+            )
+        elif not git_output(repo, "rev-list", f"{base_commit}..HEAD"):
+            problems.append("report claims resolved but no commits were produced")
+        elif not git_output(repo, "diff", "--name-only", base_commit, "HEAD"):
+            problems.append("report claims resolved but the commits change nothing")
```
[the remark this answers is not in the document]
cgsheeh: I'm leaving this as-is. `git format-patch`, which `hackbot_runtime.changes.collect` uses to build the mbox, skips empty commits, so an empty commit for a skipped source never reaches the artifact — the aggregate check plus the round-trip test covers what can actually go wrong. Happy to add a per-commit check anyway if you'd rather have the belt and braces.

services/hackbot-api/app/schemas.py
```diff
@@ -175,3 +175,72 @@ class TestPlanGeneratorInputs(BaseModel):
...
+
+
+class GitUpliftSource(BaseModel):
+    """A patch to uplift, identified by a commit already in the Firefox repo."""
+
+    # Discriminator tag selecting this variant in the `UpliftSource` union.
+    kind: Literal["git"] = "git"
+
+    # Full git commit SHA the agent fetches and cherry-picks onto the branch.
+    commit: str = Field(description="Full git commit SHA to cherry-pick.")
```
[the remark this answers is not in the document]
cgsheeh: Fixed. `commit` and `target_commit` now carry `pattern=r"^[0-9a-f]{40}$"`, mirrored on the agent's own `AgentInputs` and `GitSource`, so an abbreviation is rejected at the request boundary instead of failing at the `git fetch` partway into a run. The UI validates the same shape, so it surfaces next to the field rather than as a 422.

agents/uplift/hackbot_agents/uplift_merge_conflict_resolver/prompts/system.md
```diff
@@ -0,0 +1,109 @@
...
+1. Its diff has already been fetched for you, at the path your task lists for
+   it. Do not try to download it yourself.
+2. Fetch the base commit listed for it: `git fetch --depth=1 origin <base>`.
+   Do not skip this. The diff names the blobs it was built from, and until that
+   commit is local they are missing, so `git apply --3way` reports "repository
+   lacks the necessary blob", silently falls back to a direct apply, and fails
+   leaving you no conflict markers at all.
+3. Apply it with a three-way merge so conflicts surface as markers:
+   `git apply --3way <diff-file>`
+4. Resolve any conflicts and commit the fixed patch.
```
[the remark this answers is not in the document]
cgsheeh: Fixed. Step 4 of the Phabricator workflow now tells the agent to read the revision's title and summary with `get_phabricator_revision` and commit under those, ending the message with a `Differential Revision:` line, rather than writing a message of its own.

services/hackbot-ui/components/TriggerForm.tsx
```diff
@@ -284,6 +307,48 @@ export function TriggerForm() {
...
+          </div>
+
+          <div className="field">
+            <label htmlFor="upliftSources">
+              Sources * (JSON list, applied in order)
+            </label>
+            <textarea
+              id="upliftSources"
+              placeholder={
+                '[{"kind": "git", "commit": "abc123"},\n' +
```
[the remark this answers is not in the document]
cgsheeh: Fixed, and gone a little further: the placeholder shows a full 40-character SHA, and `parseUpliftSources` now rejects an abbreviation, so a copied example fails next to the field instead of reaching the API.

agents/uplift/hackbot_agents/uplift_merge_conflict_resolver/prompts/system.md
evgenyrp: @suhaibmujahid do you have an opinion here? We can set the identity in the runtime part once.

agents/uplift/hackbot_agents/uplift_merge_conflict_resolver/prompts/system.md
evgenyrp: The prefix caching will work regardless of whether it's a system or user prompt. It all gets concatenated into the final prompt. Since we have full control over this one-time session, there's no difference. I see your point, and it's fine to extract the fixed part and make it more explicit as a system prompt. It's just that, conceptually, a system prompt was not designed for information about a specific task. Up to you.

agents/uplift/hackbot_agents/uplift_merge_conflict_resolver/prompts/system.md
suhaibmujahid: Short answer: We need to commit here.

In this case, the agent does not need to submit anything to Phabricator and each commit will be representing a different patch in the stack. The agent only need to solve the merge conflicts. Later, Lando will collect the `changes.patch` file through the API, which already preserve the individual commits in mbox diff format.

agents/uplift/hackbot_agents/uplift_merge_conflict_resolver/prompts/system.md
suhaibmujahid: > The prefix caching will work regardless of whether it's a system or user prompt. 

True

> It's just that, conceptually, a system prompt was not designed for information about a specific task.

Yes, in most cases, the system prompt should be static content. Dynamic parts, if needed at all, could be related to tool instructions, safety guards, etc. Not the task on hand.


However, not configuring a system problem at all is not ideal.  Without one, it defaults to a minimal prompt that’s not even the one that Claude Code uses. I think if we are not providing one, we should be at least configure it to use the Claude Code prompt with `system_prompt={“type”: “preset”, “preset”: “claude_code”}`.

See https://code.claude.com/docs/en/agent-sdk/modifying-system-prompts

evgenyrp: Yes, we should use Claude's preset, as it's used locally. This is the real system prompt. Everything we specify is task-specific.

agents/uplift/hackbot_agents/uplift_merge_conflict_resolver/prompts/workflow.md
cgsheeh: Makes sense to me, switched to using the default Claude code system prompt.

services/hackbot-api/app/schemas.py
evgenyrp: I don't have any specific suggestions. If you think the input shape works well for you, I'm fine with that. Just wanted to share my experience.

agents/uplift/hackbot_agents/uplift_merge_conflict_resolver/__main__.py
```diff
@@ -0,0 +1,79 @@
...
+    bugbug_mcp_url: str
+
+    # The broker sidecar, whose Conduit proxy diffs are fetched through.
+    broker_url: str
+
+    # The originating bug, for context when a conflict's intent is unclear.
+    bug_id: int | None = None
+
+    # Overrides; `None` leaves the SDK's or the API's own default in place.
+    model: str | None = None
```
evgenyrp, line 42: I would set the defaults here explicitly instead of relying on the harness. Those can change behavior drastically and we typically want to retest things before updating the model or effort and redeploying the agent. Like model=Opus5.5, effort="high"

evgenyrp (approved):
Overall, LGTM. I left one comment related to fixing the model parameters, which I think is quite important to control. I'll leave the runtime review to @suhaibmujahid. He has rejected similar changes I made in the past.

PR #6879, opened 2026-09-19
Avoid overscheduling due to sibling manifest rule

marco-c:
r? @bhearsum


bhearsum (approved):
This seems like a plausible way to improve this.

How will we measure how well this is working?

marco-c:
@bhearsum I will write a script to retrieve backout data and analyze effectiveness (also comparing different strategies, thresholds, etc.).
We actually already had one, but it stopped working ages ago and could be improved/simplified.

PR #6883, opened 2026-09-20
Use react-diff-view instead of diff2html for patch preview

### LINKED ISSUE #6868: Consider using a React component instead of diff2html for patch preview
See https://github.com/mozilla/bugbug/pull/6856#discussion_r4029723955

goldpanth3r:
Fixes https://github.com/mozilla/bugbug/issues/6868
Replaced diff2html's dangerouslySetInnerHTML-based rendering in PatchView with react-diff-view, which renders the diff as real React elements instead of an injected HTML string.

   - parseDiff/Diff/Hunk replace diff2html's parse/html
   - Per-file headers shown when a patch touches more than one file
     (replacing diff2html's drawFileList option)
   - Dark styling for the diff, matching the app's existing color tokens
     (--green/--red at the same opacity used by .badge.succeeded/.failed),
     added in globals.css
     (replacing diff2html's colorScheme: ColorSchemeType.DARK)


suhaibmujahid (commented):
Could please share a screenshot to show how this looks after the migration?

goldpanth3r:
Pageview
I tried running it locally, but it first required Google OAuth, so I configured my own client ID and secret. It then redirected to Mozilla authentication, which requires an LDAP account which I don't have. I therefore tested the migration locally with a small test page and attached a screenshot of the `react-diff-view` rendering.

PR #6885, opened 2026-09-20
Deduplicate Slack button clicks

### LINKED ISSUE #6732: Avoid processing the same Slack button click multiple times
If the same user or different users clicked on the same button multiple times, it should execute only one. 

We might later want to allow different users to click on the same button for use cases like feedback, but that’s for later when there is a use case for that.

ayoubdiourin7:
Fixes #6732


suhaibmujahid:
Closed in favour of #6897

PR #6897, opened 2026-09-21
Use dedupe_key in Slack button to start at most one run

### LINKED ISSUE #6732: Avoid processing the same Slack button click multiple times

suhaibmujahid:
Resolves #6732

---


ayoubdiourin7 (approved):

services/hackbot-api/app/slack/listeners/actions/start_agent_run.py
```diff
@@ -35,12 +45,25 @@ async def start_agent_run_callback(
...
+    if not run.is_new:
+        logger.warning(
+            "Slack click by '%s' for key '%r' ignored because a run with the same key run already exists (id=%s)",
+            user,
+            value.dedupe_key,
+            run.run_id,
+        )
+        await respond(
+            text=(
+                f"A {value.agent_name} run is already underway for this "
```
ayoubdiourin7: For a finished or failed run, “already underway” is no that accurate, wdyt?

suhaibmujahid: I agree, fixed in 96b272a7769063b3a7511a94aa6cef8e2dc84dec.

Issue #6873, opened 2026-09-17
The apply pending actions can double-post when having concurrent requests


suhaibmujahid:
> [!NOTE]
> This did not happen yet, so no examples, but in theory it is a bug that needs to be fixed.

The `apply_all_pending` method reads a run's not-yet-applied rows, then dispatches and commits each one, with no lock. Two concurrent calls both see the same pending rows and both dispatch them, so the same Bugzilla comment can be posted twice.

Issue #6468, opened 2026-08-03
Process long-running webhooks on a queue and return early


suhaibmujahid:
`phabricator_webhook` in `services/hackbot-api/app/routers/webhooks.py` does detection and `trigger_run` inline. Enqueue the work and return promptly so Phabricator and other services aren't kept waiting.

PR #6898, opened 2026-09-21
Make the Slack start agent run button approve pending actions before it starts the run

### LINKED ISSUE #6892: Support automatically approve pending agent actions through Slack
In this way, when starting an agent run through Slack, button can be configured to approve pending actions.

suhaibmujahid:
Resolves #6892

---


services/hackbot-api/app/slack/listeners/actions/start_agent_run.py
```diff
@@ -45,6 +47,26 @@ async def start_agent_run_callback(
     user = (body.get("user") or {}).get("id")
     value = StartAgentRunValue.model_validate_json(action["value"])
 
+    if value.apply_run_id:
```
ayoubdiourin7, line 50: What if  two simultaneous clicks apply the same actions twice? The dedupe key here only protects `trigger_run`.

ayoubdiourin7 (approved):

services/hackbot-api/app/slack/listeners/actions/start_agent_run.py
suhaibmujahid: Based on the API design, this should be safe to re-apply, though concurrent clicks could have race conditions (a bug), which is out of the scope of this PR. It is tracked in https://github.com/mozilla/bugbug/issues/6873.

services/hackbot-api/app/slack/listeners/actions/start_agent_run.py
```diff
@@ -45,6 +47,26 @@ async def start_agent_run_callback(
...
+        actions = await client.apply_actions(value.apply_run_id)
```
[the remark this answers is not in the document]
suhaibmujahid: Fixing the slow return is out of the scope of this PR, it is tracked in https://github.com/mozilla/bugbug/issues/6468.

PR #6901, opened 2026-09-21
Make test rail client strongly typed

### LINKED ISSUE #6466: Make the TestRail client strongly typed
Type the TestRail client's responses and models the same way as the Phabricator client.

johnpangas:
Closes #6466 
This pull request refactors the TestRail client to return strongly-typed Pydantic models instead of unstructured dictionaries.


services/hackbot-api/app/action_handlers/testrail_handler.py
```diff
@@ -184,36 +154,23 @@ async def apply(self, params: dict[str, Any], ctx: ApplyContext) -> ActionResult
...
                 created_case_ids[generated_id] = case_id
 
             case_ids = list(created_case_ids.values())
-            run_id = _require_id(
-                await client.add_run(_run_payload(suite_id, case_ids, summary)),
-                "run",
-            )
+            run_id = (
+                await client.add_run(_run_payload(suite_id, case_ids, summary))
+            ).id
```
suhaibmujahid, line 173: Instead of having `run_id` as a var, we could access the id as property, `run.id`.

libs/testrail-client/testrail_client/client.py
```diff
@@ -2,11 +2,33 @@
...
+TestRailResource = TypeVar("TestRailResource", bound=BaseModel)
+
+
+def _validate_list(
+    response: dict[str, Any] | list[Any],
+    key: str,
+    model: type[TestRailResource],
+) -> list[TestRailResource]:
+    values = response.get(key, []) if isinstance(response, dict) else response
+    return TypeAdapter(list[model]).validate_python(values)
```
suhaibmujahid, line 31: This is kind of an anti-pattern, and it would make it harder to expand if we want to read things other than the list form the API response later.

I would suggest creating a model for each response type.

services/hackbot-api/app/action_handlers/testrail_handler.py
```diff
@@ -184,36 +154,23 @@ async def apply(self, params: dict[str, Any], ctx: ApplyContext) -> ActionResult
...
-                        _case_payload(
-                            test_case,
-                            case_type_id,
-                            template_id,
-                        ),
-                    ),
-                    "case",
+                case = await client.add_case(
+                    section_id,
+                    _case_payload(test_case, case_type_id, template_id),
```
suhaibmujahid, line 165: The client lib should form the `payload`, not the caller. If it is a big change to do here, you could file a follow-up issue for this to be fixed in a separate PR.

suhaibmujahid (changes requested):

PR #6902, opened 2026-09-21
Reject HTML tags in Phabricator revision summaries

### LINKED ISSUE #6418: HTTPStatusError: Client error '406 Not Acceptable' for url 'https://phabricator.services.mozilla.com/api/differential.setdiffproperty'
Sentry Issue: [HACKBOT-API-3](https://mozilla.sentry.io/issues/7639136499/?referrer=github_integration)

```
HTTPStatusError: Client error '406 Not Acceptable' for url 'https://phabricator.services.mozilla.com/api/differential.setdiffproperty'
For more information check: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/406
  File "hackbot_runtime/actions/handlers/phabricator_handler.py", line 280, in apply
    await _set_local_commits(
  File "hackbot_runtime/actions/handlers/phabricator_handler.py", line 151, in _set_local_commits
    await _conduit_request(
  File "hackbot_runtime/actions/handlers/phabricator_handler.py", line 43, in _conduit_request
    return await _client().conduit_request(method, **payload)
  File "phabricator_client/client.py", line 53, in conduit_request
    response.raise_for_status()
  File "httpx/_models.py", line 829, in raise_for_status
    raise HTTPStatusError(message, request=request, response=self)

Failed to update Phabricator revision D314383
```

ayoubdiourin7:
Phabricator returns a `406` error when a revision summary contains an HTML tag such as `<link>`. This change catches those tags before submission and asks the agent to rewrite the summary.

Fixes #6418


libs/hackbot-runtime/hackbot_runtime/actions/phabricator.py
```diff
@@ -32,10 +32,22 @@
...
 
+_HTML_TAG_RE = re.compile(r"<[A-Za-z/!?][^<>]*>")
+
+
 def _validate_summary(summary: str | None) -> None:
     if not summary:
         return
 
+    match = _HTML_TAG_RE.search(summary)
+    if match:
```
suhaibmujahid, line 43: Sometimes, mentioning the HTML tag is legitimate. Perhaps we could escape it instead of failing or removing it.  That way, it remains visible but without upsetting Phabricator.

WDYT?

ayoubdiourin7: I tested escaping the  HTML entity on the dev instance, but Phabricator displays the entities literally instead of rendering it; here is the [example](https://phabricator-dev.allizom.org/D4150#6340474). JSON escaping and Markdown formatting still trigger the `406`

suhaibmujahid: Hers is a working example: https://phabricator-dev.allizom.org/D4150#6340476

I posted it through the web UI. Though, if it try to edit it, it will fail with 406. I'm curious what the comment box does differently that we need to simulate here.

ayoubdiourin7: I tried submitting the same text from your working comment through the API, but it still returned a `406`.

PR #6904, opened 2026-09-21
Enable flake8-bugbear (B) linting rules

### LINKED ISSUE #6427: Enable flake8-bugbear (B) linting rules
https://docs.astral.sh/ruff/rules/#flake8-bugbear-b

goldpanth3r:
Fixes https://github.com/mozilla/bugbug/issues/6427

Enables the flake8-bugbear rule set ("B") in ruff's lint config and fixes
the violations it surfaces across the codebase.

## Config change

- Added "B" to ruff's `select` list.
- Added `[tool.ruff.lint.flake8-bugbear] extend-immutable-calls` for
  `fastapi.Depends` and `fastapi.Query`, since FastAPI's dependency
  injection relies on calling these in argument defaults — the intended,
  idiomatic pattern, which B008 otherwise flags as unsafe.

## Fixes

- **Mutable default arguments** (`to_ignore=set()`, `products=[]`,
  `support_files=[]`, etc.) replaced with `None` sentinels, initialized
  inside the function body — avoids the classic shared-mutable-default
  bug.
- **`assert False`** replaced with `raise AssertionError(...)` —
  `assert` statements are stripped when Python runs with `-O`, so the
  old code silently did nothing in that mode; `raise` always fires.
- **`return`/`break`/`continue` inside `finally`** removed where present
  (`setup_profile.py`), restructured to avoid the control-flow hazard
  without changing behavior.
- **Unused loop variables** prefixed with `_` (`_label`, `_is_pos`,
  `_i`, etc.) where confirmed unused in the loop body.
- and more

All changes are intended to be behavior-preserving; each fix was
reviewed individually rather than accepted as a blanket autofix.


bugbug/models/component.py
```diff
@@ -359,7 +359,7 @@ def generate_meaningful_tuples():
                     continue
 
                 if count > 0:
-                    for i in range(count):
+                    for _i in range(count):
```
suhaibmujahid: Since it is not used and we do not need to preserve what it represents, I suggest useing `_` instead:
```suggestion
                    for _ in range(count):
```

bugbug/models/defect_enhancement_task.py
```diff
@@ -51,6 +51,8 @@ def overwrite_classes(self, bugs, classes, probabilities):
 
     def get_extra_data(self):
         labels = self.le.inverse_transform([0, 1, 2])
-        labels_map = {str(label): index for label, index in zip(labels, [0, 1, 2])}
+        labels_map = {
+            str(label): index for label, index in zip(labels, [0, 1, 2], strict=False)
```
suhaibmujahid: This is a risky change where should be handled  case by case. Ideally thy should `strict=True` unless we intentionally need `strict=False`.

I suggest excluding `B905` for now and filing a follow-up issue to enable it later.

bugbug/rust_code_analysis_server.py
```diff
@@ -49,7 +49,9 @@ def start_process(self, thread_num: int | None = None):
                 cmd += ["-j", str(thread_num)]
             self.proc = subprocess.Popen(cmd)
         except FileNotFoundError:
-            raise RuntimeError("rust-code-analysis is required for code analysis")
+            raise RuntimeError(
+                "rust-code-analysis is required for code analysis"
+            ) from None
```
suhaibmujahid: For this case, I would use `from exc` instead of `from None` to preserve the original error, which would provide helpful info.

bugbug/model.py
```diff
@@ -96,7 +96,7 @@ def classification_report_imbalanced_values(
 
     result = {"targets": {}}
 
-    for i, label in enumerate(labels):
+    for i, _label in enumerate(labels):
```
suhaibmujahid: Simpler:
```suggestion
    for i in range(len(labels)):
```

bugbug/repository.py
```diff
@@ -1176,7 +1176,7 @@ def update_complex_experiences(
                 for i in range(len(items)):
                     exp_queues[i][day] = all_commit_lists[i] + (commit.node,)
 
-    for i, commit in enumerate(tqdm(commits)):
+    for _i, commit in enumerate(tqdm(commits)):
```
suhaibmujahid: Simpler:
```suggestion
    for commit in tqdm(commits):
```

bugbug/tools/core/platforms/phabricator.py
```diff
@@ -528,7 +528,7 @@ async def get_base_commit_hash(self) -> str:
         )
         pushes = r.json()["pushes"]
         closest_push = None
-        for push_id, push in pushes.items():
+        for _push_id, push in pushes.items():
```
suhaibmujahid: Simpler:
```suggestion
        for push in pushes.values():
```

bugbug/swarm.py
```diff
@@ -78,8 +78,10 @@ def get(
     AUTH,
     rev_ids: Collection[int] | None = None,
     modified_start: datetime | None = None,
-    version_l=[0, 1],
+    version_l=None,
```
suhaibmujahid: We could use tuple instead:

```suggestion
    version_l=(0, 1),
```

This apply to slimier cases as well.

scripts/bug_retriever.py
```diff
@@ -147,7 +147,7 @@ def retrieve_bugs(self, limit: int | None = None) -> None:
         new_bugs = bugzilla.download_bugs(all_ids)
 
         # Get regression_related_ids again (the set could have changed after downloading new bugs).
-        for i in range(7):
+        for _i in range(7):
```
suhaibmujahid:
```suggestion
        for _ in range(7):
```

scripts/bug_retriever.py
```diff
@@ -177,7 +177,7 @@ def retrieve_bugs(self, limit: int | None = None) -> None:
             include_invalid=True,
             include_additional_products=bugzilla.ADDITIONAL_PRODUCTS,
         )
-        for i in range(2):
+        for _i in range(2):
```
suhaibmujahid:
```suggestion
        for _ in range(2):
```

scripts/commit_classifier.py
```diff
@@ -252,12 +252,12 @@ def update_commit_db(self):
 
         assert db.download(repository.COMMITS_DB, support_files_too=True)
 
-        for commit in repository.get_commits():
+        for _commit in repository.get_commits():
             pass
```
suhaibmujahid: This is a bit confusing, we use `commit` but outside the loop, so using `_commit` is not accurate. 

What we need here is the last item in the iterator. We we could use `collections.deque()` for that:


```suggestion
        latest_commit = deque(repository.get_commits(), maxlen=1).pop()
```

scripts/generate_landings_risk_report.py
```diff
@@ -141,12 +141,12 @@ def __init__(self, repo_dir: str) -> None:
         assert db.download(repository.COMMITS_DB, support_files_too=True)
 
         logger.info("Updating commits DB...")
-        for commit in repository.get_commits():
+        for _commit in repository.get_commits():
             pass
```
suhaibmujahid: Same as https://github.com/mozilla/bugbug/pull/6904/changes#r4066951055

scripts/testing_policy_stats.py
```diff
@@ -30,12 +30,12 @@ def __init__(self, repo_dir: str) -> None:
         assert db.download(repository.COMMITS_DB, support_files_too=True)
 
         logger.info("Updating commits DB...")
-        for commit in repository.get_commits():
+        for _commit in repository.get_commits():
             pass
```
suhaibmujahid: Same as https://github.com/mozilla/bugbug/pull/6904/changes#r4066951055

tests/test_phabricator_trusted_filtering.py
```diff
@@ -792,7 +792,7 @@ def test_get_users_info_batch_mixed_trust():
     users_info = _get_users_info_batch(set(phids_by_username.values()))
 
     # Verify structure
-    for phid, info in users_info.items():
+    for _phid, info in users_info.items():
```
suhaibmujahid:
```suggestion
    for info in users_info.values():
```

pyproject.toml
```diff
@@ -157,5 +157,8 @@ convention = "google"
 [tool.ruff.lint.isort]
 known-first-party = ["bugbug_http"]
 
+[tool.ruff.lint.flake8-bugbear]
+extend-immutable-calls = ["fastapi.Depends", "fastapi.Query"]
```
suhaibmujahid: Why do we need this?

suhaibmujahid (changes requested):
Thank you for the nice work. I have some suggestion to make it better.

pyproject.toml
goldpanth3r: Removed now. It was added because `B008` false flags FastAPI's normal, recommended way of writing routes. So suppressed the errors following Ruff's documentation https://docs.astral.sh/ruff/settings/#lint_flake8-bugbear_extend-immutable-calls. 

Instead Now, excluded `B008`. Should I file a follow-up issue to enable it later?

bugbug/models/defect_enhancement_task.py
goldpanth3r: Noted! Reverted the `strict=False` additions and excluded `B905`. Filed #6916.

goldpanth3r:
Thanks for your support! I've made the changes you suggested — let me know if there's anything else to adjust.

This is my first contribution to mozilla/bugbug, and I'm looking forward to contributing more.

bugbug/utils.py
```diff
@@ -80,7 +80,7 @@ def _hstack(self, Xs, n_samples):
 
         transformer_names = (name for name, transformer, column in self.transformers_)
         types = []
-        for i, (f, transformer_name) in enumerate(zip(Xs, transformer_names)):
+        for _i, (f, transformer_name) in enumerate(zip(Xs, transformer_names)):
```
suhaibmujahid: Here is a another case where we do not need `enumerate()` since we don not use it:

```suggestion
        for f, transformer_name in zip(Xs, transformer_names):
```

pyproject.toml
suhaibmujahid: Yes, please file an issue.

docs/hackbot/agents.md
```diff
@@ -10,13 +10,14 @@ An agent is a Python package started as `python -m hackbot_agents.<name>`. It ow
...
     bug_id: int
 
-async def main(ctx: HackbotContext) -> BugFixResult:
-    ...
 
-run_async(main)                          # runtime takes over: config, auth, summary, exit
+async def main(ctx: HackbotContext) -> BugFixResult: ...
+
+
+run_async(main)  # runtime takes over: config, auth, summary, exit
```
suhaibmujahid: This change seems unrelated to this PR. If the linter is complaining about it, it is okay to keep it. Otherwise, it is better to drop this change.

docs/hackbot/tools.md
```diff
@@ -10,7 +10,9 @@ A tool is an `async` handler whose **first parameter is a context object**:
 
 ```python
 @tool
-async def get_bugs(ctx: BugzillaContext, bug_ids: Annotated[list[int], Field(description=...)]) -> dict:
+async def get_bugs(
+    ctx: BugzillaContext, bug_ids: Annotated[list[int], Field(description=...)]
+) -> dict:
```
suhaibmujahid: Same as https://github.com/mozilla/bugbug/pull/6904/changes#r4071970822

scripts/commit_classifier.py
```diff
@@ -252,12 +253,11 @@ def update_commit_db(self):
...
         assert db.download(repository.COMMITS_DB, support_files_too=True)
 
-        for commit in repository.get_commits():
-            pass
+        latest_commit = deque(repository.get_commits(), maxlen=1).pop()
 
         repository.download_commits(
             self.repo_dir,
-            rev_start="children({})".format(commit["node"]),
+            rev_start="children({})".format(latest_commit),
```
suhaibmujahid: It seems that `["node"]` was dropped here by mistake.
```suggestion
            rev_start="children({})".format(latest_commit["node"]),
```

scripts/generate_landings_risk_report.py
```diff
@@ -141,12 +141,12 @@ def __init__(self, repo_dir: str) -> None:
...
         logger.info("Updating commits DB...")
-        for commit in repository.get_commits():
-            pass
+
+        latest_commit = collections.deque(repository.get_commits(), maxlen=1).pop()
 
         repository.download_commits(
             repo_dir,
-            rev_start="children({})".format(commit["node"]),
+            rev_start="children({})".format(latest_commit),
```
suhaibmujahid: Same as https://github.com/mozilla/bugbug/pull/6904/changes#r4072000349

scripts/testing_policy_stats.py
```diff
@@ -30,12 +30,11 @@ def __init__(self, repo_dir: str) -> None:
...
 
         logger.info("Updating commits DB...")
-        for commit in repository.get_commits():
-            pass
+        latest_commit = collections.deque(repository.get_commits(), maxlen=1).pop()
 
         repository.download_commits(
             repo_dir,
-            rev_start="children({})".format(commit["node"]),
+            rev_start="children({})".format(latest_commit),
```
suhaibmujahid: Same as https://github.com/mozilla/bugbug/pull/6904/changes#r4072000349

bugbug/models/defect_enhancement_task.py
suhaibmujahid: Thank you!

suhaibmujahid (commented):
> Thanks for your support! 

It is my pleasure!  


> I've made the changes you suggested — let me know if there's anything else to adjust.

I left some comments.


> This is my first contribution to mozilla/bugbug, and I'm looking forward to contributing more.

You're welcome! I'm happy to help and look forward to having you as a long-term contributor. Feel free to also join us in the [bugbug](https://chat.mozilla.org/#/room/#bugbug:mozilla.org) Matrix room, it's the place to connect with the team.

bugbug/utils.py
goldpanth3r: Fixed.

docs/hackbot/agents.md
goldpanth3r: Dropped this change. It was from Ruff's formatter.

docs/hackbot/tools.md
goldpanth3r: Fixed.

scripts/commit_classifier.py
goldpanth3r: Fixed. Sorry for the oversight.

scripts/testing_policy_stats.py
goldpanth3r: Fixed.

scripts/generate_landings_risk_report.py
goldpanth3r: Fixed.

pyproject.toml
goldpanth3r: Sure! Filed #6919

docs/hackbot/agents.md
```diff
@@ -50,14 +50,14 @@ or code.
 
 ## The catalog
 
-| Agent                 | Does                                                                                                                                                                        | Source | Firefox build | Auto-applies actions |
-| --------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | :----: | :-----------: | :------------------: |
-| `bug-fix`             | Triage a Bugzilla bug and produce a candidate fix as a Phabricator revision. Also handles `@hackbot` follow-ups on an existing revision, and Bugzilla `needinfo?` requests. |  yes   |      yes      |       **yes**        |
```
suhaibmujahid: This seems an unrelated change.

suhaibmujahid (approved):

ksy36 (approved):

docs/hackbot/agents.md
```diff
@@ -50,14 +50,14 @@ or code.
...
+| `test-repair`         | Classify a CI test failure as regression or intermittent, blame the culprit commit, propose a fix.                                       |  yes   |      yes      |       **yes**        |
+| `build-repair`        | Analyze a Firefox build failure at a specific commit and produce a candidate fix.                                                        |  yes   |      yes      |          no          |
+| `frontend-triage`     | Read-only root-cause analysis and fix plan for a desktop frontend bug.                                                                   |  yes   |      no       |          no          |
+| `autowebcompat-repro` | Reproduce a web-compatibility report in headless Firefox via DevTools MCP.                                                               |   no   |      no       |          no          |
+| `test-plan-generator` | Generate Firefox QA test cases, run them through DevTools MCP, report results.                                                           |   no   |      no       |          no          |
```
suhaibmujahid:
```suggestion
| Agent                 | Does                                                                                                                                                                        | Source | Firefox build | Auto-applies actions |
| --------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | :----: | :-----------: | :------------------: |
| `bug-fix`             | Triage a Bugzilla bug and produce a candidate fix as a Phabricator revision. Also handles `@hackbot` follow-ups on an existing revision, and Bugzilla `needinfo?` requests. |  yes   |      yes      |       **yes**        |
| `test-repair`         | Classify a CI test failure as regression or intermittent, blame the culprit commit, propose a fix.                                                                          |  yes   |      yes      |       **yes**        |
| `build-repair`        | Analyze a Firefox build failure at a specific commit and produce a candidate fix.                                                                                           |  yes   |      yes      |          no          |
| `frontend-triage`     | Read-only root-cause analysis and fix plan for a desktop frontend bug.                                                                                                      |  yes   |      no       |          no          |
| `autowebcompat-repro` | Reproduce a web-compatibility report in headless Firefox via DevTools MCP.                                                                                                  |   no   |      no       |          no          |
| `test-plan-generator` | Generate Firefox QA test cases, run them through DevTools MCP, report results.                                                                                              |   no   |      no       |          no          |
```

docs/hackbot/agents.md
suhaibmujahid: Reverted in https://github.com/mozilla/bugbug/pull/6904#discussion_r4095944480

Issue #6467, opened 2026-08-03
Add common high-level operations to the client libraries


suhaibmujahid:
The client libs expose low-level calls. Add helpers for frequent use cases (e.g. fetch revision + diff, resolve reviewers) so callers don't reassemble them each time.

PR #6907, opened 2026-09-21
Fetch only the delivery's transactions from Conduit

### LINKED ISSUE #6911: Limit Phabricator webhook transaction searches to the current delivery
Hackbot currently calls `transaction.search` using only the revision PHID, which  fetch unrelated transactions from the revision's history before filtering them locally.

ayoubdiourin7:
Fixes  #6911


libs/phabricator-client/phabricator_client/client.py
```diff
@@ -119,11 +119,17 @@ async def conduit_request(self, method: str, **payload: Any) -> dict:
...
         return data["result"]
 
-    async def search_transactions(self, object_phid: str) -> list[dict]:
-        """Return the transactions (comments, status changes, ...) on an object."""
-        result = await self.conduit_request(
-            "transaction.search", objectIdentifier=object_phid
-        )
+    async def search_transactions(
+        self, object_phid: str, phids: list[str] | None = None
+    ) -> list[dict]:
```
suhaibmujahid, line 124: I would keep this method closer to the API structure, including the variable names.

Later as part of #6467, we can make more ergonomic.

suhaibmujahid (changes requested):

libs/phabricator-client/phabricator_client/client.py
ayoubdiourin7: updated in [23d7726](https://github.com/mozilla/bugbug/pull/6907/commits/23d7726365fead1d0b94749a73a35785865cbcb3)

suhaibmujahid (approved):

PR #6908, opened 2026-09-21
Return the anchor transaction PHID from detection

### LINKED ISSUE #6913: Return the anchor transaction PHID from Phabricator webhook detection
Phabricator webhook  does not expose which transaction represents the triggering.

ayoubdiourin7:
Fixes #6913


services/hackbot-api/app/phabricator_webhook.py
```diff
@@ -96,13 +97,27 @@ def find_hackbot_mentions(
...
+    """The transaction PHID that identifies this submission.
+
+    Phabricator's edit path adds at most one general ``comment`` transaction
+    per submission, so it is the natural handle; a submission made of inline
+    comments only has none, in which case the first mention stands in.
+    """
+    for mention in mentions:
+        if mention.comment_type == "regular":
+            return mention.transaction_phid
+    return mentions[0].transaction_phid
```
suhaibmujahid: Is the order here is guaranteed?

suhaibmujahid: Why not always use the first mention?

services/hackbot-api/app/phabricator_webhook.py
```diff
@@ -96,13 +97,27 @@ def find_hackbot_mentions(
...
+                    transaction_phid=transaction["phid"],
                     diff_id=diff_id,
                 )
             )
             break
     return matches
 
 
+def anchor_transaction_phid(mentions: list[HackbotMention]) -> str:
+    """The transaction PHID that identifies this submission.
```
suhaibmujahid: We might have an id through the header if it is not in the payload; that might be a more suitable choice as an identifier.

ayoubdiourin7: Here is a sanitized example of the received webhook:

```json
{
  "headers": {
    "Content-Type": "application/json",
    "X-Phabricator-Webhook-Signature": "<HMAC-SHA256>"
  },
  "body": {
    "object": {
      "type": "DREV",
      "phid": "PHID-DREV-example"
    },
    "triggers": [
      {
        "phid": "PHID-HRUL-example"
      }
    ],
    "action": {
      "test": false,
      "silent": false,
      "secure": false,
      "epoch": 1789763356
    },
    "transactions": [
      {
        "phid": "PHID-XACT-DREV-inline"
      },
      {
        "phid": "PHID-XACT-DREV-comment"
      },
      {
        "phid": "PHID-XACT-DREV-internal"
      }
    ]
  }
}
```

 I found that each phab webhook contains at most one regular-comment transaction, one or more inline-comment transactions, and internal transactions. it can also contain only inline comments. 
 And the headers provide no delivery ID

services/hackbot-api/app/phabricator_webhook.py
ayoubdiourin7: I checked the implementation [ transaction.search ](https://github.com/mozilla-conduit/phabricator/blob/0b65fc4dbed011cf2a538eb5326375d5860e4de5/src/applications/transactions/conduit/TransactionSearchConduitAPIMethod.php#L102-L123)uses the default transaction query ordering, which is transaction ID descending, so the first result is currently stable across retries

ayoubdiourin7: Yes, we can use the first mention. I chose the regular comment as the natural anchor for the submission, but we can also use the min if we don’t want to rely on ordering.

suhaibmujahid: Nice, we could use that.

suhaibmujahid: > we can also use the min if we don’t want to rely on ordering

I agree, even if it's ordered, that is an implementation detail, not part of what the API promises (unless it is documented somewhere), thus using `min()` would be safer for future changes on Phab side.

suhaibmujahid (approved):

services/hackbot-api/app/phabricator_webhook.py
```diff
@@ -203,4 +220,4 @@ async def detect_mention_and_revision(
         )
         return None
 
-    return comment, revision_id, bug_id
+    return comment, revision_id, bug_id, anchor_transaction_phid(authorized_mentions)
```
suhaibmujahid: It is long now, let us use name tuple for this.

services/hackbot-api/app/phabricator_webhook.py
ayoubdiourin7: okay , that is updated in [c0220cb](https://github.com/mozilla/bugbug/pull/6908/commits/c0220cb841602176198b5a7d3eefe533774f5ff7)

services/hackbot-api/app/phabricator_webhook.py
ayoubdiourin7: Agree, updated in [b39bf10](https://github.com/mozilla/bugbug/pull/6908/commits/b39bf1007b7b493ffbf7544650fcf7d8fab93295)

PR #6909, opened 2026-09-21
Deduplicate Phabricator webhook runs in the database

### LINKED ISSUE #6779: Add db-level deduplication for Phabricator webhooks
Similar to #6716 
Follow-up #6353 

Replace Phabricator’s process-local webhook deduplication with DB-level deduplication.

ayoubdiourin7:
Use the anchor transaction PHID as the persistent run deduplication key and remove the process-local TTL cache.

Closes #6779

---


services/hackbot-api/app/routers/webhooks.py
```diff
@@ -131,50 +124,35 @@ async def phabricator_webhook(
...
+        dedupe_key=f"phab-txn:{anchor_phid}",
     )
-    # Mark seen only after a successful trigger: if detection or the trigger call
-    # raises (transient Conduit/API failure), the delivery 500s and Phabricator's
-    # retry must be reprocessed rather than dropped as a duplicate.
-    for phid in fresh:
-        _seen_transactions[phid] = True
     log.info(
-        "Triggered bug-fix run %s for D%s (bug %s) from @hackbot mention",
+        "Triggered bug-fix run %s for D%s (bug %s) from @hackbot mention (%s)",
```
suhaibmujahid, line 161: Saying "Triggered" is not accurate in case it is duplicate. We can differentiate using `run.is_new`.

services/hackbot-api/app/routers/webhooks.py
```diff
@@ -131,50 +124,35 @@ async def phabricator_webhook(
...
         run.run_id,
         revision_id,
         bug_id,
+        anchor_phid,
     )
     return {"status": "triggered", "run_id": run.run_id}
```
suhaibmujahid, line 167: The response is not accurate now. 

We were returning `"status": "ignored"` on duplicates. I'm not sure if we need a response at all, I think nothing read it anyway.

suhaibmujahid (changes requested):

services/hackbot-api/app/routers/webhooks.py
ayoubdiourin7: nice catch! updated in [c7f1a34](https://github.com/mozilla/bugbug/pull/6909/commits/c7f1a346acf8a6f0c55b778f983d3987daa81208)

services/hackbot-api/app/routers/webhooks.py
ayoubdiourin7: i think for now we don't need it, but i updated in [c7f1a34](https://github.com/mozilla/bugbug/pull/6909/commits/c7f1a346acf8a6f0c55b778f983d3987daa81208)

services/hackbot-api/app/routers/webhooks.py
suhaibmujahid: It does not seem to be updated!

services/hackbot-api/app/routers/webhooks.py
suhaibmujahid: It does not seem to be updated!

suhaibmujahid (approved):

